PR #14, opened 2022-05-31
Update `ethers` to v5

### LINKED ISSUE #8: Use ethers v5
This is required by #7 anyway

ArmandoGraterol:
Most of the changes while migrating to v5 can be summarised with this list:

- Now `BigNumber` is imported from `ethers` instead of `ethers/utils` .

- `bigNumberify` was removed, now the way in which you transform a value into `BigNumber` is by using `BigNumber.from(value)`.

- How we decode and encode actions has changed since the `interface` object has changed. Example: From `colonyClient.interface.functions.emitDomainReputationPenalty.encode([...])` to `colonyClient.interface.encodeFunctionData([...])`

  - This `encodeFunctionData` has 2 params, the first one is where the function name goes, in the example case it should be `"emitDomainReputationPenalty"`, and in the second param we plug in the values of the function, or in our case, action like we did in the good ol' v4.

- Also, the imports of some utilities like `/constants`, `parseUnits`, `poll`, `/providers` changed. For example, frequently used `constants` as `AddressZero` can now be found in `@ethersproject/constant`.

The changes mentioned above are the most relevant to the code we currently have in the CDapp/Dapp, but a list of all the things to pay attention to when migrating from v4 to v5 can be found here: https://docs.ethers.io/v5/migration/ethers-v4/

Resolves #8


chinins (approved):
That's a lot files - nice patience! :) Looking good.

src/redux/sagas/utils/extendedInfuraProvider.ts
```diff
@@ -20,7 +20,7 @@ export class ExtendedInfuraProvider extends InfuraProvider {
       async () => {
         const result = await this._parentGetCode(addressOrName, blockTag);
         if (result === '0x') {
-          return undefined;
+          return '';
```
rdig, line 23: Why this change from `undefined` to an empty `string` ?

rdig (approved):
Lot of work went into this. Nice job @ArmandoGraterol :tada:

src/redux/sagas/utils/extendedInfuraProvider.ts
ArmandoGraterol: I saw that typescript was complaining about it returning `undefined` when according to the type definition of `getCode` it should return `Promise<String>`. 

So at first, I changed it to `Promise<String | undefined>` but then typescript would complain that the `Property 'getCode' in type 'ExtendedInfuraProvider' is not assignable to the same property in base type 'InfuraProvider'.` so, in the end, I left it as an empty string as that's still a falsy value.

This is also the case in the `ExtendedJsonRpcProvider`.

Do you think that this change could be a problem? It seems like this is an "internal" function as I didn't find places where it was used in the dapp

rdig: Nah, don't worry. Leave it in for now.

Hopefully we'll be able to remove these "custom" functions entirely at some point

PR #15, opened 2022-06-01
Upgrade react & react plugins

### LINKED ISSUE #3: Upgrade React and It's Plugins
As it says on the tin.

Please test the components inside `src/components/shared` as they're most likely affected by the upgrade

chinins:
## Description

This PR updates react & its plugins. The only 2 libraries that are not updated to their latest versions are:

- `react-router-dom`, not updating to version 6 as it doesn't have `Prompt` that we use (updated only to the latest v5). The authors of the library that they are working on `Prompt` so we'll have to update it to v6 once `Prompt` is included. We will also have to switch `useHistory` to `useNavigate`. Alternatively I can make an issue to substitute `Prompt` in the `SaveGuard` to update the package to v6. This will require the app to run though, so should be done separately later.
- `react-intl`. This requires update of typescript and there is a separate issue for that. I wanted to keep it separately as too not have PR too big.

I went through each component file that we have now to check for errors associated with updated packages.

**Changes** 🏗

- update package versions
- update some components based on the latest version of the packages

resolves #3

### DIFF
--- a/src/components/shared/BreadCrumb/SingleCrumb.tsx
+++ b/src/components/shared/BreadCrumb/SingleCrumb.tsx
@@ -1,17 +1,25 @@
-import React, { ReactElement } from 'react';
+import React, { isValidElement } from 'react';
+import { FormattedMessage } from 'react-intl';
 
 import NavLink from '~shared/NavLink';
 
+import { CrumbText } from './BreadCrumb';
 import styles from './BreadCrumb.css';
 
 interface Props {
-  crumbText: string | ReactElement;
+  crumbText: CrumbText;
   crumbLink?: string;
   lastCrumb?: boolean;
 }
 
 const SingleCrumb = ({ crumbText, crumbLink, lastCrumb }: Props) => {
   const crumbTitle = typeof crumbText === 'string' ? crumbText : '';
+  const text =
+    typeof crumbText === 'string' || isValidElement(crumbText) ? (
+      crumbText
+    ) : (
+      <FormattedMessage {...crumbText} />
+    );
 
   if (lastCrumb) {
     return (
@@ -19,10 +27,10 @@ const SingleCrumb = ({ crumbText, crumbLink, lastCrumb }: Props) => {
         <span className={styles.breadCrumble}>
           {crumbLink ? (
             <NavLink className={styles.invisibleLink} to={crumbLink}>
-              {crumbText}
+              {text}
             </NavLink>
           ) : (
-            crumbText
+            text
           )}
         </span>
       </div>
@@ -33,10 +41,10 @@ const SingleCrumb = ({ crumbText, crumbLink, lastCrumb }: Props) => {
       <span className={styles.breadCrumble}>
         {crumbLink ? (
           <NavLink className={styles.invisibleLink} to={crumbLink}>
-            {crumbText}
+            {text}
           </NavLink>
         ) : (
-          crumbText
+          text
         )}
       </span>
       <span className={styles.arrow}>&gt;</span>


dit7ya:
How do you recommend to test this one since we do not have any UI yet?

rdig:
@dit7ya  By looking at code. That's the only way you can right now.

Actual testing will come about when we start porting actual components

src/components/shared/BreadCrumb/SingleCrumb.tsx
```diff
@@ -1,19 +1,18 @@
...
 
 import NavLink from '~shared/NavLink';
 
-import { CrumbText } from './BreadCrumb';
-
 import styles from './BreadCrumb.css';
 
 interface Props {
-  crumbText: CrumbText;
+  crumbText: string | ReactElement;
```
rdig, line 8: So this will no longer accept Message Descriptors ? 

I think it should still accept them

src/components/shared/CSVUploader/CSVUploader.tsx
```diff
@@ -15,7 +15,10 @@ interface Props {
   setProcessingData: React.Dispatch<React.SetStateAction<boolean>>;
 }
 
-const MIME_TYPES = ['text/csv'];
+const MIME_TYPES = {
+  text: ['.csv'],
+};
+// ['text/csv'];
```
rdig, line 21: Leftover ?

src/components/shared/DatePicker/DatePickerContent.tsx
```diff
@@ -53,9 +53,12 @@ const DatePickerContent = ({
...
         onDayClick={handleDayPick}
-        selectedDays={isSelectedDay}
-        captionElement={(props) => <CaptionElement {...props} />}
-        navbarElement={(props) => <NavbarElement {...props} />}
+        selected={isSelectedDay}
+        components={{ Caption: (props) => <CaptionElement {...props} /> }}
+        /* Doesn't work with the new package version and we are not using it anywhere.
+         It's hard to fix it without running the app. SO leaving it here for now for
+         potential fix or removal */
+        // navbarElement={(props) => <NavbarElement {...props} />}
```
rdig, line 61: :+1:

src/components/shared/FriendlyName/FriendlyName.tsx
```diff
@@ -95,7 +95,7 @@ const FriendlyName = ({
       <div className={styles.name}>
         {userDisplay || (isColony && colonyDisplay) || (
           <MaskedAddress
-            address={userDisplayAddress || colonyDisplayAddress}
+            address={userDisplayAddress || colonyDisplayAddress || ''}
```
rdig, line 98: Maybe fallback to `AddressZero` ?

rdig (approved):
Mostly good with this, the only thing that gives me a bit of pause is the change in the Breadcrumb props.

rdig:
Also, leave React router to 5 for now. Once `Prompt` get implemented we'll upgrade as well.

ArmandoGraterol (commented):
I took a look at what's new in the v18 and it seems like there's a change that we might need to implement:

FireShot Capture 603 - How to Upgrade to React 18 – React Blog - reactjs org

We are currently using ReactDOM.render to render the root so it seems like that also needs to change in order for the CDapp to be fully in v18.

I found the info here: https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html

dit7ya:
Another cool thing about modern React is that we don't have to manually import React in each and every file (the JSX runtime is implicit). I guess we will set that up in linter rules later, so no need to remove them now.

rdig:
> Another cool thing about modern React is that we don't have to manually import React in each and every file (the JSX runtime is implicit). I guess we will set that up in linter rules later, so no need to remove them now.

Actually a good point. I totally missed this

chinins:
> Another cool thing about modern React is that we don't have to manually import React in each and every file (the JSX runtime is implicit). I guess we will set that up in linter rules later, so no need to remove them now.

I think we can keep them in older files, especially react eslint plugins haven't been updated yet (it gives errors if you remove react from imports). Maybe in the future, but not a pressing matter now. It doesn't really cost anything, just an extra line of imports in some of the files.

chinins:
> I took a look at what's new in the v18 and it seems like there's a change that we might need to implement:
> 

Good catch - slipped my mind among checking every component 🙈 Fixed now

ArmandoGraterol (approved):
Looking good! NIce work checking all the files :) Aside from what Raul pointed out and the react thing that you already fixed, this is good to go for me

src/components/shared/BreadCrumb/SingleCrumb.tsx
chinins: Done

PR #18, opened 2022-06-02
Remove provisioning script using local yarn path

### LINKED ISSUE #17: Remove provisioning script using local yarn path
## Description

The CDapp has removed the unneeded `yarn` dependency, however, the `provision_submodules.sh` script refers to a local path for `yarn`.

Remove the reference to the local `yarn` path.

arrenv:
## Description

The `provision_submodules.sh` was referencing a local `yarn` path that was not a dependency and thus not being installed. This now calls `yarn` directly, which should pick up the global path.

Resolves #17


chinins (approved):
I was able to run provision no problem.

rdig (approved):
Good with this, however maybe update the Readme so that it includes `yarn` as a prerequisite

ArmandoGraterol (approved):
LGTM :ok_hand:

PR #21, opened 2022-06-03
[Experimental] Add parcel2 dev server

dit7ya:
Parcel 2 is a modern build tool which uses [SWC](https://github.com/swc-project/swc) underneath. It covers all features that you expect from a build tool and some more.

https://parceljs.org/
https://github.com/parcel-bundler/parcel

Edit: Please ignore the whitespace-only changes in package.json - only React and Typescript have been upgraded, and it is only for testing purposes.

PR #22, opened 2022-06-03
[Experimental] Add vite as the dev server

dit7ya:
Vite is another modern build tool which uses [Esbuild](https://github.com/evanw/esbuild) underneath. 

https://vitejs.dev/guide/why.html
https://vitejs.dev/guide/features.html

For me, the crux of using modern build tooling is the speed and dev experience. Both Esbuild and SWC are blazing fast and recompilation is never flow-breaking. I also find these tools easier to use and much less complex than the older ones.


rdig:
See: https://github.com/JoinColony/colonyCDapp/pull/21#issuecomment-1145937245

rdig:
Closed by #27

PR #23, opened 2022-06-06
Updated redux & dependencies

### LINKED ISSUE #4: Update Redux and It's dependencies
As it says on the tin

`reselect` is one dependency you need to pay close attention to.

Also, make sure to check the files inside `src/redux` and especially `src/redux/sagas` as they might be affected by the update

danbr:
## Description

- Updated redux & dependencies to latest.
- Files inside src/redux and src/redux/sagas were checked. No errors were displayed, nor corrected.

Resolves #4


package.json
```diff
@@ -66,7 +67,6 @@
     "@types/react-redux": "^7.1.7",
```
dit7ya, line 67: This should be removed since react-redux 8 is written in Typescript and includes type information.

https://react-redux.js.org/using-react-redux/usage-with-typescript

dit7ya (approved):
LGTM except the comment I made.

package.json
danbr: Exactly. It was a draft as I verified the removal of the types, as per the recommendations.

rdig (approved):
LGTM

ArmandoGraterol (approved):
NIce work :ok_hand:

PR #24, opened 2022-06-08
Upgrade typescript

### LINKED ISSUE #20: Upgrade typescript version
Also upgrade `react-intl` as it requires higher version of typescript that we currently have.

chinins:
## Description

This PR upgrades typescript & react-intl (and corresponding libraries)

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

- update tsconfig file because of this:
https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-use-before-define.md#how-to-use

**Deletions** ⚰️

- delete `@formatjs/intl-unified-numberformat` as we are not using it anymore (it was deprecated and a different package created instead but we don't need it)

resolves #20


src/redux/immutable/FetchableData.ts
```diff
@@ -19,7 +19,7 @@ const defaultValues: DefaultValues<Shared> = {
 };
 
 export class FetchableDataRecord<R> extends Record<Shared>(defaultValues) {
-  get record() {
+  get getRecord() {
```
rdig, line 22: Why was this renamed ?

src/redux/immutable/FetchableData.ts
```diff
@@ -19,7 +19,7 @@ const defaultValues: DefaultValues<Shared> = {
 };
 
 export class FetchableDataRecord<R> extends Record<Shared>(defaultValues) {
```
rdig, line 21: I think there's a good case to be made to get rid of the whole fetchable data concept and it's not really relevant to us anymore

rdig (approved):
Looks good so far, but we might need to come back to it once the CDapp is up and running

src/redux/immutable/FetchableData.ts
chinins: Because it extends the interface below
```interface Shared {
  record: any;
  error?: string | void;
  isFetching?: boolean;
  lastFetchedAt?: Date;
}```
And it gives the following error
`(getter) FetchableDataRecord<R>.record: R
'record' is defined as a property in class 'Record<Shared> & Readonly<Shared>', but is overridden here in 'FetchableDataRecord<R>' as an accessor.ts(2611)`

src/redux/immutable/FetchableData.ts
chinins: Should I remove it in this PR or should I create a separate issue to do that?

rdig: I need to test it properly / think about it harder :)

So open a separate issue for it :)

PR #50, opened 2022-10-31
Port - Members page

### LINKED ISSUE #40: Port over Members page
This includes:
- components, which should live in `components/common`
- update and introduce new route
- decide on how to fetch the members, watches, reputation (from the db/lambda fn obviously)
- add required queries

Note that this should be split up into multiple issues

danbr:
## Description

This is the feature PR for porting over Members page into the CDapp.

This includes:

- components, which should live in components/common
- update and introduce new route
- decide on how to fetch the members, watches, reputation (from the db/lambda fn obviously)
- add required queries

Note the work wii be split up into multiple issues

Resolves #40


danbr:
This was incorrectly closed during resolving a merge conflict. I used VSCode to continue, instead of using the terminal, which is my usual course of action.

PR #54, opened 2022-10-31
Add GraphQL Voyager

jakubcolony:
## Description

This PR adds a `voyager` script that spins up a GraphQL Voyager, a tool for visualising and inspecting GraphQL API schemas:
image

It allows you to inspect the queries, mutations, input and return types as well as explore the relations between types. I found it quite handy when previously working with GraphQL APIs.

To run:
```
npm run voyager
```

### DIFF
--- a/temp-voyager/voyager.html
+++ b/temp-voyager/voyager.html
@@ -1,0 +1,32 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <script src="https://cdn.jsdelivr.net/npm/react@16/umd/react.production.min.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/react-dom@16/umd/react-dom.production.min.js"></script>
+
+    <link
+      rel="stylesheet"
+      href="https://cdn.jsdelivr.net/npm/graphql-voyager/dist/voyager.css"
+    />
+    <script src="https://cdn.jsdelivr.net/npm/graphql-voyager/dist/voyager.min.js"></script>
+  </head>
+  <body>
+    <div id="voyager">Loading...</div>
+    <script>
+      function introspectionProvider(introspectionQuery) {
+        return fetch('http://localhost:20002/graphql', {
+          method: 'post',
+          headers: {
+            'Content-Type': 'application/json',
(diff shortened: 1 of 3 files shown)


temp-voyager/voyager.html
```diff
@@ -0,0 +1,32 @@
...
+<html>
+  <head>
+    <script src="https://cdn.jsdelivr.net/npm/react@16/umd/react.production.min.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/react-dom@16/umd/react-dom.production.min.js"></script>
+
+    <link
+      rel="stylesheet"
+      href="https://cdn.jsdelivr.net/npm/graphql-voyager/dist/voyager.css"
+    />
+    <script src="https://cdn.jsdelivr.net/npm/graphql-voyager/dist/voyager.min.js"></script>
```
rdig, line 11: Interesting usage.

Why did you choose this approach, rather than using it as a dependency / react component ? Wouldn't this make updating it more work ?

_(I'm not saying it's wrong, I'm just curious)_

rdig (approved):
Very nice addition! :tada: 

This should prove useful down the road, as I have to disable some of the built-in quieries / mutations. So this will help see what's discoverable and what's not.

arrenv (approved):
Nice, I was just talking to Raul about using something like this as well but more as a planning tool. This looks like a good option.

temp-voyager/voyager.html
jakubcolony: I wanted it to run as a separate script/page so my first idea was to spin up a small express server that would use the official middleware provided by this package. Unfortunately the middleware's config was limited and I couldn't get it to work with Amplify. I decided to serve this static file using express instead. It didn't seem as a bad option considering it's a temporary thing.
I thought about using their official React component but unless I wanted it to become an integral page of the cdapp (which I didn't), I would have to set up a completely separate project that I could then serve with `webpack serve`, right? Do you think there's a better approach?

chinins (approved):
Very nice! 👍 good idea

danbr (approved):
Nice.... I look forward to using it. Well done.

Issue #41, opened 2022-10-25
Introduce User Roles Model


rdig:
- Decide on the schema / table model (it needs to be a Colony Model relationship)
- decide on the required lambda functions
- decide on strategies to create them initially and update them after
- add logic to block ingestor to listen for relevant events (`ColonyUserRoleSet`)
- It should also store a block number (maybe even chain?) in order to be able to do historical searches

Note that this _can_ be split into multiple issues

rdig:
This can be deffered to later, after we ported the bulk of the features, as it's not blocking anything really.

It should be done at the same time we introduce proper authentication to the lambda functions

rdig:
Closed as not relevant anymore

PR #69, opened 2022-11-02
Refactor Network Detection Logic

### LINKED ISSUE #29: Refactor Network Detection Logic
- Refactor an use required ENV variables
- constants that outline the supported chains (multiple)
- `isNetworkAllowed` checks need to account for multiple chains
- `isNetworkAllowed` moved into the app context

rdig:
This PR will refactor the logic we previously used in the Dapp to determine if a network was "allowed" or now, in the context that now the Dapp is multichain, deployed in one place, but supporting chains on multiple networks, and displaying colonies from multiple networks _(even though the user can only interact with networks which their wallet is connected to)_

# New General Purpose Hooks

1. `useUserAccountRegistered()` 
Takes no arguments and returns a boolean noting if the currently connected wallet has a wallet registered or not _(returns `false` if no wallet connected)_

2. `useCanInteractWithNetwork()`
Takes no arguments and returns a boolean noting if the currently connected wallet can interact with the contracts on the current chain. This is done via two checks: if it has an account registered, and if the wallet's provider is on the same chain as any of the available contracts we have deployed _(basically what chains we support)_. Note that it returns `false` if no wallet is connected

3. `useCanInteractWithColony(colony: Colony)`
Takes the colony object as an argument are returns a boolean noting if the user can interact with the currently loaded colony. It does this by checking if the current colony is on the same chain as the user's wallet provider. It also checks for the wallet being connected _(will return early if not)_, and if the user has an account registered.

## Changes

- [x] added new helper hooks: `useUserAccountRegistered`, `useCanInteractWithNetwork`, `useCanInteractWithColony`
- [x] added `Metadata` props in the database for the `Colony` and `Token` models
- [x] removed `deployed` prop from the `Colony` model as it's and artifact of a bygone age
- [x] refactored an cleaned up all components that had used `isNetworkAllowed` checks, to use the new hooks
- [x] Fixed `AppContext` to properly log out the user when disconnecting the wallet
- [x] Refactored `AvatarDropdown` to remove reliance on most external props

## Testing

Easiest way to test this is to go to the landing page http://localhost:9091/landing

### 1. Without a wallet connected you should **NOT** see:

- any colonies subscribed to (left sidebar)
- the create colony button -- plus sign icon (left sidebar)
- anything in the user dropdown menu except the help links and "create account" (get started)
- entry to create a new colony on the page (above "explore metacolony")

This is because
- no wallet connected, so we can't determine your current wallet network
- no wallet connected so we can't determine if you have your account registered or not

Screenshot from 2022-11-03 20-06-28

### 2. With the first `Dev Wallet 1` connected you **should** see:

- colonies subscribed to (left sidebar)
- ability to create new colony -- plus sign icon (left sidebar)
- link to the user profile in the user dropdown menu
- link to the user serttings in the user dropdown menu
- you should **NOT** see the entry to register a new account
- link to create a new colony in the user dropdown menu
- entry to create a new colony on the page (above "explore metacolony")

This is because
- you have a wallet connected
- the wallet is on the same network as the deployed contracts (so you can create a new colony)
- you have an account registered so you can manage your account
- you have an account registered so you can't register a new one

Screenshot from 2022-11-03 20-06-34

Resolves #29


src/components/shared/ColorSelect/ColorSelect.tsx
```diff
@@ -2,7 +2,9 @@ import React, { useCallback, useMemo, ReactNode, ComponentProps } from 'react';
 import { defineMessages } from 'react-intl';
 
 import { Appearance, Select, SelectOption } from '~shared/Fields';
-import ColorTag, { Color } from '~shared/ColorTag';
+import ColorTag from '~shared/ColorTag';
+
```
jakubcolony, line 6: is the empty line necessary?

jakubcolony (approved):
That was long! All looks good to me, left a few minor comments 👍

src/types/transactions.ts
```diff
@@ -34,3 +34,51 @@ export interface TransactionResponse {
...
+  FAILED = 'FAILED',
+  SUCCEEDED = 'SUCCEEDED',
+}
+
+export enum TRANSACTION_METHODS {
+  DeployToken = 'deployToken',
+  DeployTokenViaNetwork = 'deployTokenViaNetwork',
+  DeployTokenAuthority = 'deployTokenAuthority',
+  Approve = 'approve',
+}
```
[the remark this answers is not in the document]
rdig: Yes. The value for these enum are actual function names on the contract side. We use the enums to call them directly.

Think something like:
```
await contractClient[TRANSACTION_METHODS.DeployToken](...params);
```

So if we change them to camel case, this whole thing breaks.

src/components/frame/GasStation/MetaMaskWalletInteraction/MetaMaskWalletInteraction.tsx
```diff
@@ -0,0 +1,51 @@
...
+    id: 'users.GasStation.MetaMaskWalletInteraction.metamaskPromptText',
+    defaultMessage: `Please finish this {isMetatransaction, select,
+      true {metatransaction}
+      false {transaction}
+      other {action}
+    } on MetaMask`,
+  },
+});
+
+const displayName = 'users.GasStation.MetaMaskWalletInteraction';
```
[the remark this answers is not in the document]
rdig: Good catch

src/components/shared/ColorSelect/ColorSelect.tsx
rdig: Can't remember why it's in here. I'll remove it.

src/components/common/ColonyHome/ColonyMembers/MembersSubsection.tsx
```diff
@@ -85,14 +85,14 @@ const MembersSubsection = ({
 }: // maxAvatars = MAX_AVATARS,
 Props) => {
   const colonyWatchers = useMemo(() => watchers?.items || [], [watchers]);
-  const { user, wallet } = useAppContext();
```
Julianahlk, line 88: I notice you are removing the wallet retrieved from the useAppContext but it looks like we need the wallet address when using getAllUserRoles. Is there another way to get the wallet address or maybe getAllUserRoles will be removed?

rdig: > I notice you are removing the wallet retrieved from the useAppContext

I am not removing it. I am in fact adding it. `useAppContext` does not exist in the Dapp, it's something new, only available in the CDapp, and it's the new way to get the user's `wallet` and `user` object.

> getAllUserRoles

This is not relevant anymore in the CDapp, and will be removed / refactored by #41

rdig: PS: 

In this particular instance I may have removed the hook, since I didn't think it was required for this component _(Again, I did a rough pass over the components, just to get them render something)_

Feel free to put it back in, if you need values from it.

willm30 (commented):
Logged out, I am seeing the "add colony" icon on the left hand side:

loggedout

And logged in with Dev 1, my user menu is missing some of the items. I also don't see the `create a colony` menu item above `Explore the metacolony`

dev1

willm30:
Also, having connected dev wallet 1 and clicking "Sign Out", I still see the colonies list (which does disappear after a refresh). Will we address that, and if so, in this issue or elsewhere?

loggout

willm30:
I'm seeing the following warning in the console after clicking the `dev wallet 1` login:

err

rdig:
@willm30 

> I'm seeing the following warning in the console after clicking the `dev wallet 1` login:
> 
> err

This is the subscriptions query not returning correct data. But I can't replicate this locally, which makes me believe it's either something with your setup, or the graphql queries haven't re-generated properly.

Please test again, but stop everything (including the backend dev envrionment), make sure you're on the latest changes from the branch, then start the environment back up again + run the data generation script.

This will make sure that amplify generates the types for current schema properly.

Let me know if you can still reproduce that, and we'll investigate further then.

willm30 (approved):
@rdig Followed your advice and all is working smoothly. Good stuff!

rdig:
> @rdig Followed your advice and all is working smoothly. Good stuff!

Pfew! I was actually worried when I couldn't replicate. Thanks for checking again! :+1:

Julianahlk (approved):
This is working as expected! Nice work 🙌
No wallet connected
Screen Shot 2022-11-07 at 7 04 09 PM
Connected to dev wallet 1
Screen Shot 2022-11-07 at 7 05 39 PM

PR #73, opened 2022-11-03
Fix generating local config files

rdig:
This PR fixes lambda function creation _(well, actually fixes the whole `amplify` script, which would crash otherwise)_ when running `npm run amplify`

It does so, by generating some temporary, local, config files which `amplify` expects to exist prior to running.

Besides that, I've also made sure that if the CDapp doesn't have a local `.env` file already, it will generate one from the example file


jakubcolony (approved):
Works well 👍

chinins (approved):
Looking good!

PR #74, opened 2022-11-03
Port over Funds Logic

### LINKED ISSUE #39: Port over Funds Page
This includes:
- components, which should live in `components/common`
- update and introduce new route
- decide on how to fetch the token balance (from the db/lambda fn obviously)
- add required queries

Note that this should be split up into multiple issues

rdig:
This PR's will refactor the way we fetch balances for tokens, port over the `funds` components from the Dapp and wire them all together.

To fetch token balances we're going to use a lambda function which will connect to the chain via a provider and fetch current balances for the required tokens, for the required domains, since these are ever changing values which cannot be properly cached.

Another issue this PR will solve is the chain's native token balance, in the context of multi chain, as the `0x00000....0000` token will not be used in the context of the colony's specific chain.

The final piece of the puzzle here is going to be the block ingestor, which will be updated so that it can keep track of, and update the database with all the transactions a colony will need to claim

**In action:**
cool-tx-ingestor-demo

cdapp-demo-funds

## Testing ERC20 transfers

Start the `truffle` console:
```bash
npm run truffle console
```
Run the following inside the `truffle` console, one command at a time:
_(without comments -- it doesn't know how to interpret them)_
```js
// local dev ether router address
n = await IColonyNetwork.at('0x5CC4a96B08e8C88f2c6FC5772496FeD9666e4D1F')
// first local dev colony
c = await IColony.at('<YOUR-COLONY-ADDRESS>')
// token of the first local dev colony
t = await Token.at('<YOUR-TOKEN-ADDRESS>')
// mint 100 tokens
await c.mintTokens('100000000000000000000')
```
After this, you'll see funds available to be claims. Claim them, then refresh the page _(data won't refresh automatically)_, then check your new balance

**NOTE** If `c.mintTokens` errors out for you then go the alternative route and mint tokens from the Token contract and then transfer them into the colony:
```js
// mint 100 tokens
await t.mint('100000000000000000000')
// transfer them into the colony
await t.transfer(c.address, '100000000000000000000')
```

## Testing Native Chain transfers (ETH/XDAI)

Connect one of the ganache test accounts to your Metamask and send some ETH to your colony. After a page refresh you'll see a ETH claim showing up

**NOTE** Due to the nature of how chain native tokens work, multiple incoming claims (for ETH/XDAI) will show up just as one big incoming transaction (all the values added together), while for the ERC20 tokens you'll see individual incoming transactions _(although when claiming one, all of them will be claimed automatically)_

Resolves #39


src/components/common/ColonyHome/ColonyFundingWidget/TokenBalanceItem.tsx
```diff
@@ -2,35 +2,28 @@ import React from 'react';
...
-  // const domainBalance = balances.find(
-  //   ({ domainId }) => domainId === currentDomainId,
-  // );
-  // const balance = domainBalance && domainBalance.amount;
-  const balance = 0;
-
-  return typeof balance === 'undefined' ? null : (
+  balance,
+}: Props) =>
+  typeof balance === 'undefined' ? null : (
```
jakubcolony, line 26: Balance is defined as `string` but here we're checking if it's undefined... Something wrong with types along the way?

jakubcolony: Digging deeper, it can indeed be undefined (since it's an array element at an arbitrary index), in which case I think we should change the prop type

src/components/common/ColonyHome/ColonyUnclaimedTransfers/ColonyUnclaimedTransfers.tsx
```diff
@@ -42,31 +41,58 @@ const MSG = defineMessages({
...
+  /*
+   * Claims data needs to be merged, both ERC20's and Native Chain Tokens
+   *
+   * Also, we have to sort in-client since Apollo's cache is being a bitch
+   * not allowing more than 3 local field entries without breaking, forcing us
+   * to do the sorting / merging here, rather than at the time we fetch data.
+   * This kinda sucks!
+   */
+  const sortedFundsClaims = [...claims, chainClaimWithToken]
+    .filter((claim) => !!claim)
```
jakubcolony, line 76: We have an util method for that: `notNull` in `~utils/arrays`

src/components/common/TokenCard/TokenCard.tsx
```diff
@@ -0,0 +1,120 @@
...
+  },
+});
+
+const TokenCard = ({ domainId, token }: Props) => {
+  const { colony } = useColonyContext();
+  const { balances, nativeToken, status } = colony || {};
+  const { nativeToken: nativeTokenStatus } = status || {};
+
+  const currentTokenBalance =
+    getBalnceForTokenAndDomain(
```
jakubcolony, line 46: Typo

src/components/common/UnclaimedTransfers/UnclaimedTransfers.tsx
```diff
@@ -0,0 +1,87 @@
+import React from 'react';
+import { defineMessages, FormattedMessage } from 'react-intl';
+
+import { useColonyContext } from '~hooks';
+import { ADDRESS_ZERO } from '~constants';
+
+import UnclaimedTransfersItem from './UnclaimedTransfersItem';
+
+import styles from './UnclaimedTransfers.css';
+import { ColonyFundsClaim } from '~gql';
```
jakubcolony, line 10: Should be higher up

src/components/common/UnclaimedTransfers/UnclaimedTransfers.tsx
```diff
@@ -0,0 +1,87 @@
...
+const UnclaimedTransfers = () => {
+  const { colony } = useColonyContext();
+  const { profile, fundsClaims, chainFundsClaim, tokens } = colony || {};
+  const { items: claims = [] } = fundsClaims || {};
+
+  /*
+   * @NOTE We have to do some very heavy lifting (more or less) here due to us
+   * not being able to use Apollo's cache, so we want to short-circuit early
+   * in order to not waste any computing resources unecesarily
+   */
```
jakubcolony, line 34: This logic seem to be duplicated here and in `ColonyUnclaimedTransfers`. Could be extracted into a hook potentially?

src/components/common/UnclaimedTransfers/UnclaimedTransfers.tsx
```diff
@@ -0,0 +1,87 @@
...
+          {...MSG.title}
+          values={{
+            colony: profile?.displayName || 'colony',
+          }}
+        />
+      </div>
+      <ul>
+        {sortedFundsClaims.map((claim) => (
+          <UnclaimedTransfersItem
+            claim={claim as unknown as ColonyFundsClaim}
```
jakubcolony, line 76: The problem with this type is that you declare it as `ColonyFundsClaim`, which is the GraphQL type containing all the possible fields, whereas `claim` is only the subset of those fields that we actually query for. I suggest creating a fragment, exporting it from `~types/graphql.ts` and using that as the type in `UnclaimedTransfersItem`

src/context/apolloCache.ts
```diff
@@ -0,0 +1,23 @@
+import { InMemoryCache } from '@apollo/client';
+
+import { colonyCache } from '~cache';
+
+/*
+ * @NOTE Make any modifications to GraphQL query data in here (replaces old client-only resolvers)
+ *
+ * See: https://www.apollographql.com/docs/react/local-state/managing-state-with-field-policies
+ */
+const localOnlyFields = {
```
jakubcolony, line 10: I think calling this `localOnlyFields` is a bit misleading. What we seem to be doing is modifying the fields returned from the API, rather than adding fields that are only present locally (which is also possible, and may be the case in the future)

jakubcolony (commented):
Nice work! The claiming seems to work when I change the lambda functions timeout.

Apart from the long response times I've been experiencing, the testing instructions didn't work for me - if I try to mint tokens for the first colony (created by `temp-create-data` script), I get the unauthorized error and it only works with the manually created colonies (similarly to actions)

src/components/common/ColonyHome/ColonyUnclaimedTransfers/ColonyUnclaimedTransfers.tsx
rdig: I know, and I preferred to not use it :P I think I even removed it in some place. But can't really remember my reasoning back then for it.

Anyway, I'll refactor this to make use of it.

src/components/common/UnclaimedTransfers/UnclaimedTransfers.tsx
rdig: Make sense. Will do that.

src/context/apolloCache.ts
rdig: Sure. Will rename

rdig:
Cleaned up / fixed everything that Jakub pointed out in the first round of reviews, the biggest change of all being the introduction of the `useColonyFundsClaims` hook, along with it's composed types

src/components/common/ColonyHome/ColonyFundingWidget/ColonyFundingWidget.tsx
```diff
@@ -63,29 +91,27 @@ const ColonyFundingWidget = (/* { currentDomainId }: Props */) => {
           <FormattedMessage {...MSG.title} />
         </NavLink>
       </Heading>
-      {/* {data && !isLoadingTokenBalances ? ( */}
       <ul data-test="availableFunds">
         {tokens?.map((token) => (
-          <li key={token.tokenAddress}>
+          <li key={token?.tokenAddress}>
```
willm30, line 96: This `li` could be given a name and moved into own component (`AvailableFundItem`)?

src/components/common/ColonyHome/ColonyFundingWidget/TokenBalanceItem.tsx
```diff
@@ -2,35 +2,28 @@ import React from 'react';
...
-  // const domainBalance = balances.find(
-  //   ({ domainId }) => domainId === currentDomainId,
-  // );
-  // const balance = domainBalance && domainBalance.amount;
-  const balance = 0;
-
-  return typeof balance === 'undefined' ? null : (
+  balance,
+}: Props) =>
+  balance ? (
```
willm30, line 26: Since it doesn't make sense to render this component without `balance` defined, my sense is that this should be rendered as 

`balance && <TokenBalanceItem balance={balance} />`, in the parent. Then you can remove the ternary in here

src/components/common/ColonyHome/ColonyUnclaimedTransfers/ColonyUnclaimedTransfers.tsx
```diff
@@ -126,7 +117,7 @@ const ColonyUnclaimedTransfers = () => {
           <li>
             <Link
               className={styles.manageFundsLink}
-              to={`/colony/${name}/funds`}
+              to={`/colony/${colony?.name}/funds`}
```
willm30, line 120: The `ActionButton` above is not visibly disabled when it's actually disabled. It could benefit from being grey out or similar, or hidden entirely.

It and the tooltip should also probably live in their own component (`ClaimButton`) (there's some duplication with `UnclaimedTransferItem`).

src/components/common/ColonyTotalFunds/ColonyTotalFunds.tsx
```diff
@@ -82,24 +66,30 @@ const ColonyTotalFunds = () => {
...
 
   return (
     <div className={styles.main}>
       <div className={styles.selectedToken}>
         <Numeral
           className={styles.selectedTokenAmount}
           decimals={getTokenDecimalsWithFallback(currentToken?.token?.decimals)}
-          // value={currentToken.balances[COLONY_TOTAL_BALANCE_DOMAIN_ID].amount}
-          value={0}
+          value={totalTokenBalance?.balance as string}
```
willm30, line 92: Perhaps leave a note here as to why we can type cast this safely (i.e. null domain is applied in the cache so we know `find` above will return)

src/cache/colony/index.ts
```diff
@@ -0,0 +1,13 @@
+import tokensField from './tokens';
+import balancesField from './balances';
+
+const colony = {
+  Colony: {
+    fields: {
```
willm30, line 6: You mentioned apollo only permits 3 local fields. What's the third one stopping us from doing the claims logic here too? (re your comment about the claims logic (`@NOTE We have to do some very heavy lifting...`)

willm30 (commented):
Just mentioning here for others' benefit, this branch started worked nicely for me after creating a second colony. Wasn't behaving on the first one I created. I also couldn't test funds in different domains for the same reason as Jakub.

Other than that, functionality is all as expected. I've left comments mostly on code splitting.

One other thing is that we could hide the `incoming funds` heading when there are none, or at least leave a message below it. Else it looks like the actual funds are "incoming". 
willsworld

Couple of final thoughts (not necessarily for this pr):
1. Funds route needs mobile styles applied
2. Could we use redux to store balance state, so that when a claim is made, we could implement the optimistic ui pattern and not need to reload the browser to see the updates in the ui? This way, you could share the state between the funds route and the Colony Home page, and update it optimisitcally whenever a user makes a claim.

src/components/common/ColonyFunding/ColonyFunding.tsx
```diff
@@ -1,149 +1,141 @@
...
-
-  const selectedDomainLabel = 'Dummy Domain Label';
-  // const selectedDomainLabel: string = useMemo(() => {
-  //   const { label = '' } =
-  //     domainChoices.find(
-  //       ({ value }) => value === selectedDomainId.toString(),
-  //     ) || {};
-  //   return typeof label === 'string' ? label : formatMessage(label);
-  // }, [domainChoices, formatMessage, selectedDomainId]);
+  const domainChoices = useMemo<
```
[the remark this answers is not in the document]
rdig: Already done, see 574ac3ffa3fd2bb92cb129cb42229610a0e848fd

src/components/common/ColonyFunding/ColonyFunding.tsx
```diff
@@ -34,112 +34,103 @@ const MSG = defineMessages({
...
-  // }
-
-  // if (!colonyName || error || !data?.processedColony) {
-  //   console.error(error);
-  //   return <NotFoundRoute />;
-  // }
-
-  // const { processedColony: colony } = data;
+  // eslint-disable-next-line react/no-unstable-nested-components
+  const Aside = () => (
```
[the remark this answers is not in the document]
rdig: While I appreciate this, and agree with you that is should live outside, we're going to change the whole UI in like 2-3 months. 

I'm not going to waste time improving **Dapp** component logic _(which this is, since every component I've added was copy pasted from there)_, only to have to remove it in the very near future

src/components/common/ColonyFunding/ColonyFunding.tsx
```diff
@@ -1,149 +1,141 @@
...
-  //       <LoadingTemplate loadingText={MSG.loadingText} />
-  //     </div>
-  //   );
-  // }
+  const selectedDomainLabel: string = useMemo(() => {
+    const { label = '' } =
+      domainChoices.find(
+        ({ value }) => value === selectedDomainId.toString(),
+      ) || {};
+    return typeof label === 'string' ? label : formatMessage(label);
```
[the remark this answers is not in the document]
rdig: Legacy from the Dapp

src/components/common/ColonyFundingBanner/ColonyFundingBanner.tsx
```diff
@@ -0,0 +1,50 @@
...
+    id: `${displayName}.tipText`,
+    defaultMessage: `Tip: to fund your colony, send tokens to your colony's address:`,
+  },
+});
+
+const ColonyFundingBanner = () => {
+  const { colony } = useColonyContext();
+  const { colonyAddress = '' } = colony || {};
+
+  const formattedAddress = useMemo(() => {
```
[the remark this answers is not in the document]
rdig: 1. See https://github.com/JoinColony/colonyCDapp/pull/74#discussion_r1131145391
2. Not really, since this is the only place in the whole Dapp/Cdapp this is being used

willm30:
Also, storing name as an empty string when transferring the native token is giving me a dynamo db error
eth

err

src/components/common/ColonyFunding/ColonyFunding.tsx
```diff
@@ -1,149 +1,141 @@
...
-                onChange={(value) => setSelectedDomainId(Number(value))}
-                options={domainChoices}
+            {isMobile ? (
+              <ColonyDomainSelector
+                filteredDomainId={selectedDomainId}
+                onDomainChange={setSelectedDomainId}
               />
-            </Form> */}
+            ) : (
+              <Form
```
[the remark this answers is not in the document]
rdig: Again, good improvement, but not going to do it, since this potentially wasted time. See: https://github.com/JoinColony/colonyCDapp/pull/74#discussion_r1131145391

src/components/common/ColonyFundingMenu/ColonyFundingMenu.tsx
```diff
@@ -0,0 +1,169 @@
...
+    const isSupportedColonyVersion = true;
+
+    interface FIProps {
+      text: SimpleMessageValues;
+      handleClick: () => void;
+      disabled: boolean;
+    }
+
+    // eslint-disable-next-line react/no-unstable-nested-components
+    const FundingItem = ({ text, handleClick, disabled }: FIProps) => (
```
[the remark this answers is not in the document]
rdig: Good suggestion, but not doing it. See: https://github.com/JoinColony/colonyCDapp/pull/74#discussion_r1131145391

src/components/common/ColonyFundingMenu/ColonyFundingMenu.tsx
```diff
@@ -0,0 +1,169 @@
...
+          disabled={
+            !canInteractWithColony || !isSupportedColonyVersion || disabled
+          }
+        />
+      </li>
+    );
+
+    return (
+      <ul className={styles.main}>
+        <FundingItem
```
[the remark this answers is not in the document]
rdig: Why would this improvement be warranted here? 

It's a list of 3 links, which didn't change, and won't change until we tear it down, so why I add the extra logic ?

src/components/common/ColonyHome/ColonyFundingWidget/ColonyFundingWidget.tsx
rdig: Good suggestion, not doing it: https://github.com/JoinColony/colonyCDapp/pull/74#discussion_r1131145391

rdig:
> Also, storing name as an empty string when transferring the native token is giving me a dynamo db error

I don't really understand what you're saying here? You bypassed validation and created a token without a name string? Or is the validation broken and allowed you to create one?

src/components/common/ColonyTotalFunds/ColonyTotalFunds.tsx
rdig: Did one better, I made a better fallback for `totalTokenBalance` removing the need to forcefully cast it

willm30:
> I don't really understand what you're saying here? You bypassed validation and created a token without a name string? Or is the validation broken and allowed you to create one?

@rdig By native token, I mean when transferring Eth via metamask.

This happens when I:
-> Login to a wallet via metamask
-> Transfer eth to the colony
-> Log back in via a dev wallet (currently unable to join a colony when logged in via metamask because `web3-onboard` is currently set up to validate that metamask users are on gnosis chain)
-> Claim funds

The entry the block ingestor submits to the db has no name, which is generating the error.

rdig:
@willm30 Something's off here

I understand the process you are describing, but ETH transfers / claims have no bearing on the block ingestor. They are not handled like normal ERC20 transfers, they are fetched, directly from the chain by a lambda function.

I never experienced what you're describing when claiming ETH, but am curious to find out what's going on. So if you want we can have a call tomorrow to debug this.

src/components/common/ColonyTotalFunds/ColonyTotalFunds.tsx
```diff
@@ -82,24 +66,30 @@ const ColonyTotalFunds = () => {
...
+           */
+          .filter((balance) => balance?.domain === null)
+          .find(
+            (balance) =>
+              balance?.token?.tokenAddress === currentToken.token.tokenAddress,
+          ) || { balance: '0' }
+      );
+    }
+    return { balance: '0' };
+  }, [balances, currentToken]);
```
willm30, line 84: Could all the logic from `useColonyContext` down to here become `useColonyTotalFunds`, like `useColonyFundsClaims`? 

Not a big point, especially if we'll be revisiting this all again shortly. More trying to get a sense of whether that's a pattern we like in general for components with more than a few lines of logic above the view (and thus something we can implement generally going forward).

willm30 (approved):
Working for me (aside from the warning I mention above. Happy to look at that together tomorrow, I'll ping you) 👍

src/cache/colony/index.ts
rdig: That's exactly the problem, any other field I add to that object, it crashes everything.

src/components/common/ColonyTotalFunds/ColonyTotalFunds.tsx
rdig: Sure, I'm happy to extract as much logic as I can into hooks

rdig:
Did a new round of fixes, mostly prompted by the issue that @willm30 found with the chain native token not having proper data set. Turns out it was part intentional, I just forgot to follow trough in the cache update. All should be fixed now.

If you want to re-review these are the changes that I've added:
- the cache update fix: 2361d4a43f9691a624f983e8e75f0ce5ecaba90e which is the most important one of the bunch
- extracted some totals / tokens logic into specialized hooks: 58a9797ef6968f853e50ea328c83d16b54940371 and 6251ec4c23b31143facbde4b9993a256f94853e1

@jakubcolony I know about that but I wanted to avoid tackling it here since I want to find a proper solution for that _(and not just tack apollo query triggers at t he end of the saga)_, so maybe we can have a call about that sometime. It's no rush, since this will come up in other PRs as well, and once we do it, we can fix all instances of this occurring

src/cache/colony/balances.ts
```diff
@@ -0,0 +1,109 @@
...
+       * in order to dynamically contruct this (since multiple colonies
+       * can be deployed to multiple chains)
+       *
+       * @WARNING we have to get the full domain, and basically recreate the
+       * whole balance object (including the typename) since, if the alternative
+       * is true, and we pass the reference along, it will actually prevent
+       * us from returning the modified data, and will overwrite our returned
+       * results with the ones already available in the cache.
+       */
+      const allDomainBalances = domainBalanceRefs.items.map((balanceRef) => {
```
willm30, line 51: I wonder if this also could be defined as a function that lives outside this main function, e.g. `const getAllDomainBalances = (domainRefs, readField, cache) => {...}`

Mainly from a readability perspective. Seeing four consecutive `return`s followed by an object is a little hard to follow.

src/cache/colony/balances.ts
```diff
@@ -0,0 +1,109 @@
...
+        const domain = cache.data.data[`Domain:${domainId}`];
+        return {
+          __typename: BALANCE_TYPE_NAME,
+          id,
+          domain,
+          token: getTokenDataWithNative(token),
+          balance,
+        };
+      });
+      const totalTokenBalances = allDomainBalances.reduce((totals, balance) => {
```
willm30, line 76: A similar thought here, e.g. 


``` js
const totalTokenBalances = getTotalTokenBalances(allDomainBalances)
```

willm30 (commented):
I'm still seeing this dynamo db error when claiming Eth (`id cannot be empty string`). Looks to be caused by the block ingestor trying to mutate the db in response to the claim event. (am definitely on latest branch)

errdyanmo

And also not seeing the name of the native token in the db (can't remember if this was intentional or not):

funds

(Here's the exact process I'm following in case it helps reproduce:
1. Build dev env on this branch
2. Manually create colony. Mint tokens via truffle. Claim didn't show up so I created a second colony manually.
3. Mint tokens via truffle and claim them in ui. Works fine.
4. Send eth from metamask directly to colony using a dev wallet I just reset in metamask ui. Claim eth via the ui and then I see the error.)

rdig:
> I'm still seeing this dynamo db error when claiming 

@willm30 two things here:
- You're not claiming ETH in that error, you'r claiming and ERC20 token, ETH shows up as `0x00000...` in that log output, so it's something else at play
- It's actually impossible, using the commit the block ingestor is running in this PR, to claim ETH: https://github.com/JoinColony/block-ingestor/blob/master/src/eventProcessor.ts#L166-L171 _(unless you changed the commit in the docker compose config manually)_

If you want, we can have a call to debug this again, but this is not ETH related

src/cache/colony/balances.ts
rdig: Don't care pro or con here, so If you insist, sure.

PR #56, opened 2022-11-01
Clean up Colony Home

jakubcolony:
## Description

This PR attempts to clean up the `ColonyHome` and related components as much as currently possible.

**Changes** 🏗

* Ported over getting the filtered domain id from search params
* Renamed the aside components to avoid confusion with identically named pages

Resolves #36

PR #76, opened 2022-11-04
Port over member reputation

### LINKED ISSUE #62: Port over user's reputation
Port over the components responsible for displaying user's reputation:
image

jakubcolony:
## Description

This PR ports over member reputation. This includes:
 - Creating a new `getUserReputation` lambda function that generates a GraphQL query
 - Porting over and refactoring `MemberReputation` component
 - Creating a `useUserReputation` hook to encapsulate reputation fetching logic
 - In response to Raul's comment, this PR also adds a functions-specific eslint config to disable a conflicting eslint rule

## To test
As there's no reputation in any of the colonies yet, you'll see the following:
Screenshot 2022-11-04 at 21 19 36

You can edit the lambda function to return ~13.81% reputation by applying the following patch:
```
diff --git a/amplify/backend/function/getUserReputation/src/index.js b/amplify/backend/function/getUserReputation/src/index.js
index b725cd6..b5babf2 100644
--- a/amplify/backend/function/getUserReputation/src/index.js
+++ b/amplify/backend/function/getUserReputation/src/index.js
@@ -3,6 +3,7 @@ const { getColonyNetworkClient, Network } = require('@colony/colony-js');
 const {
   providers,
   utils: { Logger },
+  constants: { AddressZero },
 } = require('ethers');
 
 Logger.setLogLevel(Logger.levels.ERROR);
@@ -35,16 +36,20 @@ exports.handler = async (event) => {
 
   const { skillId } = await colonyClient.getDomain(domainId ?? ROOT_DOMAIN_ID);
 
-  try {
-    const { reputationAmount } = await colonyClient.getReputationWithoutProofs(
-      skillId,
-      walletAddress,
-      rootHash,
-    );
+  return walletAddress === AddressZero
+    ? '11904644120813032491751554'
+    : '1644530790622401088627970';
 
-    return reputationAmount.toString();
-    // eslint-disable-next-line no-empty
-  } catch {}
+  // try {
+  //   const { reputationAmount } = await colonyClient.getReputationWithoutProofs(
+  //     skillId,
+  //     walletAddress,
+  //     rootHash,
+  //   );
 
-  return null;
+  //   return reputationAmount.toString();
+  //   // eslint-disable-next-line no-empty
+  // } catch {}
+
+  // return null;
 };

```

Once you refresh, you should see the following:
Screenshot 2022-11-04 at 21 19 45

## TODO

- [x] Lambda function
- [x] MemberReputation component

Resolves #62


rdig:
@jakubcolony 

 > Please note: Only the 3 newest commits are related to this PR, the rest are shown because this branch is based off https://github.com/JoinColony/colonyCDapp/pull/56

Then change the base branch of your PR _(to that of the branch you're based from)_ and you'll only see the relevant ones. It also makes it easier to review.

I'll wait on my review until you do that.

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -81,18 +89,17 @@ type Query {
...
-  ): User @function(name: "createUniqueColony")
-  createUniqueDomain(
-    input: CreateUniqueDomainInput
-  ): Domain @function(name: "createUniqueDomain")
+  createUniqueUser(input: CreateUniqueUserInput): User
+    @function(name: "createUniqueUser")
+  createUniqueColony(input: CreateUniqueColonyInput): User
+    @function(name: "createUniqueColony")
+  createUniqueDomain(input: CreateUniqueDomainInput): Domain
+    @function(name: "createUniqueDomain")
```
rdig, line 102: Did you change these or was it automated ?

docker/files/amplify/amplify-meta.json.base
```diff
@@ -35,11 +35,16 @@
...
+    },
     "createUniqueDomain": {
       "build": true,
       "providerPlugin": "awscloudformation",
       "service": "Lambda"
+    },
+    "getUserReputation": {
+      "build": true,
+      "providerPlugin": "awscloudformation",
+      "service": "Lambda"
```
rdig, line 47: Very nice!

amplify/backend/function/getUserReputation/src/index.js
```diff
@@ -0,0 +1,40 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
```
rdig, line 1: I think we to think about putting a eslint config file at the `/amplify/backend/function/` level so that it has rules just for the contained files.

Otherwise we'll just litter the lambda function code full of `//eslint-disable`'s

src/components/frame/RouteLayouts/UserNavigation/UserNavigation.tsx
```diff
@@ -39,6 +39,15 @@ const MSG = defineMessages({
...
 const UserNavigation = () => {
+  /**
+   * TODO: Refactor after #67 is done
+   */
+  const { colonyName } = useParams<{ colonyName: string }>();
+  const { data } = useQuery(gql(getFullColonyByName), {
+    variables: { name: colonyName },
+  });
+  const [colony] = data?.getColonyByName?.items || [];
+
```
rdig, line 50: _PS: I'm not saying this because you introduced this here, I'm just ranting!_

This is such bad component architecture. Colony data should not live in here.

But it's nothing we can do right now. Because this is how the design "is".

Hopefully, when we get to the redesign part of it, we'll be able to properly separate user/colony concerns.

src/components/shared/MemberReputation/MemberReputation.tsx
```diff
@@ -41,24 +42,35 @@ const MemberReputation = ({
...
-    variables: { address: walletAddress, colonyAddress, domainId, rootHash },
+  const { data: userReputationData } = useQuery(gql(getUserReputation), {
+    variables: {
+      input: {
+        colonyAddress,
+        walletAddress,
+        domainId,
+        rootHash,
+      },
+    },
```
rdig, line 53: This is is a `shared` _(old Dapp `core`)_ type component.

They should be re-usable, and mostly for display purposes. They should not handle fetching data internally. All that data fetching should happen in the `common` type components of the CDapp.

Now I know this was this way when you brought it over from the Dapp, but this would be a good opportunity to fix that.

Please refactor this, so that the data is fetched outside from it _(most likely in `UserNavigation`)_, and this component just ingests it and displays it (maybe handle formatting

rdig (changes requested):
Very nice work! :100: 

This is all good, I've just rejected it since I want you to also refactor the shared `MemberReputation` component. Once you do that, this will be good to.

Also, for testing, those values you used are not realistic. Here's something more relevant:
total colony rep amount = `11904644120813032491751554`
user rep amount =  `1644530790622401088627970`

These, plugged in, should give the user an approx reputation of `13.81` percent inside the colony

amplify/backend/function/getUserReputation/src/index.js
```diff
@@ -0,0 +1,40 @@
...
+  });
+  const colonyClient = await networkClient.getColonyClient(colonyAddress);
+
+  const { skillId } = await colonyClient.getDomain(domainId ?? ROOT_DOMAIN_ID);
+
+  const { reputationAmount } = await colonyClient.getReputationWithoutProofs(
+    skillId,
+    walletAddress,
+    rootHash,
+  );
```
rdig, line 37: Please wrap this call inside a try/catch block and handle "rep not available" errors as not errors, since they aren't really, it's just the reputation oracle throwing them off.

Screenshot from 2022-11-07 13-43-03

Otherwise it will pollute the log output with useless errors.

amplify/backend/function/getUserReputation/src/index.js
```diff
@@ -0,0 +1,40 @@
...
+const { getColonyNetworkClient, Network } = require('@colony/colony-js');
+const { providers } = require('ethers');
+
+const ROOT_DOMAIN_ID = 1; // this used to be exported from @colony/colony-js but isn't anymore
+const RPC_URL = 'http://network-contracts.docker:8545'; // this needs to be extended to all supported networks
+
+/**
+ * @type {import('@types/aws-lambda').APIGatewayProxyHandler}
+ */
+exports.handler = async (event) => {
```
rdig, line 11: Please set `ethers`'s log level to `ERROR` only, to remove the "Duplicate contract definition" warnings in the console:
Screenshot from 2022-11-07 13-43-14

You do this by importing the `Logger` class from `ethers` and passing it a error config level:
```js
import { utils } from 'ethers/utils';

const { Logger } = utils;

Logger.setLogLevel(Logger.levels.ERROR);
```

rdig (commented):
Some some more things, but these are just minor console output cleanups

src/hooks/useUserReputation.ts
```diff
@@ -0,0 +1,53 @@
...
+    variables: {
+      input: {
+        colonyAddress: colonyAddress ?? '',
+        walletAddress: walletAddress ?? '',
+        domainId,
+        rootHash,
+      },
+    },
+    fetchPolicy: 'cache-and-network',
+    skip: !colonyAddress || !walletAddress,
```
rdig, line 29: I never noticed the use of `skip` when I first reviewed this. Nice addition :+1:

src/hooks/useUserReputation.ts
```diff
@@ -0,0 +1,53 @@
...
+  const { data: userReputationData } = useQuery(gql(getUserReputation), {
+    variables: {
+      input: {
+        colonyAddress: colonyAddress ?? '',
+        walletAddress: walletAddress ?? '',
+        domainId,
+        rootHash,
+      },
+    },
+    fetchPolicy: 'cache-and-network',
```
rdig, line 28: I'm not sure `cache-and-network` works here, as I think it should be `network-only` to get the freshest reputation always, but I don't have enough arguments for one of the other.

So for the time being leave it as it is, and if we notice it doesn't work as we expect it, then we'll address it.

amplify/backend/function/getUserReputation/src/index.js
```diff
@@ -30,11 +35,16 @@ exports.handler = async (event) => {
...
+      walletAddress,
+      rootHash,
+    );
+
+    return reputationAmount.toString();
+    // eslint-disable-next-line no-empty
+  } catch {}
 
-  return reputationAmount.toString();
+  return null;
```
rdig, line 48: I don't think this disable is required. I think you can just catch the error, but not do anything with the variable

```suggestion
} catch (error) {
  return null;
}
```

But it works either way, so I don't really care. Leave it as it is.

amplify/backend/function/getUserReputation/src/index.js
```diff
@@ -0,0 +1,49 @@
...
+  const colonyAddress = input?.colonyAddress;
+  const walletAddress = input?.walletAddress;
+  const domainId = input?.domainId;
+  const rootHash = input?.rootHash;
+
+  const provider = new providers.JsonRpcProvider(RPC_URL);
+
+  const {
+    etherRouterAddress: networkAddress,
+    // eslint-disable-next-line global-require
```
rdig, line 26: I'd also disable this rule in the function-level eslint, since we use it in all functions basically.

rdig (approved):
Good to go!

src/hooks/useUserReputation.ts
jakubcolony: I only added it in one of the latest commits as part of moving it to its own hook

src/hooks/useUserReputation.ts
jakubcolony: I believe `cache-and-network` will eventually get the freshest reputation with the only exception being it will show the cached value (which I feel like would be correct most of the time) while it performs the network request. I didn't actually specify this myself but rather copied it from the dapp

chinins (approved):
Looking good - great job!

Issue #68, opened 2022-11-02
Port Members:Create `getUsersReputationForColony' lambda function


danbr:
## Description

<Partially written to organise my thoughts, maybe it will help others when they embark on porting their first feature to the CDapp>
Create a lambda function to obtain members of a Colony.

`getUsersReputationForColony' 
parameter: Colony address
output: [{ user id, reputation }]

The function should:

- Look up Colony
- get users of Colony (graphQL)
- get reputation of each user from RPC endpoint (reputation miner)
- return clean & organised data

The function skeleton is created using the template provided Amplify.

rdig:
Is this planning to get the reputation for users in all domains at once ? or just for one domain at a time ?

There's also two things I want to mention:
- Once #67 gets implement, this can be merged into the colony object in that provider, and be available everywhere the colony data is available
- we'll actually get some caching benefits out of the box, as since this query will run through Apollo, it will be cached locally by default

rdig:
I also have some ideas on how to deffer member reputation data loading and leverage part of the database to store members, but that involves many more moving parts and should be an optimization for down the line, after everything is implemented.

willm30:
MERGED #111

Issue #63, opened 2022-11-02
Cleanup and Refactor `UserAvatar` component


rdig:
- cleanup imports and remove what's no longer required
- remove all traces of IPFS / `withDataFetcher` logic
- Remove the `Hooked` higher order function
- make sure styles are up to date with the latest ones from the Dapp
- introduce logic to account for thumbnails, as we now have that value in the db

rdig:
This will also clean up, and wire the user tooltip

PR #77, opened 2022-11-06
Refactor colony home members/watchers aside

### LINKED ISSUE #38: Clean up Colony Home Members/Watchers aside
A lot of commented out code got left behind my original colony wiring.

Please take time and clean it up, refactor it properly. This includes component, styles, and everything else that's needed.

Please inspect the component before starting it, as there might be features not available yet, like #29 and decide when / how to handle that

Julianahlk:
## Description

This is an attempt to clean up the colony home members/watchers aside and other related components.

**New stuff** ✨

* Ported `InviteLinkButton` component
* Ported `ClickableHeading` component

Resolves #38

### DIFF
--- a/src/components/common/ColonyHome/ColonyMembersWidget/MembersSubsection.tsx
+++ b/src/components/common/ColonyHome/ColonyMembersWidget/MembersSubsection.tsx
@@ -6,9 +6,8 @@ import Heading from '~shared/Heading';
 import { Tooltip } from '~shared/Popover';
 import UserAvatar from '~shared/UserAvatar';
 // import Icon from '~shared/Icon';
-// import InviteLinkButton from '~dashboard/InviteLinkButton';
+import InviteLinkButton from '~shared/Button/InviteLinkButton';
 
-// import HookedUserAvatar from '~users/HookedUserAvatar';
 // import { useAvatarDisplayCounter, useTransformer } from '~hooks';
 // import {
 //   Colony,
@@ -16,9 +15,7 @@ import UserAvatar from '~shared/UserAvatar';
 //   ColonyWatcher,
 // } from '~data/index';
 import { COLONY_TOTAL_BALANCE_DOMAIN_ID } from '~constants';
-// import { useUserAccountRegistered } from '~hooks';
-// import { getAllUserRoles } from '~modules/transformers';
-// import { hasRoot, canAdminister } from '~modules/users/checks';
+import { useAppContext } from '~hooks';
 import { Colony, User } from '~types';
 
 import styles from './ColonyMembersWidget.css';
@@ -72,8 +69,6 @@ interface Props {
   isContributorsSubsection: boolean;
 }
 
-// const UserAvatar = HookedUserAvatar({ fetchUser: true });
-
 // const MAX_AVATARS = 12;
 
 const MembersSubsection = ({
@@ -85,12 +80,8 @@ const MembersSubsection = ({
 }: // maxAvatars = MAX_AVATARS,
 Props) => {
   const colonyWatchers = useMemo(() => watchers?.items || [], [watchers]);
-  // const userHasAccountRegistered = useUserAccountRegistered();
-
-  // const allUserRoles = useTransformer(getAllUserRoles, [
-  //   colony,
-  //   currentUserWalletAddress,
-  // ]);
+  const { user } = useAppContext();
+  // const hasRegisteredProfile = user?.name;
   // const canAdministerComments =
   //   userHasAccountRegistered &&
   //   (hasRoot(allUserRoles) || canAdminister(allUserRoles));
@@ -131,16 +122,15 @@ Props) => {
             />
           </NavLink>
         </Tooltip>
-        {/* {!isContributorsSubsection && (
+        {!isContributorsSubsection && (
           <InviteLinkButton
-            colonyName={colonyName}
-            buttonAppearance={{ theme: 'blueWithBackground' }}
+            colonyName={name}
+            buttonAppearance={{ theme: 'blue' }}
           />
-        )} */}
+        )}
       </div>
     ),
-    // [members, membersPageRoute, isContributorsSubsection, colonyName],
-    [isContributorsSubsection, membersPageRoute, colonyWatchers],
+    [isContributorsSubsection, membersPageRoute, colonyWatchers, name],
   );
 
   if (!colonyWatchers) {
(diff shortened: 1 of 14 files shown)


rdig (changes requested):
This is not yet ready. :disappointed: 

It's missing:
- the user avatar display counter
- hover styles around the user avatar which are currently broken:
Screenshot from 2022-11-06 13-33-28

It does **not** have to include:
- user banned logic -- will be handled alongside comments
- user avatar tooltip -- will be handled by #63

Julianahlk:
> * the user avatar display counter

@rdig , For the user avatar display counter, I was thinking this was dependent on  #40 because the `MembersSubsection` needs a replacement for useContributorsAndWatchersQuery in `ColonyMembers`. Do you think it's better to handle this query as part of this task?

> * hover styles around the user avatar which are currently broken:

For the user avatar hover styles, I was working on the refactoring of the Avatar component in #63 and I it looks like that work should handle the avatar hover styles as that cleans up the various Popover components.

rdig:
>  For the user avatar display counter, I was thinking this was dependent on https://github.com/JoinColony/colonyCDapp/issues/40 because the MembersSubsection needs a replacement for useContributorsAndWatchersQuery in ColonyMembers. Do you think it's better to handle this query as part of this task?

It's not about the the query, it's about the logic to display them, and display the counter. That should work regardless of what user data you plug into it (be it watchers or members). So it can very well be done now, using just watchers.

> For the user avatar hover styles, I was working on the refactoring of the Avatar component in https://github.com/JoinColony/colonyCDapp/issues/63 and I it looks like that work should handle the avatar hover styles as that cleans up the various Popover components.

Sure. If you want to handle it in #63 i don't have a problem, as long as it gets fixed in the end.

danbr:
I thought it would be useful to have a lambda function `colonyMembersWithReputation` that returns [Member] that have some reputation for a Colony. (without details about amount of reputation, nor domains)
I am adding this in #68 while trying to organise my thoughts for a more details func.

Julianahlk:
> I thought it would be useful to have a lambda function `colonyMembersWithReputation` that returns [Member] that have some reputation for a Colony. (without details about amount of reputation, nor domains) I am adding this in #68 while trying to organise my thoughts for a more details func.


For my first pass, I have things working using just the watchers as Raul suggested. I think its best if I created a separate task to go over this again using your `colonyMembersWithReputation` function.

src/components/common/ColonyHome/ColonyMembersWidget/MembersSubsection.tsx
```diff
@@ -119,28 +102,29 @@ Props) => {
...
-        )} */}
+        )}
       </div>
     ),
-    // [members, membersPageRoute, isContributorsSubsection, colonyName],
-    [isContributorsSubsection, membersPageRoute, colonyWatchers],
+    [isContributorsSubsection, membersPageRoute, colonyWatchers, name],
   );
 
   if (!colonyWatchers) {
```
jakubcolony, line 130: This hasn't been committed by you but I don't think this check works:
```
const colonyWatchers = useMemo(() => watchers?.items || [], [watchers]);
```
`colonyWatchers` can be either an array of watchers or an empty array, neither of those will ever make `!colonyWatchers` true. Please have a look into this

src/components/common/ColonyHome/ColonyMembersWidget/MembersSubsection.tsx
```diff
@@ -72,33 +63,25 @@ interface Props {
...
 }
 
-// const UserAvatar = HookedUserAvatar({ fetchUser: true });
-
-// const MAX_AVATARS = 12;
+const MAX_AVATARS = 12;
 
 const MembersSubsection = ({
   colony: { name, watchers },
   // members,
```
jakubcolony, line 70: Is that still needed?

jakubcolony (changes requested):
Nice work @Julianahlk!
image

All seems to work well, I left a few in-code comments

src/components/shared/Button/InviteLinkButton.tsx
```diff
@@ -0,0 +1,71 @@
+import React, { useState, useEffect, useCallback } from 'react';
```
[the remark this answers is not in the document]
Julianahlk: I have been thinking more about where to place components I think it may be better to organize components by their intended use instead of if they are already used in multiple places or not.

1. Group single purpose components by their relationship to other components. Example, many of the components in common/ColonyHome. For example: NewActionButton is good example as the intention is that there would only be 1 action button in the CDapp (although I feel it may make more sense if this button was in a subfolder of ColonyHomeLayout).
- Pros
    - Easy to see the hierarchy of components
    - Easy to determine where to put non-shared components
- Cons
    - The component hierarchy will be missing shared components
    - Lots of moving components around to different folders if they become shared

2. Group all components with the intention of being generalist/shareable components in a shared folder. Since these components are meant to be re-used they always have the potential of being shared across multiple components.
- Pros
    - A component’s location does not need to move after creation
    - This way of thinking is more likely to lead to components with a single responsibility and shareable from the start
    - Components that are designed to be shared instead of used for only one place in the CDapp should lead to a more consistent design and less work to update the design
- Cons
    - Could be more work to generalize components

Overall, I think we will have reasons to build both types of components but I think it will lead to more maintainability if we try to use shared components first before using specific purpose components. For the InviteLinkButton, I think it may have potential to be re-used in other places in the CDapp and it doesn’t necessary feel designed just for this one use.

src/components/common/ColonyHome/ColonyMembersWidget/MembersSubsection.tsx
Julianahlk: Leaving this in place as it will be used when the members data is eventually passed in from the ColonyMembersWidget

src/components/common/ColonyHome/ColonyMembersWidget/MembersSubsection.tsx
Julianahlk: good catch, will fix this

jakubcolony (approved):
LGTM, nice work!

danbr (approved):
LGTM. Well done.

Julianahlk:
@rdig Can you please take a look again as this PR is ready. Thanks

rdig (approved):
Yep, this seems to work as expected now.

Nice work.

Screenshot from 2022-11-10 20-54-49

PR #78, opened 2022-11-06
Clean up and refactor `UserAvatar` component

Julianahlk:
## Description

This PR takes care of the cleaning and refactoring of the `UserAvatar` component.

**New stuff** ✨

* Added new lambda function `getReputationForTopDomains `
* Added `TokenInfoPopoverWrapper `
* Created a new  `UserReputationItem` component that handles each item in the user reputation list

**Changes** 🏗

* Cleaned up `UserAvatar` component
* Combined `UserInfoPopover` and `MemberInfoPopover` into one component
* Simplified the `Tag` component

**Deletions** ⚰️

* `InfoPopOver`
* `MemberInfoPopover`
* `UserInfoPopover`

## Testing

1. Add reputation using truffle to a user and a domain to show reputation in the UserInfoPopover:
```c = await IColony.at('0x3a157280ca91bB49dAe3D1619C55Da7F9D4438c3')```
```c.emitDomainReputationReward(1, accounts[0], '1000000000000000000')```

2. Make sure the reputation monitor is ON:
`http://127.0.0.1:3001/reputation/monitor/toggle`
`http://127.0.0.1:3001/reputation/monitor/status`

Without Reputation
Screen Shot 2022-11-23 at 4 58 31 PM

With Reputation
Screen Shot 2022-11-23 at 4 55 10 PM

## TODO

- [x] Cleanup imports and remove what's no longer required
- [x] Remove all traces of IPFS / withDataFetcher logic
- [x] Remove the Hooked higher order function
- [x] Make sure styles are up to date with the latest ones from the Dapp
- [x] Introduce logic to account for thumbnails, as we now have that value in the db
- [x] Clean up, and wire the user tooltip
- [x] Fix the user avatar hover styles
- [x] Create task for the refactoring of getMainClasses in a separate [PR](https://github.com/JoinColony/colonyCDapp/issues/89)

Resolves #63

PR #79, opened 2022-11-08
Port over user profile

### LINKED ISSUE #33: Port Over User Profile
- Both the user Profile component
- As well as the user Profile advanced component
- Add new routes and queries to be able to load user profile through address as well: `/user/0x0.....0`
- Introduce URL input fields on the advanced tab (ask Product about them)
- Introduce centralized / decentralized switch on the advanced tab

chinins:
## Description

This PR ports over the User profile & settings.

Several things to note:

- user avatar/thumbnail will be stored in the DB so I changed the display in the Avatar (assuming the same will happen for the Colony avatar as well)
- figma specs are not exactly implemented (aligned with Arren) as the redesign is coming and it's not worth to invest time in this right now
- not cleaning up UserAvatar as Juliana is doing it

There are a couple of questions left for Raul, I'll update the PR as soon as I know more:

- right now the user is updated in a bit of a hacky way. Should I do query refetching or updating cache?
- loading state of the user profile settings (while loading it's showing redirection to the landing page) - is that ok?
- is the colony avatar stored in the DB - I've updated the colony avatar assuming this is the case
- there is an error about user avatar size (it's coming from  the  DB, usually when the image is above 500kb & below 1Mb). Do we need to update DB settings for the profile or is it just local dev environment?

## TODO

- [ ] Add TODOs

resolves #33


src/components/common/UserProfileEdit/UserAdvanceSettings.tsx
```diff
@@ -0,0 +1,243 @@
...
+              labelValues={{
+                isOn: values[SlotKey.Metatransactions],
+              }}
+              name={SlotKey.Metatransactions}
+              disabled={!metatransasctionsToggleAvailable}
+            />
+            <QuestionMarkTooltip
+              tooltipText={MSG.metaTooltip}
+              /* @ts-ignore */
+              tooltipTextValues={{ br: <br /> }}
```
jakubcolony, line 165: Could you look if this `/* @ts-ignore */` can be fixed? It seems like `QuestionMarkTooltip` defines this prop as `SimpleMessageValues` but looking at the comments on the different types:
image
And with the fact that `QuestionMarkTooltip` renders a `FormattedMessage` maybe this could be changed to `ComplexMessageValues` or `UniversalMessageValues` instead?

src/components/common/UserProfileEdit/UserAvatarUploader.tsx
```diff
@@ -0,0 +1,102 @@
...
+import { FileReaderFile } from '~shared/FileUpload';
+import AvatarUploader from '~shared/AvatarUploader';
+import UserAvatar from '~shared/UserAvatar';
+import { InputStatus } from '~shared/Fields';
+
+import { User } from '~types';
+import { updateUser } from '~gql';
+
+import styles from './UserAvatarUploader.css';
+import { useAppContext } from '~hooks';
```
jakubcolony, line 14: Should be higher up

src/components/common/UserProfileEdit/UserAvatarUploader.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+interface Props {
+  /** Current user */
+  user: User;
+}
+
+const UserAvatarUploader = ({ user, user: { walletAddress } }: Props) => {
+  const appContext = useAppContext();
+
+  const [updateAvatar, { error, called, loading }] = useMutation(
```
jakubcolony, line 37: Leaving a comment to change this to a custom hook once #90 is merged

src/components/common/UserProfileEdit/UserMainSettings.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+import { updateUser } from '~gql';
+import { User } from '~types';
+import { useAppContext } from '~hooks';
+
+import styles from './UserProfileEdit.css';
+
+const MSG = defineMessages({
+  heading: {
+    id: 'users.UserProfileEdit.UserMainSettings.heading',
```
jakubcolony, line 30: Should use the `${displayName}` notation

src/components/common/UserProfileEdit/UserAdvanceSettings.tsx
```diff
@@ -0,0 +1,243 @@
...
+  const [showSnackbar, setShowSnackbar] = useState<boolean>(false);
+  useEffect(() => {
+    if (showSnackbar) {
+      const timeout = setTimeout(() => setShowSnackbar(true), 200000);
+      return () => {
+        clearTimeout(timeout);
+      };
+    }
+    return undefined;
+  }, [showSnackbar]);
```
jakubcolony, line 100: I'm not sure about the purpose of this `useEffect`. The timeout for hiding the Snackbar seems to already be present inside that component. What is this doing?

src/components/common/UserProfileEdit/UserMainSettings.tsx
```diff
@@ -0,0 +1,214 @@
...
+  const [showSnackbar, setShowSnackbar] = useState<boolean>(false);
+  useEffect(() => {
+    if (showSnackbar) {
+      const timeout = setTimeout(() => setShowSnackbar(true), 200000);
+      return () => {
+        clearTimeout(timeout);
+      };
+    }
+    return undefined;
+  }, [showSnackbar]);
```
jakubcolony, line 106: Please have a look at `UserAdvanceSettings` and my comment regarding this `useEffect`. If it's actually needed, please consider moving it into its own hook

src/components/common/UserProfileEdit/UserProfileEdit.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { Tab, TabList, TabPanel, Tabs } from '~shared/Tabs';
+
+import ProfileTemplate from '~frame/ProfileTemplate';
+import LandingPage from '~frame/LandingPage';
+
+import UserProfileSpinner from '../UserProfile/UserProfileSpinner';
+import Sidebar from './Sidebar';
+import UserMainSettings from './UserMainSettings';
+import UserAdvanceSettings from './UserAdvanceSettings';
+import { useAppContext } from '~hooks';
```
jakubcolony, line 13: Should be higher up

src/components/shared/Button/Button.tsx
```diff
@@ -99,7 +99,10 @@ const Button = ({
   const titleText =
     typeof title == 'string' ? title : title && formatMessage(title);
   const buttonText =
-    typeof text == 'string' ? text : text && formatMessage(text, textValues);
+    typeof text == 'string'
+      ? text
+      : text && <FormattedMessage {...text} values={textValues} />;
+  //  formatMessage(text, textValues);
```
jakubcolony, line 105: Can this be removed?

src/components/shared/FileUpload/FileUpload.tsx
```diff
@@ -269,10 +268,9 @@ const FileUpload = ({
   );
 };
 
-const enhance = compose<
-  AsFieldArrayEnhancedProps<Props> & ForwardedRefProps,
-  Props
->(
+FileUpload.displayName = displayName;
+
+const enhance = compose(
```
jakubcolony, line 273: I'm wondering if the `asFieldArray` could be refactored into a HOC/hook which would allow us to go away from this `compose` pattern?

jakubcolony (commented):
Nice work @chinins! 

Based on your questions to Raul I guess there will be some further changes to this so only commenting for now.

Could the avatar uploading be improved? Currently it doesn't update after loading finishes and you have to refresh the page. And when you remove it, there's no loading at all and it looks like nothing's happening:

https://user-images.githubusercontent.com/112586815/203412183-679f252d-8602-4ca8-9ad0-51442e3892b9.mov

When you type something in the email field and then remove it, you are able to submit the form. I guess it's not a big issue since email isn't a required field but maybe we could add a check to see if there were any changes before enabling submit:
image

src/components/common/ColonyGrid/ColonyGrid.tsx
```diff
@@ -0,0 +1,81 @@
...
+import Link from '~shared/Link';
+
+import { CREATE_COLONY_ROUTE } from '~routes/index';
+import { WatchedColonies } from '~types/index';
+
+import ColonyGridItem from './ColonyGridItem';
+
+import styles from './ColonyGrid.css';
+
+const displayName = 'common.ColonyGrid';
```
rdig, line 18: I'd argue this should live under `shared` _(yeah, I know it lived under `dashboard` in the Dapp)_, but then again, it will be removed rather soon-ish once we refactor the user profile, so don't really care all that much if it stays here for now.

src/components/common/ColonyGrid/ColonyGrid.tsx
```diff
@@ -0,0 +1,81 @@
...
+    <div className={styles.main}>
+      <div className={styles.sectionTitle}>
+        <Heading text={MSG.title} appearance={{ size: 'medium' }} />
+      </div>
+      <div className={styles.colonyGrid}>
+        {colonies.map((colony) =>
+          colony ? (
+            <ColonyGridItem
+              colony={colony?.colony}
+              key={colony?.colony.colonyAddress}
```
rdig, line 71: Can you please destructure this, `colony.colony` is a really awful variable name

src/components/common/UserProfile/UserMeta.tsx
```diff
@@ -0,0 +1,92 @@
...
+      </div>
+      <div className={styles.headingContainer}>
+        {displayName && (
+          <Heading
+            appearance={{ margin: 'none', size: 'medium', theme: 'dark' }}
+            text={displayName}
+            data-test="userProfileName"
+          />
+        )}
+        {currentUser?.walletAddress === walletAddress && (
```
rdig, line 57: I'd still assign this to a variable just to make it more readable _(even though, I agree, it's a bit of waste if we use it just for one instance)_

src/components/common/UserProfile/UserProfile.tsx
```diff
@@ -0,0 +1,39 @@
...
+import UserMeta from './UserMeta';
+import UserProfileSpinner from './UserProfileSpinner';
+import UserColonies from './UserColonies';
+
+import styles from './UserProfile.css';
+
+const displayName = 'common.UserProfile';
+
+const UserProfile = () => {
+  const { username } = useParams();
```
rdig, line 17: See my comments about renaming this

src/components/common/UserProfile/UserProfile.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const { user, loading, error } = useUserByNameOrAddress(username || '');
+
+  if (loading) {
+    return <UserProfileSpinner />;
+  }
+
+  if (error || !user) {
+    return <NotFoundRoute />;
+  }
```
rdig, line 26: Until you address the rest of the changes, I'll test this some more locally, as to maybe find some solution to the "redirect to landing page" issue

src/components/common/UserProfileEdit/UserAdvanceSettings.tsx
```diff
@@ -0,0 +1,243 @@
...
+      validationSchema={validationSchema}
+      onSubmit={onSubmit}
+    >
+      {({ isValid, values, submitForm }) => (
+        <div className={styles.main}>
+          <Heading
+            appearance={{ theme: 'dark', size: 'medium' }}
+            text={MSG.heading}
+            textValues={{
+              // @ts-ignore
```
rdig, line 143: Why is this ignored ?

src/components/common/UserProfileEdit/UserAdvanceSettings.tsx
```diff
@@ -0,0 +1,243 @@
...
+              learnMoreLink: (
+                <ExternalLink
+                  text={{ id: 'text.learnMore' }}
+                  href=""
```
rdig, line 147: Shouldn't this have a link ?

src/components/common/UserProfileEdit/UserAdvanceSettings.tsx
```diff
@@ -0,0 +1,243 @@
...
+              {...MSG.customEndpoints}
+              values={{ isOn: values[SlotKey.DecentralizedMode] }}
+            />
+          </div>
+          <p className={stylesAdvance.descriptions}>
+            <FormattedMessage {...MSG.metaDescription} />
+          </p>
+          <div className={stylesAdvance.toggleContainer}>
+            <Toggle
+              label={MSG.labelGnosisRPC}
```
rdig, line 193: Just RPC, it's not Gnosis specific

src/components/common/UserProfileEdit/UserAdvanceSettings.tsx
```diff
@@ -0,0 +1,243 @@
...
+              labelValues={{ isOn: decentralizedModeEnabled }}
+              name={SlotKey.DecentralizedMode}
+            />
+            <QuestionMarkTooltip
+              tooltipText={MSG.gnosisRPCTooltip}
```
rdig, line 198: Same as above

src/components/common/UserProfileEdit/UserMainSettings.tsx
```diff
@@ -0,0 +1,214 @@
...
+const validationSchema = yup.object({
+  email: yup.string().email().nullable(),
+  bio: yup.string().nullable(),
+  displayName: yup.string().nullable(),
+  location: yup.string().nullable(),
+  website: yup.string().url().nullable(),
+});
+
+const UserMainSettings = ({ user }: Props) => {
+  const appContext = useAppContext();
```
rdig, line 94: Why not destructure `updateUser` directly ?

```suggestion
  const { updateUser } = useAppContext();
```

Also, please check with @willm30 's changes as to not have differences since both of you exported that function from the context independently of each other

src/components/common/UserProfileEdit/UserProfileEdit.tsx
```diff
@@ -0,0 +1,66 @@
...
+const UserProfileEdit = () => {
+  const { user, userLoading } = useAppContext();
+
+  if (userLoading) {
+    return <UserProfileSpinner />;
+  }
+
+  if (!user) {
+    return <LandingPage />;
+  }
```
rdig, line 37: Oh, this is where the landing redirect actually happens. Yeah. I'll look into this to see if we don't have a proper solution for it.

rdig: One version to fix this, of the top of my head, would be to have a 3rd value for the user:
- "notset" (just a string)
- the actual user returned object (if the user is found)
- null (if the user is not found)

But this might end up being messy code-wise, so I'll thin on this some more.

src/components/shared/Avatar/Avatar.tsx
```diff
@@ -34,15 +34,15 @@ const displayName = 'Avatar';
 
 const Avatar = ({
```
rdig, line 35: Please make sure your changes are more-or-less compatible with @Julianahlk 's changes in #78

src/components/shared/FileUpload/FileUpload.tsx
rdig: I'd like that very much :)

src/components/shared/InfoPopover/MemberInfoPopover/MemberInfoPopover.tsx
```diff
@@ -4,26 +4,28 @@ import { BigNumber } from 'ethers';
...
-  useUserBalanceWithLockQuery,
-  useUserReputationForTopDomainsQuery,
-} from '~data/index';
+// import {
+//   AnyUser,
+//   useColonyNativeTokenQuery,
+//   Colony,
+//   useUserBalanceWithLockQuery,
+//   useUserReputationForTopDomainsQuery,
+// } from '~data/index';
```
rdig, line 13: This can be removed entirely, as the `~data` import / functionality is not coming back

src/components/shared/InfoPopover/MemberInfoPopover/MemberInfoPopover.tsx
```diff
@@ -40,63 +42,63 @@ const MemberInfoPopover = ({
...
-    );
-  }
-  const nativeToken = userBalanceData?.user.userLock.nativeToken;
-  const userLock = userBalanceData?.user.userLock;
-  const inactiveBalance = BigNumber.from(userLock?.nativeToken?.balance || 0);
-  const lockedBalance = BigNumber.from(userLock?.totalObligation || 0);
-  const activeBalance = BigNumber.from(userLock?.activeTokens || 0);
-
-  const totalBalance = inactiveBalance.add(activeBalance).add(lockedBalance);
+  // const { data: userBalanceData, loading: loadingUserBalance } =
```
rdig, line 56: Again, please check with @Julianahlk so that this aligns (and maybe you don't run into conflicts) with the changes she's doing with the avatars / popovers

src/components/shared/UserMention/UserMention.tsx
```diff
@@ -29,51 +28,39 @@ interface Props {
 }
 
 const UserMention = ({
-  username,
+  user,
```
rdig, line 31: I guess this could be destructured as well, but it's more of a nice to have rather than an objective change.

src/context/AppContext.tsx
```diff
@@ -45,6 +48,7 @@ export const AppContextProvider = ({ children }: { children: ReactNode }) => {
         const query = apolloClient.query({
           query: gql(getCurrentUser),
           variables: { address },
+          fetchPolicy: 'network-only',
```
rdig, line 51: Can we change this to `cache-and-network` ?

I think it will have the same result that you're looking for, while also allow us to leverage some of Apollo's client side caching _(think instances where you log out / log in with the same wallet and user)_

rdig (changes requested):
A lot of work went into this, and it;s pretty good.

I've left a round of comments and changes that need to be done.

I'll also look into the "redirect to landing page" issue when loading the user data, so that we can have a proper solution for it prior to you merging this.

src/components/common/UserProfileEdit/UserProfileEdit.tsx
willm30: @rdig In my port User wizard branch, I suggested something similar. 

User if the user is found.
Null if the user isn't found. 
Undefined for the default value (as it is currently). Thus we can distinguish between user being null (not found) versus user being undefined (haven't completed fetching yet).

rdig: Have you already implemented that ? 

Anyway, I agree with your approach.

src/components/common/UserProfileEdit/UserAdvanceSettings.tsx
```diff
@@ -0,0 +1,243 @@
...
+              className={stylesAdvance.tooltipContainer}
+              tooltipClassName={stylesAdvance.tooltipContent}
+              tooltipPopperOptions={{
+                placement: 'right',
+              }}
+            />
+          </div>
+          <Input
+            name={SlotKey.CustomRPC}
```
willm30, line 207: Since this is an input specific to a url, we could specify formatting options to prevent the user from inputting whitespace. I copied a url from the browser and it came with a bunch of whitespace at the end, which resulted in the "invalid url" error showing, which was confusing given that I'd entered a valid url (but of course, was because of the trailing whitespace).

src/components/common/UserProfileEdit/UserAdvanceSettings.tsx
```diff
@@ -0,0 +1,243 @@
...
+            disabled={!values[SlotKey.DecentralizedMode]}
+          />
+          <div className={stylesAdvance.validateButtonContainer}>
+            <Button
+              text={MSG.validate}
```
willm30, line 212: Button functionality will be added in a future issue?

willm30 (changes requested):
Nice stuff! Adding my notes here.

If you upload a photo that is too large or of the wrong type, the "chose" button stops working -- i.e. doesn't reopen the os's file explorer (on windows 11, at least)

choosebroken

The profile photo doesn't persist when submitting the form:  
loseicon

Subscribed colonies list is showing on mobile, and the wallet overlaps with the "Profile Picture" heading:
mobile

Validation message should be customised (e.g. "Enter a valid URL") and save button shouldn't be enabled if there's an error in the input:   

saveenabled

jakubcolony:
Hi @chinins, my PR for GraphQL typings has now been merged and you can apply those changes to your PR. Let me know if you have any questions

src/components/common/UserProfileEdit/UserAdvanceSettings.tsx
```diff
@@ -0,0 +1,243 @@
...
+
+const validationSchema = yup.object({
+  [SlotKey.Metatransactions]: yup.bool(),
+  [SlotKey.DecentralizedMode]: yup.bool(),
+  [SlotKey.CustomRPC]: yup.string().url(),
+});
+
+const displayName = 'users.UserProfileEdit.UserAdvanceSettings';
+
+const UserAdvanceSettings = () => {
```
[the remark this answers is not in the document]
chinins: good catch!

src/context/AppContext.tsx
chinins: Not allowing me because of
`export declare type FetchPolicy = 'cache-first' | 'network-only' | 'cache-only' | 'no-cache' | 'standby';`
and 
`Invariant Violation: The cache-and-network fetchPolicy does not work with client.query, because client.query can only return a single result. Please use client.watchQuery to receive multiple results from the cache and the network, or consider using a different fetchPolicy, such as cache-first or network-only.`

src/components/shared/FileUpload/FileUpload.tsx
chinins: I think I have tried to modify it but there were some problems but don't recall exactly. I have opened another issue for it not to be a drag on this PR #106

src/utils/hoc/index.ts
```diff
@@ -21,3 +21,12 @@ export const withForwardingRef = <Props extends Record<string, any>>(
   forwardRef<ForwardedRefProps, Props>((props, ref) =>
     createElement(BaseComponent, { ...props, forwardedRef: ref }),
   );
+
+// eslint-disable-next-line @typescript-eslint/ban-types
+export const compose = (...funcs: Function[]) =>
```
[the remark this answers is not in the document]
chinins: To be explored in #106  as per my other comment

src/components/common/UserProfileEdit/UserAdvanceSettings.tsx
chinins: yeah, I was waiting for it.

src/components/common/UserProfileEdit/UserAdvanceSettings.tsx
chinins: indeed

src/components/common/UserProfileEdit/UserAdvanceSettings.tsx
chinins: Your are correct, that was a leftover from the dapp

chinins:
I've implemented all the changes or created separate issues to solve some of the comments.

@willm30, Raul said to leave in the unresponsiveness after you upload a file that's too large. It's been like this in the dapp and we are changing user profile interface anyway.

@jakubcolony, the app refreshes after I upload the avatar and reflects it properly for me (please, see below the gifs). Please, test again and if you still experience it not showing updated avatar after the upload/remove, let's have a call to debug it together as I can't reproduce.

 
remove-av
upload-av

src/components/common/UserProfile/UserMeta.tsx
```diff
@@ -0,0 +1,94 @@
...
+  user: User;
+}
+
+const UserMeta = ({ user: { walletAddress, profile }, user }: Props) => {
+  const { displayName, bio, website, location } = profile || {};
+  const { user: currentUser } = useAppContext();
+
+  const currentWalletAddress = currentUser?.walletAddress;
+
+  const isMobile = useMediaQuery({ query: query.query700 });
```
jakubcolony, line 39: You could change it to @Julianahlk's `useMobile` hook when you rebase: https://github.com/JoinColony/colonyCDapp/pull/100/files#diff-e08c126439ed0646aed221a78f7a2e69fffda573abfde3d671292e790dcb1248

jakubcolony (commented):
Accidentally submitted before I finished reviewing... Please see the comment below

src/graphql/mutations.ts
```diff
@@ -0,0 +1,722 @@
+/* tslint:disable */
```
jakubcolony, line 1: This file should be deleted

src/graphql/queries.ts
```diff
@@ -0,0 +1,542 @@
+/* tslint:disable */
```
jakubcolony, line 1: This file should be deleted

src/graphql/subscriptions.ts
```diff
@@ -0,0 +1,619 @@
+/* tslint:disable */
```
jakubcolony, line 1: This file should be deleted

src/graphql/custom-queries.ts
```diff
@@ -0,0 +1,173 @@
+/* tslint:disable */
```
jakubcolony, line 1: This file should be deleted

jakubcolony (approved):
Great work @chinins, I really like the improvements! 💪 

Thumbs up from me, there's a few minor things I noticed but nothing major

The gear icon seems slightly off the middle to me:
image


>@willm30, Raul said to leave in the unresponsiveness after you upload a file that's too large. It's been like this in the dapp and we are changing user profile interface anyway.

Could it be the same issue that I fixed in this PR?
https://github.com/JoinColony/colonyDapp/pull/3929

>@jakubcolony, the app refreshes after I upload the avatar and reflects it properly for me (please, see below the gifs). Please, test again and if you still experience it not showing updated avatar after the upload/remove, let's have a call to debug it together as I can't reproduce.

Looks fixed to me 👍 

>When you type something in the email field and then remove it, you are able to submit the form. I guess it's not a big issue since email isn't a required field but maybe we could add a check to see if there were any changes before enabling submit:
image

This still seems to be the case for me

Julianahlk (commented):
Looking good! 🙌 I would like to re-review this once you've rebased. 

@rdig, just curious about Olga's question regarding the avatar error: 
> There is an error about user avatar size (it's coming from the DB, usually when the image is above 500kb & below 1Mb). Do we need to update DB settings for the profile or is it just local dev environment?

Since I am still seeing that error when uploading png files above 500kb & below 1Mb:
uploadAvatar

willm30 (approved):
User profile image is now persisting after a form submission and I see that you've created a separate issue to handle the mobile styling. Good job.

src/context/AppContext.tsx
```diff
@@ -27,7 +27,7 @@ export const AppContext = createContext<AppContextValues>({});
 
 export const AppContextProvider = ({ children }: { children: ReactNode }) => {
   let initialWallet;
-  let initialUser;
+  const initialUser = null;
```
willm30, line 30: Initial user should not be initialised to null, else it will never be `undefined` and thus there's no benefit to introducing the `null` type to the `User` object. The logic I implemented for the User wizard is now on master so a rebase should resolve.

src/components/common/UserProfileEdit/UserProfileEdit.tsx
```diff
@@ -0,0 +1,67 @@
...
+  headingAdvance: {
+    id: `${displayName}.headingAdvance`,
+    defaultMessage: 'Advanced settings',
+  },
+});
+
+const UserProfileEdit = () => {
+  const { user, userLoading } = useAppContext();
+
+  if (userLoading || user === null) {
```
willm30, line 32: After rebasing,  `null` and `undefined` should be swapped so it loads when `undefined` and redirects if it returns without a user (`null`)

src/hooks/useUserByNameOrAddress.ts
```diff
@@ -0,0 +1,24 @@
+import { isAddress } from 'ethers/lib/utils';
+
+import { useCombinedUserQueryQuery } from '~gql';
```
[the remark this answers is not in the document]
chinins: no - good catch :)

src/components/common/UserProfile/UserProfile.tsx
```diff
@@ -0,0 +1,41 @@
...
+    usernameOrAddress || '',
+  );
+
+  if (loading) {
+    return <UserProfileSpinner />;
+  }
+
+  if (error || !user) {
+    return <NotFoundRoute />;
+  }
```
rdig, line 28: Does this need to implement the user `undefined` / `null` logic at all ?

src/components/common/UserProfileEdit/UserAvatarUploader.tsx
```diff
@@ -0,0 +1,107 @@
...
+  /** Current user */
+  user: User;
+}
+
+const UserAvatarUploader = ({
+  user,
+  user: { walletAddress, profile },
+}: Props) => {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const { __typename, ...userProfile } = profile || {};
```
rdig, line 38: hmm.... why  was there a need to remove `__typename` here ? It doesn't really make sense to me

src/components/common/UserProfileEdit/UserMainSettings.tsx
```diff
@@ -0,0 +1,225 @@
...
+});
+
+const UserMainSettings = ({
+  user: { walletAddress, profile },
+  user,
+}: Props) => {
+  const { updateUser } = useAppContext();
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const { __typename, ...userProfile } = profile || {};
```
rdig, line 105: Same question as as above

src/components/common/UserProfileEdit/UserProfileEdit.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+const UserProfileEdit = () => {
+  const { user, userLoading } = useAppContext();
+
+  if (userLoading || user === undefined) {
+    return <UserProfileSpinner />;
+  }
+
+  if (user === null) {
+    // return <LandingPage />;
```
rdig, line 39: Is this still needed ?

src/components/shared/Button/Button.tsx
```diff
@@ -100,7 +100,9 @@ const Button = ({
   const titleText =
     typeof title == 'string' ? title : title && formatMessage(title);
   const buttonText =
-    typeof text == 'string' ? text : text && formatMessage(text, textValues);
+    typeof text == 'string'
+      ? text
+      : text && <FormattedMessage {...text} values={textValues} />;
```
rdig, line 105: Why was this change needed ? Why was `formatMessage` not adequate here ?

rdig (approved):
This is mostly good, but this is just me pre-approving it as to indicate that.

Before merging this, please:
- fix your graphql user update error _(it's something you'll have to figure out yourself, as my availability is limited this week)_
- I left some questions in this review which need to be clarified

src/components/common/UserProfileEdit/UserAvatarUploader.tsx
chinins: it gives an error otherwise
Screenshot 2022-12-07 at 16 46 20

jakubcolony: Apollo will, by default, add `__typename` to queries as it needs it for caching and normalisation. And we will need to remove it, presumably not just here, but wherever we want to use the object from the response in a mutation. So I suggest we create an util function to do it for us so that we don't have to destructure and disable eslint every time.

chinins: great idea @jakubcolony - implemented

src/components/common/UserProfile/UserProfile.tsx
chinins: This one doesn't as it doesn't matter if we are logged in or not.

src/components/common/UserProfileEdit/UserProfileEdit.tsx
chinins: It was work in progress as I was trying to fix the loading. Left if as I wanted to discuss it with you. I've updated it all now to the state that I think it should be with the current limitations.

It's still not working perfectly. Either we have landing page showing up while the wallet is connecting or we don't have landing page shown at all if we the wallet is not connected and we just see the loading spinner for a long time and then the `Oh noes! We tried and tried but couldn't get any response from the service.` message.
I've opted for the first option.

Maybe I'm missing something - open to suggestions.

chinins:
@rdig, I've fixed everything but I just want you to check the `AppContext` changes & the comment I left separately (putting it again here). It's about loading of the edit-profile route:

It's still not working perfectly. Either we have landing page showing up while the wallet is connecting or we don't have landing page shown at all if the wallet is not connected and we just see the loading spinner for a long time and then the Oh noes! We tried and tried but couldn't get any response from the service. message.
I've opted for the first option.

Maybe I'm missing something - open to suggestions.

willm30:
@chinins I have written a hook that I believe solves the problem of navigating to `edit-profile` when a wallet can't be found in context.

I have tested with the following actions:

1. Navigate to `/edit-profile` with no wallet connected. --> redirects to `/landing`
2. Connect wallet to a wallet with a registered user, then navigate to `edit-profile`. --> displays Profile Edit screen
3. While on `edit-profile`, refresh the browser. --> shows loading spinner, then displays Profile Edit screen
4. Disconnect wallet while at `edit-profile` --> redirects to `/landing`
5. Navigate to `edit-profile` with a wallet that doesn't have a user associated. --> shows loading spinner, then redirects to `/landing` 

I believe I have covered all the scenarios, but would be grateful if you could confirm.

(The behaviour at 5. could be improved if we also kept `lastUser` in local storage, like `lastWallet`. In this case, if we had a `lastWallet`, but no `lastUser`, we could redirect instantly to `/landing` instead of running a query to check if there's a user associated with the wallet). 


**~hooks/useCanEditProfile**

``` tsx
import { useEffect } from 'react';
import { useNavigate } from 'react-router';

import { LANDING_PAGE_ROUTE } from '~routes';
import { getLastWallet } from '~utils/autoLogin';

import useAppContext from './useAppContext';

const useCanEditProfile = () => {
  const { user, userLoading } = useAppContext();

  // If there's a last wallet label, it means we're logged in.
  // If null, we're logged out and should be redirected to landing page.
  const lastWalletLabel = getLastWallet();

  useEffect(() => {
    // cannot edit profile if we're logged out or no there's no registered user in db.
    if (lastWalletLabel === null || user === null) {
      navigate(LANDING_PAGE_ROUTE);
    }
  }, [user]);

  return {
    // If lastWalletLabel isn't null (i.e. we're logged in), but the user is undefined, it means we're about to try to hydrate
    // the user into app context. Therefore, show loading.
    loadingProfile:
      userLoading || (user === undefined && lastWalletLabel !== null),
  };
};

export default useCanEditProfile;

```
And in the `UserProfileEdit` component:

```tsx

const UserProfileEdit = () => {
  const { loadingProfile } = useCanEditProfile();

  if (loadingProfile) {
    return <UserProfileSpinner />;
  }

  // By this point, if user is null or undefined, we'll be redirected to /landing.
 // (More just to assure typescript that user will be defined when the component renders)
  if (!user) {
    return;
  }

return (
    <ProfileTemplate
....
```

And in `AppContext`, ensure that `initialUser` is not set to `null` if a wallet cannot be found in context.

```tsx
 /*
   * Wallet
   */
  try {
    initialWallet = getContext(ContextModule.Wallet);
  } catch (error) {
    ~~initialUser = null;~~ <-- delete
    // silent
    // It means that it was not set in context yet
  }
```

src/context/AppContext.tsx
```diff
@@ -35,6 +35,7 @@ export const AppContextProvider = ({ children }: { children: ReactNode }) => {
   try {
     initialWallet = getContext(ContextModule.Wallet);
   } catch (error) {
+    initialUser = null;
```
willm30, line 38: This is the line I suggest needs removing if you go with `useCanEditProfile`

src/components/common/UserProfileEdit/UserAvatarUploader.tsx
```diff
@@ -0,0 +1,110 @@
...
+  },
+  avatarFileError: {
+    id: `${displayName}.avatarFileError`,
+    defaultMessage: 'This filetype is not allowed or file is too big',
+  },
+});
+
+interface Props {
+  /** Current user */
+  user: User;
```
jakubcolony, line 31: Could you not access it via the `appContext` in line 39?

src/components/common/UserProfileEdit/UserProfileEdit.tsx
```diff
@@ -0,0 +1,67 @@
...
+});
+
+const UserProfileEdit = () => {
+  const { user, userLoading } = useAppContext();
+
+  if (userLoading || user === undefined) {
+    return <UserProfileSpinner />;
+  }
+
+  if (user === null) {
```
jakubcolony, line 36: I think this should redirect user (by replacing the URL) rather than just render the landing page

jakubcolony (approved):
I'm wondering if we're not overcomplicating things unnecessarily. It feels like just two variables, `userLoading` and `user` should be enough to display the correct UI, just like with any other query:

```js
if(userLoading) {
  // loading
}

if(!user) {
  // there's definitely no user, since loading has now finished
}

// if we get here, loading has finished and there's a user
```

I got it to work in all cases, just by setting the default value for `userLoading` in `AppContext` to `true` instead:
```js
const [userLoading, setUserLoading] = useState(true);
```

src/graphql/mutations/user.graphql
```diff
@@ -1,5 +1,17 @@
...
+mutation UpdateUserProfile ($input: UpdateProfileInput!) {
+  updateProfile(input: $input) {
+    id
+    avatar
+    bio
+    displayName
+    location
+    website
+    email
+  }
```
Julianahlk, line 16: I noticed that the thumbnail is not getting updated when uploading a new profile picture. Should the thumbnail be added here to the UpdateUserProfile mutation?

Julianahlk (approved):
Great work Olga! ✨ Just left a comment below.

src/graphql/mutations/user.graphql
chinins: Shouldn't we have some functionality for creating thumbnails? As it's outside of this PR I left it out. It seems that it's not worth to just store the same data in thumbnail as in avatar...

src/components/common/UserProfileEdit/UserProfileEdit.tsx
chinins: I agree, but for the current implementation that wouldn't work. If when loading is fixed I will redirect.

chinins:
@jakubcolony, the userLoading set to true straight away was not working in one of the scenarios - I tried it and had to dismiss.

@jakubcolony  @willm30 , thank you for suggestions. I have created a separate issue #134 as I don't want to delay this PR any more. I will explore and test it more and will try to fix it.

PR #80, opened 2022-11-08
Introduce Colony Provider

### LINKED ISSUE #67: Introduce `ColonyProvider`
- will be a react context behind the scenes
- wrap around the `ColonyHome` _(or `ColonyLayout` if you need to pass in the colony identifier)_
- set up a hook for it _(`useColony` or something similar)_
- will handle colony data loading, typings, and data merging _(if you need to combine more data sources inside the colony object)_

jakubcolony:
## Description

This PR introduces a new `ColonyContext` and related components. The main purpose is to streamline colony fetching, handling loading and error states and provide better typing.

All the components making the `getFullColonyByName` query or getting the colony via props have been refactored to use the new context.
Additionally, the boolean returned by `useCanInteractWithColony` hook is also exposed via that context. 

~~@rdig a couple of questions before I mark this as ready:~~
 - ~~Instead of "encapsulating" the logic from `useCanInteractWithColony` hook I just called it from the context - is that fine or should I move the logic entirely?~~
 - ~~There's a bunch of components that get colony passed as a prop. I held on for now with refactoring them to use context as it could lead to us having to write more code checking if `colony` returned by the context isn't undefined - whereas when the colony is passed via props this check would have already been done by the parent component. What's your thoughts on that?~~
 - ~~For the `getFullColonyByName` query, should I change the fetch policy to `network-only`? Or is it okay for it to returned cached colony data? Alternatively we could use `cache-and-network`.~~

Resolves #67

### DIFF
--- a/src/components/common/ColonyHome/ColonyDomainSelector/CreateDomainButton.tsx
+++ b/src/components/common/ColonyHome/ColonyDomainSelector/CreateDomainButton.tsx
@@ -3,15 +3,12 @@ import { defineMessages, useIntl } from 'react-intl';
 
 import Icon from '~shared/Icon';
 // import { useDialog } from '~core/Dialog';
-
 // import CreateDomainDialog from '~dialogs/CreateDomainDialog';
-
-// import { Colony } from '~data/index';
-import { Colony } from '~types';
+import { useColonyContext } from '~hooks';
 
 import styles from './CreateDomainButton.css';
 
-const displayName = 'common.ColonuHome.ColonyDomainSelector.CreateDomainButton';
+const displayName = 'common.ColonyHome.ColonyDomainSelector.CreateDomainButton';
 
 const MSG = defineMessages({
   buttonCreateNewDomain: {
@@ -20,12 +17,11 @@ const MSG = defineMessages({
   },
 });
 
-interface Props {
-  colony: Colony;
-}
-
-const CreateDomainButton = ({ colony }: Props) => {
+const CreateDomainButton = () => {
   const { formatMessage } = useIntl();
+
+  const { colony } = useColonyContext();
+
   // const openCreateDomainDialog = useDialog(CreateDomainDialog);
 
   // const handleClick = useCallback<MouseEventHandler<HTMLButtonElement>>(
--- a/src/components/common/ColonyHome/ColonyExtensions/ColonyExtensions.tsx
+++ b/src/components/common/ColonyHome/ColonyExtensions/ColonyExtensions.tsx
@@ -2,34 +2,33 @@ import React from 'react';
 import { defineMessages, FormattedMessage } from 'react-intl';
 
 import Heading from '~core/Heading';
-import { Colony, useColonyExtensionsQuery } from '~data/index';
+import { useColonyExtensionsQuery } from '~data/index';
 import { MiniSpinnerLoader } from '~core/Preloaders';
 import extensionData from '~data/staticData/extensionData';
-
-import styles from './ColonyExtensions.css';
 import NavLink from '~core/NavLink';
 import ExtensionStatus from '~dashboard/Extensions/ExtensionStatus';
+import { useColonyContext } from '~hooks';
+
+import styles from './ColonyExtensions.css';
+
+const displayName = 'dashboard.ColonyHome.ColonyExtensions';
 
 const MSG = defineMessages({
   title: {
-    id: 'dashboard.ColonyHome.ColonyExtensions.title',
+    id: `${displayName}.title`,
     defaultMessage: 'Enabled extensions',
   },
   loadingData: {
-    id: 'dashboard.ColonyHome.ColonyMembers.loadingData',
+    id: `${displayName}.loadingData`,
     defaultMessage: 'Loading extensions data ...',
   },
 });
 
-interface Props {
-  colony: Colony;
-}
+const ColonyExtensions = () => {
+  const { colony } = useColonyContext();
 
-const displayName = 'dashboard.ColonyHome.ColonyExtensions';
-
-const ColonyExtensions = ({ colony: { colonyName, colonyAddress } }: Props) => {
   const { data, loading } = useColonyExtensionsQuery({
-    variables: { address: colonyAddress },
+    variables: { address: colony?.colonyAddress },
   });
 
   if (loading) {
@@ -60,7 +59,7 @@ const ColonyExtensions = ({ colony: { colonyName, colonyAddress } }: Props) => {
               <li key={address} className={styles.extension}>
                 <NavLink
                   className={styles.invisibleLink}
-                  to={`/colony/${colonyName}/extensions/${extensionId}`}
+                  to={`/colony/${colony?.name}/extensions/${extensionId}`}
                   text={extensionData[extensionId].name}
                 />
                 <ExtensionStatus
--- a/src/components/common/ColonyHome/ColonyFinishDeployment/ColonyFinishDeployment.tsx
+++ b/src/components/common/ColonyHome/ColonyFinishDeployment/ColonyFinishDeployment.tsx
@@ -4,11 +4,12 @@ import { defineMessages, FormattedMessage } from 'react-intl';
 import Alert from '~core/Alert';
 import { ActionButton } from '~core/Button';
 
-import { Colony, useNetworkContracts } from '~data/index';
+import { useNetworkContracts } from '~data/index';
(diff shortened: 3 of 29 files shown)


rdig:
@jakubcolony 
> Instead of "encapsulating" the logic from useCanInteractWithColony hook I just called it from the context - is that fine or should I move the logic entirely?

Fine

> There's a bunch of components that get colony passed as a prop. I held on for now with refactoring them to use context as it could lead to us having to write more code checking if colony returned by the context isn't undefined - whereas when the colony is passed via props this check would have already been done by the parent component. What's your thoughts on that?

Kinda of the point of this was to not have to pass the colony object around everywhere and just get it where it needs to be used.

I guess we'll just have to treat each component on a case by case basis then.

> For the getFullColonyByName query, should I change the fetch policy to network-only? Or is it okay for it to returned cached colony data? Alternatively we could use cache-and-network.

Cache and network is ok

jakubcolony:
>Kinda of the point of this was to not have to pass the colony object around everywhere and just get it where it needs to be used.

>I guess we'll just have to treat each component on a case by case basis then.

I have refactored all the Colony Home components to use the context rather than props

src/components/common/ColonyHome/ColonyDomainSelector/CreateDomainButton.tsx
```diff
@@ -3,15 +3,12 @@ import { defineMessages, useIntl } from 'react-intl';
...
-
 // import CreateDomainDialog from '~dialogs/CreateDomainDialog';
-
-// import { Colony } from '~data/index';
-import { Colony } from '~types';
+import { useColonyContext } from '~hooks';
 
 import styles from './CreateDomainButton.css';
 
-const displayName = 'common.ColonuHome.ColonyDomainSelector.CreateDomainButton';
```
rdig, line 14: Haha. Nice catch!

src/components/common/ColonyHome/ColonyExtensions/ColonyExtensions.tsx
```diff
@@ -2,34 +2,33 @@ import React from 'react';
...
-
-const displayName = 'dashboard.ColonyHome.ColonyExtensions';
+const ColonyExtensions = () => {
+  const { colony } = useColonyContext();
 
-const ColonyExtensions = ({ colony: { colonyName, colonyAddress } }: Props) => {
   const { data, loading } = useColonyExtensionsQuery({
-    variables: { address: colonyAddress },
+    variables: { address: colony?.colonyAddress },
   });
```
rdig, line 32: This won't exist anymore, as extensions will be fetched alongside the main colony object.

But you can leave it here for now. We'll refactor it when we get to it.

src/components/common/ColonyHome/ColonyFinishDeployment/ColonyFinishDeployment.tsx
```diff
@@ -18,31 +19,29 @@ import { getAllUserRoles } from '~modules/transformers';
...
-}: Props) => {
+const ColonyFinishDeployment = () => {
   const { version: networkVersion } = useNetworkContracts();
   const { wallet } = useAppContext();
   const userHasAccountRegistered = useUserAccountRegistered();
 
+  const { colony } = useColonyContext();
+  const { colonyAddress } = colony || {};
+  /** @TODO: This should be a property of colony */
+  const isDeploymentFinished = false;
```
rdig, line 43: It actually isn't since we won't use it anymore. 

(it will be removed along with the whole "check if deployment finished" logic)

src/components/common/ColonyHome/ColonyHomeLayout.tsx
```diff
@@ -48,30 +45,34 @@ const ColonyHomeLayout = ({
   showActions = true,
   onDomainChange = () => null,
 }: Props) => {
+  const { colony } = useColonyContext();
+
+  if (!colony) {
+    return null;
+  }
```
rdig, line 52: Agree with you that this pattern is a bit annoying. 

I like the cleanliness of using the context hook everywhere, it cuts down on imports a lot.

Now if there was a way to not require this check and we would be golden _(but no, there's isn't really a good way around it -- other  than force casting types, or returning default values, none of which are "good enough")_

src/components/common/ColonyHome/ColonyUpgrade/ColonyUpgrade.tsx
```diff
@@ -109,7 +106,8 @@ const ColonyUpgrade = ({ colony }: Props) => {
                       <ExternalLink
                         text={{ id: 'text.learnMore' }}
                         href={getNetworkRelaseLink(
-                          parseInt(colony.version, 10) + 1,
+                          // parseInt(colony.version, 10) + 1,
+                          10,
```
rdig, line 110: I think we have an oversight as in not tracking the version of the colony in the database.

I'll have to add it at some point. (probably when I refactor the ids to avoid forked chain collision)

src/context/ColonyContext.tsx
```diff
@@ -0,0 +1,77 @@
...
+    return null;
+  }
+
+  if (loading) {
+    return <LoadingTemplate loadingText={MSG.loadingText} />;
+  }
+
+  if (!colony || error) {
+    return <NotFoundRoute />;
+  }
```
rdig, line 66: Very nice putting these in the context.

src/routes/Routes.tsx
```diff
@@ -133,9 +136,11 @@ const Routes = () => {
...
           path={COLONY_HOME_ROUTE}
           element={
-            <Default routeProps={{ hasBackLink: false }}>
-              <ColonyHome />
-            </Default>
+            <ColonyContextProvider>
+              <Default routeProps={{ hasBackLink: false }}>
+                <ColonyHome />
+              </Default>
+            </ColonyContextProvider>
```
rdig, line 143: Why in here, and not inside `Entry`  nested under the `AppContextProvider` ?

That way it would wrap around, and be available to all routes

rdig (approved):
Quality work! I left some comments here and there, but overall very good! :taco:

src/routes/Routes.tsx
chinins: Agree with Raul, I think it would be better to have under the `AppContextProvider`, this way all the contexts will also be grouped together (on top of not repeating code around each route)

src/context/ColonyContext.tsx
```diff
@@ -0,0 +1,77 @@
...
+      name: colonyName ?? '',
+    },
+    fetchPolicy: 'cache-and-network',
+  });
+
+  const [colony] = data?.getColonyByName?.items || [];
+
+  const canInteractWithColony = useCanInteractWithColony(colony);
+
+  const colonyContext = useMemo<ColonyContextValue>(
```
chinins, line 51: a nitpick but I think useMemo is not really necessary here, you can just create an object when passing the value. Or am I missing something?

chinins (approved):
Very nice!

src/context/ColonyContext.tsx
rdig: Use of memo is correct here. Since this is passed as a prop to the Provider component, which itself is a top level component, if not memoized, will re-render like crazy. 

And not just the first child, but the whole tree, which in this case is more or less the whole app.

chinins: didn't think about it - good to know!

src/components/common/ColonyHome/ColonyHomeLayout.tsx
jakubcolony: Yes, I was facing exactly the same issue in my previous job :)

src/routes/Routes.tsx
jakubcolony: The only reason for putting it here is that there are some routes (not many but landing page, create colony, user pages) that don't need access to this context. 
I have refactored the context to return plain `children` if there's no `colonyName` in the path and will use it in `Entry` instead

jakubcolony: ^ I'm taking this back. In order to access the `colonyName` param in the path, `ColonyContextProvider` needs to be inside a route that specifies such param. 
@rdig Any thoughts?

rdig: I guess that's a good argument to have it in the route then, since we'll use route params quite a bit

danbr (approved):
Very good. Thats a lot less Colonies being passed around. well done.

PR #81, opened 2022-11-09
Remove colony deployment components

### LINKED ISSUE #57: Remove ColonyFinishDeployment
As discussed with @rdig, this component is no longer needed. Anything related to it should be removed - including message descriptors, alerts and sagas.

jakubcolony:
## Description

This PR removes all things related to colony deployment.

@rdig I noticed there are actions related to Token and OneTx deployment. Should those also be removed?

Resolves #57


rdig:
@jakubcolony 

> @rdig I noticed there are actions related to Token and OneTx deployment. Should those also be removed?

Probably not. Can you point me to the files so I can check ?

jakubcolony:
@rdig
`/src/redux/sagas/transactions/getMetatransactionPromise.ts` has some references of it as well as the `colonyCreate` saga

rdig:
> /src/redux/sagas/transactions/getMetatransactionPromise.ts has some references of it as well as the colonyCreate saga

No, don't remove those. Those are actual "real" deployments, the thing you're supposed to remove is just the one we've added in the dapp, but which doesn't have a correspondent on the contracts side.

This is a case of unfortunate naming.

So basically:
- colony finish deployment -- remove
- everything else "deployment" stays

_PS: if you're really curious, ask me next time we're in a call, and I'll explain how we ended up with this_

jakubcolony:
@rdig thank you, in that case I think this is ready for review.

danbr (approved):
LGTM

chinins (approved):
Some good quality deletion :)

Julianahlk (approved):
LGTM 👍

src/i18n/en.json
```diff
@@ -1,414 +1,406 @@
...
-    "label.optional": "Optional",
-    "label.email": "Email address",
-    "label.password": "Password",
-    "label.force": "Force",
-    "label.banned": "Banned",
-    "label.loading": "Loading",
-    "label.waiting": "Waiting",
-    "placeholder.email": "handle@domain.com",
-    "placeholder.password": "Password",
+  "home": "Home",
```
rdig, line 2: So I think the json were set at 4 spaces by prettier.

I don't mind the change to 2, but please make sure that the rule is enforced globally, otherwise, when we pull the updated file. our linters will (potentially) change it back

jakubcolony: Actually, it's a tricky one. There's an `.editorconfig` which specifies json files indentation as 4 spaces but there are a few files (`package.json`, `tsconfig.{dev,prod}.json`) that do not adhere to this which makes me think it's not effectively enforced.
Let's touch this tomorrow and I can address it separately. For now I have reverted the indentation changes I made.

PR #82, opened 2022-11-10
Clean up `ColonyAvatar`

### LINKED ISSUE #64: Cleanup and Refactor `ColonyAvatar` component
- cleanup imports and remove what's no longer required
- remove all traces of IPFS / `withDataFetcher` logic
- Remove the `Hooked` higher order function
- make sure styles are up to date with the latest ones from the Dapp
- introduce logic to account for thumbnails, as we now have that value in the db

jakubcolony:
## Description

It seems like most of the work described in the issue has already been done by Raul, I just cleaned up the typings and imports and applied a fix from [colonyDapp#3830](https://github.com/JoinColony/colonyDapp/pull/3830). That is unless I missed something 🙂

Resolves #64

### DIFF
--- a/src/components/shared/ColonyAvatar/ColonyAvatar.tsx
+++ b/src/components/shared/ColonyAvatar/ColonyAvatar.tsx
@@ -1,33 +1,21 @@
 import React from 'react';
 
 import { Address, Colony } from '~types';
-
-import Avatar from '~shared/Avatar';
+import Avatar, { Props as AvatarProps } from '~shared/Avatar';
 import NavLink from '~shared/NavLink';
 
-export interface Props {
+export interface Props
+  extends Pick<AvatarProps, 'avatarURL' | 'className' | 'notSet' | 'size'> {
   /** Address of the colony for identicon fallback */
   colonyAddress: Address;
-
-  /** Avatar image URL (can be a base64 encoded url string) */
-  avatarURL?: string;
-
-  /** Is passed through to Avatar */
-  className?: string;
-
-  /** Avatars that are not set have a different placeholder */
-  notSet?: boolean;
 
   /** If true the UserAvatar links to the user's profile */
   showLink?: boolean;
 
-  /** Avatar size (default is between `s` and `m`) */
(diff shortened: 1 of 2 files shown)


src/components/shared/ColonyAvatar/ColonyAvatar.tsx
```diff
@@ -1,33 +1,21 @@
...
   showLink?: boolean;
 
-  /** Avatar size (default is between `s` and `m`) */
-  size?: 'xxs' | 'xs' | 's' | 'm' | 'l' | 'xl';
-
   /** The corresponding user object if available */
   colony?: Colony;
 
-  prefferThumbnail?: boolean;
+  preferThumbnail?: boolean;
```
danbr, line 18: nice!

danbr (approved):
Looks good. Short & sweet.

Julianahlk (approved):
Looks very clean 🧼👍

chinins (approved):
Very good!

src/components/shared/ColonyAvatar/ColonyAvatar.tsx
```diff
@@ -36,21 +24,17 @@ const ColonyAvatar = ({
...
+  colony,
   notSet,
   size,
   showLink,
-  prefferThumbnail = true,
+  preferThumbnail = true,
 }: Props) => {
-  const imageString = prefferThumbnail ? profile?.thumbnail : profile?.avatar;
+  const { profile, name } = colony || {};
+  const imageString = preferThumbnail ? profile?.thumbnail : profile?.avatar;
```
rdig, line 34: I'm ok with this addition, especially since we're going to have to use this component in places where we'll want to showcase the full size image rather than the thumbnail _(eg: @chinins 's user profile)_

Just please make sure this pattern makes it's way to the other two components that deal with avatars: user and token.

Note that @Julianahlk is currently working on the user avatar part, so make sure to sync up on this, so that this gets into her changes as well.

rdig (approved):
Looks good. Nice work!

Left some comments, but the're more about disseminating information, rather than code changes.

Issue #49, opened 2022-10-31
Refactor / Cleanup Colony Watcher Subscriptions


rdig:
This issue will cover:
- porting over, cleaning up / adapting the `ThreeDotsButton` shared component
- cleaning up / ensuring styles are correct for the colony popover (the one that shows up when clicking the three dots button)
- writing a custom mutation to join a colony's watchlist (or using the existing one if it's suitable)
- writing a custom mutation to leave a colony's watchlist (or using the existing one if it's suitable)
- wiring up the above and cleaning the components it touches along the way

PR #83, opened 2022-11-11
Port Colony Creation Wizard

### LINKED ISSUE #32: Port Over Colony Creation Wizard
- Port over the components
- Should live in `components/common`
- Determine if the wizard design refresh (ask Product about it) can be integrated in this issue, or will it require a separate one

Note that this is blocked by https://github.com/JoinColony/colonyCDapp/issues/30

willm30:
## Description

This PR ports over the Colony Creation wizard from the Dapp.

**New stuff** ✨

* `~common/ColonyCreationWizard`: Colony Creation Wizard UI incl. mobile styles
* `~frame/WizardTemplateColony`: Wizard template UI incl. mobile styles

**Changes** 🏗

* Wired up `ColonyCreate` saga, incl. db mutations
* convert `withWizard` to functional component
* Exposed `updateUser` in AppContext, so `SubscribedColoniesList` can be uptodate as soon as we arrive to home from wizard
* Moved validation in `ColonyStepName` to its own file and out of `useEffects`
* Refactored `TokenSelector` to reduce props required and use data available inside component
* Updated old colour variables
* Added custom yup methods
* Aligned styles for `ExternalLink`, `DecisionOption/Hub` and `InputLabel` with Dapp

**Deletions** ⚰️

* I have not ported `StepUserName` because [my understanding](https://github.com/JoinColony/colonyCDapp/issues/32#issuecomment-1305425542) is that username creation is no longer being handled by the Colony Creation wizard. Any references to a username in the other steps take either the logged in user's display name or name. It is also wired such that you are unable to use the wizard if you're logged in with an account that doesn't have an associated user (you can test this by logging in with one of the wallets for which an account does not get created by the temp script, e.g. dev wallet 16), so we can assume any user using the wizard will have a name. 

* The revealing of the `token name` and `token symbol` inputs at `StepSelectToken` after failing to fetch a token address. This doesn't work correctly in the dapp (i.e. form won't validate), and in the event the user decides they want to create a token instead, they are able to click "I want to create a New Token"; ergo, unnecessary.

## TODO

- [x] port over UI components
- [x] wire up colony creation saga, incl. db mutations 
- [x] connect add existing ERC20 token
- [x] implement existing colony name check
- [x] ensure styles are same as in dapp, including mobile
- [x] Fix [error](https://github.com/JoinColony/colonyCDapp/blob/de212a5f69368e695c27d31c705c8c0333310fed/src/redux/sagas/colony/colonyCreate.ts#L300) inside createUniqueColony resolver ~~(I suggest this be tackled in a follow up issue.)~~

## Other notes

I'm presuming the revamp to the wizard flow will change the differences between the dapp and the cdapp version, e.g. the registering of the ENS address. So have not touched in this issue. 

I also named the files `ColonyCreationWizard` instead of `CreateColonyWizard` because the other files in `common` all began with `Colony` (i.e. I thought was a naming convention). But looking at the User wizard, I've realized it isn't. So I will change back, but will do after the pr is approved to avoid all the difs that will result. 

Resolves #32


src/components/common/ColonyCreationWizard/StepColonyName.tsx
```diff
@@ -0,0 +1,170 @@
...
+import { DEFAULT_NETWORK_INFO } from '~constants';
+import { useAppContext } from '~hooks';
+
+import { validationSchema } from './StepColonyNameValidation';
+import { FormValues } from './ColonyCreationWizard';
+
+import { query700 as query } from '~styles/queries';
+import styles from './StepColonyName.css';
+
+const displayName = 'common.CreateColonyWizard.StepColonyName';
```
Julianahlk, line 22: The folder name is ColonyCreationWizard so that would make the display name common.ColonyCreationWizard.StepColonyName

src/components/common/ColonyCreationWizard/StepColonyNameValidation.ts
```diff
@@ -0,0 +1,87 @@
...
+
+import { ContextModule, getContext } from '~context';
+import {
+  getColonyByName,
+  GetColonyByNameQuery,
+  GetColonyByNameQueryVariables,
+} from '~gql';
+import { intl } from '~utils/intl';
+
+const displayName = 'common.CreateColonyWizard.StepColonyNameValidation';
```
Julianahlk, line 13: common.ColonyCreationWizard.StepColonyNameValidation

src/components/common/ColonyCreationWizard/StepCreateToken.tsx
```diff
@@ -0,0 +1,172 @@
...
+import { multiLineTextEllipsis } from '~utils/strings';
+import { intl } from '~utils/intl';
+
+import { FormValues } from './ColonyCreationWizard';
+import { keyRequired } from './StepColonyNameValidation';
+import { switchTokenInputType } from './StepSelectToken';
+
+import styles from './StepCreateToken.css';
+
+const displayName = `common.CreateColonyWizard.StepCreateToken`;
```
Julianahlk, line 19: common.ColonyCreationWizard.StepCreateToken

src/components/common/ColonyCreationWizard/TokenSelector.tsx
```diff
@@ -0,0 +1,208 @@
...
+  getTokenByAddress as getTokenByAddressDocument,
+  GetTokenByAddressQuery,
+  GetTokenByAddressQueryVariables,
+  Token,
+} from '~gql';
+import { DEFAULT_NETWORK_INFO } from '~constants';
+
+import styles from './TokenSelector.css';
+
+const displayName = 'common.CreateColonyWizard.TokenSelector';
```
Julianahlk, line 21: common.ColonyCreationWizard.TokenSelector

src/components/common/ColonyCreationWizard/TokenSelector.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+const MSG = defineMessages({
+  inputLabel: {
+    id: `${displayName}.label`,
```
Julianahlk, line 25: id: `${displayName}. inputLabel `,

src/components/shared/DecisionHub/DecisionOption.tsx
```diff
@@ -5,18 +5,22 @@ import {
...
 import { getMainClasses } from '~utils/css';
+
 import Icon from '../Icon';
 import Link from '../Link';
 import { Tooltip } from '../Popover';
 import Heading from '../Heading';
+
 import styles from './DecisionOption.css';
 
+const displayName = 'DecisionOption';
```
Julianahlk, line 19: const displayName = 'DecisionHub.DecisionOption';

src/components/shared/Redirect/Redirect.ts
```diff
@@ -0,0 +1,19 @@
+import { useEffect } from 'react';
+import { useNavigate } from 'react-router-dom';
+
+interface Props {
+  to: string;
+}
+
+const displayName = 'shared.Redirect';
```
Julianahlk, line 8: const displayName = 'Redirect';

Julianahlk (changes requested):
Great work Will! I only left a few comments re displayNames and I noticed when selecting the option to use an existing token, 1. the continue button is already enabled without inputing a token address 2. there is no error message if the continue button is pressed.

test

jakubcolony (changes requested):
Nice work porting this over @willm30!

A couple of things I noticed:

- If you switch from using existing token to creating new token using the "I want to create a New Token" link and then you go back and choose the "Use an existing ERC20 token" option, it takes you to creating new token screen instead (it also works vice-versa). Please see the video:

https://user-images.githubusercontent.com/112586815/202541594-ba48bb48-e7c2-4678-9b20-f4e0b8b8dc03.mov
- If you switch between existing and new token options, and back to existing, the button isn't disabled despite the form not being valid:

https://user-images.githubusercontent.com/112586815/202540852-72e8b5ea-42b8-4cbd-b4b8-056b61df218e.mov


>Redirect: React Router v6 doesn't export a Redirect component like v5, so I've created one to do the same thing.

I'm not sure this is needed as react-router's `Redirect` was replaced by the `Navigate` component: https://reactrouter.com/en/main/components/navigate. One caveat is that the old `Redirect` replaces the current URL whereas `Navigate` pushes the new URL onto the history stack but there's a `replace` prop you can set to control that.

.eslintrc.js
```diff
@@ -38,8 +38,9 @@ module.exports = {
     },
   ],
   rules: {
-    // Using the typescript-eslint version for that
+    // Using the typescript-eslint version for these
     'no-unused-vars': 'off',
+    'no-undef': 'off',
```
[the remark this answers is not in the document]
willm30: I am going by the following advice:

> We strongly recommend that you do not use the no-undef lint rule on TypeScript projects. The checks it provides are already provided by TypeScript without the need for configuration - TypeScript just does this significantly better.

[Typescript-Eslint](https://github.com/typescript-eslint/typescript-eslint/blob/main/docs/linting/TROUBLESHOOTING.md#i-get-errors-from-the-no-undef-rule-about-global-variables-not-being-defined-even-though-there-are-no-typescript-errors)

src/components/common/ColonyCreationWizard/StepColonyName.tsx
willm30: I am going to change the folder name back, I made an error when changing the name from 'CreateColonyWizard'. I will do it after approval since it will generate diffs in virtually every file.

willm30:
@Julianahlk  @jakubcolony Thanks both, believe I have fixed. Please confirm that I haven't inadvertantly introduced anything else. My profuse clicking backwards, forwards and sideways suggests not but can never be sure.

willm30:
> I'm not sure this is needed as react-router's Redirect was replaced by the Navigate component: 

Sounds good to me, have swapped.

src/components/common/ColonyCreationWizard/StepColonyName.tsx
Julianahlk: ok sounds good 👍

Julianahlk (approved):
Everything is working as expected now. Nice work Will 😉.

src/components/common/ColonyCreationWizard/StepConfirmAllInput.tsx
```diff
@@ -0,0 +1,109 @@
+import React from 'react';
+import { defineMessages, FormattedMessage } from 'react-intl';
+
+import { WizardStepProps } from '~shared/Wizard';
+import Heading from '~shared/Heading';
+import Button from '~shared/Button';
+import { ActionForm, FormStatus } from '~shared/Fields';
```
rdig, line 7: Was Formik updated to the latest version ? Can you please check ?

src/components/common/ColonyCreationWizard/StepConfirmTransactions.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+  // Cancel the saga when the component unmounts
+  useEffect(
+    () => () => {
+      dispatch({ type: ActionTypes.CREATE_CANCEL });
+    },
+    [dispatch],
+  );
+
+  const normalizedColonyName = ensNormalize(chosenColonyName);
```
rdig, line 95: See above comments about ens

src/components/common/ColonyCreationWizard/StepTokenChoice.tsx
```diff
@@ -0,0 +1,179 @@
...
+import { Form } from '~shared/Fields';
+import { multiLineTextEllipsis } from '~utils/strings';
+import { SELECT_NATIVE_TOKEN_INFO } from '~constants';
+
+import { FormValues, Step2 } from './ColonyCreationWizard';
+
+import { query700 as query } from '~styles/queries';
+import styles from './StepTokenChoice.css';
+
+const LEARN_MORE_URL = SELECT_NATIVE_TOKEN_INFO;
```
rdig, line 18: Is this needed ? can't you just use `SELECT_NATIVE_TOKEN_INFO` directly ?

Or if you really care about naming, do something like:
```ts
import { SELECT_NATIVE_TOKEN_INFO as LEARN_MORE_URL } from '~constants';
```

src/components/common/ColonyCreationWizard/StepTokenChoice.tsx
```diff
@@ -0,0 +1,179 @@
...
+    defaultMessage: `Good for projects that already have their own token or want to use an existing one like DAI.`,
+  },
+});
+
+const options = [
+  {
+    value: 'create',
+    title: MSG.createTokenTitle,
+    subtitle: MSG.createTokenSubtitle,
+    icon: 'question-mark',
```
rdig, line 77: We really need to refactor `Icon`s to use `enum`s rather then string names.

Can you please open a maintenance  issue for that?

src/components/common/ColonyCreationWizard/TokenSelector.tsx
```diff
@@ -0,0 +1,212 @@
...
+        {...getStatusText({
+          addressFieldInput: tokenAddress,
+          isFetchingAddress,
+          hasError,
+          tokenData: { name, symbol },
+        })}
+        onChange={(e) => {
+          fetchToken(e);
+          // Timeout ensures pesky formik validates with correct state
+          setTimeout(() => validateField(addressField), 0);
```
rdig, line 199: Do we still need this hack ?

src/components/frame/SubscribedColoniesList/SubscribedColoniesList.tsx
```diff
@@ -21,9 +21,16 @@ const MSG = defineMessages({
...
+  const { user, userLoading, updateUser }: AppContext = useAppContext();
   const canInteractWithNetwork = useCanInteractWithNetwork();
 
+  /* Ensures colony list is up-to-date post create colony flow. */
+  useEffect(() => {
+    if (updateUser) {
+      updateUser(user?.walletAddress);
+    }
+  }, [user, updateUser]);
+
```
rdig, line 33: Why not call this from inside the saga at the end of creation flow ?

src/components/frame/WizardTemplateColony/WizardTemplateColony.tsx
```diff
@@ -0,0 +1,95 @@
...
+                  </div>
+                ) : (
+                  <div className={styles.yeihMoney}>
+                    <Numeral
+                      suffix={DEFAULT_NETWORK_TOKEN.symbol}
+                      value={ethBalance}
+                    />
+                  </div>
+                )}
+              </div>
```
rdig, line 82: Screenshot from 2022-11-21 22-19-17

Going forward we won't display the balance anymore, as the CDapp should not be in charge of the user's wallet. Metamask can handle that.

So please remove this part.

src/components/shared/Fields/InputLabel/InputLabel.tsx
```diff
@@ -68,9 +69,11 @@ const InputLabel = ({
       id={inputId ? `${inputId}-label` : undefined}
       htmlFor={inputId || undefined}
     >
-      <span className={styles.labelText}>{labelText}</span>
-      {helpText && <span className={styles.help}>{helpText}</span>}
-      {extra && <span className={styles.extra}>{extra}</span>}
+      <div>
+        <span className={styles.labelText}>{labelText}</span>
+        {helpText && <span className={styles.help}>{helpText}</span>}
+      </div>
```
rdig, line 75: Why was this wrapped inside a block ?

src/context/AppContext.tsx
```diff
@@ -37,14 +39,15 @@ export const AppContextProvider = ({ children }: { children: ReactNode }) => {
...
-  const updateUser = useCallback((address) => {
+  const updateUser = useCallback((address?: string) => {
     if (address) {
       try {
         setUserLoading(true);
         const apolloClient = getContext(ContextModule.ApolloClient);
         const query = apolloClient.query({
           query: gql(getCurrentUser),
           variables: { address },
+          fetchPolicy: 'network-only',
```
rdig, line 50: I'm ok with this here, since `updateUser` is conditioned when being called.

But going forward we need to avoid using `network-only` as much as possible, and to try to leverage Apollo's cache.

src/routes/Routes.tsx
```diff
@@ -44,7 +46,7 @@ import {
   // CLAIM_TOKEN_ROUTE,
 } from './routeConstants';
 import NotFoundRoute from './NotFoundRoute';
-import { ColonyContextProvider } from '~context/ColonyContext';
```
rdig, line 47: :+1:

rdig (approved):
Very high quality of work for the amount of changes you had to do. Kudos :100: 

I did however leave comments, some of which I want you to address prior to merging.

But at any rate, I'm pre-approving it.

Nice work!

chinins (approved):
Very nice work!

src/redux/sagas/transactions/transactionChannel.ts
```diff
@@ -33,17 +33,27 @@ const parseEventData = (
...
+  let parsedLogs: (LogDescription | null)[] = [];
+  if (receipt && receipt.logs) {
+    parsedLogs = receipt.logs.map((log) => {
+      try {
+        return client.interface.parseLog(log);
+      } catch {
+        /*
+         * `parseLog` will throw if the event the log refers to is not among the set of events
+         *  contained in the client interface. In this event, we just return null.
+         */
```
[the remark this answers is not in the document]
willm30: @rdig Yep, it's a change in `ethers`. Before it returned `null`, now it throws, so I'm swallowing it and returning null if it errors to simulate the old behaviour. I mentioned this to Chris and he suggested we investigate the source of the non-network client logs and parse with the correct ABI, hence my raising the issue in colonyJS.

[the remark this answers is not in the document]
rdig: That's not feasible. Any contract can emit events, some of which you really can't know about, so there's no "good way" of parsing all events in there.

Your approach is the best we have for right now. Just parse the events from the contracts you need, and filter out the rest.

src/components/common/ColonyCreationWizard/StepConfirmAllInput.tsx
willm30: @rdig Yep, latest is 2.2.9 which is our version in the Cdapp. 

As an aside, Formik is no longer being maintained (or at least, that's the impression I get from its last update being a year old, and there being over 600 outstanding issues in the repo. Plus Jared Palmer has moved on to turbopack, which suggests he's got other things to care about). 

I've noticed a few things that make formik a little annoying to work with, the most notable of which being that it doesn't reliably validate the latest form state. There are a number of issues in the formik repo that refer to this, some of which have been addressed, some are still open. [He essentially says it's a problem with hooks and can't be fixed](https://github.com/jaredpalmer/formik/issues/529#issuecomment-571294217). I've calling `validate` inside a `setTimeout` as a workaround, which does the job, but is a bit ugly. 

I understand that it's not currently high priority, but may be worth considering migrating to something else at some point like [react hook form](https://react-hook-form.com/), which has 30+k stars on github and is currently maintained (0 outstanding issues). Or using it to build new forms, and considering formik as legacy.

rdig: I'm happy to migrate, but we need to have a call an figure out what the exact scope of this undertaking will be.

Also, I really appreciate you took time to research this. :+1:

willm30: Awesome thanks. I have sent an invite for Thursday.

src/components/frame/SubscribedColoniesList/SubscribedColoniesList.tsx
willm30: @rdig I don't believe I can access React context outside of a component (AppContext in this case), so while I could get the latest user while in the saga, I'm not able to call `updateUser`.

The dapp pattern of updating the apollo cache at the end of the saga won't affect the values in context, which is what both the SubscribedColoniesList and ColonySubscription are using to render the colony avatars and display the `Join this colony` btn respectively. 

So I don't believe we have an alternative?

src/components/shared/Fields/InputLabel/InputLabel.tsx
willm30: @rdig This was a change I made when doing the mobile styles in the dapp. It's to be able to use `flex`'s `space-between` to align the label elements, which made correct alignment automatic without need for margins/padding. It ensures you get this:
existing Instead of this:
space-between

src/components/common/ColonyCreationWizard/StepTokenChoice.tsx
willm30: Sure, https://github.com/JoinColony/colonyCDapp/issues/92

src/components/common/ColonyCreationWizard/TokenSelector.tsx
willm30: I think so unfortunately. (referenced in comment above)

src/utils/intl.ts
```diff
@@ -0,0 +1,14 @@
+import { createIntl, createIntlCache } from '@formatjs/intl';
+
+// https://formatjs.io/docs/intl
+
+const cache = createIntlCache();
+
+// For use outside of React components
+export const intl = createIntl(
```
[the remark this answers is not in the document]
willm30: @rdig Just want to check you had something like this in mind? https://github.com/JoinColony/colonyCDapp/pull/83/commits/79ec4a0882be12fe19aa3d668dd6cfc692e66fb5#diff-151319ce8167c54c18d1d2f343ba2df500e65b24ee675ba59b57318f31fdd4fb

I.e a function you can pass messages to that will return you a new intl object each time it's called? Instead of adding the various sets of messages to the intl obj that I had previously created in `utils/intl`?

willm30:
@rdig Thanks! Have implemented. Will merge today, just left a few responses to confirm I've understood correctly.

src/components/frame/SubscribedColoniesList/SubscribedColoniesList.tsx
rdig: Bah. You're correct. I was thinking about the other context we have, but this is React's. Leave it like it is for now, although it might need to move somewhere inside the subscribe/unsubscribe component once @arrenv  gets around to #49 

For now leave it as it is

src/components/shared/Fields/InputLabel/InputLabel.tsx
rdig: Right. I wasn't "accusing" you of this being a wrong change, I was just curious as to why.

All good then.

rdig (approved):
Ok with the most recent changes. Feel free to merge!

PR #85, opened 2022-11-14
Make reputation monitor available on port 3001

area:
## Description

The reputation monitor wasn't available in the CDapp. This PR puts it on 3001. I think while setting up the Docker environments there was some confusion between the miner/oracle (which remains on 3002) and this monitor (which is now on 3001), and this seems to work from my (very) cursory testing without breaking access to the oracle, but it's possible there's some subtlety's I'm unaware of.


rdig (approved):
Good to go!

PR #87, opened 2022-11-15
Fix ganache address used for mining in local dev

area:
It's specifically the address at [index 5](https://github.com/JoinColony/colonyDapp/blob/d9211aabff928e2d0f9fc8908adad492112b6f37/scripts/start_all.js#L46) in the array that's set up for mining, so that's the address that should be used.

In the current state, the environment (and miner) comes up, but the address being used for mining has no CLNY staked, and so is unable to mine.

### DIFF
--- a/docker/files/network/run.sh.base
+++ b/docker/files/network/run.sh.base
@@ -16,7 +16,7 @@ DISABLE_DOCKER=true yarn truffle migrate
 # Export deployed addresses
 
 ETHER_ROUTER_ADDRESS=$(jq -r .etherRouterAddress etherrouter-address.json)
-MINER_ACCOUNT_ADDRESS=$(jq -r '.addresses | to_entries | .[16] | .value' ganache-accounts.json)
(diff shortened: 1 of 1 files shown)


docker/files/network/run.sh.base
```diff
@@ -16,7 +16,7 @@ DISABLE_DOCKER=true yarn truffle migrate
 # Export deployed addresses
 
 ETHER_ROUTER_ADDRESS=$(jq -r .etherRouterAddress etherrouter-address.json)
-MINER_ACCOUNT_ADDRESS=$(jq -r '.addresses | to_entries | .[16] | .value' ganache-accounts.json)
+MINER_ACCOUNT_ADDRESS=$(jq -r '.addresses | to_entries | .[5] | .value' ganache-accounts.json)
```
rdig, line 19: I moved this to the back of the array when selecting the address in order to not mess with the accounts  we use for testing.

It's ok for now, but I'll make sure to remove the 6th account from the available wallets in app, just to avoid confusion.

rdig (approved):
:+1:

PR #88, opened 2022-11-17
Port User Creation Wizard

### LINKED ISSUE #31: Port over User Creation Wizard
- Port over the components
- Should live in `components/common`

~~Note that this is blocked by #30~~

willm30:
## Description

This PR ports over the user creation wizard. It also adds a step for the user to input their email address.

To test, connect a wallet that does not have a username created by the temp script (i.e. not dev wallet 1-3). Then go through the create user flow via `/landing`. 

**New stuff** ✨

* `UserCreationWizard`: ported over and wired up to Network and DB. 
* `StepUserEmail`: gives user option to add an email and set permissions. This step is optional, as are the permissions, hence will validate if either is empty. Note that [design](https://www.figma.com/file/xGRNwVyii0qaCkWs4hASUn/Onboarding-2.0?node-id=757%3A13627&t=hewgB33GPlT9dsxd-0) is not pixel perfect. Awaiting design refresh. 
* `ConfusableWarning`: ported over and installed `@colony/unicode-confusables-noascii`

**Changes** 🏗

* Updated `User` model to include email permissions (**note**: you may need to re-run `npm run dev` to ensure your DynamoDB instance is correctly synced with the updated Graphql api)
* Removed ens domain extension string from username input
* Lightly refactored CreateColony validation to bring into line with CreateUser validation

**Deletions** ⚰️
* `createUserWithSecondAttempt`: I am presuming this is no longer necessary, but I don't have the context for why we had it. Something to do with IPFS?

Resolves #31


.eslintrc.js
```diff
@@ -102,7 +102,7 @@ module.exports = {
     'jsdoc/check-indentation': 'off',
     'no-restricted-exports': 'off',
     'no-use-before-define': 'off',
-    '@typescript-eslint/no-use-before-define': ['error'],
+    '@typescript-eslint/no-use-before-define': ['error', 'nofunc'],
```
willm30, line 105: Can we allow this? It means we can take advantage of function hositing, which is nice in, for example, `CreateUserWizardValidation.ts`, as all the helper functions can be defined at the bottom of the file, leaving the actual validation schemas at the top. 

It's probably only going to be relevant in files like this one, since we don't generally use `function` declarations (which are necessary inside yup tests to be able to access the helpers that yup provides on `this`. )

src/utils/yup/customMethods.ts
```diff
@@ -1,64 +1,44 @@
...
     name: 'hasHexPrefix',
     message: msg || en.string.hasHexPrefix,
-    test(value) {
+    test(value: string) {
       return value ? value.toLowerCase().startsWith('0x') : true;
     },
   });
 }
 
-addMethod(string, 'colonyName', colonyName);
```
willm30, line 61: This was changed from `ensAddress` in #83, becoming less general in the process. I've moved it to `StepColonyNameValidation.ts`, because I don't imagine it being reused.

src/components/common/CreateUserWizard/CreateUserWizardValidation.tsx
```diff
@@ -0,0 +1,100 @@
...
+import { intl } from '~utils/intl';
+import { runQuery } from '~utils/yup';
+
+/*
+ * The username regex is composed of
+ * ^[A-Za-z0-9] starts with upper case, lower case or numerals
+ * [A-Za-z0-9-_.] can include upper case, lower case, numerals, underscore, hyphen or period
+ * {0,99}$ match the preceding set at least 0 and at most 99 times from the end (so excluding the first char)
+ */
+const USERNAME_REGEX = '^[A-Za-z0-9][A-Za-z0-9-_.]{0,99}$';
```
willm30, line 20: In the dapp, the validation message for usernames is: "'Only characters a-z, 0-9, - and . are allowed'". I have followed this, but included underscores like in the colony name.

src/components/common/CreateUserWizard/StepUserEmail.tsx
```diff
@@ -0,0 +1,208 @@
...
+});
+
+enum EmailPermissions {
+  isHuman = 'isHuman',
+  sendNotifications = 'sendNotifications',
+}
+
+type Permission = `${EmailPermissions}`;
+
+class PermissionDescription {
```
willm30, line 63: Using a class reduces the boilerplate of creating object literals, especially given I can't use property value shorthand here. Probably not that noticeable here given there are only two items in the array, but still strikes me as "correct".

src/components/frame/LandingPage/LandingPage.tsx
```diff
@@ -51,19 +82,11 @@ const LandingPage = () => {
           />
         </div>
         <ul>
+          {wallet && user === null && (
```
willm30, line 85: I have added a `null` option to the AppContext's user state, i.e. it can now either be `undefined`, `null`, or `User` (as opposed to `undefined` or `User`). This gives us a way of distinguishing between the case where there is no user registered, and the case where we are still in the process of determining if there is a user registered (previously both would've been `undefined`). 

I was using `!userAccountRegistered`, but the `useUserAccountRegistered` hook returns a bool, and thus isn't flexible enough to indicate the distinction outlined above. Without it, the Landing page will always render the `Create a username` item, before rendering the `Create a colony` item if appropriate (which leads to ui flicker). 

This way it only loads the correct item, not one then the other straight after.

src/redux/sagas/utils/index.ts
```diff
@@ -1,7 +1,6 @@
 export { default as getColonyManager } from './getColonyManager';
 export { default as getNetworkClient } from './getNetworkClient';
 export { default as getGasPrices } from './getGasPrices';
-export { default as createUserWithSecondAttempt } from './createUserWithSecondAttempt';
```
willm30, line 4: I am presuming this is no longer necessary, but I don't have the context for why we have it. Something to do with IPFS?

src/utils/intl.ts
```diff
@@ -14,7 +19,14 @@ const cache = createIntlCache();
 export const intl = (messages: Record<string, string> = {}, locale = 'en') =>
   createIntl(
     {
-      messages,
+      messages: {
+        ...colonyMessages,
```
willm30, line 23: So that the intl object created outside a react component has access to all the same messages as one created inside a react component

Julianahlk (approved):
Excellent work Will 🙌 and thank you for leaving nice comments in your code to better understand what you are doing.
usercreation

src/components/common/CreateUserWizard/CreateUserWizardValidation.tsx
```diff
@@ -0,0 +1,100 @@
+import { string, object, array, ValidationError } from 'yup';
```
jakubcolony, line 1: Shouldn't this file name be in camelCase, considering it contains no component?

src/components/common/CreateUserWizard/StepConfirmTransaction.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import { useSelector } from 'react-redux';
+
+import { defineMessages } from 'react-intl';
```
jakubcolony, line 4: These two lines should be swapped

src/components/common/CreateUserWizard/StepUserEmail.tsx
jakubcolony: I have nothing against this pattern but the file has grown a bit large, maybe it's worth extracting the class to a separate file?

src/components/common/CreateUserWizard/CreateUserWizard.tsx
```diff
@@ -0,0 +1,80 @@
...
+  initialValues,
+  steps,
+})(WizardTemplate, { hideQR: true });
+
+interface UserStepTemplateProps {
+  heading: MessageDescriptor | string;
+  description: MessageDescriptor;
+  descriptionValues?: FormattedMessageProps['values'];
+  Input: JSX.Element;
+  Button: JSX.Element;
```
jakubcolony, line 36: I was trying to find examples of similar props and I think the general pattern is to use camelCase despite them accepting a JSX.Element. Please have a look at React Router's `Route` and its `element` prop, or, in the dapp, there's `ClaimFunds` and `VestingPageLayout` components that use camelCase for JSX.Element type props.

src/components/shared/ConfusableWarning/ConfusableWarning.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+const MSG = {
+  warningText: {
+    id: 'ConfusableWarning.warningText',
+    defaultMessage: `<span>Warning.</span> This username has confusable characters; it may be trying to impersonate another user.`,
+  },
+  warningCurrentUserText: {
+    id: 'ConfusableWarning.warningCurrentUserText',
+    defaultMessage: `<span>Warning.</span> Your username has confusable characters. This will show a warning when selected by users.`,
+  },
```
jakubcolony, line 22: These could use the `${displayName}` notation

src/components/frame/LandingPage/LandingPage.tsx
jakubcolony: Could you not do this check using `userLoading` from `AppContext`?
```
(!userLoading && !user)
```
I don't think your approach is wrong (and it still might make sense to allow `User` to be null), just wondering if this would be a bit clearer about your intention, which I understand as "if the loading is finished and there's still no user"

jakubcolony (approved):
Some really high quality work here @willm30!

All seems to be working well and the validation is robust:
Screenshot 2022-11-24 at 08 49 09
Screenshot 2022-11-24 at 08 49 26
Screenshot 2022-11-24 at 08 49 36

Please consider adding debouncing/throttling to the query checking for colony name uniqueness, currently it's being sent per every keystroke:

https://user-images.githubusercontent.com/112586815/203751877-3468fbbd-6cbd-4d2b-a873-7ca8ec8ba640.mov

src/components/common/CreateUserWizard/CreateUserWizardValidation.tsx
willm30: @jakubcolony Yes it should. Although it does look weird having a camel-cased file beginning with the component name sandwiched between the Pascal-cased components. How's this for a convention: naming validation files with `validation[ComponentName]`, e.g. `validationCreateUserWizard.ts`? That way validation files don't start with the component's name and will likely be grouped together near the bottom of the directory.

src/components/common/CreateUserWizard/CreateUserWizard.tsx
willm30: Yep happy with that

danbr (approved):
Looking good - Well done..... 👍

src/components/frame/LandingPage/LandingPage.tsx
willm30: @jakubcolony You're right. Thanks! I will revert the introduction of a null state given it's no longer relevent to this pr. It may be reintroduced if it's material to Olga's user settings pr or similar.

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -10,6 +10,8 @@ input CreateUniqueUserInput {
   id: ID!
   name: String!
   profile: ProfileInput
+  email: String!
+  emailPermissions: [String!]!
```
rdig, line 14: I'd also put this in some object field like `status` or `meta`

src/components/common/CreateUserWizard/CreateUserWizardValidation.tsx
rdig: I guess this should be in line with colonies and be up to 255 chars

src/components/common/CreateUserWizard/StepUserEmail.tsx
rdig: Agree with Jakub, should go in a separate file (inside the current Component's folder)

src/components/frame/LandingPage/LandingPage.tsx
rdig: Cool, so this is actually implemented, so that @chinins can pull in your changes directly

rdig (changes requested):
Mostly good with this, I'm just not happy with adding a second `email` filed on the `User` model. Please use the one inside `profile`. 

About this:

> createUserWithSecondAttempt: I am presuming this is no longer necessary, but I don't have the context for why we had it. Something to do with IPFS?

No. It was because we had a race condition that in very rare cases prevent the user from being created in the mongo database. This actually ran on login as well, and if it detected you had a user registered on chain, but not in the database, it would had attempted to re-create it.

But I agree, it can be removed safely.

src/components/frame/LandingPage/LandingPage.tsx
willm30: Ok, will leave the addition of null as is inside AppContext, but change check in `Landing` as Jakub suggests for readablility

src/components/common/CreateUserWizard/CreateUserWizardValidation.tsx
jakubcolony: Since the validation will most likely not be exported outside the component, I think it'd be fine to just call it `validation.ts`. We seem to follow this pattern with other component-specific files, such as `types.ts` in many components (Wizard, Popover, FileUpload to name a few) or `utils.ts` in SingleUserPicker. Obviously that would mean the validation files will be in between component files but I think that actually helps to create a clear structure of what depends on what
I'll leave that with you

willm30:
@rdig Profile is now a model with the email as an index. I have updated the create user and colony resolvers to reflect this change. I'm not getting any problems running the temp script nor creating users and colonies via the ui, so I think I've caught everything involved in the change. But pls confirm.

src/utils/yup/index.ts
```diff
@@ -9,3 +24,98 @@ export const validateSync =
...
+ * @param query Query to be executed
+ * @param isOptional Is field required? False by default
+ * @param circuitBreaker False by default. If true, will not run test. Can take a test function from a previous test. In this case,
+ * will not run if the function resolves to false. Previous tests should therefore assert positively, e.g. isUsernameValid. If it
+ * is not valid, do not run this test...
+ * @returns yup TestFunction, that runs the provided query using the input value
+ * as the query's input variable. Function will return true if there are no results (i.e. value doesn't exist in db);
+ * false if there are (value does exist in db); and an appropriate error message if the query errors.
+ */
+export function createYupTestFromQuery({
```
willm30, line 78: I've created this utility function to abstract a pattern I've used in both colony and user creation wizard. It's useful if you need to run a check against the db using the value of the input field, (like is this username/email already taken?).

willm30:
@jakubcolony Debouncing is currently a bit awkward because our input component takes the onChange method you pass to it and calls the Formik onChange handler before then calling the method you've passed. This behaviour invalidates any debouncing you may be attempting. It's solveable by modifying the input component so that if you want debouncing functionality, you manually call setFieldValue (which is what the Formik onChange handler does), however, since we're migrating from Formik, I don't think it's worth changing now only to re-change soon.

The other way to do it is to turn off validation on change and have it only on blur, but I don't like this in the username input because it feels like the ui isn't responsive. It works better in the email input, so implemented this solution there. 

It's also worth noting that apollo won't send duplicate requests, e.g. if  the user types something they've already typed, and that I've implemented it so that requests aren't sent if the input's invalid. These help cut down needless requests.

But for now, the username input will fire a request for every unique, valid input, something I agree can be improved with debouncing, and will look at when implementing the hooks-forms version of this form.

jakubcolony:
>@jakubcolony Debouncing is currently a bit awkward because our input component takes the onChange method you pass to it and calls the Formik onChange handler before then calling the method you've passed. This behaviour invalidates any debouncing you may be attempting. It's solveable by modifying the input component so that if you want debouncing functionality, you manually call setFieldValue (which is what the Formik onChange handler does), however, since we're migrating from Formik, I don't think it's worth changing now only to re-change soon.

>The other way to do it is to turn off validation on change and have it only on blur, but I don't like this in the username input because it feels like the ui isn't responsive. It works better in the email input, so implemented this solution there.

>It's also worth noting that apollo won't send duplicate requests, e.g. if the user types something they've already typed, and that I've implemented it so that requests aren't sent if the input's invalid. These help cut down needless requests.

>But for now, the username input will fire a request for every unique, valid input, something I agree can be improved with debouncing, and will look at when implementing the hooks-forms version of this form.

Sounds good to me, no point in spending time on it considering it'll be refactored before too long

jakubcolony:
Hi @willm30, my PR for GraphQL typings has now been merged and you can apply those changes to your PR. Let me know if you have any questions

willm30:
> Hi @willm30, my PR for GraphQL typings has now been merged and you can apply those changes to your PR. Let me know if you have any questions

Thanks, will add.

rdig (approved):
Yeah, this looks very good now. Nice work @willm30 :tada: 

_(just make sure to thoroughly check it after fixing all the merge conflicts)_

PR #94, opened 2022-11-23
Add icons enum

### LINKED ISSUE #92: Create enum for use with Icon component rather than string literals
Currently, when using the `Icon` component, you pass the name of the svg you want to use as a string literal to the `name` prop, like so:

iconeg

Going forward, we should use an `enum` to capture all of the options we currently have available and use that to reference the desired svg. Something like:

``` jsx
<Icon name={ICONS.caretRight} />
```

Per this comment: https://github.com/JoinColony/colonyCDapp/pull/83#discussion_r1028524688

chinins:
## Description

This PR adds an `Icons` enum to be used instead of passing strings directly into the `Icon` component. 

It also removes SVGs that are not being used in the dapp or that are not being ported.

Changes:
- pass an enum value to `Icon` instead of a string

Deletions:
- unused svgs

resolves #92


willm30:
Taking this one over but is not that high priority so will come back to it in between future prs

willm30:
Re Raul's comment in the issue: 

> We also need to investigate if the loader we're using to process SVGs is adequate going forward, as we haven't looked at that in ages

SVGO is being maintained and I haven't noticed any problems. Is there anything in particular you had in mind when thinking about whether it's adequate, @rdig?

src/images/icons.json
```diff
@@ -1,114 +1,66 @@
...
-    "Alessandra Aphid",
-    "Terry Termite",
-    "Ava Ant",
-    "Bong Joon Bee",
-    "Werner Wasp",
-    "Berthold Beetle",
-    "Sanjeet Shrimp",
-    "Norman Naked Mole Rat"
-  ]
+    "//Note": "Be sure to update the enum at ~constants/icons when adding SVGs to this file.",
```
willm30, line 2: No other way to comment in json but I think this is handy.

rdig:
Nah, if SVGO is still maintained we're still good. Part of the issue was to have a look into all the libs that we use in the SVG/Icons stack

jakubcolony (commented):
The code looks good, but I'm not sure if this is a straight-up improvement...
Yes, we now have autocompletion, but:
- We need to maintain 2 lists of icons
- We need to import a separate type wherever we want to use an icon

Can we not get rid of `icons.json` and store everything in one place? There are different ways of achieving this, but one of them could be:
```js
  const icons = ['arrow-right', 'arrow-rotate-ccw'] as const;
  type IconName = typeof icons[number]; // this creates a union: 'arrow-right' | 'arrow-rotate-ccw';
```

Now we can use `icons` array to load the required icons, and `IconName` to type the `Icon`'s component name prop. 

That way we only keep one list and we don't have to import anything to get autocompletion.

What do you think?

PR #95, opened 2022-11-23
Port extensions

### LINKED ISSUE #42: Port Over Extensions
- Port over components, refactor and update them. Should live in `components/common`
- Decide on the schema/table model (needs to be a relationship of Colony)
- decide on what historic data we need to keep, or if we want to (this is a larger discussion, as it gives us the ability to show motions from extensions that have been uninstalled -- and if we want to do that)
- refactor and re-introduce required routes
- add relevant queries and lambda functions
- decide if we refactor, keep or remove the static data file, as it's no longer that relevant

Note that this _should_ be split into multiple issues

jakubcolony:
## Description
This is a feature PR for porting extensions. It also contains the changes from #74 since it was based on the work Raul has done in that branch.

## Testing
Make sure to run the `temp-create-data` script. It will take care of setting the current known version of OneTxPayment / VotingReputation for you - otherwise the extensions won't show in the CDapp. This won't be necessary once block-ingestor#7 is ready.

Play around with both available extensions: install, uninstall, deprecate, re-enable, etc.
Please pay particular attention to these parts of the CDapp:
- Extensions widget in colony home:
image

- Extensions page:
image

- Extension details page:
image

- Extension setup page:
image

Resolves #42


Julianahlk (approved):
This is working well for me!  Good work Jakub 👍
Screen Shot 2022-12-06 at 5 07 10 PM

src/components/common/ColonyExtensions/ColonyExtensions.tsx
```diff
@@ -0,0 +1,184 @@
+import React, { useMemo } from 'react';
+import { defineMessages, FormattedMessage } from 'react-intl';
+import { getExtensionHash } from '@colony/colony-js';
+
+import {
+  InstallableExtensionData,
+  InstalledExtensionData,
```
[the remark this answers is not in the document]
jakubcolony: Good spot, fixed!

danbr (approved):
Looks good. OneTxPayment is installed and CDapp also displays. 
I presume that the UI stuff like Extension details will be added in another PR, functionality is the priority.

jakubcolony:
After talking to Raul, this branch will be the feature branch for extensions porting which all other PRs should be merged with. Switching it back to draft.

rdig (approved):
I'm adding my review here as I went with Jakub over the important bits here and agree with what was implemented, even though this branch has now become a feature branch.

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -291,3 +310,27 @@ type ContractEvent @model {
...
   encodedArguments: String # basically a JSON string
 }
+
+type ColonyExtension @model {
+  id: ID!
+  colonyId: ID!
+    @index(
+      name: "byColony"
+      sortKeyFields: ["hash"]
+      queryField: "getExtensionByColonyAndHash"
```
willm30, line 320: What does the `ByColonyAndHash` mean here, as opposed to just `getExtensionByColony`?

jakubcolony: A single colony can have multiple extensions (or, more specifically, extension installations). This is also setting up an index on the extension hash for fast retrieval of a particular extension installation in a colony

src/hooks/useExtensionsData.ts
```diff
@@ -0,0 +1,117 @@
...
+        return {
+          ...extensionConfig,
+          ...extension,
+          availableVersion: version ?? 1,
+        };
+      })
+      .filter(notNull);
+  }, [colonyExtensions, extensionVersions]);
+
+  const availableExtensionsData = useMemo<InstallableExtensionData[]>(() => {
```
willm30, line 78: Do we need to memoise these? This feels to me like it belongs in a named function instead.

amplify/backend/api/colonycdapp/schema.graphql
jakubcolony: >I ask because at the moment we've got this distinction between extension and extensions

I believe this is a very much standard approach for any model - there are separate get all and get single queries:
image

You could use one query and perform filtering on the client-side but I think the idea with the CDapp is to offload as much processing to the backend as possible. The benefits of having 2 separate queries are:
- Querying on the indexed fields (`byColonyAndHash`) is faster than filtering 
- When you're viewing an individual extension page, you can make a query to retrieve just the one extension instead of all of them
- Whenever dealing with `items`, Amplify will automatically introduce pagination and only return up to 100 items. Filtering happens after the pagination, so if the first item matching the filter is say 103rd in the database, it won't be returned by the query you proposed
While those might not be apparent at the start, I think they still outweigh the downsides.

jakubcolony: Do you think renaming `useExtensionData` to `useExtensionByIdData` would make the distinction clearer?

willm30: @jakubcolony 

> I believe this is a very much standard approach for any model

Sorry, I was referring primarily to the hooks of similar name. 

Currently, I only see one main requirement:

1. get all installed extensions at a given colony id

From this, we can derive non-installed (i.e. available) extensions, and we can also derive whether a specific extension  is installed or not, thus my sense we only need one query. Maybe I could put my thoughts into code as it would be easier for you to review?

> Querying on the indexed fields (byColonyAndHash) is faster than filtering

When I run this query `getExtensionByColonyAndHash` in the graphql explorer, it gets all extensions installed at the given colony id, and then optionally if I provided an extension hash filters them by that hash. What is the difference between that and the query I wrote above? I understand the need to index the colonyId so we can search extensions at a given colony, but in both cases it's filtering that retrieves a specific extension at a specific colony? The website also alludes to `sortKeyField` as filtering: 

> [The example client query below allows you to query for "Customer" based on their name and filter based on phoneNumber:](https://docs.amplify.aws/cli/graphql/data-modeling/#configure-a-secondary-index)

My thinking is that if they are functionally equivaent, we can use the same query for both getting all extensions and getting a specific extension and simplfy/combine these two hooks.

willm30: > Whenever dealing with items, Amplify will automatically introduce pagination and only return up to 100 items.

If I understand correctly per my comment above, we'd have the same problem with `getExtensionByColonyAndHash`?

jakubcolony: >Currently, I only see one main requirement:
>1. get all installed extensions at a given colony id

>From this, we can derive non-installed (i.e. available) extensions, and we can also derive whether a specific extension is installed or not, thus my sense we only need one query. Maybe I could put my thoughts into code as it would be easier for you to review?

I think it's fair to add a requirement of getting a single extension (with a specified hash) for a given colony. That way we can benefit of what I mentioned earlier:
>When you're viewing an individual extension page, you can make a query to retrieve just the one extension instead of all of them

I don't disagree this _could_ be done using a single query. Similarly to how we could call `listColonies` and filter them on the client for one with a given name instead of calling `getColonyByName`. But this is far from being optimal and it seems like all we're gaining is minimally simplified design.

From Amplify docs:
> Amazon DynamoDB is a key-value and document database that delivers single-digit millisecond performance at any scale but making it work for your access patterns requires a bit of forethought. DynamoDB query operations may use at most two attributes to efficiently query data. The first query argument passed to a query (the hash key) must use strict equality and the second attribute (the sort key) may use gt, ge, lt, le, eq, beginsWith, and between. DynamoDB can effectively implement a wide variety of access patterns that are powerful enough for the majority of applications.

Unfortunately they're quite useless but this suggests to me there is some underlying difference in how the data is stored and retrieved. 

This [SO thread](https://stackoverflow.com/questions/63226777/dynamodb-filter-and-sort-the-same-field) suggests to me the syntax of filtering and using sort key is the same (hence the understandable confusion), but the way filtering works is narrowing the entire partition of results whereas using sort key allows DynamoDB to select just what we're interested in. So it's not only faster, but we should not experience the same issue related to pagination if I understand this correctly.

Looks like there are a few ways to go about this (for example, we could use `getExtensionByColonyAndHash` without the `hash` parameter to retrieve all extensions in a given colony instead of getting them using Colony's `extensions` field) but I still believe there should be two queries - for all and a single extension, just like any other model and to ensure we're not querying more data than we need - it will at some point slow things down and cost more money.

Yes, we're probably talking distant future and/or small performance benefits but if the primary downside remains just slightly simpler design, I still think this is the right way to go

willm30: Thanks for pointing me to the SO article, it in turn led me to the dynamo db docs which help explain [primary keys](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.CoreComponents.html#HowItWorks.CoreComponents.PrimaryKey), and I think you're right about preferring a sort key to a filter. 

> Looks like there are a few ways to go about this (for example, we could use getExtensionByColonyAndHash without  the hash parameter to retrieve all extensions in a given colony instead of getting them using Colony's extensions field)

I agree, but happy to leave as is for now since it's fairly minor.

Thanks for taking the time to have this discussion and help me understand dynamo db better. All good from me 👍

scripts/temp-create-data.js
```diff
@@ -429,6 +450,9 @@ const createUserAndColonyData = async () => {
   await subscribeUserToColony(secondUser.address, thirdColonyAddress);
   await subscribeUserToColony(thirdUser.address, firstColonyAddress);
   await subscribeUserToColony(thirdUser.address, secondColonyAddress);
+
+  await setExtensionVersion('OneTxPayment', 3);
+  await setExtensionVersion('VotingReputation', 7);
```
jakubcolony, line 455: This will be removed when block-ingestor#7 is ready

rdig (approved):
I didn't do a full in-depth review of this as it was already reviewed via it's individual PRs, I just kept an eye out if anything sticks out.

From my point of view, this is good to be merged.

src/redux/sagas/extensions/extensionEnable.ts
```diff
@@ -0,0 +1,84 @@
...
+  try {
+    if (!isInitialized && initializationParams) {
+      const initParams = modifyParams(initializationParams, payload);
+
+      const {
+        channels,
+        transactionChannels,
+        transactionChannels: { initialise },
+        createGroupTransaction,
+      } = yield setupEnablingGroupTransactions(
```
willm30, line 38: Just a note about the `call` pattern often used in sagas (here e.g. would be `call(setupEnablingGroupTransactions, ...args)`. 

In the event we ever want to test the sagas, it's designed with that in mind. It doesn't affect the functionality, but may be worth sticking to it a) for consistency and b) should we ever wish to test them.

[Source](https://redux-saga.js.org/docs/basics/DeclarativeEffects/)

willm30 (approved):
Nice job! Working smoothly for me, all as expected. 

Just some pretty minor notes from clicking around in general (not a comment on the porting):

1. We should probably format the voting extn inputs to accept numbers only (i.e via the `formattingOptions` prop). Validation works fine but still looks a bit odd being able enter letters:

nums-only

2. When you click "Confirm" at the end of the voting rep configuration, the scroll to top feels a bit off. It's not smooth and also doesn't reach all the way to the top. Also from memory I don't think it works at all if there's an error, which would leave someone thinking it's just loading for a long time if it errored in production.

3. When you click re-enable or uninstall it disables the button, which results in this grey rectangle appearing (see right of modal below). I don't think that's necessary. You can't interact with those buttons anyway while the modal is open and I don't think the grey rectangle is doing the ui any favours.

disabled


4. The OneTxPayment deprecated label looks a little misaligned (a fraction indented)

deprecated

src/redux/sagas/extensions/extensionEnable.ts
jakubcolony: I'm not yet very familiar with redux-saga, but aren't we supposed to use `call` for functions returning Promises? 
If I understand correctly, it's still testable the way the docs describe (it yields, so the value can be accessed with `iterator.next().value`) but is a synchronous step.
I'll go ahead and merge this PR but if you can confirm your suggestion is still relevant for synchronous functions, I'll change it later

jakubcolony:
@willm30 Thanks for your review!

>When you click re-enable or uninstall it disables the button, which results in this grey rectangle appearing (see right of modal below). I don't think that's necessary. You can't interact with those buttons anyway while the modal is open and I don't think the grey rectangle is doing the ui any favours.

I think this is `ActionButton`'s default behaviour. It sets the `loading` prop on the button while action is in progress.****

src/redux/sagas/extensions/extensionEnable.ts
willm30: @jakubcolony Sorry I thought `setupEnablingGroupTransactions` was async since it's called using `yield`. If it's synchronous presumably it could be called directly. That said, an effect may still be useful for synchronous functions, in the sense that you don't need to test whether the function itself returns the correct data, only that it's being called correctly with the correct arguments. Unit testing of the function itself presumably wouldn't happen as part of the saga test. But I'm really just thinking out loud. Feel free to leave as is.

PR #97, opened 2022-11-23
Port over The UAC :tm:

### LINKED ISSUE #58: Port over action creation wizard
Port over the `ColonyHomeActions` component that is responsible for setting up the action creation wizard. 
This issue can be split into multiple issues as needed

ArmandoGraterol:
Ported over all of the Dialogs used in the UAC from the Dap + some misc dialogs that are not necessarily from the UAC.

The stuff that needed the most changes were the forms in the dialogs. I feel like the code of the UAC didn't really need a drastic rework or rewrite as it's a part of the Dapp that has been constantly worked and improved on, nevertheless, I checked for stuff that could be simplified or improved but perhaps I missed something as it's a lot of code, and there may be some stuff that is done differently here in the CDapp that I'm not aware of, so I would appreciate if you guys point places that could be improved further.

Most, if not all, of the stuff I commented out I believe is not implemented yet and it's dependant on other stuff like extensions or motions;  or was commented out to facilitate testing the dialogs.

FireShot Capture 776 - Colony CDapp - localhost
FireShot Capture 777 - Colony CDapp - localhost
FireShot Capture 778 - Colony CDapp - localhost
FireShot Capture 779 - Colony CDapp - localhost
FireShot Capture 780 - Colony CDapp - localhost
FireShot Capture 781 - Colony CDapp - localhost

Resolves #58

### DIFF
--- a/src/utils/lodash.ts
+++ b/src/utils/lodash.ts
@@ -1,3 +1,7 @@
 export { default as once } from 'lodash/once';
 export { default as now } from 'lodash/now';
 export { default as isNil } from 'lodash/isNil';
+export { default as sortBy } from 'lodash/sortBy';
+export { default as isEmpty } from 'lodash/isEmpty';
+export { default as toFinite } from 'lodash/toFinite';
+export { default as isEqual } from 'lodash/isEqual';


willm30:
Are you able to split this up into logical chunks, e.g. each item in the modal (create expenditure, manage funds etc.), and make a separate pr for each one? Don't know about everyone else but I'd prefer to review ten 1,000 diff prs than one 12,000 diff pr. It would also follow the same pattern here and in dapp of having a feature branch, which progressively incorporates smaller prs until the whole thing is complete.

PR #109, opened 2022-11-30
Migrate UserCreationWizard from Formik to React Hook Form

### LINKED ISSUE #103: Refactor User Creation Wizard to use React Hooks Forms
The user creation wizard is a small form with both form and action form components, so strikes me as a good guinea pig for the first attempt at migrating from Formik to React Hooks Forms. 

## Issue complete when

- [ ] User creation wizard uses React Hooks Forms instead of Formik
- [ ] All traces of Formik removed from the User creation flow

## Notes 
- [ ] Look at ease of adding debouncing to input fields (was tricky with formik)

willm30:
**Note: large diff due to `package-lock.json`** (~4,200 of the additions)

## Description

This PR migrates the User Creation Wizard from Formik to React Hook Forms. 

It also adds functionality to debounce yup tests. 

**To test**: Go through the user wizard flow. Should work as expected.

**New stuff** ✨

 I have created React Hook Form specific versions of:
 - Form
 - ActionForm
 - Checkbox 
 - Input
 - Input Component
 - Formatted Input Component (aka the Cleave logic. Split out into own component for readability).

I envisage these eventually replacing their Formik counterparts when migration completes.

**Changes** 🏗

* Have modified the UserCreationWizard to use React Hook Form api.
* Have added some custom debounce logic specifically designed to debounce yup `TestFunction`s. 

Resolves  #103

## Some notes on React Hook Forms

**All references below to components, (e.g. `Form`, `Checkbox`) refer to the Hook Form version introduced in this pr.**

I've abstracted some of the implentation details via the `Form` and `ActionForm` components, so use of Hook Form should feel similar to creating a Formik form.

Some things to bear in mind:

Currently, I'm using [useForm](https://react-hook-form.com/api/useform) and [useFormContext](https://react-hook-form.com/api/useformcontext). `useForm` hooks into your inputs and keeps track of their state. It does so via refs. All you need to do is provide the input a `name` corresponding to the value in the form (just like formik), and I've handled the rest inside the input component.

`useFormContext` gives you access to form helpers without having to prop drill. It's available at any level within the `Form` or `ActionForm` components. Helpers include getting/setting values, and accessing form state. 

When using the `Form`, you can provide an `options` object specifying things like how to resolve the validation schema and the validation strategy. I've implemented it so you can also just provide the validation schema directly via a prop and the `Form` component will assume it's a `yup` schema and resolve it as such. I've also provided direct access to `defaultValues` and `mode` (a.k.a validation strategy), since these seem like other frequently needed options. 

**Differences**

The biggest difference is that your components are not going to automatically re-render on change. This is because they are uncontrolled, i.e. are responible for their own state. Note that they will still validate, depending on your choice of validation strategy (default is same as we had with formik, i.e. onChange once the input has been touched). If you need to display something based on what the user is typing (e.g. a confusable warning), you need to define a function that accepts the change event and then update state yourself using `event.target.value`. See my `useConfusable` hook for an example. 

Another api difference is that Hook Form doesn't expose a `setFieldTouched` function, so I have created one for us. See line 125 in `ConfirmEmail.tsx` for usage.  

Another is the use of Field Arrays . The Hook Form equivalent, `useFieldArray`, is only needed for dynamic forms, i.e when the number of fields rendered depends on some user input (like in Safe Control, for example). For the Email Permissions Checkboxes in Step 1 of the User Creation Wizard, we don't need a field array helper. All you do is give both checkboxes the same name, set the form value to an array, and Hook Form will automatically add/remove them from the array. I've modified the `Checkbox` component accordingly (i.e. by removing the `asFieldArray` wrapper). It still expects the form value to be an array though, even if you only have one checkbox.


src/components/common/CreateUserWizard/CreateUserWizard.tsx
```diff
@@ -5,23 +5,15 @@ import { Props as FormattedMessageProps } from 'react-intl/src/components/messag
...
-import { EmailPermissions } from '~gql';
+import withWizard from '~shared/Wizard/withWizard';
 
+import { FormValues, initialValues } from './validation';
 import StepUserName from './StepUserName';
 import StepUserEmail from './StepUserEmail';
 
 import styles from './CreateUserWizard.css';
 
-export interface FormValues {
```
willm30, line 16: Tried to import these into `./validation`, but results were `undefined`. Unclear why. Importing from `./validation` works no prob.

src/components/common/CreateUserWizard/StepUserName.tsx
```diff
@@ -37,63 +41,65 @@ const MSG = defineMessages({
...
+              disabled={isSubmitting || isSubmitted}
+              isValid={isValid}
+            />
           }
           button={
             <ContinueWizard
-              disabled={!isValid || !dirty || isSubmitting}
-              loading={isSubmitting}
+              disabled={!isValid || !isDirty || isSubmitting || isSubmitted}
+              loading={isSubmitting || isSubmitted}
```
willm30, line 96: Using both `isSubmitting` and `isSubmitted` ensures the form shows its loading state while the saga executes in the background. `isSubmitting` is true while the form is being validated, and then `isSubmitted` is true once the saga has been initiated.

src/components/shared/Button/Button.css
```diff
@@ -80,6 +80,12 @@
   background-color: disabledBackground;
 }
 
+/* Focused-visible for key tabbing */
+
+.main:focus-visible {
+  border: 1px var(--colony-black) solid;
```
willm30, line 86: This ensures that there is some visual feedback when tabbing through inputs. Any suggestions on nicer alternative welcome, but I think some feedback is desirable.

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -18,7 +18,7 @@ enum EmailPermissions {
 }
 
 input ProfileMetadataInput {
-  emailPermissions: [EmailPermissions!]!
+  emailPermissions: [String!]!
```
willm30, line 21: With `[EmailPermissions!]!` I was intending a type of `("isHuman" | "sendNotifications")[]`, but instead it generates `EmailPermissions | Array<EmailPermisisons>` which is not the same thing. 

Any one know if you can get the intended effect in GraphQL? Or I just need to leave it as a general "String" array?

Also, anyone know why codegen turns a type of `[x]!` into `x | x[]` ? Don't see how this leads to the implication that there won't be an array.

src/components/shared/Fields/Checkbox/HookForm/Checkbox.css
```diff
@@ -0,0 +1,63 @@
+@value checkmark: "data:image/svg+xml;charset=utf8,%3Csvg xmlns='http://www.w3.org/2000/svg' version='1.1' viewBox='0 0 30 30'%3E%3Cg transform='translate(0,-350.38584)' fill='%2368D2DF'%3E%3Cpath d='M 26.558594,355.88672 11.042969,372.00195 3.4414062,364.10547 2,365.49219 11.042969,374.88477 28,357.27344 Z' /%3E%3C/g%3E%3C/svg%3E";
```
willm30, line 1: I think this is a better approach than the approach in the other Checkbox, which is drawing a checkbox using the border of a span element.

However, ideally the svg would live in the icons folder and be imported in here using `url(~images/icons/check-mark.svg)`. Webpack didn't like my attempts so I'm using this workaround for now. Anyone know how to appease webpack here?

src/components/shared/Fields/Form/HookForm.tsx
```diff
@@ -0,0 +1,51 @@
...
+}
+
+const HookForm = <FormData extends Record<string, any>>({
+  children,
+  defaultValues,
+  mode = 'onTouched',
+  onSubmit,
+  onError,
+  options,
+  validationSchema,
```
willm30, line 32: Mode (i.e. validation strategy), validation schema and default values actually belong in the options object passed to the `useForm`. However, seeing as they're likely some of the more used options, I've pull them out and added them as props for convenience.

src/components/shared/Fields/Form/HookForm.tsx
```diff
@@ -0,0 +1,51 @@
...
+  return (
+    <FormProvider {...formHelpers}>
+      <form onSubmit={handleSubmit(onSubmit, onError)}>
+        {typeof children === 'function' ? children(formHelpers) : children}
+      </form>
+    </FormProvider>
+  );
+};
+
+HookForm.displayName = displayName;
```
willm30, line 50: Modelled on the style of our Formik form, this wraps a form element in the React Hook Form context and passes in the helpers to a child function that returns JSX, render-props style.

src/components/shared/Fields/Input/HookForm/FormattedInputComponent.tsx
```diff
@@ -0,0 +1,119 @@
+import React, { useState } from 'react';
```
willm30, line 1: In the Formik version, this lives inside `inputComponent`, but it's very messy in there so I've split this out into it's own component for use in the Hook Form version.

src/utils/yup/index.ts
```diff
@@ -135,3 +154,78 @@ function createUnknownError(createError: TestContext['createError']) {
     message: formatMessage({ id: 'error.unknown' }),
   });
 }
+
+/**
+ * Debounce yup validation test function, inspired by lodash's debounce function.
```
willm30, line 159: Added a custom yup validation debounce function. Lodash's debounce doesn't work for this use case as explained below.

src/redux/sagas/users/index.ts
```diff
@@ -141,9 +138,6 @@ function* usernameCreate({
       },
       meta,
     });
-
-    /* Replace: true so you can't get back to the User wizard after completion */
-    yield routeRedirect(LANDING_PAGE_ROUTE, navigate, { replace: true });
```
willm30, line 146: Moved into `onSuccess` of ActionHookForm.

src/utils/yup/tests/index.ts
```diff
@@ -0,0 +1,219 @@
...
+    return !(result as object)[cleanQueryName(queryName)]?.items.length;
+  }
+}
+
+interface YupDebounceOptions {
+  isOptional: NonNullable<CustomTestConfig['isOptional']>;
+  circuitBreaker: NonNullable<CustomTestConfig['circuitBreaker']>;
+}
+/**
+ * Debounce yup validation test function, inspired by lodash's debounce function.
```
willm30, line 129: Added a custom yup validation debounce function. Lodash's debounce doesn't work for this use case as explained below.

src/components/common/CreateUserWizard/CreateUserWizard.tsx
rdig: That's weird, as the loader should be able to handle that. Anyway, your solution is good enought.

src/components/common/CreateUserWizard/StepUserName.tsx
```diff
@@ -36,62 +38,59 @@ const MSG = defineMessages({
...
-      appearance={{ theme: 'fat' }}
-      name="username"
-      label={MSG.label}
-      formattingOptions={{ blocks: [100] }}
-      disabled={disabled}
-    />
-    {username && isConfusing(username) && <ConfusableWarning />}
-  </>
+
+const formattingOptions = { blocks: [255] };
```
rdig, line 44: The change to 255 chars was so the user name is the same length as the colony name ?

src/components/shared/Fields/Checkbox/HookForm/Checkbox.css
rdig: Does the svg loader know to parse css files as well ?

src/components/shared/Fields/Input/HookForm/FormattedInputComponent.tsx
```diff
@@ -0,0 +1,116 @@
...
+const setCleaveRawValue = (
+  cleave: ReactInstanceWithCleave,
+  maxAmount: string,
+) => {
+  const decimalValue = new Decimal(maxAmount);
+  if (decimalValue.lt(0.00001) && decimalValue.gt(0)) {
+    cleave.setRawValue(maxAmount);
+  } else {
+    cleave.setRawValue(
+      new Decimal(maxAmount).toDP(5, Decimal.ROUND_DOWN).toString(),
```
rdig, line 38: Why is `5` digits hard coded here ? Shouldn't it be a prop with a default value ?

src/utils/yup/tests/index.ts
rdig: Thanks for leaving that detailed comment explaining the phenomenon

rdig (approved):
Wow. Some quality work right here. :100: 

Nothing major stuck out for me, just minor nitpicks

src/utils/yup/tests/index.ts
```diff
@@ -0,0 +1,241 @@
...
+
+    /*
+     * If we're still waiting, return a promise that will resolve after wait has elapsed.
+     * Its resolve value depends on how many timeouts are in the queue.
+     * If there's only one left, resolve with the result of the Yup test function.
+     * Else, resolve as true and remove one timeout from the queue.
+     */
+    const result = new Promise((resolve) => {
+      const id = setTimeout(() => {
+        if (timeouts.length > 1) {
```
jakubcolony, line 223: Could you explain in what case there will be more than one element in `timeouts`?

jakubcolony (approved):
Great work Will, thanks for taking the extra time to make things similar to the Formik setup - this will definitely make the migration go smoother!

Left a couple of questions but overall I'm positive about this change, especially that we've now got you as an expert 😉

danbr (approved):
Good job @willm30, very comprehensive. Looking forward to using it......

src/utils/intl.ts
```diff
@@ -31,3 +35,17 @@ export const intl = (messages: Record<string, string> = {}, locale = 'en') =>
     },
     cache,
   );
+
+const isMessageDescriptor = (message?: Message): message is MessageDescriptor =>
+  typeof message === 'object' &&
+  ('id' in message || 'description' in message || 'defaultMessage' in message);
```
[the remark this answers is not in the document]
chinins: can't agree more

chinins (approved):
Great work!

src/components/shared/Fields/Form/HookForm.tsx
willm30: Potentially if you have a large form and you don't want it to validate on change before it's been submitted for performance reasons. This is the hook form default ('onSubmit'). 

But I think generally it'll be 'onTouched' since this was the standard in the dapp.

src/components/common/CreateUserWizard/StepUserName.tsx
willm30: Yes, I think when porting the UserCreationWizard we said we'd make them consistent, but after updating the username validation regex I forgot to update the formatting options here.

src/components/shared/Fields/Checkbox/HookForm/Checkbox.tsx
```diff
@@ -0,0 +1,104 @@
...
+  onChange,
+  value,
+  tooltipText,
+  tooltipPopperOptions,
+  dataTest,
+}: Props) => {
+  const [inputId] = useState(nanoid());
+  const { getValues, register } = useFormContext();
+  const isInitiallyChecked = getValues(name).indexOf(value) >= 0;
+  const [isChecked, setIsChecked] = useState(isInitiallyChecked);
```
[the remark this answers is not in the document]
willm30: Good spot, think I missed this.

src/components/shared/Fields/Input/HookForm/Input.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+  const containerClasses = classnames(styles.container, {
+    [styles.containerHorizontal]: appearance.direction === 'horizontal',
+  });
+
+  const showConfusableWarning =
+    showConfusable && isConfusing && isValid && touched;
+  const handleInputChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    onChange?.(e);
+    setConfusableInput(e.target.value);
```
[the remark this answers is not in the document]
willm30: The component doesn't re-render after the first couple of keystrokes if the input is valid, so if you were trying to get the field's input value from `getValues`, you'd only get the value at the last render, not at the last onchange event. So we need to hold the state ourselves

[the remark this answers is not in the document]
jakubcolony: Could the [watch](https://react-hook-form.com/api/useform/watch/) method help here?

src/utils/yup/tests/index.ts
jakubcolony: Appreciate the detailed explanation! Can we not just keep track of the most recent call and discard any previous timeouts? This is the pattern I saw the most when doing debouncing, hence my question in the first place. 
Or if this is the best way to implement it, I'm perfectly happy with it. Just curious 🙂

willm30: If you clear previous timeouts, then the validation will not complete because its promises won't resolve. 

Say you clear the previous timeout on every keystroke in the delay period. You still need to return a promise. But now it never resolves because you cleared the timeout that resolves it. And you won't be able to submit the form because the validation will never complete. 

I think that's what you're alluding to? I tried it this way first so know where that road leads haha 😄

Issue #86, opened 2022-11-14
Update colony home members/watchers aside to correctly show both contributors and watchers


Julianahlk:
This requires a query to retrieve the members data for a colony. This should be ready to work on when the members page is ported.

PR #75, opened 2022-11-04
Port Members:Create `getMembersForColony' lambda function

danbr:
## Description

This PR adds a lambda function to retrieve members of a Colony.

Naming definitions:
- Member = User watching a Colony, with or without reputation
- Contributor = User watching a Colony WITH reputation
- Watcher = User watching a Colony WITHOUT reputation

The function is intended to be used for the Members page, but can also be used for #86.

Not included, but can be added in the future when support is included for:
- Permissions
- Banned
- Whitelist

Is there an efficient way to share an enum defined in (Amplify) schema.graphql with the CDapp, without defining several times?

Resolves #68

UPDATE:
Rebase with #104 had problems so need to update and resubmit.

The issue resolved by this PR has evolved and is now covered in #111.

PR #111, opened 2022-12-01
Port Members: Create `getMembersForColony' lambda function & support Members aside on Colony home

### LINKED ISSUE #68: Port Members:Create `getUsersReputationForColony' lambda function

danbr:
## Description

This PR now covers 2 issues.

1. 'getMembersForColony' lambda function (This was originally implemented in #75 - but work continued while working on Members aside)
2. Update colony home members/watchers aside to correctly show both contributors and watchers.

Screenshot 2022-12-02 at 18 24 06

Resolves #68 & #86


jakubcolony (commented):
@danbr I'm seeing this:
image

Together with errors from amplify container:
image

Am I doing something wrong?

(At least we know error handling is working 😉)

From what I was able to test, loading looks a bit ugly (I guess not a big deal since we've got a UI refresh coming): 
Screenshot 2022-12-06 at 15 55 39

rdig:
@jakubcolony this is not @danbr 's fault, it's because `colonyJS` doesn't support colony v11 yet

Apply this patch, restart your dev environement (the network docker image will re-build) and everything will start working again.
```diff
diff --git a/docker/colony-cdapp-dev-env-network b/docker/colony-cdapp-dev-env-network
index 3222ab2..2b5a2d2 100644
--- a/docker/colony-cdapp-dev-env-network
+++ b/docker/colony-cdapp-dev-env-network
@@ -1,7 +1,7 @@
-ARG NETWORK_HASH=develop
-
 FROM colony-cdapp-dev-env/base:latest
 
+ARG NETWORK_HASH=b4c34d3982fd3c1bf0e32a421cdfe8e24c665e7f
+
 # Declare volumes to set up metadata
 VOLUME [ "/colonyCDapp/amplify/mock-data" ]
 ```

This patch basically pins the colonyNetwork repo to a specific commit hash, right before they released colony v11

**Comment by danbr** [comment]
> From what I was able to test, loading looks a bit ugly (I guess not a big deal since we've got a UI refresh coming): Screenshot 2022-12-06 at 15 55 39

I didn't touch styling in this PR. I didn't want to spend time on it as the UI refresh is on its way......

**Comment by jakubcolony** [comment]
@rdig still getting the same error after changing the hash, I also tried the one from #74 (`e2194001e88dc129213a626a7d38943cbb0c9ef1`) but it didn't work either

File: amplify/backend/api/colonycdapp/schema.graphql [hunk 1]
```diff
@@ -130,28 +147,34 @@ enum ColonyType {
   METACOLONY
 }
 
+enum SortingMethod {
+  BY_HIGHEST_REP
+  BY_LOWEST_REP
```
chinins, line 152: I know that this comes from the dapp but now it seems that we have the camel case for enums so I would update

chinins (commented):
It's all good, actually working fine for me - but i've been also using this branch in my own PR.

The only thing is that there is a little bug. When you have reputation in subdomain and you query this subdomain, no contributors are returned.

amplify/backend/api/colonycdapp/schema.graphql
jakubcolony: We're using CamelCase for names but the members still seem to be all caps - have a look at `TokenType`

chinins: Well, then we are not consistent with the FE, because I checked there and it's camel case. I don't really mind either way but I think it would be great to have consistency across everything.

chinins (approved):
Working properly now.

willm30 (commented):
For me the patch works and the watchers / contributors show up in the correct section.

However, I don't think there should be a failure message when there are 0 watchers / contributors? I would expect something like `This colony has no contributors (/watchers)`. (I can't remember what we used in the dapp.)

failedto
failed

I'm also seeing the same popover for each different user (User BORIS).
allsame

danbr:
> For me the patch works and the watchers / contributors show up in the correct section.
> 
> However, I don't think there should be a failure message when there are 0 watchers / contributors? I would expect something like `This colony has no contributors (/watchers)`. (I can't remember what we used in the dapp.)
> 

Agreed. I've updated to show zero (0) or the fetch error.

Screenshot 2022-12-15 at 22 26 47


>I'm also seeing the same popover for each different user (User BORIS).

This was a bug with the wrong user being passed.
I've corrected, although in the screen captures the user name is not displayed. This is due to the User->Profile data being empty. UserInfoPopover.UserInfo is currently using displayName (in Profile). 

Screenshot 2022-12-15 at 22 23 46
Screenshot 2022-12-15 at 22 23 55


and if i force to use User.name
Screenshot 2022-12-15 at 22 11 07
Screenshot 2022-12-15 at 22 11 13

amplify/backend/function/getMembersForColony/src/graphql.js
```diff
@@ -0,0 +1,47 @@
...
+                createdAt
+                profile {
+                  avatar
+                  bio
+                  displayName
+                  email
+                  location
+                  thumbnail
+                  website
+                }
```
willm30, line 39: Query for `profileId` here as well to fix the problem with profile being `null` when consumed by the front end.

willm30 (approved):
Nice one. 

You can fix the problem with the profile showing up as `null` in the front end by adding `profileId` to your query in the back end. (I have left a comment)

userLiz

Pre-approving 👍

amplify/backend/function/getMembersForColony/src/graphql.js
danbr: @willm30 Great thanks, its working well now.

PR #119, opened 2022-12-06
Add `postcss-svg` plugin to handle svgs in CSS files

willm30:
## Description

This PR adds the `postcss-svg` [plugin](https://www.npmjs.com/package/postcss-svg), and configures it so that we can access svgs from our `images/icons` folder within our `.css` files. It also runs them through `svgo` for optimisation. 

Being able to use svgs directly from our `icons` folder is more convenient than manually running the svgs we want to use through a URL encoder each time so we can use them directly in the css file. (see the deletion in the `Checkbox.css` file for reference)

**New stuff** ✨

* `postcss-svg` installed
* Created a duplicate of the check-mark svg in the colony primary colour for use in the `Checkbox` component (see comment below)

**Changes** 🏗

* webpack & postcss configs
* updated `Checkbox.css` to use 'check-mark-primary.svg'

### Note
This is a follow-up pr to #109, as there I was looking for a way to import svgs into our css files. 

The slight annoyance is that the option to modify the svg's colour from within the css file doesn't appear to work. Not at least from my testing, although if you fancy trying for yourself, I link to the (sparse) [docs](https://www.npmjs.com/package/postcss-svg). Hence my duplicating the checkmark icon in a different colour.

I looked into using [postcss-inline-svg](https://www.npmjs.com/package/postcss-inline-svg) instead, however, it required postcss v8. I did install postcss v8 to see what would happen, but webpack emitted various warnings regarding incompatible postcss dependencies/plugins, and resolving these seemed more effort than it's worth.

If we end up with lots of the same svg only in different colours, perhaps it'd be worth revisiting this. But for now, will this do?
To be clear, this would only affect importing svgs into css files (which only occurs once currently in the HookForm Checkbox component); it does not affect our ability to style svgs imported via the `Icon` component (how we use all the rest).


jakubcolony (approved):
Cool stuff! 👍

danbr (approved):
LGTM 👍 - I'm sure we'll figure out the colour issue in the future.

Julianahlk (approved):
Nice addition! 👍

PR #121, opened 2022-12-07
Port extension sagas

### LINKED ISSUE #112: Port over Extension sagas
This issue is to port over all extensions related saga, including:
 - Installing extension
 - Enabling extension
 - Uninstalling extension
 - Deprecating extension
 - Upgrading extension

This may be split into multiple issues.

jakubcolony:
## Description

This PR ports extension sagas and related components from the Extension Details page.

### What is done
- Sagas for installing, uninstalling and deprecating extensions
- Buttons triggering the above in `ExtensionDetails` component
- Util saga for refreshing extensions (see notes below)

### What is left to do
- Enabling extensions: #126 
- Upgrading extensions: #127 
- Extending the checks for whether an extension can be installed, enabled, deprecated by checking if colony version is supported, user can interact with network, etc.

### Testing
1. Follow the general steps from #95.
1. Navigate to your Colony > Extensions and choose an extension you want to test.
2. Play around with installing, uninstalling, deprecating, re-enabling.

### Note on `refreshExtensions` saga
~~In the Dapp, this was effectively refetching the subgraph queries relating to extensions, therefore the results were almost instantaneously visible. Now, the transaction has to be picked up by block-ingestor and stored in Amplify which takes some time and creates a race condition that usually results in the updated extension status not being visible straightaway, without refreshing a page.
I will discuss this with Raul today but some ideas to mitigate this could be:~~
- ~~Polling until the extension is updated~~
- ~~Updating the Apollo cache manually to simulate the expected response~~

I opened a separate issue for that - #138 

Resolves #112


danbr (commented):
Looks good. Nothing sticks out code-wise.

I'm not sure what to expect when testing, i cannot install (via button), nor though truffle. I'm not sure if thats expected or something is amiss with my dev-env.

Screenshot 2022-12-13 at 20 41 44

willm30 (changes requested):
In clicking "Install", the `installExtension` saga doesn't complete. 

The UI sticks here with no error-related feedback:

stuck

(For context, I have started the dev env from this branch with the block ingestor patch applied and am running the block ingestor from your [`block-ingestor` pr](https://github.com/JoinColony/block-ingestor/pull/1) in another terminal window. Is that correct?) 

I looked into why the saga wasn't completing and:

TL:DR: You can fix the saga's getting stuck by adding `metatransactions` to the destructured properties inside the `coreTransactionsReducer` (`~redux/reducers/transactions.ts`, line 39), and to the object being passed to `Transaction` (`TransactionRecordProps` has it down as required so it should be there.)

More details: 
The `createTransaction` function sets `metatransaction: true`  by default on the actions it creates. However, it appears in the `transactionsAndMessageGroups` as `false`, which it shouldn't.

The consequence of this is that the `TRANSACTION_SEND` action doesn't get dispatched by the effect in `GasStationControls`, and so the `sendTransaction` function on line 76 of `createTransaction.ts` never gets called. Instead, `TRANSACTION_ESTIMATE_GAS` gets dispatched, but this isn't being listened for in `createTransaction.ts` because metatransactions are enabled. And so the saga doesn't resolve.

I believe this inconsistency in the `metatransactions` property is because the `coreTransactionsReducer`  does not destructure the `metatransactions` property on the action.

When I add `metatransactions` in the reducer, things start to flow as expected, but then hit the following error. This comes from the call to `getMetaTransactionPromise` on line 89 of `sendTransaction.ts`.

I will leave to you to debug further! 😅

errc

I'm also seeing the following react intl error:

intlerror

rdig:
@willm30 @jakubcolony note that metatransactions are not currently supported by `colonyJS`, so anything related to them is going to break, hence why Will had to jump trough hoops to get it working.

Julianahlk (approved):
Nice work Jakub! Overall this is working well for me with a few todos that you are already aware of but I noticed a few of them are without specific tasks created to handle them.

For the `ExtensionDetailsAside`, I noticed this is the first time a component has been called with “aside”, In all other cases we use the term “Widget”. I could understand this change if you are arguing for renaming all components that are on the “aside” for consistency. However, in the current state of the CDapp it feels like widget is more appropriate.

I also wanted to mention when I tried to Install an extension more than 2 times, the install button remains in a loading state. At this point, I am not sure if this is normal behavior or this is something worth to to investigate further. 
extensionstesting

willm30:
It seems odd that we've all had slightly different experiences. Perhaps @jakubcolony you could outline the correct steps to set up the env and then I can retry before you look too much into the errors I encountered?

jakubcolony:
@Julianahlk thanks for your review!

>For the ExtensionDetailsAside, I noticed this is the first time a component has been called with “aside”, In all other cases we use the term “Widget”. I could understand this change if you are arguing for renaming all components that are on the “aside” for consistency. However, in the current state of the CDapp it feels like widget is more appropriate.

Actually, it was me who introduced the "Widget" naming for `ColonyHomeLayout` aside so let me explain my intentions here. I renamed `ColonyFundingWidget` and `ColonyMembersWidget` to differentiate them from `ColonyFunding` and `ColonyMembers` respectively - they used to have the same name before which was a bit confusing.

The way I saw it, Widgets form a part of aside/sidebar rather than replace it entirely:
image

`ExtensionDetailsAside` is an aside/sidebar, rather than a widget, and while it could consist of some widgets, it feels like it's limited to just the single task of showing the extension details.

Hopefully that makes sense, let me know if you still think this could be improved.

>I also wanted to mention when I tried to Install an extension more than 2 times, the install button remains in a loading state. At this point, I am not sure if this is normal behavior or this is something worth investigating further.

Good spot, looks like the transaction popover isn't opening at all... Let me investigate it further

Julianahlk:
> @Julianahlk thanks for your review!
> 
> > For the ExtensionDetailsAside, I noticed this is the first time a component has been called with “aside”, In all other cases we use the term “Widget”. I could understand this change if you are arguing for renaming all components that are on the “aside” for consistency. However, in the current state of the CDapp it feels like widget is more appropriate.
> 
> Actually, it was me who introduced the "Widget" naming for `ColonyHomeLayout` aside so let me explain my intentions here. I renamed `ColonyFundingWidget` and `ColonyMembersWidget` to differentiate them from `ColonyFunding` and `ColonyMembers` respectively - they used to have the same name before which was a bit confusing.
> 
> The way I saw it, Widgets form a part of aside/sidebar rather than replace it entirely: image
> 
> `ExtensionDetailsAside` is an aside/sidebar, rather than a widget, and while it could consist of some widgets, it feels like it's limited to just the single task of showing the extension details.
> 
> Hopefully that makes sense, let me know if you still think this could be improved.
> 

Yes, it makes sense now. Thank you for explaining what your intentions were with the renaming. 👍

willm30 (approved):
Not sure what happened last time, must have had metatransactions on without realising. Anyway, install, deprecate and uninstall all playing nicely 🙂

I'm seeing the following warning from Apollo:
apollowarn

I suppose can be fixed when you come to a conclusion on how to refresh to show latest state?

jakubcolony:
Thanks @willm30, exactly, the warning is related to the refreshExtensions saga, for which I opened a separate issue - #138

danbr:
> Not sure what happened last time, must have had metatransactions on without realising. Anyway, install, deprecate and uninstall all playing nicely 🙂
> 
> I'm seeing the following warning from Apollo: apollowarn
> 

@jakubcolony Not sure if you updated anything, but after i deleted branches, cleaned the dev env, i'm finally seeing the CDapp as other colleagues.....
Same as @willm30's experience, 'install, deprecate and uninstall all playing nicely 🙂', and the Apollo warning. Well done. Approving.

danbr (approved):
Working well - 👍

PR #122, opened 2022-12-07
Migrate Colony Creation Wizard to use React Hook Forms

### LINKED ISSUE #116: Migrate Colony Creation Wizard to use React Hook Form
See #103 -> same idea but for colony creation wizard

willm30:
## Description

This PR migrates the Colony Creation wizard from Formik to React Hook Form. It should work the same as before, just now it's using the React Hook Form components.

**New stuff** ✨

* HookForm-specific `InputStatus`
* `utils/network`: some helpers to see what network speed the user has. May be useful to conditionally show a loading status based on this information. Faster connections may make a loading status redundant, e.g. if we're querying our own db, it's typically very fast and thus no need to display a loading status. The loading status also appears to flash in this case, because of the connection speed, which is also good to avoid.

**Changes** 🏗

* `ColonyCreationWizard`: now using React Hook Form components.
* Tidied up imports in `UserCreationWizard`

Resolves  #116

### DIFF
--- a/src/components/common/CreateColonyWizard/TokenSelector.tsx
+++ b/src/components/common/CreateColonyWizard/TokenSelector.tsx
@@ -1,13 +1,15 @@
-import React, { useEffect, ReactNode } from 'react';
-import { defineMessages, MessageDescriptor, useIntl } from 'react-intl';
-import { useField, useFormikContext } from 'formik';
+import React from 'react';
+import { defineMessages } from 'react-intl';
+import { useFormContext } from 'react-hook-form';
 
-import { Input } from '~shared/Fields';
-import { Appearance } from '~shared/Fields/Input/Input';
-
-import { usePrevious } from '~hooks';
+import {
+  HookFormInput as Input,
+  HookFormInputProps as InputProps,
+} from '~shared/Fields';
 import { isAddress } from '~utils/web3';
-import { Token, useGetTokenByAddressLazyQuery } from '~gql';
+import { formatText } from '~utils/intl';
+import { connectionIs4G } from '~utils/network';
+import { GetTokenByAddressQuery, Token, useGetTokenByAddressQuery } from '~gql';
 import { DEFAULT_NETWORK_INFO } from '~constants';
 
 import styles from './TokenSelector.css';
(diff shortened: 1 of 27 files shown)


src/components/common/CreateColonyWizard/TokenSelector.tsx
```diff
@@ -83,92 +60,54 @@ const getStatusText = ({
...
 
+const getTokenData = (tokenQuery?: GetTokenByAddressQuery) => {
+  const tokenData = tokenQuery?.getTokenByAddress?.items;
+  const { name, symbol } = tokenData?.[0] || {};
+  return { name, symbol };
+};
+
+const getLoadingState = (isLoading: boolean) =>
+  // Don't show loading state if connection is 4G or unknown
+  connectionIs4G() !== false ? false : isLoading;
```
willm30, line 81: The thought here is that on quick connections, the loading state is so brief that it looks better not to show it (else it flashes for a microsecond and then disappears).

src/components/common/CreateColonyWizard/TokenSelector.tsx
```diff
@@ -83,92 +60,54 @@ const getStatusText = ({
...
-  ] = useGetTokenByAddressLazyQuery();
-
-  const tokenData = tokenQuery?.getTokenByAddress?.items;
-  const { name, symbol } = tokenData?.[0] || {};
-  const prevTokenAddress = usePrevious(tokenAddress);
+    skip: !isValid,
+    onCompleted: handleComplete,
+  });
 
-  const fetchToken = (event: React.ChangeEvent<any>) => {
```
willm30, line 124: Moved to validation schema

src/components/shared/Fields/InputStatus/HookForm/InputStatus.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const Element = appearance.direction === 'horizontal' ? 'span' : 'p';
+  return (
+    <Element
+      className={getMainClasses(appearance, styles, {
+        error: !!error && !isLoading,
+        hidden:
+          (!text && !isLoading) || (!!error && !isNil(touched) && !touched),
+      })}
+    >
+      {isLoading ? loadingText : text}
```
willm30, line 52: Bringing the loading state into this component means we don't have to think about it higher up the component hierarchy, we can just pass in an `isLoading` prop.

rdig (approved):
Nothing sticks out and follows the same lines as you did for the user creation wizard refactor.

Nice work.

src/components/shared/Fields/Input/HookForm/InputComponent.tsx
```diff
@@ -22,6 +22,7 @@ const LengthWidget = ({ length, maxLength }: LengthWidgetProps) => (
 export interface HookFormInputComponentProps
   extends Omit<InputProps, 'placeholder'> {
   placeholder?: string;
+  inputValueLength: number;
 }
 
 const displayName = 'HookFormInputComponent';
```
Julianahlk, line 28: This display name could be fixed.

src/components/shared/Fields/InputStatus/HookForm/InputStatus.tsx
```diff
@@ -0,0 +1,59 @@
...
+    'appearance' | 'status' | 'statusValues' | 'isLoading' | 'loadingAnnotation'
+  > {
+  /** Error text (if applicable) */
+  error?: Message;
+
+  /** Has input field been touched? */
+  touched?: boolean;
+}
+
+const displayName = 'HookFormInputStatus';
```
Julianahlk, line 24: This display name could be fixed.

Julianahlk (approved):
All good, nice work! Just left comments about the display names.

chinins (approved):
Great job!

PR #130, opened 2022-12-09
Port enabling extensions saga

### LINKED ISSUE #126: Port over enabling extensions
Breaking #112 into smaller chunks to unblock people, this issue is to port enabling extensions, including `ExtensionActionButton`, sagas and any necessary changes to block-ingestor.

jakubcolony:
## Description
This PR ports over enabling extensions. At the moment it only focuses on the 2 extensions we initially support, OneTxPayment and VotingReputation.

In order to unblock other issues that depend on this, I restricted the scope of this PR to port just the saga (and related components) and will be working on the block-ingestor part under #154.

## Changes
- Extensions without initialization params (e.g. OneTxPayment) are considered initialized by default
- Enable extension saga
- `ExtensionSetup` component

## Testing
1. After installing OneTxPayment, it should be displayed as "Enabled" straightaway.
2. After installing VotingReputation, you should see "Enable" button that will redirect you to ExtensionSetup. You should see a similar form with config params and be able to submit it using the Confirm button at the bottom. This should trigger the `enableExtension` saga which should complete successfully. You will not see the status updating as this requires changes to block-ingestor.

Resolves #126


src/components/common/Extensions/ExtensionSetup/ExtensionSetup.tsx
```diff
@@ -0,0 +1,197 @@
...
+                />
+              </p>
+              {complementaryLabel && (
+                <span className={styles.complementaryLabel}>
+                  <FormattedMessage {...MSG[complementaryLabel]} />
+                </span>
+              )}
+            </div>
+          )}
+        </div>
```
Julianahlk, line 146: This looks like a good opportunity to extract into a separate component for the InitializationParam, what do you think?

Julianahlk (approved):
I tested it and works as expected, nice work! Just left a comment about a potentially extracting a component.

src/utils/extensions.ts
```diff
@@ -1,9 +1,42 @@
...
 /**
- * Type guard to distinguish installed extension data from extension config
+ * Type guard to distinguish installed extension data from installable extension data
  */
 export const isInstalledExtensionData = (
   extension: AnyExtensionData,
 ): extension is InstalledExtensionData =>
   (extension as InstalledExtensionData).address !== undefined;
+
+export const mapToInstallableExtensionData = (
```
willm30, line 17: For me, these functions make it less obvious what's happening in the hooks they're used in than if I were to see just the:

```
return [
          ...availableExtensions,
          ...extensionConfig,
          availableVersion: version,
        ];
```

Especially since it includes "map" in the title, which makes me think of a `.map` utility.

There is some code duplication within and between the `useExtension(s)Data` hooks, which may present a better opportunity to improve code readability. Or did you have something else in mind with these?

src/constants/extensions.ts
```diff
@@ -1,10 +1,10 @@
 import { ColonyRole, Extension } from '@colony/colony-js';
 import { defineMessages } from 'react-intl';
+import * as yup from 'yup';
+import toFinite from 'lodash/toFinite';
```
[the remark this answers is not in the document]
jakubcolony: I remembered seeing this in one of your PRs, I'm definitely up for making this consistent, I will add a TODO and come back to it when we rebase current master onto #95

src/components/common/Extensions/ExtensionSetup/ExtensionSetup.tsx
jakubcolony: Agree, thanks!

src/components/common/Extensions/ExtensionSetup/ExtensionSetup.tsx
```diff
@@ -0,0 +1,197 @@
...
+    extensionId,
+    initializationParams,
+  },
+}: Props) => {
+  const navigate = useNavigate();
+  const { colony } = useColonyContext();
+
+  const transform = pipe(
+    mapPayload((payload) => {
+      if (extensionId === Extension.VotingReputation) {
```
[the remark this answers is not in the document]
jakubcolony: Good shout, done

src/components/common/Extensions/ExtensionSetup/ExtensionSetup.tsx
```diff
@@ -0,0 +1,197 @@
...
+    >
+      {(formikProps) => {
+        const { isSubmitting, isValid, status } = formikProps;
+        return (
+          <div>
+            <Heading
+              appearance={{ size: 'medium', margin: 'none' }}
+              text={MSG.title}
+            />
+
```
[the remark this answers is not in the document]
jakubcolony: Personal habit, removed them

src/utils/extensions.ts
jakubcolony: I agree both of these hooks are similar yet there are distinctions in what they do:
1. Get data from different queries
2. Map the GraphQL data (`ColonyExtension`) to a custom CDapp entity (`Installed/InstallableExtensionData`) - I don't know a better word for it, "object mapping" sounds like it describes the process best - see https://automapper.org/ or even the `mapPayload` function a few comments above, it has nothing to do with `Array.map()` too. Open to suggestions though
3. In case of `useExtensionData`, return a single extension data, otherwise perform some filtering to split the data into installed and installable sets before returning

I think it's only step 2) that is common for both, so I extracted that into the mapping utils. Hopefully this makes it clearer to you but let me know if you think this can be improved further

willm30: Thanks for explaining, all good

willm30 (approved):

PR #140, opened 2022-12-16
Fix user wizard `index.ts` export order

willm30:
## Description

**Today I Learned** that the order of exports matters in `index.ts` files. This PR corrects the export order in the `UserCreationWizard` so imports are defined when the wizard loads.

**Changes** 🏗

* Move "main" export to bottom of `index.ts`
* I also update a tooltip in the ColonyCreationWizard which I meant to include in the previous commit

### DIFF
--- a/src/components/common/CreateColonyWizard/StepColonyNameInputs.tsx
+++ b/src/components/common/CreateColonyWizard/StepColonyNameInputs.tsx
@@ -28,7 +28,7 @@ const MSG = defineMessages({
   },
   tooltip: {
     id: `${displayName}.tooltip`,
-    defaultMessage: `We use ENS to create a .{displayENSDomain} subdomain for your colony. You can use this to create a custom URL and invite people to join your colony.`,
+    defaultMessage: `You can use this to create a custom URL and invite people to join your colony.`,
   },
 });
 
--- a/src/components/common/CreateUserWizard/index.ts
+++ b/src/components/common/CreateUserWizard/index.ts
@@ -1,4 +1,3 @@
-export { default, ContinueWizard, UserStepTemplate } from './CreateUserWizard';
 export { default as StepUserName } from './StepUserName';
 export { default as StepUserEmail } from './StepUserEmail';
 export {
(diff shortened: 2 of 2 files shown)


src/components/common/CreateColonyWizard/StepColonyNameInputs.tsx
```diff
@@ -28,7 +28,7 @@ const MSG = defineMessages({
   },
   tooltip: {
     id: `${displayName}.tooltip`,
-    defaultMessage: `We use ENS to create a .{displayENSDomain} subdomain for your colony. You can use this to create a custom URL and invite people to join your colony.`,
+    defaultMessage: `You can use this to create a custom URL and invite people to join your colony.`,
```
willm30, line 31: This I meant to add to the previous commit for the Colony Creation Wizard. We've removed the ens extension so this tooltip needs to be updated to reflect that.

src/components/common/CreateUserWizard/index.ts
```diff
@@ -10,3 +9,4 @@ export {
   UserWizardStep2,
 } from './validation';
 export { default as ConfirmEmail } from './ConfirmEmail';
+export { default, ContinueWizard, UserStepTemplate } from './CreateUserWizard';
```
willm30, line 12: Moving the "main" export, i.e. the component entry point (`CreateUserWizard`), to the bottom of the file ensures that all of its dependencies are defined (e.g. `StepUserEmail, StepUserName`), before it loads.

jakubcolony:
Hey Will, out of curiosity, what were the symptoms of the wrong order import?

willm30:
> Hey Will, out of curiosity, what were the symptoms of the wrong order import?

@jakubcolony The route threw an error before anything had loaded, which I believe is because Webpack loaded the main wizard component, which depended on imports beneath it in the index.ts (which hadn't yet themselves been loaded). So when trying to use these components, the wizard was telling me the steps weren't implemented. 

crash1

This helped me debug it: https://stackoverflow.com/questions/49116860/webpack-es6-modules-circular-dependencies-when-using-index-files

Would you mind confirming that this also happens on your machine if you move the main export back to the top of the file?

jakubcolony (approved):
Thanks for the explanation, yep, that fixed it for me 👍

rdig (approved):
Very nice catch, and fix.

PR #143, opened 2022-12-16
Migrate & Cleanup Colony Subscriptions

### LINKED ISSUE #49: Refactor / Cleanup Colony Watcher Subscriptions

arrenv:
## Description

This issue includes the migration of users being able to subscribe "Join" a Colony. Including the components used along the way. This includes the `ThreeDotsButton` shared component.

**Changes** 🏗

* Changes go here

## TODO

- [ ] - porting over, cleaning up / adapting the ThreeDotsButton shared component
- [ ] - cleaning up / ensuring styles are correct for the colony popover (the one that shows up when clicking the three dots button)
- [ ] - writing a custom mutation to join a colony's watchlist
- [ ] - writing a custom mutation to leave a colony's watchlist
- [ ] - wiring up the above and cleaning the components it touches along the way

Resolves #49


src/hooks/useCanInteractWithColony.ts
```diff
@@ -48,10 +48,24 @@ export const useCanInteractWithColony = (colony?: Colony): boolean => {
...
+
+  /*
+   * Check if connected to the same chain
+   */
   const [{ id: walletHexChainId }] = wallet.chains;
   const colonyChain = colony?.meta?.chainId || DEFAULT_NETWORK_INFO.chainId;
   const userWalletChain = parseInt(walletHexChainId.slice(2), 16);
-  return colonyChain === userWalletChain && canInteractWithNetwork;
+
+  /*
```
arrenv, line 59: I felt this check was relevant here as it is also relevant to whether can interact with a Colony or not.

src/components/frame/SubscribedColoniesList/SubscribedColoniesList.tsx
```diff
@@ -20,17 +20,10 @@ const MSG = defineMessages({
 });
 
 const SubscribedColoniesList = () => {
-  const { user, userLoading, updateUser } = useAppContext();
+  const { user, userLoading } = useAppContext();
   const canInteractWithNetwork = useCanInteractWithNetwork();
   const isMobile = useMobile();
```
arrenv, line 26: Removed as of this comment as it is now redundant - https://github.com/JoinColony/colonyCDapp/pull/83/files#r1029539075

src/components/common/NewActionButton/NewActionButton.tsx
```diff
@@ -64,7 +64,7 @@ const MSG = defineMessages({
 // }
 
 const NewActionButton = (/* { colony, ethDomainId }: Props */) => {
-  const canInteractWithColony = useColonyContext();
+  const { canInteractWithColony } = useColonyContext();
```
arrenv, line 67: I noticed this issue, although, feel free to correct me if this was not the intended purpose.

src/context/AppContext.tsx
```diff
@@ -94,23 +97,46 @@ export const AppContextProvider = ({ children }: { children: ReactNode }) => {
     }
   }, [updateUser, wallet]);
 
+  const asyncFunction = useAsyncFunction({
```
arrenv, line 100: Moved this here to allow it to be used by other components, like I have done within the `Join this Colony` button. Allowing us to prompt a user to connect, as opposed to direct them to the Colony creation page.

This is also more relevant considering the user creation happens on connection.

src/components/frame/RouteLayouts/UserNavigation/Wallet/Wallet.tsx
```diff
@@ -32,31 +31,9 @@ const MSG = defineMessages({
 });
 
 const Wallet = () => {
-  const { wallet, updateWallet, walletConnecting, setWalletConnecting } =
-    useAppContext();
+  const { wallet, walletConnecting, connectWallet } = useAppContext();
   const isMobile = useMobile();
 
-  const asyncFunction = useAsyncFunction({
```
arrenv, line 39: Removed as of the below comment in `AppContext`

src/components/frame/SubscribedColoniesList/SubscribedColoniesList.tsx
```diff
@@ -72,7 +64,7 @@ const SubscribedColoniesList = () => {
             );
           })}
       </div>
-      {canInteractWithNetwork && !isMobile && (
```
arrenv, line 75: Removed this so that the create a Colony button still appears in the sidebar when wallet is not connected.

src/components/frame/RouteLayouts/UserNavigation/Wallet/Wallet.tsx
```diff
@@ -32,31 +31,9 @@ const MSG = defineMessages({
...
-      walletConnectSuccess = true;
-    } catch (error) {
-      console.error('Could not connect wallet', error);
-    }
-    if (updateWallet && walletConnectSuccess) {
-      updateWallet();
-    }
-    setWalletConnecting?.(false);
-  }, [asyncFunction, updateWallet, setWalletConnecting]);
-
```
arrenv, line 59: Removed as of the below comment in `AppContext`

src/components/frame/SubscribedColoniesList/SubscribedColoniesList.tsx
```diff
@@ -20,17 +20,9 @@ const MSG = defineMessages({
...
+  const { user, userLoading } = useAppContext();
   const isMobile = useMobile();
 
-  /* Ensures colony list is up-to-date post create colony flow. */
-  useEffect(() => {
-    if (updateUser) {
-      updateUser(user?.walletAddress);
-    }
-  }, [user, updateUser]);
-
```
arrenv, line 33: Removed as of the following comment as it is now redundant - https://github.com/JoinColony/colonyCDapp/pull/83/files#r1029539075

src/context/AppContext.tsx
```diff
@@ -94,23 +97,46 @@ export const AppContextProvider = ({ children }: { children: ReactNode }) => {
...
+      walletConnectSuccess = true;
+    } catch (error) {
+      console.error('Could not connect wallet', error);
+    }
+    if (updateWallet && walletConnectSuccess) {
+      updateWallet();
+    }
+    setWalletConnecting?.(false);
+  }, [asyncFunction, updateWallet, setWalletConnecting]);
+
```
arrenv, line 120: Moved this here to allow it to be used by other components, like I have done within the `Join this Colony` button. Allowing us to prompt a user to connect, as opposed to direct them to the Colony creation page.

I feel this prompt is more relevant considering the user creation happens on connection now.

src/hooks/useCanInteractWithColony.ts
```diff
@@ -48,10 +48,24 @@ export const useCanInteractWithColony = (colony?: Colony): boolean => {
...
+   * Checking if watching (following) or not
+   */
+  const isWatching = !!(user?.watchlist?.items || []).find(
+    (item) => (item?.colony as Colony)?.colonyAddress === colony?.colonyAddress,
+  );
+
+  return (
+    colonyChain === userWalletChain && canInteractWithNetwork && isWatching
+  );
```
arrenv, line 68: I felt this check was relevant. Feel free to correct me if it is not the intended purpose of `useCanInteractWithColony`.

src/hooks/useCanInteractWithColony.ts
```diff
@@ -48,10 +48,24 @@ export const useCanInteractWithColony = (colony?: Colony): boolean => {
...
   const userWalletChain = parseInt(walletHexChainId.slice(2), 16);
-  return colonyChain === userWalletChain && canInteractWithNetwork;
+
+  /*
+   * Checking if watching (following) or not
+   */
+  const isWatching = !!(user?.watchlist?.items || []).find(
+    (item) => (item?.colony as Colony)?.colonyAddress === colony?.colonyAddress,
+  );
+
```
willm30, line 65: I notice this `.find` logic is also being used in `ColonySubscription`, so could be wrapped up into a little utility function and shared by both

src/hooks/useCanInteractWithColony.ts
willm30: Should we update this copy as well? e.g. connect wallet, register user, _join the colony_ ...

colonyinteract

src/context/AppContext.tsx
willm30: > user creation happens on connection now

Not sure what you mean here?

willm30 (commented):
Good stuff! Just left a few questions / observations:

* The ColonyMembersWidget doesn't automatically update when I join / leave a colony. Could we add this to this PR so it doesn't require a page refresh?

* If I logout, click 'joinColony' and then log in with a user that does not have a registered user account, I should probably be redirected to /create-user? It currently does this:

jc

* I'm also noticing the actions / extensions list is moving up & down with the presence / absence of the joinColony btn. I think these should have a fixed position.
menu-jumps

* When my wallet is not connected, clicking the add colony button in the colonies list just takes me to the landing page. Should this prompt me to connect my wallet first, then redirect me to the create colony route (or similar)?

jakubcolony (commented):
Nice work 👍 ! Will got here first so I'm mostly just seconding his comments

src/hooks/useCanInteractWithColony.ts
arrenv: Yeah, good point. I did not even see that come up.

src/context/AppContext.tsx
willm30: Not as far as I'm aware. We have the username and email inputs in the user creation flow so I don't think we could implement this (how would we select a username for the wallet, for instance?). Users are created when you run the temp script, maybe that's what's happening?

arrenv:
> * The ColonyMembersWidget doesn't automatically update when I join / leave a colony. Could we add this to this PR so it doesn't require a page refresh?

Yeah, sure thing.

> * If I logout, click 'joinColony' and then log in with a user that does not have a registered user account, I should probably be redirected to /create-user? It currently does this:

It creates a user entry automatically, right? In the new UI/UX we will likely handle this with a popup or callout.

> * I'm also noticing the actions / extensions list is moving up & down with the presence / absence of the joinColony btn. I think these should have a fixed position.

I agree.

> * When my wallet is not connected, clicking the add colony button in the colonies list just takes me to the landing page. Should this prompt me to connect my wallet first, then redirect me to the create colony route (or similar)?

Going to the add Colony page first I think is the right flow, otherwise, it is a doing an action that was not intended. And the Create a colony page is not limited to only connected users.

src/context/AppContext.tsx
arrenv: Perhaps a misunderstanding, this is based off a conversation with Raul, which I thought I recall him saying that this was how it was setup.

willm30:
> Going to the add Colony page first I think is the right flow, otherwise, it is a doing an action that was not intended. And the Create a colony page is not limited to only connected users.

At the moment, the create colony route is only accessible if a wallet is connected and a user account registered, so would need refactoring if the intention is to be able to use it without either of these. You also, however, need your wallet connected to interact with the network contracts (e.g. to deploy the colony). Is the intention that one should be able to create a colony without a username registered or a wallet connected?

src/context/AppContext.tsx
willm30: @rdig Perhaps you're able to clarify here.

arrenv:
> At the moment, the create colony route is only accessible if a wallet is connected and a user account registered, so would need refactoring if the intention is to be able to use it without either of these. You also, however, need your wallet connected to interact with the network contracts (e.g. to deploy the colony). Is the intention that one should be able to create a colony without a username registered or a wallet connected?

My bad, you're right, I was thinking of the landing page and not the first step of the Create Colony process. So, what I mean is that it would go to the landing page when not connected. Then the Create a colony button on that screen would prompt them to login.

src/context/AppContext.tsx
arrenv: @willm30 Either way, I still feel it makes sense to extract it.

arrenv:
> The ColonyMembersWidget doesn't automatically update when I join / leave a colony. Could we add this to this PR so it doesn't require a page refresh?

@willm30 Just confirming in case I am missing something, this requires a bit of refactoring of Colony Context to have it update, but most likely it is setting up the Member list queries, right? Which seems a little more involved.

> If I logout, click 'joinColony' and then log in with a user that does not have a registered user account, I should probably be redirected to /create-user? It currently does this:

Yeah, this is what I thought would be resolved when the user account is account created on first connection, I can refractor if not the case. Will just wait on confirmation from @rdig.

willm30:
> @willm30 Just confirming in case I am missing something, this requires a bit of refactoring of Colony Context to have it update, but most likely it is setting up the Member list queries, right? Which seems a little more involved.

@arrenv It may be the case that you can pass down an `updateColony` function from ColonyContext that updates the colony held in context -> calling this, say, after clicking 'join', would update the members widget because currently it's reading watchers from the colony object it receives as a prop (which comes from context). However, also happy for it to be a separate issue if it strays a bit too far from the intent of this pr.

src/context/AppContext.tsx
rdig: Will is correct here. We did in fact had a conversation with Arren about this, but the user creation thing was merely suggested, not implemented.

It can be theoretically done, however it's not really strait forward and will require some work to implemenent.

Eventually we might want to have it, if we decide for it, but for the time being it's not going to be implemented.

src/components/common/ColonyHome/ColonySubscription/ColonySubscription.tsx
```diff
@@ -33,43 +38,76 @@ const MSG = defineMessages({
...
+    if (updateUser && updateColony) {
+      updateUser(user?.walletAddress);
+      updateColony();
+    }
+  }, [user, updateUser, watchState, unwatchState, updateColony]);
+
+  const handleJoinClick = () => {
+    if (user) {
+      watch();
+    } else if (wallet && !user) {
```
arrenv, line 82: There is a modal that we planned to add with the new onboarding for creating a user account without navigating away from a Colony. The design is here - https://www.figma.com/file/xGRNwVyii0qaCkWs4hASUn/Onboarding-2.0?node-id=2208%3A45823&t=s6z7HdeHEKIkptnb-1

I have allowed for it here.

src/components/common/ColonyHome/ColonySubscription/ColonySubscription.tsx
```diff
@@ -81,35 +119,19 @@ const ColonySubscription = () => {
                 data-test="colonyMenuPopover"
                 title={MSG.colonyMenuTitle}
               />
-            )} */}
-            <div>***</div>
+            )}
           </ColonySubscriptionInfoPopover>
         )}
-        {!isSubscribed && (
+        {!canInteractWithColony && !walletConnecting && !initConnect && (
```
arrenv, line 125: Added `initConnect` to stop jumping in of Join button

src/context/ColonyContext.tsx
```diff
@@ -43,13 +60,50 @@ export const ColonyContextProvider = ({
...
+      } catch (e) {
+        console.error(e);
+        setColonyError(e);
+      } finally {
+        setUpdateLoading(false);
+      }
+    }
+  }, [colonyName]);
+
+  const colony = latestColony ?? data?.getColonyByName?.items?.[0] ?? undefined;
```
arrenv, line 94: Added so Colony context updates.

src/utils/newUser.ts
```diff
@@ -0,0 +1,4 @@
+/* Handle new user connecting */
+export const newUser = () => {
+  // TO DO: Add new user modal
+};
```
arrenv, line 4: Added here, will be finished in another PR as mentioned in other comment.

src/components/common/ColonyHome/ColonySubscription/ColonySubscription.tsx
willm30: Perhaps we just remove the spinner above? We wouldn't need the additional `initConnect` logic in that case, and because we're just querying our own db, it's pretty fast and makes a spinner redunant. 

Testing with / without spinner and `initConnect` seem very similar, even when throttling to slow 3G.

src/utils/watching.ts
```diff
@@ -0,0 +1,11 @@
+import { Colony, User } from '~types';
+
+/* Find in users watchlist if watching a specific colony */
+export const watchingColony = (
```
willm30, line 4: May I suggest `findWatchedColony` or `getWatchedColony`?

willm30 (commented):
> So, what I mean is that it would go to the landing page when not connected. Then the Create a colony button on that screen would prompt them to connect.

Currently, the create a colony button isn't showing on landing when a wallet isn't connected (i.e. after clicking the add colony button on colony home). I guess that should also be included in this pr?

src/components/common/ColonyHome/ColonySubscription/ColonySubscription.tsx
willm30: That said, leaving the spinner in and just commenting out `initConnect`  in `ColonySubscription`, I don't notice the Join button jumping. So not sure of the intention of `initConnect`.

src/components/common/ColonyHome/ColonySubscription/ColonySubscription.tsx
```diff
@@ -33,43 +38,76 @@ const MSG = defineMessages({
...
+      variables: {
+        input: {
+          colonyID: colony?.colonyAddress || '',
+          userID: user?.walletAddress || '',
+        },
+      },
+    });
+
+  /* Unwatch (unfollow) a colony */
+  const [unwatch, { data: unwatchState, loading: loadingUnwatch }] =
```
jakubcolony, line 66: How about `unwatchData` instead of `unwatchState`?

src/context/ColonyContext.tsx
```diff
@@ -43,13 +60,50 @@ export const ColonyContextProvider = ({
     fetchPolicy: 'cache-and-network',
   });
 
-  const colony = data?.getColonyByName?.items?.[0] ?? undefined;
+  const updateColony = useCallback(() => {
```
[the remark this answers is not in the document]
jakubcolony: I think we should take a step back - is the intention to just refetch the colony query following the watch/unwatch mutation so that the UI updates with the newest data?

If so, there are two solutions already provided that Apollo that require minimal/no changes to ColonyContext:
- On the mutation itself, you can specify the `refetchQueries` field to include the colony query:
```js
const [watch, { data: watchState, loading: loadingWatch }] =
    useCreateWatchedColoniesMutation({
      variables: {
        input: {
          colonyID: colony?.colonyAddress || '',
          userID: user?.walletAddress || '',
        },
       refetchQueries: [GetFullColonyByNameDocument]
      },
    });
```
The `GetFullColonyByNameDocument` is generated automatically by codegen and exported from `~gql`.
- `useGetFullyColonyByNameQuery` also returns a `refetch` function, which you can expose to the context consumers

That way you shouldn't have to replicate all the existing logic

src/utils/watching.ts
```diff
@@ -0,0 +1,11 @@
...
+  colony: Colony | undefined,
+  user: User | null | undefined,
```
jakubcolony, line 6: Why not mark them as optional with `?`?

src/utils/watching.ts
```diff
@@ -0,0 +1,11 @@
...
+): any => {
```
jakubcolony, line 7: You can get the return type for example by accessing `NonNullable<User['watchlist']>['items'][number]`. Alternatively create a GraphQL fragment for the fields you're querying on User > watchlist and use the fragment here, like so (`fragments/user.graphql`)
```gql
fragment WatchlistItem on WatchedColonies {
  id
  colony {
    ..WatchedColony
  }
  createdAt
}

# Later in the user fragment:
watchlist {
  items {
    ..WatchlistItem
  }
}
```

This will generate a `WatchlistItemFragment` which, in line with other types, you can export from `types/graphql` and use here

jakubcolony (changes requested):

src/components/common/ColonyHome/ColonySubscription/ColonySubscription.tsx
arrenv: If you reload the app with auto login the `Join this Colony` flashes in and out, or if you connect a wallet. Basically, the async requests happening between `updateUser` and `updateWallet` means that there is a state where loading isn't happening but the user and wallet don't yet exist.

The `initConnect` was intended to bridge the the whole process and avoid this issue.

src/context/ColonyContext.tsx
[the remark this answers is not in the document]
arrenv: Very nice! Works great.

arrenv:
@willm30 & @jakubcolony thanks for the reviews. Updated based on feedback.

willm30 (commented):
@arrenv Only registered users should be able to create colonies, right? 

Currently you can create a colony without a registered account. 

registered-user

Which causes the query in `ColonyContext` to fail when trying to access the new colony's route:

arrenv:
> @arrenv Only registered users should be able to create colonies, right?
> 
> Currently you can create a colony without a registered account.

I am actually able to create a Colony without a registered account. As far as I am aware, there is also no reason from a contract perspective to limit creating a Colony to having a registered user account. The 404 at the end of the creation process seems to be a UI imposed limit.

image

src/components/common/ColonyHome/ColonySubscription/ColonySubscription.tsx
jakubcolony: @arrenv @willm30 
As far as I understand, what we want to achieve here is not show the Join button if wallet or user is loading (and if they're not already a watcher, but I'll leave it out).

We should be able to check for `walletConnecting` and `userLoading` and I tried that but was still experiencing the Join button flicker.

I think there's a bug in `AppContext`'s `updateUser`. We set `userLoading` to true, make the query, then set `userLoading` back to false, but we are not awaiting the query, hence userLoading is false even while the query is still loading.

Here's how I fixed this:
- Make `updateUser` async and await the `query` method
- Change `!initConnect` in the line above to `!userLoading` (destructure that from `useAppContext`)
- Remove `initConnect` from AppContext

jakubcolony (changes requested):
Great progress, thanks for addressing my previous comments!

>Maybe this is a matter of architecture in the way watching works. The decentralized version also seems to run into a bit of contradiction here. Perhaps it needs more discussion.

I'd suggest we restrict the ability of non-registered users to create a colony until we've resolved that to avoid the error that Will mentioned

src/components/common/ColonyHome/ColonySubscription/ColonySubscription.tsx
willm30: Thanks for looking into it @jakubcolony. I also feel we could do without `initConnect`. On master, `updateUser` is async as I added it some prs ago, so perhaps a rebase plus your suggestions here would solve the problem.

src/components/common/ColonyHome/ColonySubscription/ColonySubscription.tsx
```diff
@@ -33,43 +39,86 @@ const MSG = defineMessages({
...
+    });
+
+  /* Update user on watch/unwatch */
+  useEffect(() => {
+    if (updateUser) {
+      updateUser(user?.walletAddress);
+    }
+  }, [user, updateUser, watchData, unwatchData]);
+
+  const handleSubscribe = useCallback(() => {
```
jakubcolony, line 91: Is there any benefit in putting this into `useCallback`? @willm30 you might be able to tell

src/components/common/ColonyHome/ColonySubscription/ColonySubscription.tsx
```diff
@@ -33,43 +39,86 @@ const MSG = defineMessages({
...
+    if (user) {
+      watch();
+    } else if (wallet && !user) {
+      handleNewUser();
+      // TO Do: update to new user modal
+      navigate(CREATE_USER_ROUTE, { replace: true });
```
jakubcolony, line 97: Should we use replace here? We might want to allow people to go back to colony home from the user creation flow

src/components/common/CreateColonyWizard/CreateColonyWizard.ts
```diff
@@ -1,16 +1,26 @@
-import { ComponentType } from 'react';
```
jakubcolony, line 1: The idea here is to show the user creation wizard before colony creation if user isn't registered, right? It might be easier just to hide the create colony button so there's only one option for unregistered users, but alternatively, there could be a way of achieving what you want without modifying the existing wizards too heavily:
On Create colony button click, if user is unregistered, store a flag that tells you they want to create colony (it could be in context or local storage). Then simply redirect them to create user wizard. Modify the user wizard submission handler to check for the flag, and if it's present, redirect them to creating colony. 
That way, create colony doesn't need to know anything about user creation

src/components/common/CreateUserWizard/StepUserName.tsx
```diff
@@ -54,13 +55,25 @@ const UsernameInput = ({ disabled }: UsernameInputProps) => (
...
+  /* Handle if user flow is within Colony create */
   /* Replace: true so you can't get back to the User wizard after completion */
-  const handleSuccess = () => navigate(LANDING_PAGE_ROUTE, { replace: true });
+  const handleSuccess = () => {
+    if (!matchPath(pathname, CREATE_USER_ROUTE)) {
+      if (updateUser && wallet) {
+        updateUser(wallet?.address);
+      }
+      return nextStep(wizardValues);
+    }
```
jakubcolony, line 74: Re my previous comment, here you could check for the presence of the flag and redirect them to colony creation

jakubcolony (commented):
This is looking really solid now @arrenv. I shared a few thoughts on the way the connection between user and colony creation wizard is implemented. As a side note, it'd be great if it was a separate issue with a separate PR, that way I could approve the colony subscriptions part and we could carry on discussing the other issue

src/components/common/ColonyHome/ColonySubscription/ColonySubscription.tsx
willm30: No, could remove

src/components/common/ColonyHome/ColonySubscription/ColonySubscription.tsx
```diff
@@ -33,43 +39,86 @@ const MSG = defineMessages({
 });
 
 const ColonySubscription = () => {
-  const { colony } = useColonyContext();
-  const { colonyAddress } = colony || {};
+  const { colony, canInteractWithColony } = useColonyContext();
+  const {
```
willm30, line 43: Could these hooks be wrapped up into a, say, `useColonySubscription` hook that only returns the methods/data the view needs, e.g. the handleSubscribe/unsubscribe and the loading state?

src/components/common/ColonyHome/ColonySubscription/ColonySubscription.tsx
```diff
@@ -81,35 +130,19 @@ const ColonySubscription = () => {
                 data-test="colonyMenuPopover"
                 title={MSG.colonyMenuTitle}
               />
-            )} */}
-            <div>***</div>
+            )}
           </ColonySubscriptionInfoPopover>
         )}
-        {!isSubscribed && (
+        {!canInteractWithColony && !walletConnecting && !userLoading && (
```
willm30, line 136: Perhaps move this to a named variabled e.g `canSubscribe && Button...`

willm30 (changes requested):
The create colony wizard flow is not working for me. When I get to create token, I cannot move forward or back. The continue button doesn't respond, neither does the go back button in the ui.

newToken

I also second Jakub's suggestion of splitting this pr so we can review the wiring in of the ColonySubscription button separately from ensuring the create user / create colony flows work as expected.

src/components/common/CreateColonyWizard/CreateColonyWizard.ts
arrenv: I will remove from this PR completely. While I agree hiding it is easier from a development perspective, it removes value from the user experience, as a users goal may be to create a colony and not just a user account. Essentially, hiding it removes a strong call to action.

jakubcolony: Yeah, I thought so, hence my alternative suggestion. Are you still planning to implement it in a separate PR?

arrenv: Yes, I will put it in a different PR.

src/components/common/ColonyHome/ColonySubscription/ColonySubscription.tsx
```diff
@@ -1,30 +1,26 @@
...
 
-import { Colony } from '~types';
-import { useAppContext, useColonyContext } from '~hooks';
-import { CREATE_USER_ROUTE } from '~routes/index';
-
 import ColonySubscriptionInfoPopover from './ColonySubscriptionInfoPopover';
 
 import styles from './ColonySubscription.css';
+import { useColonyContext } from '~hooks';
+import useColonySubscription from '~hooks/useColonySubscription';
```
jakubcolony, line 12: Could be exported from ~hooks directly as above

src/utils/watching.ts
```diff
@@ -0,0 +1,11 @@
+import { Colony, User } from '~types';
+
+/* Find in users watchlist if watching a specific colony */
+export const getWatchedColony = (
+  colony?: Colony,
+  items?: NonNullable<User['watchlist']>['items'],
+): NonNullable<User['watchlist']>['items'][number] | undefined => {
```
jakubcolony, line 7: I think extracting that to its own type would make this a bit cleaner

jakubcolony (approved):
Nice work @arrenv! I think this is good to go now 👍

willm30 (commented):
Am I right in thinking the desired changes to the create user / create colony flow are being handled in a separate issue? 

In which case, I think we should revert the functionality back to how it was originally in this pr (i.e. can't see create colony btn until there's a registered user), so we don't inadvertantly introduce bugs. For instance, the ability to create a colony without a registered user looks strange, and results in a 404 redirect:

create-colony

Also noting that if I click "Join this Colony" while logged in to a wallet that doesn't have a registered user account, I get redirected to the user creation wizard (as expected), but then don't get redirected back to the colony at the end, nor do I get joined to the colony automatically at the end. I need to hit the back button in the browser to get back to the colony page, but we should probably do one or the other so this isn't necessary. (Happy for this to be included in a different pr, just mentioning here for completeness).

src/context/AppContext.tsx
```diff
@@ -43,6 +46,7 @@ export const AppContextProvider = ({ children }: { children: ReactNode }) => {
     // It means that it was not set in context yet
   }
 
+  // initConnect used for smoother handling async processes
```
willm30, line 49: can remove this comment I believe

src/hooks/useColonySubscription.ts
```diff
@@ -0,0 +1,86 @@
...
+    walletConnecting,
+    connectWallet,
+    userLoading,
+  } = useAppContext();
+  const navigate = useNavigate();
+
+  const watchedItem = getWatchedColony(colony, user?.watchlist?.items);
+
+  /* Watch (follow) a colony */
+  const [subscribe, { data: watchData }] = useCreateWatchedColoniesMutation({
```
willm30, line 29: nitpick, but subscribe may be misleading since it's not actually a subscription in the apollo sense of the term. I think `watch()` \ `unwatch()` captures the same intent?

Issue #99, opened 2022-11-24
Port mobile styles for Colony Home


willm30:
Currently, colony home page isn't mobile repsonsive like in the dapp. These styles should be ported over. 

Related components that should also have their styles brought across:

- [x] Colony Home handled in #101 
- [x] UserNavigation handled in #101
- [x] SubscribedColonies Dropdown handled in #136 
- [x] ColonyTotalFunds

PR #144, opened 2022-12-18
Tidy components at `edit-profile`

willm30:
## Description

While migrating the components at `edit-profile` to use React Hook Form, I spotted a few opportunities to improve code reuse and readability. This PR implements those changes. 

It also makes consistent the form submission behaviour between the "main" and "advanced" settings tabs, so that in both, the snackbar appears and the loading spinner does not flash. Additionally, it removes the flashing of the loading spinner after uploading a new user avatar. 

It also adds the "Go to Profile" button in the top left hand corner.

I have checked that form state persists correctly when switching tabs, routes, and refreshing the browser; but would be grateful if you could confirm when reviewing. This applies to the user avatar and both forms.

**New stuff** ✨

* Abstracted: `AdvancedSettingsRow`, `SubmitForm` and `UserInfo`

**Changes** 🏗

* Added loading spinner to `UserColonies` at "My profile" -> avoids text flashing in the event there are no colonies
* Form submission is now consistent in both tabs -- i.e. no loading spinner
* Avatar also now doesn't display loading spinner after selecting a new one
* Added "Go to profile" button in top left and correct mobile styling for `EditProfile`.
* CustomRPC url won't update if decentralizedNode if switched off. See comment in code for rationale. Does this feel right?


src/components/frame/RouteLayouts/UserNavigation/Wallet/Wallet.tsx
```diff
@@ -32,8 +32,8 @@ const MSG = defineMessages({
 });
 
 const Wallet = () => {
-  const [walletConnecting, setWalletConnecting] = useState<boolean>(false);
```
willm30, line 35: Moving this to `appContext` makes it accessible elsewhere, for example in `UserProfile`. Here it's useful because the text rendered in the event there are no colonies is dependent on which wallet is connected, and so we can show a loading state while the wallet is connecting.

src/components/common/UserProfile/UserColonies.tsx
```diff
@@ -20,23 +21,39 @@ const displayName = 'common.UserProfile.UserColonies';
...
   user: { walletAddress, watchlist, name, profile },
 }: Props) => {
-  const { user: currentUser } = useAppContext();
+  const { user: currentUser, walletConnecting } = useAppContext();
   const isCurrentUser = currentUser?.walletAddress === walletAddress;
+
+  if (walletConnecting) {
+    return (
+      <div className={styles.loader}>
+        <SpinnerLoader
```
willm30, line 49: When the wallet is connecting, this will default to MSG.otherUserNoColonies, and so we get this flashing of text in the event the other message should be shown once the wallet connects. I think it's preferable to show a loading state while the wallet is connecting, hence the inclusion of a spinner here and moving the `walletConnecting` state into appContext.

src/components/common/UserProfileEdit/UserAdvancedSettings.tsx
```diff
@@ -109,134 +104,58 @@ const UserAdvancedSettings = () => {
...
+        resetOnSubmit
+      >
+        {({ formState: { isValid, defaultValues }, watch }) => {
+          const values = watch();
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+          const isDirty = getIsDirty(values, defaultValues!);
+          return (
+            <div className={styles.main}>
+              {advancedSettingsRows.map((row) => (
+                <AdvancedSettingsRow
```
willm30, line 138: Each row should use the same template so have moved into own component.

src/components/common/UserProfileEdit/UserAdvancedSettings.tsx
```diff
@@ -109,134 +104,58 @@ const UserAdvancedSettings = () => {
...
+                  paragraphText={row.paragraphText}
+                  toggleDisabled={!metatransactionsAvailable}
+                  toggleLabel={row.toggleLabel}
+                  tooltipText={row.tooltipText}
+                  tooltipTextValues={row.tooltipTextValues}
+                  extra={row.extra}
+                  key={row.name}
+                />
+              ))}
+              <SaveForm
```
willm30, line 149: This is the same across both tabs so have moved into own component also.

src/components/common/UserProfileEdit/UserAvatarUploader.tsx
```diff
@@ -28,49 +27,58 @@ const UserAvatarUploader = ({
...
+        input: { id: walletAddress, avatar: null },
       },
     });
-    updateUser?.(user.walletAddress);
+    setAvatar(null);
   };
 
+  // on dismount
+  useEffect(
+    () => () => updateUser?.(user.walletAddress),
```
willm30, line 58: Moving the call to `updateUser` to the component dismount avoids the `loadingSpinner` from displaying after uploading a new avatar. In the meantime we set the avatar to component state so it can be displayed as the placeholder for the `AvatarUploader`.

src/components/common/UserProfileEdit/UserMainSettings.tsx
```diff
@@ -94,127 +73,99 @@ const validationSchema = object({
...
-    const valuesToSubmit = {
-      email: updatedProfile.email === '' ? null : updatedProfile.email,
-    };
-
-    editUser({
-      variables: {
-        input: {
-          id: walletAddress,
-          ...userProfile,
-          ...updatedProfile,
```
willm30, line 125: Here we can just spread the updatedProfile, we don't need to also spread the previous user profile.

src/hooks/useUserByNameOrAddress.ts
```diff
@@ -8,6 +8,7 @@ const useUserByNameOrAddress = (usernameOrAddress: string) => {
       name: usernameOrAddress,
       address: usernameOrAddress,
     },
+    fetchPolicy: 'cache-and-network',
```
willm30, line 11: This ensures the avatar is uptodate when you change it in "Settings" and then navigate to "My Profile"

src/components/common/UserProfileEdit/UserAdvancedSettings.tsx
```diff
@@ -28,78 +29,74 @@ const MSG = defineMessages({
...
+  );
+
+const setFormValuesToLocalStorage = (
+  values: FormValues,
+  setSettingsKey: UserSettingsHook['setSettingsKey'],
+) => {
+  Object.entries(values).forEach(
+    ([key, value]: [SlotKey, string | boolean]) => {
+      if (
+        // don't save changes to custom rpc if decentralized mode is off
```
willm30, line 86: The thinking here is to prevent someone changing the custom rpc by accident -> for example, with decentalized mode on, they change the rpc url, but then change their mind and turn decentralized mode off and submit the form without reverting the rpc url to its original value.

In this case, presumably they didn't mean to change the rpc url (because why change it only to have decentralized mode off?). So this check would prevent any updates to the custom rpc if decentralized mode is off.

src/context/AppContext.tsx
```diff
@@ -43,6 +45,7 @@ export const AppContextProvider = ({ children }: { children: ReactNode }) => {
   const [wallet, setWallet] = useState(initialWallet);
   const [user, setUser] = useState<User | null | undefined>(initialUser);
   const [userLoading, setUserLoading] = useState(false);
+  const [walletConnecting, setWalletConnecting] = useState(false);
```
willm30, line 48: See comment in Wallet for rationale

src/routes/Routes.tsx
```diff
@@ -149,9 +150,16 @@ const Routes = () => {
         <Route
           path={USER_EDIT_ROUTE}
           element={
-            <UserLayout routeProps={{ hasBackLink: true }}>
+            <Default
```
willm30, line 153: Adds "Go to Profile" button like in dapp in `edit-profile`

src/components/shared/Fields/Form/HookForm.tsx
```diff
@@ -40,7 +43,25 @@ const HookForm = <FormData extends FieldValues>({
...
+    reset,
+    formState: { isSubmitting },
+  } = formHelpers;
+  const values = watch();
+
+  /*
+   * Effect resets default values to latest values, which resets the isDirty prop.
+   * Useful in user settings.
+   */
+  useEffect(() => {
```
willm30, line 59: This is so that when you hit save, the form re-disables, which it should because technically the form is now not dirty any longer.

src/components/common/UserProfileEdit/UserProfileEdit.tsx
```diff
@@ -44,10 +44,10 @@ const UserProfileEdit = () => {
     >
       <Tabs>
         <TabList>
-          <Tab>
+          <Tab tabIndex={0}>
```
willm30, line 47: I was noticing the `focus-visible` outline was occasionally being triggered on click, which I don't think is intended (normally reversed for tabbing). Anecdotally, this has resolved it for me.

src/components/common/UserProfileEdit/UserMainSettings.tsx
```diff
@@ -113,57 +86,35 @@ const UserMainSettings = ({
...
         defaultValues={defaultValues}
         onSubmit={handleSubmit}
         validationSchema={validationSchema}
+        resetOnSubmit
       >
-        {({ formState: { isSubmitting, isDirty, isValid } }) => (
+        {({ formState: { isDirty, isValid } }) => (
           <div className={styles.main}>
-            <FieldSet>
-              <InputLabel label={MSG.labelWallet} />
```
willm30, line 153: "UserInfo" is more descriptive in my opinion as to what this portion of the code is doing.

src/components/common/UserProfileEdit/UserAdvancedSettings.tsx
```diff
@@ -109,134 +106,53 @@ const UserAdvancedSettings = () => {
...
+          [SlotKey.DecentralizedMode]: decentralizedModeEnabled,
+          [SlotKey.CustomRPC]: customRpc,
+        }}
+        validationSchema={validationSchema}
+        onSubmit={(values) => handleSubmit(values, setSettingsKey)}
+        resetOnSubmit
+      >
+        {({ formState: { isValid, defaultValues }, watch }) => {
+          const values = watch();
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
willm30, line 130: Even though `defaultValues` correctly infers the values inputted to the Form prop, it still types it as `...values | undefined`. Hence need to assert non-null.

This invites a question of when to use the `defaultValues` provided by HookForm like here, and when to move the defaultValues provided as props to a variable and use those (like in `UserMainSettings`). 

I would prefer to use the one provided by HookForm, but it involves disabling the linter if you use a non-null assertion like here, which may get a bit ugly if you do it a lot. Hence my opting for the other pattern in `UserMainSettings`, since I'm using the default values in more than one place.

You could also do `defaultValues || {}` but this is unintuitive since we know the default Values are defined because we're defining them.

src/components/common/UserProfile/UserColonies.tsx
rdig: I actually appreciate you explaining the more "funky" parts of the implementation. 

I personally leave actual comments inside the file, but I think your way of doing it might be preferable overall as it enables discussion. _(ie: comments chain)_

I want to encourage this going forward as a git / PR style @jakubcolony @ArmandoGraterol @Julianahlk

src/components/common/UserProfileEdit/AdvancedSettingsRow.tsx
```diff
@@ -0,0 +1,190 @@
...
+      <div className={stylesAdvance.validateButtonContainer}>
+        <Button
+          text={MSG.validate}
+          disabled={!decentralizedMode || !isValid}
+          // @TODO: validation to be added later
+        />
+      </div>
+    </>
+  );
+};
```
rdig, line 113: I think this needs to be generic.

While currently we only a single input on the _Advanced Tab_, the plan is to have multiple _(RPC endpoint, subgraph, IPFS, block ingestor, maybe even a custom db, etc)_

src/components/common/UserProfileEdit/AdvancedSettingsRow.tsx
```diff
@@ -0,0 +1,190 @@
...
+          tooltipPopperOptions={{
+            placement: 'right',
+          }}
+        />
+      </div>
+      <p className={stylesAdvance.descriptions}>
+        <FormattedMessage {...paragraphText} />
+      </p>
+      <div className={stylesAdvance.settingsRowExtra}>{extra}</div>
+      <hr />
```
rdig, line 184: Not really a big fan of `hr` elements, but then again this might be just a personal nitpick rather than an objective complaint

src/components/common/UserProfileEdit/SaveForm.tsx
```diff
@@ -0,0 +1,59 @@
...
+      show={showSnackbar}
+      setShow={setShowSnackbar}
+      msg={error ? MSG.snackbarError : MSG.snackbarSuccess}
+      type={error ? SnackbarType.Error : SnackbarType.Success}
+    />
+  </>
+);
+
+SaveForm.displayName = displayName;
+export default SaveForm;
```
rdig, line 59: Maybe add a empty line between these two ?

src/components/common/UserProfileEdit/AdvancedSettingsRow.tsx
```diff
@@ -0,0 +1,190 @@
...
+    </>
+  );
+};
+
+AdvancedSettingsRow.displayName = displayName;
+export default AdvancedSettingsRow;
```
rdig, line 190: Same here ?

src/components/common/UserProfileEdit/UserAdvancedSettings.tsx
rdig: I think you're overthinking this :)

So the point of this feature is to be super power user feature. Stuff like "by accident" should not apply here since we're dealing with people who know how to set up, sync and configure their own archive node.

I'm ok with both versions of this interaction, but you shouldn't be too concerned with the UX in this case.

rdig (approved):
Very nice work.

Left some comments here and there, but overall quality work right here

src/components/common/UserProfileEdit/UserAdvancedSettings.tsx
willm30: Ok, I think simpler is better so have removed

src/components/common/UserProfileEdit/SaveForm.tsx
willm30: Should be showing as outdated as have added empty line

src/components/common/UserProfileEdit/AdvancedSettingsRow.tsx
willm30: Same as above

src/components/common/UserProfileEdit/UserInfo.tsx
```diff
@@ -0,0 +1,56 @@
...
+      <CopyableAddress appearance={{ theme: 'big' }} full>
+        {walletAddress}
+      </CopyableAddress>
+      <div className={styles.usernameContainer}>
+        <InputLabel label={MSG.labelUsername} />
+        <UserMention
+          user={user}
+          title={user.name || walletAddress}
+          hasLink={false}
+          data-test="userProfileUsername"
```
jakubcolony, line 47: What's the general idea around these? Should we be porting that over, even though there are no tests?

src/routes/Routes.tsx
```diff
@@ -149,9 +150,16 @@ const Routes = () => {
...
+              routeProps={{
+                hasBackLink: true,
+                hasSubscribedColonies: isMobile,
```
jakubcolony, line 156: Is that supposed to be showing colonies list somewhere in the /edit-profile route on mobiles? Couldn't see it anywhere

jakubcolony (approved):
Great work, the comments definitely helped 👍

Julianahlk (approved):
Nice works! Can confirm that the form state persists correctly when switching tabs, routes, and refreshing the browser 👍

src/components/common/UserProfileEdit/UserInfo.tsx
willm30: @jakubcolony My sense is that test coverage from dapp will be ported over eventually. Is that the plan @rdig?

src/routes/Routes.tsx
willm30: Yes, when the mobile version of the subscribed colonies list gets ported, it'll show. I envision this being resolved by https://github.com/JoinColony/colonyCDapp/issues/99

src/routes/Routes.tsx
```diff
@@ -49,19 +49,20 @@ import CreateColonyWizard from '~common/CreateColonyWizard';
...
-//     defaultMessage: 'Go to profile',
-//   },
-//   loadingAppMessage: {
-//     id: 'routes.Routes.loadingAppMessage',
-//     defaultMessage: 'Loading App',
-//   },
-// });
+const MSG = defineMessages({
+  userProfileEditBack: {
+    id: 'routes.Routes.userProfileEditBack',
```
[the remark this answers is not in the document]
willm30: I'm sure it can't hurt to add it

PR #148, opened 2022-12-19
Port ExtensionUpgrade component

### LINKED ISSUE #60: Port over Extension Upgrade
Port over the `ExtensionUpgrade` component rendered by `ColonyHomeLayout`.

Julianahlk:
## Description

This PR Ports the `ExtensionUpgrade` component rendered by ColonyHomeLayout and `oneTxMustBeUpgraded`.

Resolves #60


src/components/common/ColonyHome/ExtensionUpgrade/ExtensionUpgrade.tsx
```diff
@@ -3,13 +3,13 @@ import { defineMessages, FormattedMessage } from 'react-intl';
...
-import { useColonyExtensionsQuery } from '~data/index';
-import { oneTxMustBeUpgraded } from '~modules/dashboard/checks';
-import { useColonyContext } from '~hooks';
+import Alert from '~shared/Alert';
+import Button from '~shared/Button';
+import { oneTxMustBeUpgraded } from '~utils/checks';
+import { useColonyContext, useExtensionData } from '~hooks';
 
 import styles from './ExtensionUpgrade.css';
+import { InstalledExtensionData } from '~types';
```
jakubcolony, line 12: Should be higher up

src/utils/checks/oneTxMustBeUpgraded.ts
```diff
@@ -0,0 +1,13 @@
+import { Extension, ONE_TX_PAYMENT_VERSION_LATEST } from '@colony/colony-js';
```
jakubcolony, line 1: @rdig I didn't know this constant exists, how does it relate to storing the most recent version available in the network in the DB and exposing them on extension data via `availableVersion` field?
In the dapp it was yet something different, we were checking for a specific OneTxPayment version...

jakubcolony, line 1: On a different note, would it make sense to put this in the existing `utils/extensions` file or to create a separate subfolder for extensions utils rather than scatter them around?

jakubcolony (commented):
Just commenting for now, until I get an answer from Raul on how we should check for extensions version...

src/utils/checks/oneTxMustBeUpgraded.ts
Julianahlk: Sure, since `utils/extension` is small right now, I moved `oneTxMustBeUpgraded` into that file. I think we can create a subfolder later if necessary.

jakubcolony: Good point @willm30. Raul wanted to keep the versions in the database. I guess we'll find out once he's back

Julianahlk: @rdig any thoughts on the comments above?

Julianahlk: @jakubcolony @willm30 discussed with Raul and the extension version should come from the DB but for now we can leave it as is until #163 is merged. I left a comment in the code.

willm30: @Julianahlk #163 is keeping track of the Colony version, you should be able to get the extension version from the db in this branch

Julianahlk: Oh okay cool. I will look into it. Thanks 👍

Julianahlk: @jakubcolony and @willm30 I used `GetCurrentExtensionVersion` query as suggested. Can you both please take a look again? Thank you.

src/hooks/useOneTxMustBeUpgraded.ts
```diff
@@ -0,0 +1,29 @@
...
+      extensionHash,
+    },
+    fetchPolicy: 'cache-and-network',
+  });
+
+  const { version } = versionData?.getCurrentVersionByItem?.items?.[0] || {};
+
+  return version &&
+    extensionData?.currentVersion &&
+    isInstalledExtensionData(extensionData)
```
Julianahlk, line 24: Using isInstalledExtensionData so that this notifies the user to upgrade both if the extension is deprecated or enabled.

src/hooks/useOneTxMustBeUpgraded.ts
```diff
@@ -0,0 +1,29 @@
+import { getExtensionHash } from '@colony/colony-js';
+import { InstalledExtensionData } from '~types';
+import { isInstalledExtensionData } from '~utils/extensions';
+import { useGetCurrentExtensionVersionQuery } from '~gql';
+
+export const useOneTxMustBeUpgraded = (
+  extensionData: InstalledExtensionData,
```
willm30, line 7: Type definition needs to include the `| null` since that's how `extensionData` is defined in `ExtensionUpgrade`

src/components/common/ColonyHome/ExtensionUpgrade/ExtensionUpgrade.tsx
```diff
@@ -26,25 +29,19 @@ const MSG = defineMessages({
...
 
   const isExtensionDetailsRoute = extensionId === Extension.OneTxPayment;
 
-  const oneTxPaymentExtension = data?.processedColony?.installedExtensions.find(
-    ({ details, extensionId: extensionName }) =>
-      details?.initialized &&
-      !details?.missingPermissions.length &&
-      extensionName === Extension.OneTxPayment,
+  const mustUpgrade = useOneTxMustBeUpgraded(
+    extensionData as InstalledExtensionData,
```
willm30, line 43: Just checking this is how it works in the Dapp? That the banner only appears if the extension has been installed?

Is it not also possible to upgrade an extension that hasn't been installed yet? I.e. so that you install the most up to date version?

willm30: Should also include the `| null` typing from `extensionData`

willm30 (changes requested):
If you set `mustUpgrade` to true (line 46 of `ExtensionUpgrade`), when clicking the "Actions" navigation item, we get an error:

appcrash

The `ExtensionUpgrade` component presumably needs to be generalised so it can be shared across extensions? I.e. if the voting extension needs upgrading, we should be able to use the same component. Currently it's hardcoded for `OneTx`. 

And I'm seeing the Upgrade button even without the banner telling me to upgrade. I think these should be synchronised (i.e. we shouldn't see the upgrade button if there's nothing to upgrade to). 

upgrade-no-banner

jakubcolony (changes requested):
I feel like we should rename this component. It suggests it covers upgrading of (possibly any) extension, while in reality it only seems to be an alert informing of an outdated version of OneTxPayment (and no other extension).

I thought it would be helpful to provide some testing instructions (which would involve manually updating the current version of OneTxPayment in the db to 4), but this is currently blocked by the bug I fixed in #163. If it gets merged before, I will come back and provide the instructions

src/components/common/ColonyHome/ExtensionUpgrade/ExtensionUpgrade.tsx
```diff
@@ -26,25 +29,19 @@ const MSG = defineMessages({
...
   const { extensionId } = useParams<{
     extensionId: string;
   }>();
 
-  const { data } = useColonyExtensionsQuery({
-    variables: { address: colonyAddress },
-  });
+  const { extensionData } = useExtensionData(Extension.OneTxPayment);
 
   const isExtensionDetailsRoute = extensionId === Extension.OneTxPayment;
```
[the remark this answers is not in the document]
Julianahlk: @jakubcolony maybe the variable name doesn't make sense but based on the dapp it seems the purpose of isExtensionDetailsRoute is to determine if the user is currently on the details p[age for OneTxPayment to determine whether to show a button to navigate to OneTxPayment details page

src/hooks/useOneTxMustBeUpgraded.ts
Julianahlk: Removed useOneTxMustBeUpgraded so should no longer be relevant

Julianahlk:
> If you set `mustUpgrade` to true (line 46 of `ExtensionUpgrade`), when clicking the "Actions" navigation item, we get an error:
> 
This should be resolved.

> The `ExtensionUpgrade` component presumably needs to be generalised so it can be shared across extensions? I.e. if the voting extension needs upgrading, we should be able to use the same component. Currently it's hardcoded for `OneTx`.

@willm30 My assumption is that OneTx is a required extension therefore it must be upgraded. Are you saying that the other extensions have the same upgrade requirement?

> And I'm seeing the Upgrade button even without the banner telling me to upgrade. I think these should be synchronised (i.e. we shouldn't see the upgrade button if there's nothing to upgrade to).
> 
The upgrade button wasn't in the scope of this PR so this could be an existing issue in the extensions feature? @jakubcolony

jakubcolony:
>The upgrade button wasn't in the scope of this PR so this could be an existing issue in the extensions feature?

That sounds right, I did some fixes to the button in #158 but it probably needs another look when the entire branch is more or less ready

willm30:
> My assumption is that OneTx is a required extension therefore it must be upgraded. Are you saying that the other extensions have the same upgrade requirement

I would have thought so. I.e. if we change the contracts of the voting reputation extension, it's going to require "upgrading". 
@jakubcolony Is that your understanding?

jakubcolony:
@willm30 The upgrading functionality is already ported, this component's role was to alert people on an outdated OneTxPayment version - hence my request to change its name to be more specific. I checked with Raul this morning and there's no intentions to extend it to other extensions

willm30 (approved):
All looking good to me 👍

src/components/common/ColonyHome/ExtensionUpgrade/ExtensionUpgrade.tsx
jakubcolony: @Julianahlk I think this is still outstanding. The check is one-line, so we shouldn't need a separate function making extra query calls

src/components/common/Extensions/ExtensionStatusBadge/ExtensionStatusBadge.tsx
```diff
@@ -44,6 +44,10 @@ const ExtensionStatusBadge = ({
   let status;
   let theme;
 
+  if (!extensionData) {
+    return null;
+  }
+
```
jakubcolony, line 50: Why? It should never be `null`/`undefined`

jakubcolony (approved):
LGTM!

Issue #142, opened 2022-12-16
Generate optimised thumbnail from user avatar


willm30:
Creating this issue so we don't lose visibility of this. 

Currently when uploading a user avatar at `edit-profile`, the thumbnail doesn't update. 

I believe this is because we would like to implement an automatic thumbnail generator, presumably in a custom lambda function designed for that purpose. [details tbc]

So, the flow would be to upload an avatar, and then the lambda function will upload the avatar to the db and also generate an optimised thumbnail, and upload that also to the user's profile. This would ensure the thumbnail img is updated at the same time as the user uploads an avatar. 

## TODO

- [x] Research library for thumbnail optimisation 

### [Squoosh api](https://github.com/GoogleChromeLabs/squoosh/tree/dev/libsquoosh)

Probably my no.1 if its bundle size weren't 12mb. Maintained by google, I've actually used it before in the browser, it's very good; however, 12mb is too big, surely. It's also marked as experimental so perhaps best to avoid in production for now.

### [Sharp](https://www.npmjs.com/package/sharp)

Heard of this when using Gatsby as their official image plugins are built on top of it. Looks like it is node only though.

### [Jimp](https://www.npmjs.com/package/jimp)

1mn weeky downloads, 12k stars, and can run in the browser. But 3mb bundle.

### [Pica](https://www.npmjs.com/package/pica)

3k stars. 216kb bundle.

willm30:
@rdig Can I confirm I understand this correctly? I.e. the intention is to create a lamda function to handle the thumbnail generation / optimisation based on the avatar the user uploads?

rdig:
So I actually didn't give this a proper think.

We can generate the thumbnail both in the lambda function, on upload, and both in the client, and then sending both files. _(client side works in this case since it's a one-time thing, so won't cause mass processing overhead)_

Both are ok, and both are doable, and I don't really have a strong argument towards one or the other.

You have any thoughts on this ?

willm30:
@rdig If it were as simple as the following (pseudo-code)

``` tsx
import generateOptimisedThumbnail from 'some-external-library-or-local-util'

const newAvatar = formValues.avatar
const optimisedThumbnail = generateOptimisedThumbnail(newAvatar)

updateProfileMutation({
   avatar: newAvatar
   thumbnail: optimisedThumbnail
})

```

Then I would actually opt for client for the sake of simplicity. Especially if the call to optimise the thumbnail was small and quick.

[...]

PR #149, opened 2022-12-20
Add thumbnail optimiser for avatar upload

### LINKED ISSUE #141: Address Item size error in Avatar Upload
In uploading a file 529kb in size, I get the following error from apollo:

Image

Either we should change the file size validation and label to something within the limit we have with amplify, or increase that limit to permit files up to 1mb as is currently intended.

willm30:
## Description

This PR adds support for thumbnail optimisation when a user uploads an avatar to their profile in User Settings. 

avatar upload

It also refactors the AvatarUploader to use a new component `SingleFileUpload`, instead of the original `FileUpload` (see heading below for explanation). 

## Test

Upload an avatar. The thumbnail should automatically update per GIF above. None of the `AvatarUpload` functionality should have changed so please also double check this, e.g. upload an image that's too large / wrong file type, remove the avatar etc.  

**New stuff** ✨

* `getThumbnail` (thumbnail optimiser)
* `SingleFileUpload`

**Changes** 🏗

* Mainly made changes to the `AvatarUploader` and `UserAvatarUploader`, to accomodate the new components above.
 
**Deletions** ⚰️

* `AvatarUploadItem` -- no longer necessary.
* `Sidebar` -- was a fairly superfluous component and has been merged with `UserAvatarUploader`.

### Thumbnail Optimisation 

I looked into a few different libraries, but in the end went with [Pica](https://github.com/nodeca/pica), which has a bundle size of 216kb and 3.2k stars on github. Specifically, I'm using a [wrapper](https://github.com/nodeca/image-blob-reduce) around Pica also created by the Pica author, to resize images from Blobs/Files (as opposed to image / canvas elements). The image optimisation algorithm used is [`mks2013`](https://johncostella.com/magic/) and I am only using the pica defaults.
 
### SingleFileUpload

`FileUpload` is designed with the CSV uploader in mind, or in other words, designed to support the uploading of multiple files and the dynamic rendering of a component per file uploaded (see `CSVUploaderItem` for an example). 

This is not the same use case as the `AvatarUploader`, which only needs one file uploading and nothing to be dynamically rendered once uploaded. I think this distinction is sufficient to warrant two separate components (not least because it makes it easier and simpler to understand what's happening now when uploading an avatar). 

In order to avoid duplication of the actual uploading logic, I have created the `useDropzoneWithFileReader` hook, which separates the file reading/uploading logic from the file reading view. Not only does it make the upload process easier to follow, it also makes the file uploading indifferent as to whether it's rendered inside of a form or not (which I expect, again, is more useful for uploading multiple files than a single file).

Although I haven't done it in this pr, I envisage `useDropzoneWithFileReader` also being used in a refactored `FileUpload` when the `CSVUploader` is implemented (becoming `MultipleFileUpload` in the process).  

Ultimately, the intention here is to abstract uploading logic so it's separated from the upload view, which allows us to have separate components for single file uploads with no dynamic content versus multiple file uploads with dynamic content. 

Resolves  #141, #142


src/components/shared/UserAvatar/UserAvatar.tsx
```diff
@@ -35,7 +35,6 @@ const UserAvatar = ({
   address,
   banned = false,
   colony,
-  notSet = true,
```
willm30, line 38: Removing this default as more often than not it was being overridden

src/hooks/useDropzoneWithFileReader.ts
```diff
@@ -0,0 +1,50 @@
+import { DropzoneOptions, DropzoneProps, useDropzone } from 'react-dropzone';
+
+import {
+  DEFAULT_MIME_TYPES,
+  DEFAULT_MAX_FILE_SIZE,
+} from '~shared/FileUpload/limits';
+import { HandleFileAccept } from '~shared/FileUpload/types';
+import getFileReader from '~utils/fileReader';
+
+const useDropzoneWithFileReader = ({
```
willm30, line 10: This hook makes the use of dropzone more flexible and separates the logic from a specific view.

src/images/optimisation/index.ts
```diff
@@ -0,0 +1,31 @@
+import ImageBlobReduce, { Options } from 'image-blob-reduce';
```
willm30, line 1: `image-blob-reduce` is a wrapper around the [pica library](https://github.com/nodeca/pica). Under the hood, it uses the [`mks2013`](https://johncostella.com/magic/) algorithm to handle image optimisation, and handles resizing via the use of a `canvas` element.

src/types/definitions/image-blob-reduce/index.d.ts
```diff
@@ -0,0 +1,33 @@
+declare module 'image-blob-reduce' {
```
willm30, line 1: Manually added types as `image-blob-reduce` doesn't come with any

src/utils/fileReader/index.ts
```diff
@@ -34,22 +39,31 @@ const fileReaderFactory = (options: Partial<FileReaderOptions>) => {
           );
         }
 
-        const { name, type, size, lastModified } = file;
+        const { type, size } = file;
+        let name: File['name'] | undefined;
+        let lastModified: File['lastModified'] | undefined;
+
+        if (file instanceof File) {
```
willm30, line 46: fileReader can also handle blobs, and in the case of thumbnail optimisation, a `Blob` is returned by the optimiser. Blobs don't have `name` or `lastModified` fields, hence adding this check.

src/utils/fileReader/index.ts
```diff
@@ -34,22 +39,31 @@ const fileReaderFactory = (options: Partial<FileReaderOptions>) => {
...
+        }
+
         resolve({
           name,
           type,
           size,
           lastModified,
           uploadDate: Date.now(),
           data: evt.target.result,
+          file,
```
willm30, line 58: Having access to the original input file is useful for thumbnail optimisation. (see `UserAvatarUploader: handleFileUpload`)

src/utils/fileReader/index.ts
```diff
@@ -21,7 +26,7 @@ const fileReaderFactory = (options: Partial<FileReaderOptions>) => {
     return file && config.maxFileSize && file.size > config.maxFileSize;
   }
 
-  function readFilePromise(file: File): Promise<FileReaderFile | Error> {
+  function readFile(file: File | Blob): Promise<FileReaderFile> {
```
willm30, line 29: Renaming as after adding types to this document, including the return type in the function name is no longer helpful

src/components/shared/FileUpload/SingleFileUpload.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+import { withForwardingRef, ForwardedRefProps } from '~utils/hoc';
+import useDropzoneWithFileReader from '~hooks/useDropzoneWithFileReader';
+import { SingleFileUploadProps } from './types';
+
+import styles from './SingleFileUpload.css';
+
+const displayName = 'SingleFileUpload';
+
+const SingleFileUpload = ({
```
willm30, line 12: I believe this is easier to follow than using the `FileUpload` component, which is tailored more for the CSV uploader than the Avatar uploader.

src/components/common/UserProfileEdit/UserAvatarUploader.tsx
```diff
@@ -1,20 +1,27 @@
...
 
 const MSG = defineMessages({
+  heading: {
+    id: `${displayName}.heading`,
+    defaultMessage: 'Profile Picture',
+  },
   uploaderLabel: {
     id: `${displayName}.uploaderLabel`,
-    defaultMessage: 'At least 250x250px, up to 1MB, .png or .svg',
+    defaultMessage: 'At least 250x250px, up to 280KB, .png or .svg',
```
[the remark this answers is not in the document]
Julianahlk: @willm30 280kb is a bit small and seems like an inconvenience. Couldn't the image be compressed  more so a large initial file size could be used?

src/components/shared/AvatarUploader/AvatarUploader.tsx
```diff
@@ -34,115 +34,146 @@ const ACCEPTED_MIME_TYPES = {
   'image/png': [],
```
Julianahlk, line 34: @willm30  Its always been like this but since you are making changes, is there any reason not to allow uploading jpeg?

Julianahlk (approved):
Had a few comments but overall looks good and working well for me so approving. 💯

willm30:
@Julianahlk I think that's a good suggestion. I've added jpg and webp support, and removed the max file size since we can compress the avatar in the browser before uploading to the db. I've also removed the  label above the avatar since it seems redundant (e.g. is there any reason to say min 250px x 250? This wasn't being enforced in the dapp.)

Julianahlk:
> @Julianahlk I think that's a good suggestion. I've added jpg and webp support, and removed the max file size since we can compress the avatar in the browser before uploading to the db. I've also removed the label above the avatar since it seems redundant (e.g. is there any reason to say min 250px x 250? This wasn't being enforced in the dapp.)

@willm30 after your changes, what happens now is if I attempt to upload a png file that is larger than ~1.1MB I am getting the error "Accepted file types are: svg, jpg, png and webp". However, jpeg files of all sizes seem to work fine (I have tried up to 20MB).

willm30:
@Julianahlk Hmm, I think I'll need to test with the image you're using to see exactly what's happening since I've been testing with a PNG of 10mb and having no problem. I've even resized it to 1.1MB just to be sure and am having no trouble. 

Sometimes the compression actually makes the file larger, which may be the case here. 

If you want you could also go to `UserAvatarUploader` and log out the errors in `onError` on line 40 and `handleFileReject` on line 72. Both get passed the error as the first parameter and these are the only two places an error could be coming from.

src/components/common/UserProfileEdit/UserAvatarUploader.tsx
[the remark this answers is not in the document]
jakubcolony: I agree with Juliana this limit is quite restrictive. In longer term we might want to consider storing static assets in S3, there already seems to be a dedicated Amplify library for this - https://docs.amplify.aws/lib/storage/getting-started/q/platform/js/.

jakubcolony (approved):
Some really good quality work @willm30! 

The only thing that stood out is that if you try to upload a corrupted file (e.g. change the extension of a file from pdf to jpg), there's an uncaught ImageBlobReduce error in the console and the spinner just keeps on going:
```
Uncaught (in promise) Error: ImageBlobReduce: failed to create Image() from blob
```
Maybe we could also handle this case and show a nice error message?

willm30:
@jakubcolony @Julianahlk Thanks both for helping find these edge cases! Given that the upload size error doesn't seem to be an issue with removing the max limit, but instead the unpredictable result of compressing a given image, I've opted to improve the error messaging instead of reimposing the old limit of 1mb. 

For images that fail to compress sufficiently:
toolarge

For images that are the incorrect file type : 
unsupported

And in all other cases: 
corrupt

jakubcolony (approved):
Great stuff, everything works now! Or if it doesn't, it still looks nice:
image
image

Julianahlk:
@willm30 Tested again and getting the same results as Jakub so its all good now. Overall, it seems like jpegs compress much better than png images.

PR #155, opened 2022-12-24
Port `useEnabledExtensions` hook

### LINKED ISSUE #128: Port over `useEnabledExtensions` hook
Blocked by #126

jakubcolony:
## Description

This small PR ports over the `useEnabledExtensions` hook. I spotted an opportunity to extract and encapsulate some logic related to checking whether an extension is enabled so I added a new field `isEnabled` to `InstalledExtensionData`. 

This also brings in a small change to `ExtensionStatusBadge` which helps us avoid weird statuses like the following:
image

From now, there will only be a maximum of one status shown (and deprecated will take precedence over enabled).

## Testing
It doesn't look like anything previously relying on this hook has been ported over, so you'll just have to call it anywhere in the CDapp and log out the results to verify that they're correct. For the `ExtensionStatusBadge`, try to deprecate an enabled extension, and it should show as just "Deprecated".

Resolves #128

### DIFF
--- a/src/components/common/Extensions/ExtensionStatusBadge/ExtensionStatusBadge.tsx
+++ b/src/components/common/Extensions/ExtensionStatusBadge/ExtensionStatusBadge.tsx
@@ -10,10 +10,6 @@ import styles from './ExtensionStatusBadge.css';
 const displayName = 'common.Extensions.ExtensionStatusBadge';
 
 const MSG = defineMessages({
-  installed: {
-    id: `${displayName}.installed`,
-    defaultMessage: 'Installed',
-  },
   notInstalled: {
     id: `${displayName}.notInstalled`,
     defaultMessage: 'Not installed',
@@ -30,8 +26,8 @@ const MSG = defineMessages({
     id: `${displayName}.enabled`,
     defaultMessage: 'Enabled',
   },
-  notEnabled: {
-    id: `${displayName}.notEnabled`,
+  disabled: {
+    id: `${displayName}.disabled`,
     defaultMessage: 'Disabled',
   },
 });
@@ -62,7 +58,8 @@ const ExtensionStatusBadge = ({
     status = MSG.enabled;
     theme = 'primary';
   } else {
-    status = MSG.installed;
+    status = MSG.disabled;
+    theme = 'golden';
   }
 
   const isDeprecated =
--- a/src/components/common/Extensions/ExtensionStatusBadge/ExtensionStatusBadge.tsx
+++ b/src/components/common/Extensions/ExtensionStatusBadge/ExtensionStatusBadge.tsx
@@ -50,28 +50,29 @@ const ExtensionStatusBadge = ({
 
   if (!isInstalledExtensionData(extensionData)) {
     status = MSG.notInstalled;
-  } else if (!extensionData.isInitialized) {
-    status = MSG.notEnabled;
-    theme = 'golden';
+  } else if (extensionData.isDeprecated) {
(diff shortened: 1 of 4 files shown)


src/hooks/useEnabledExtensions.tsx
```diff
@@ -1,56 +1,25 @@
...
-//       : null,
-//     isOneTxPaymentExtensionEnabled,
-//     installedExtensionsAddresses,
-//     isLoadingExtensions: loading,
-//   };
-// };
-//
-
-const useEnabledExtensions = (...args) => args;
+  return {
```
willm30, line 18: May I suggest this be generalised, so we needn't manually copy / paste the above each time we add a new extn?

E.g.
``` js
const enabledExtns = installedExtensionsData.reduce(
    (extns, extn) => ({
      ...extns,
      [`is${extn.extensionId}Enabled`]: extn.isEnabled,
    }),
    {},
  );

  return {
    loading,
    ...enabledExtns,
  };

File: src/components/common/Extensions/ExtensionStatusBadge/ExtensionStatusBadge.tsx [hunk 2]
```diff
@@ -50,28 +50,29 @@ const ExtensionStatusBadge = ({
 
   if (!isInstalledExtensionData(extensionData)) {
     status = MSG.notInstalled;
-  } else if (!extensionData.isInitialized) {
-    status = MSG.notEnabled;
```
willm30, line 54: Just checking you intended to remove the "Disabled" tag? I believe it's supposed to represent the state after installation but before it's been enabled. 

If so, the message can be removed from the MSG object above.

jakubcolony: Good spot. Installed and disabled feel very similar but it might make more sense to make disabled the default status

src/hooks/useEnabledExtensions.tsx
jakubcolony: I'm just wondering if doing this won't cause more trouble down the line.
You're definitely right that we wouldn't have to modify this file when adding new supported extension but the way these are set up still requires doing quite a bit of copy pasting in other places.
My concern is that we'll lose strong typing, in your example if you try to access `enabledExtns.isOneTxPaymentEnabled` you will get the following type error: `Property 'isOneTxPaymentEnabled' does not exist on type '{}'` so I guess we would have to type it as `Record<string, boolean>` at best which I think isn't ideal. It also feels like we will be using this hook more often than we will be adding new extensions, so what are your thoughts on that?

willm30: @jakubcolony How's this?

``` js
import { Extension } from '@colony/colony-js';

type Extensions = `${Extension}`;
type EnabledExtensions = Partial<Record<Extensions, boolean>>
```
then

``` js
     ...
      [`is${extn.extensionId}Enabled`]: extn.isEnabled,
    }),
    {} as EnabledExtensions,
  );
```
That's probably the most correct way to handle the typing?

jakubcolony: I like your approach, thanks for the suggestion

willm30 (approved):
Nice one, especially like the removal of double statuses like "enabled deprecated"

Julianahlk (commented):
Overall looks good, just missing the disabled tag after installing like @willm30 already mentioned

Julianahlk (approved):

PR #156, opened 2023-01-03
Fix `/edit-profile` loading state

### LINKED ISSUE #134: Fix the loading states for the user profile edit
As per discussion in #79

willm30:
## Description

This PR fixes [the problem](https://github.com/JoinColony/colonyCDapp/pull/79#issuecomment-1341046043) with the loading state at `/edit-profile`.

## Testing

I have covered the following scenarios:

1. Navigate to /edit-profile with no wallet connected. --> redirects to /landing
2. Connect wallet to a wallet with a registered user, then navigate to edit-profile. --> displays Profile Edit screen
3. While on edit-profile, refresh the browser. --> shows loading spinner, then displays Profile Edit screen
4. Disconnect wallet while at edit-profile --> redirects to /landing
5. Navigate to edit-profile with a wallet that doesn't have a user associated. --> shows loading spinner, then redirects to /landing

Please double-check each and any others you can think of.

**New stuff** ✨

* `useCanEditProfile.ts` to manage loading states and redirect to landing

Resolves #134

### DIFF
--- a/src/Entry.tsx
+++ b/src/Entry.tsx
@@ -5,8 +5,8 @@ import { IntlProvider } from 'react-intl';
 import { ApolloProvider } from '@apollo/client';
 
 import layout from '~styles/layout.css';
-import { DialogProvider } from '~shared/Dialog';
 import '~utils/yup/customMethods'; // ensures custom yup methods are available when components load
+import { DialogProvider } from '~shared/Dialog';
 // import { TokenActivationProvider } from '~users/TokenActivationProvider';
 
 import messages from './i18n/en.json';
--- a/src/components/common/ColonyHome/ColonyMembersWidget/MembersSubsection.tsx
+++ b/src/components/common/ColonyHome/ColonyMembersWidget/MembersSubsection.tsx
@@ -160,7 +160,6 @@ const MembersSubsection = ({
                 // banned={canAdministerComments && banned}
                 banned={false}
                 showInfo
-                notSet={false}
                 colony={colony}
                 user={user}
                 popperOptions={{
--- a/src/components/common/UserProfileEdit/UserAvatarUploader.tsx
+++ b/src/components/common/UserProfileEdit/UserAvatarUploader.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useState } from 'react';
 import { defineMessages } from 'react-intl';
 
 import AvatarUploader from '~shared/AvatarUploader';
@@ -6,7 +6,6 @@ import UserAvatar from '~shared/UserAvatar';
 
 import { User } from '~types';
 import { useUpdateUserProfileMutation } from '~gql';
-import { useAppContext } from '~hooks';
 import { FileReaderFile } from '~utils/fileReader/types';
 
 const displayName = 'common.UserProfileEdit.UserAvatarUploader';
@@ -27,7 +26,6 @@ const UserAvatarUploader = ({
   user,
   user: { walletAddress, profile },
 }: Props) => {
-  const { updateUser } = useAppContext();
   const [avatar, setAvatar] = useState<string | null>(profile?.avatar || null);
   const [updateAvatar, { error }] = useUpdateUserProfileMutation();
   const updatedUser = {
@@ -57,12 +55,6 @@ const UserAvatarUploader = ({
     });
     setAvatar(null);
   };
-
-  // on dismount
-  useEffect(
-    () => () => updateUser?.(user.walletAddress),
-    [updateUser, user.walletAddress],
-  );
 
   return (
     <AvatarUploader
--- a/src/components/common/UserProfileEdit/UserMainSettings.tsx
+++ b/src/components/common/UserProfileEdit/UserMainSettings.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useState } from 'react';
 import { defineMessages } from 'react-intl';
 import { string, object, InferType } from 'yup';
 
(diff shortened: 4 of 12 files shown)


src/Entry.tsx
```diff
@@ -5,8 +5,8 @@ import { IntlProvider } from 'react-intl';
 import { ApolloProvider } from '@apollo/client';
 
 import layout from '~styles/layout.css';
-import { DialogProvider } from '~shared/Dialog';
 import '~utils/yup/customMethods'; // ensures custom yup methods are available when components load
+import { DialogProvider } from '~shared/Dialog';
```
willm30, line 9: Ordering matters here to avoid circular module dependencies

src/hooks/index.ts
```diff
@@ -34,6 +34,7 @@ export { default as useTitle } from './useTitle';
 export { default as useTokenInfo, TokenInfoProvider } from './useTokenInfo';
 export { default as useUserAvatarImageFromIPFS } from './useUserAvatarImageFromIPFS';
 export { default as useUserSettings, SlotKey } from './useUserSettings';
+export { default as useCanEditProfile } from './useCanEditProfile';
```
willm30, line 37: Order matters here, needs to go after './useUserSettings' so `SlotKey` is defined

src/components/common/UserProfileEdit/UserMainSettings.tsx
```diff
@@ -94,15 +94,10 @@ const UserMainSettings = ({
         },
       },
     });
+
+    updateUser?.(walletAddress, true);
   };
 
-  useEffect(
```
willm30, line 99: With the addition of a "background update" option, updating the user on dismount is no longer necessary and can be performed in the form submit handler.

src/context/AppContext.tsx
```diff
@@ -47,34 +49,37 @@ export const AppContextProvider = ({ children }: { children: ReactNode }) => {
...
-        query.then(({ data }) => {
+  const updateUser = useCallback(
+    async (address?: string, shouldBackgroundUpdate = false) => {
+      if (address) {
+        try {
+          if (!shouldBackgroundUpdate) {
+            setUserLoading(true);
+          }
+          const apolloClient = getContext(ContextModule.ApolloClient);
+          const { data } = await apolloClient.query<
```
willm30, line 60: Using `async / await` here instead of  `.then` forces React to pause code execution until the query has completed, which stops React from batching the `setUserLoading` calls. This ensures `userLoading` actually gets set to true, which wasn't happening before.

src/context/AppContext.tsx
```diff
@@ -47,34 +49,37 @@ export const AppContextProvider = ({ children }: { children: ReactNode }) => {
...
-          GetCurrentUserQuery,
-          GetCurrentUserQueryVariables
-        >({
-          query: GetCurrentUserDocument,
-          variables: { address },
-          fetchPolicy: 'network-only',
-        });
-        query.then(({ data }) => {
+  const updateUser = useCallback(
+    async (address?: string, shouldBackgroundUpdate = false) => {
```
willm30, line 53: Per the comment below, now that `userLoading` is being set to true as expected, it's useful that we have some way of performing context updates in the "background", i.e. without prompting the display of a loading spinner. E.g when updating user settings or uploading a user avatar.

src/context/AppContext.tsx
```diff
@@ -37,7 +40,6 @@ export const AppContextProvider = ({ children }: { children: ReactNode }) => {
   try {
     initialWallet = getContext(ContextModule.Wallet);
   } catch (error) {
-    initialUser = null;
```
willm30, line 40: If this is set to `null` here, user will never be `undefined` and thus we lose the distinction between a wallet without a registered user (`null`) and the state in which we're checking if there's a registered user (`undefined`)

Julianahlk:
@willm30 Overall looks good and all the scenarios you mentioned worked well for me. 

Only the I notice is that when you are redirected to the landing page after logging out from /edit-profile, the subscribed colonies list is still showing and the user avatar still has the thumbnail of the previously logged in user. If you logout from other routes these are cleared.

willm30:
@Julianahlk Thanks, I had the relevant lines in my other PR as was working on these together and didn't realise but have moved over to here.

src/components/common/ColonyHome/ColonyMembersWidget/MembersSubsection.tsx
```diff
@@ -160,7 +160,6 @@ const MembersSubsection = ({
                 // banned={canAdministerComments && banned}
                 banned={false}
                 showInfo
-                notSet={false}
```
willm30, line 163: UserAvatar had notSet default to true, even though more or less everywhere it was used it was being passed a false prop, like here. I have removed the default and removed these props where they were used.

src/components/common/UserProfileEdit/UserAvatarUploader.tsx
```diff
@@ -58,12 +56,6 @@ const UserAvatarUploader = ({
     setAvatar(null);
   };
 
-  // on dismount
-  useEffect(
```
willm30, line 62: Effect no longer necessary

src/components/frame/AvatarDropdown/AvatarDropdown.tsx
```diff
@@ -97,8 +97,8 @@ const AvatarDropdown = ({ preventTransactions = false, spinnerMsg }: Props) => {
           <UserAvatar
             address={wallet?.address || ''}
             size="s"
-            notSet={false}
             user={user}
+            notSet={!wallet}
```
willm30, line 101: Ensures notSet is true when user is logged out.

jakubcolony (approved):
Great to see a solution to this @willm30! All working well for me 👍

willm30:
@Julianahlk Are you able to re-review when you get a moment? Thank you

Julianahlk (approved):
Everything working now 👍 Looks good.

PR #159, opened 2023-01-06
Implement Custom Rpc Validation

### LINKED ISSUE #145: Implement custom rpc endpoint validation
Currently we have a `@TODO` in the `onClick` method of the `validate` button to add validation:

Image

Creating this issue for visibility.

TODO:

-- How are we planning to validate the url?

willm30:
## Description

This PR implements validation for a custom gnosis chain rpc in the user advanced settings. 

## Testing

Go to user advanced settings. Enter a url. Only a valid gnosis  rpc endpoint will validate (you can find some at [chainlist](https://chainlist.org/chain/100)) 

If URL is not a valid RPC: 

unable

If RPC is not gnosis:

wrongid

**New stuff** ✨

* Added gnosis chain rpc validation

**Deletions** ⚰️

* Removed the validate button since validation normally occurs automatically.

Resolves #145

### DIFF
--- a/src/components/common/UserProfileEdit/UserAdvancedSettings.tsx
+++ b/src/components/common/UserProfileEdit/UserAdvancedSettings.tsx
@@ -11,11 +11,13 @@ import useUserSettings, {
   UserSettingsHook,
 } from '~hooks/useUserSettings';
 import { canUseMetatransactions } from '~utils/checks';
+import { yupDebounce } from '~utils/yup/tests';
 
 import AdvancedSettingsRow, {
   getAdvancedSettingsRows,
 } from './AdvancedSettingsRow';
 import SaveForm from './SaveForm';
+import { isValidURL, validateCustomGnosisRPC } from './validation';
 
 import styles from './UserAdvancedSettings.css';
 
@@ -26,9 +28,13 @@ const MSG = defineMessages({
     id: `${displayName}.heading`,
     defaultMessage: 'Advanced settings {learnMoreLink}',
   },
-  rpcErrorMessage: {
-    id: `${displayName}.rpcErrorMessage`,
+  invalidURLError: {
+    id: `${displayName}.invalidURLError`,
     defaultMessage: 'Enter a valid URL',
+  },
+  invalidRPCError: {
+    id: `${displayName}.invalidRPCError`,
+    defaultMessage: 'Unable to validate RPC endpoint',
   },
 });
 
@@ -39,7 +45,17 @@ const validationSchema = object({
     .defined()
     .when(`${[SlotKey.DecentralizedMode]}`, {
       is: true,
-      then: string().url(() => MSG.rpcErrorMessage),
+      then: string()
+        .required(() => MSG.invalidURLError)
+        .url(() => MSG.invalidURLError)
+        .test(
+          'gnosisRpc',
+          () => MSG.invalidRPCError,
+          yupDebounce(validateCustomGnosisRPC, 200, {
+            isOptional: false,
+            circuitBreaker: isValidURL,
+          }),
+        ),
     }),
 }).defined();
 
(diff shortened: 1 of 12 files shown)


willm30:
@rdig Before I publish this pr for review, can I confirm this is what you have in mind by gnosis chain rpc validation? We're validating that the rpc url is legitimate, right? To do this I'm using the [chainList list](https://chainlist.org/chain/100), but let me know if there's a better dataset or if I've misunderstood the nature of the validation we're implementing.

willm30:
@rdig Just following up re the intended nature of custom rpc validation. I'm querying the chainlist dataset in this branch but let me know if validation here refers to something else or there's a better dataset to use.

rdig:
Hmm. I can't remember what this is all about. Is this about the RPC input on the user's advanced page ?

willm30:
@rdig Yes, there was a "Validate" button beneath the url input, so I assumed there was some plan to perform additional validation on the inputted rpc url (aside from the basic valid url check). 

If that's not the case I can just remove and close the pr.

rdig:
I think this should be just a "simple" url check, basically make a `eth_version` call and see if it returns.

This is because RPC urls are intended to be local nodes _(ie: 127.0.0.1:8545)_, so we can't really validate them against a set list.

src/components/shared/Fields/InputStatus/HookForm/InputStatus.tsx
```diff
@@ -45,8 +44,7 @@ const HookFormInputStatus = ({
     <Element
       className={getMainClasses(appearance, styles, {
         error: !!error && !isLoading,
-        hidden:
-          (!text && !isLoading) || (!!error && !isNil(touched) && !touched),
```
willm30, line 49: This `isNil` was causing the status to appear before the field had been touched.

src/components/common/UserProfileEdit/validation.ts
```diff
@@ -0,0 +1,46 @@
...
+    return true;
+  } catch {
+    return false;
+  }
+};
+
+async function validateCustomRPC(rpcURL: string, expectedChainId: string) {
+  const provider = new ethers.providers.JsonRpcProvider(rpcURL);
+  try {
+    const chainId = await provider.send('net_version', []);
```
willm30, line 29: We request the network id from the rpc url given by the user.

If it's either an invalid rpc or on the wrong chain, we return an error message. Else we validate the url successfully.

willm30:
@rdig I've gone with `net_version` because it gives us information not only about whether the url is an rpc, but also if it's on the correct chain.

src/components/common/UserProfileEdit/UserAdvancedSettings.tsx
```diff
@@ -26,20 +28,34 @@ const MSG = defineMessages({
...
+      then: string()
+        .required(() => MSG.invalidURLError)
+        .url(() => MSG.invalidURLError)
+        .test(
+          'gnosisRpc',
+          () => MSG.invalidRPCError,
+          yupDebounce(validateCustomGnosisRPC, 200, {
+            isOptional: false,
+            circuitBreaker: isValidURL,
+          }),
```
rdig, line 57: I'm good with this, at least for now. But we'll most likely have to change it prior to going live.

At the very least we'll need to be able to validate Ganache for dev mode _(which might have any chain id -- we can agree on a "proper" one)_.

Later, it will need to support basically every chain the CDapp will get deployed to

rdig (approved):
I'm good with it now, however I'd like us to be able to validate our Ganache dev chain as well.

jakubcolony (approved):
All working well, including debouncing - nice work!

Julianahlk (approved):
Nice implementation! Validation is working well:
Screen Shot 2023-01-26 at 2 44 34 PM
Screen Shot 2023-01-26 at 2 46 40 PM
Screen Shot 2023-01-26 at 2 44 23 PM

src/components/common/UserProfileEdit/UserAdvancedSettings.tsx
willm30: Ok, opened an issue to keep track: https://github.com/JoinColony/colonyCDapp/issues/230

PR #160, opened 2023-01-09
Add `refreshEnabledExtension` util

### LINKED ISSUE #152: Add refreshEnabledExtension util function
Similarly to util functions added in #147, we'll need one for enabled extensions. 

Marking as blocked by #130

jakubcolony:
## Description

This small PR adds a `refreshEnabledExtension` util, similarly to other functions from #147 

For testing, try to enable `VotingReputation`. You should see the status updated immediately. Please note, the status will revert back to Disabled when you refresh unless you follow the first two steps from [block-ingestor#3](https://github.com/JoinColony/block-ingestor/pull/3). Otherwise, the block ingestor simply won't pick up the enable event and won't update the db.

Resolves #152

### DIFF
--- a/src/redux/sagas/extensions/extensionEnable.ts
+++ b/src/redux/sagas/extensions/extensionEnable.ts
@@ -72,7 +72,7 @@ function* extensionEnable({
     return yield putError(ActionTypes.EXTENSION_ENABLE_ERROR, error, meta);
   }
 
-  yield call(refreshEnabledExtension, colonyAddress, extensionId);
+  refreshEnabledExtension(colonyAddress, extensionId);
 
   initChannel.close();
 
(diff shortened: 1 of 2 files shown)


src/redux/sagas/extensions/extensionEnable.ts
```diff
@@ -71,7 +72,7 @@ function* extensionEnable({
     return yield putError(ActionTypes.EXTENSION_ENABLE_ERROR, error, meta);
   }
 
-  // yield call(refreshExtensions);
+  yield call(refreshEnabledExtension, colonyAddress, extensionId);
```
willm30, line 75: I notice you dropped the `yield call(...)` pattern when doing the refresh in `extensionInstall.ts`. Not certain if there's a preference for one over the other (I'm used to seeing it this way) but should probably keep them consistent.

willm30 (approved):
Nice, working for me 👍

src/redux/sagas/extensions/extensionEnable.ts
jakubcolony: Good spot, fixed!

Julianahlk (approved):
Looking Good 👍

Issue #178, opened 2023-01-11
Refactor NewAction and NewDecision buttons to reuse logic


willm30:
When the new action button has been wired in, the new decision button can borrow its logic instead of copy / pasting it.

willm30:
Blocked by the work on actions

Issue #96, opened 2022-11-23
Standardise code formatting 


jakubcolony:
Currently we don't have a standardised approach towards code formatting.
In both Dapp and CDapp, we rely on `.editorconfig` and eslint rules (which is actually discouraged: https://typescript-eslint.io/docs/linting/troubleshooting/formatting), in block-ingestor there are just eslint rules which also seem to be different than the ones in the CDapp.
We don't have a project specific prettier config which is one popular solution.
This issue is to investigate and provide a better code formatting in the CDapp and block-ingestor repositories. The way this should be done is that there's no manual action from the developer required to fix those and code that doesn't adhere to the rules is prevented from being pushed.
@rdig fyi, feel free to comment

willm30:
FYI, I've been using a `.prettierrc` file with the following:

```json
{
  "singleQuote": true,
  "trailingComma": "all"
}
```
and a `.prettierignore` with:

```
*.css
*.graphql
*.json
```
Which does the trick.

jakubcolony:
Hey @willm30, my setup is very similar, I think it does a good job together with prettier defaults regarding line lengths, etc. We'd just have to make sure it doesn't conflict with the existing linting.

I also have a similarly looking `.prettierignore` but the reason I set it up is because without it I was introducing many formatting changes to existing code, maybe this is also something to have a look into and make `.graphql` and `.json` files compatible with whatever formatting setup we end up with?

PR #180, opened 2023-01-11
Wire in New Decision modal

### LINKED ISSUE #167: Wire in ColonyNewDecisions
- [x] Clicking New Decision should open the modal

newdecisions

willm30:
**Note** Actual diff is about 1,000. 850 is package-lock and 200 is prettifying icons.json.

## Description

This PR wires in the NewDecisionButton such that the New Decision Modal with Rich Text editor appears at `/decisions`.

decisionmodal

Note that the preview route is yet to be wired in, clicking "preview" takes you to a blank screen. Go back to /decisions, and click New Decision again. Since nothing else has been wired in yet, you should just see your saved decision. If you want to remove this, you need to remove it directly in local storage (awaiting #188).

## Testing

When testing, I've been comparing to the new decisions modal currently live on xdai.colony.io. Please double check validation and that all the toolbar features work correctly. 

The domain selector is yet to be implemented so that will come later. 

And there's a bug with tiptap's placeholder component  (#185), so there is currently no placeholder in the text area.  

**New stuff** ✨

* New Decision Modal, Rich Text Editor

Resolves #167


src/components/common/ColonyDecisions/NewDecisionButton/NewDecisionButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+const NewDecisionButton = ({ ethDomainId }: Props) => {
+  const { user } = useAppContext();
+  const { editor } = useRichTextEditor();
+
+  //   const { isVotingExtensionEnabled, isLoadingExtensions } =
+  //     useEnabledExtensions({
+  //       colonyAddress: colony.colonyAddress,
+  //     });
+
+  // @TODO: This is copied from NewActionButton, extract instead.
```
willm30, line 26: #178

src/hooks/useRichTextEditor.ts
```diff
@@ -0,0 +1,41 @@
+import { Extension, useEditor, Node } from '@tiptap/react';
+// @TODO: Investigate fix...
```
willm30, line 2: Currently unable to import `Placeholder` component from tiptap due to an internal key naming conflict with tiptap core. #185

src/images/icons.json
```diff
@@ -1,114 +1,117 @@
...
-    "Alessandra Aphid",
-    "Terry Termite",
-    "Ava Ant",
-    "Bong Joon Bee",
-    "Werner Wasp",
-    "Berthold Beetle",
-    "Sanjeet Shrimp",
-    "Norman Naked Mole Rat"
-  ]
+    "icons": [
```
willm30, line 2: Related to #96, this is what my prettier does to this file and we should probably commit a simple prettier config to the repo so it doesnt get changed back by accident.

src/utils/decisions.ts
```diff
@@ -0,0 +1,22 @@
+import { DecisionDialogValues } from '~common/ColonyDecisions/DecisionDialog/DecisionDialog';
+import { Address, Decision } from '~types';
+
+const getLocalStorageDecisionKey = (walletAddress: Address) =>
+  `decision:${walletAddress}`;
```
willm30, line 5: Having a key that contains the wallet address saves us having to implement logic to detect whether the current user is the same user as the user that stored the decision. We will just save one decision per user, and retreive that decision if the user wants to make an edit to their decision in decision preview.

src/components/common/ColonyDecisions/NewDecisionButton/NewDecisionButton.tsx
```diff
@@ -1,28 +1,30 @@
...
+const displayName = 'common.ColonyDecisions.NewDecisionButton';
 
 interface Props {
   ethDomainId: number;
 }
 
 const NewDecisionButton = ({ ethDomainId }: Props) => {
-  const { colony } = useColonyContext();
   const { user } = useAppContext();
+  const { editor } = useRichTextEditor();
```
willm30, line 19: Instantiating the editor here and passing it down ensures it's defined and available before the dialog opens. This prevents the dialog from opening before the editor object is available, which results in a jiggery ui.

src/components/common/ColonyDecisions/NewDecisionButton/NewDecisionButton.tsx
```diff
@@ -33,32 +35,33 @@ const NewDecisionButton = ({ ethDomainId }: Props) => {
...
   return (
     <DialogButton
-      loading={false /*isLoadingData*/}
+      loading={false /* isLoadingData */}
       text={{ id: 'button.newDecision' }}
-      handleClick={() => console.log('Wire up New Decision Button')}
-      //   disabled={
+      handleClick={handleClick}
+      disabled={!user}
       //     mustUpgrade ||
```
willm30, line 65: #189

src/components/shared/RichTextEditor/RichTextEditor.tsx
```diff
@@ -0,0 +1,85 @@
...
+}: RichTextEditorProps) => {
+  const {
+    getFieldState,
+    setValue,
+    formState: { isSubmitting },
+  } = useFormContext();
+  const { error: contentError, isTouched: contentTouched } =
+    getFieldState(name);
+
+  useEffect(() => {
```
willm30, line 40: Tiptap does not expose the editors state directly. Instead, you need to hook into events that the base class emits. 

We hook into the "selectionUpdate" event which fires whenever the editor content changes, and also the "blur" event.

This needs to be done inside an effect to ensure clean up happens on dismount and also to avoid setting the form state (via setValue) while the `RichTextEditor` is being rendered.

src/types/definitions/index.ts
```diff
@@ -0,0 +1,14 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
+
+interface Brand {
+  brand: string;
+  version: string;
+}
+
+interface Navigator {
+  userAgentData?: {
```
willm30, line 9: For the `cmd` or `ctrl` keyboard shortcut check in `toolbarConfig`

jakubcolony (approved):
Nice work!
I've got one question: Currently, if you enter some decision details and go to the Preview, then go back and open the dialog again, you see the whatever details you previously entered. However, the Update button is disabled until you make some changes. Does that mean there's no way of going back to the Preview without making any changes?

willm30:
@jakubcolony When this gets wired in you'll be able to get back to the preview via the modal: https://github.com/JoinColony/colonyCDapp/pull/198

willm30:
There's also the list item #200 which shows up when you have a draft

src/images/icons/codeBlock.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="17" height="17" viewBox="0 0 17 17" fill="none" xmlns="http://www.w3.org/2000/svg">
```
ArmandoGraterol, line 1: A question, in the Dapp we always tried to use icons with a viewbox of `0 0 30 30`, has that changed in the CDapp? In the icons added in this PR I see that there are a few with `0 0 20 20` and others like this one that are different.

src/utils/decisions.ts
```diff
@@ -0,0 +1,22 @@
+import { DecisionDialogValues } from '~common/ColonyDecisions/DecisionDialog/DecisionDialog';
```
ArmandoGraterol, line 1: We should add `DecisionDialogValues` to the `index` file of the folder, to make this import cleaner

ArmandoGraterol (approved):
That's a proper dialog, innit?

I just have a question regarding the icons, but as far as I know the validation and functionality of the dialog is working as expected

src/components/common/ColonyDecisions/DecisionDialog/DecisionControls.css
```diff
@@ -0,0 +1,3 @@
+.main button[type=submit] {
```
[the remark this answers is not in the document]
willm30: The way that component is set up, adding class names to `Button` overrides the appearance object.

src/components/common/ColonyDecisions/DecisionDialog/DecisionDialog.tsx
```diff
@@ -0,0 +1,103 @@
...
+    close();
+  };
+
+  return (
+    <Dialog cancel={cancel}>
+      <Form<DecisionDialogValues>
+        defaultValues={{
+          motionDomainId: draftDecision?.motionDomainId ?? domainId,
+          title: draftDecision?.title,
+          description: draftDecision?.description || '<p></p>',
```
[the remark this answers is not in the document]
willm30: I have added trim to the title. Trimming the description would require me to implement a custom yup method because we're not dealing with plain text. We're storing the decision text as an html string, which when empty defaults to `<p></p>`. Empty spaces thus might look like `<p>  </p>` and so the built in trim would see nothing wrong.

Going to leave as is for the time being.

src/images/icons/codeBlock.svg
willm30: These are taken from the dapp. I think if it's important, we should create a separate issue for enforcing the same viewbox on all svgs in the repo as there are a few exceptions that've been ported over from the dapp. 

What's the rationale for wanting them all to be `0 0 30 30`, btw?

ArmandoGraterol: As far as I remember, the reason for all of them to be like that was to make sure they are all consistently formatted the same way

Issue #196, opened 2023-01-19
Wire in `DecisionNotFound`


willm30:
DecisionNotFound contains a button to create a new decision. Needs wiring in.

willm30:
MERGED #197

Issue #193, opened 2023-01-18
Wire in "Edit" decision


willm30:
On the decision preview. Base off #191

willm30:
MERGED #197

PR #197, opened 2023-01-19
Wire in Decision Preview Controls

### LINKED ISSUE #169: Wire in "Delete" Decision
- [x] Clicking "delete" (see #168) shows this modal:
- [ ] Clicking the "Delete" button in the modal removes the preview held in localstorage and returns the user to `colony/.../decisions`

deleteDecision

willm30:
## Description

This PR wires in the delete and edit decision modals, and well as the "Create new decision" button at `/preview` when there is 
no stored decision. "Publish" is being handled separately.

Note: It's worth reviewing https://github.com/JoinColony/colonyCDapp/pull/201 first, since it updates the state-related logic introduced here to account for the DraftDecisionItem.

Delete decision

deleteme

Create new decision

no-decision-saved

Edit existing decision

ddetails

## Testing 

Clicking delete should: 
1. Open the delete decision modal
2. Clicking "Delete" in the modal should: 
3. Navigate to `/decisions` and remove the decision from local storage.

Click edit should: 
1. Open the DecisionDialog
2. Making an update should close the modal and update the preview. 
3. If you go back to the main `/decisions` page, open the dialog, and submit it, you should still be redirected to `decisions/preview`. 

Click create new decision should:
1. Open a clean modal
2. Submitting the dialog should update decisions preview.

**New stuff** ✨

* DeleteDecisionModal
* Wire in Edit Decision
* Wire in "Create new decision"

Resolves  #169 , #193, #196

### DIFF
--- a/src/components/common/ColonyDecisions/DecisionPreview/DecisionPreview.tsx
+++ b/src/components/common/ColonyDecisions/DecisionPreview/DecisionPreview.tsx
@@ -1,10 +1,11 @@
-import React, { useEffect, useMemo, useState } from 'react';
+import React, { useState } from 'react';
 import { defineMessages } from 'react-intl';
 
 import Tag from '~shared/Tag';
 import { getDecisionFromLocalStorage } from '~utils/decisions';
 import LoadingTemplate from '~frame/LoadingTemplate';
 import { useAppContext } from '~hooks';
+import { Decision } from '~types';
 
 import {
   DecisionData,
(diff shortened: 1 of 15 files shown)


src/components/common/ColonyDecisions/DecisionPreview/DecisionPreview.tsx
```diff
@@ -32,8 +33,13 @@ const DecisionPreview = () => {
   const { user, walletConnecting, userLoading } = useAppContext();
   const walletAddress = user?.walletAddress || '';
 
+  const [decisionPreview, setDecisionPreview] = useState<Decision>();
   const decision = getDecisionFromLocalStorage(walletAddress);
 
+  if (!decisionPreview && decision) {
+    setDecisionPreview(decision);
+  }
```
jakubcolony, line 41: I always thought you're supposed to only set state inside callbacks or `useEffect`. While this doesn't trigger render loop because of the condition, I think it might be better to put this inside a `useEffect` - see https://stackoverflow.com/a/74036012

jakubcolony (approved):
LGTM 👍

src/components/common/ColonyDecisions/DecisionPreview/DecisionPreview.tsx
willm30: @jakubcolony I change this in a subsequent pr to use redux so all the decisions components can access the same state. 

However, on the subject of setting state, I'm using a pattern recommended by the React docs: https://beta.reactjs.org/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

Also enjoyed [this talk](https://www.youtube.com/watch?v=bGzanfKVFeU&t) which is more or less covers what's included in the link above (i.e when to use useEffect and how to use it less).

rdig (approved):
Yep, this one's good as well.

But this should have included the redux/dispatch changes from #201 as it would have made review more pertinent

src/components/common/ColonyDecisions/DecisionPreview/DecisionPreview.tsx
jakubcolony: @willm30 Thanks for sharing, looks like I'll have some reading for the weekend!

PR #199, opened 2023-01-19
Add refreshUpgradedExtension util function

### LINKED ISSUE #151: Add refreshUpgradedExtension util function
Similarly to util functions added in #147, we'll need one for upgraded extensions. 

Marking as blocked by #132 and #148

jakubcolony:
## Description

Small PR, adds a refreshUpgradedExtension function similar to all other refresh functions.

Testing is tricky, since we still haven't figured out how to install a newer version of extension in the network (I wasted some time on it yesterday with no luck), so I think we'll just have to "trust" that it works and test it comprehensively when testing the main extensions branch.

Resolves #151


Julianahlk (approved):
Looks good 👍

ArmandoGraterol (approved):
Looks like in theory, it should work :crossed_fingers:

PR #201, opened 2023-01-20
Port draft decision list item

### LINKED ISSUE #200: Wire in Draft Decision List Item
When there is a draft decision, the corresponding list item should appear in the decisions list.

Image

willm30:
## Description

This PR wires in the DraftDecisionItem. It should be present whenever you have a draft decision saved.

decision-item

It also moves decision state to the redux store so we don't have many layers of passing component state to the various components that require access to the current decision draft. The NewDecisionButton, DraftDecisionItem and the DecisionPreview all need access to the state of the current draft, and moving it to redux provides a clean way for them all to access it (note that these components are not neatly colocated, making props unsuitable). 

## Testing

When you have a draft decision saved, it should show up as the draft decision item in `ColonyDecisions`. All aspects of this should work as expected, e.g. edit, delete, the avatar icon, onClick etc.

**New stuff** ✨

* `DraftDecisionItem`
* `ListItem` base component which extracts the view from `ActionListItem` for use in `DraftDecisionItem`. Haven't refactored ActionListItem as it's out of scope for this pr but expect `ListItem` to be used there also.

**Changes** 🏗

* Moved Draft decision state to redux store

**Deletions** ⚰️

* We had this old component `ItemsList` that isn't being used anywhere in the dapp, so have removed.

Resolves  #200


src/components/common/ColonyDecisions/DecisionDialog/DecisionDialog.tsx
```diff
@@ -48,27 +48,27 @@ const validationSchema = object()
...
   const navigate = useNavigate();
   const domainId = getDomainId(ethDomainId);
   const walletAddress = user?.walletAddress || '';
-  const draftDecision = getDecisionFromLocalStorage(walletAddress);
-  const content = draftDecision?.description;
+  const content = decision?.description;
 
   const handleSubmitDialog = (values: DecisionDialogValues) => {
-    handleSubmit?.(values);
+    dispatch({ type: ActionTypes.DECISION_DRAFT_CREATED, payload: values });
```
willm30, line 71: Dispatch an action to the store when we create a decision. This tells redux to store the decision so that it's available to the entire app.

src/components/common/ColonyDecisions/DecisionPreview/DecisionPreview.tsx
```diff
@@ -30,21 +33,10 @@ const MSG = defineMessages({
 
 const DecisionPreview = () => {
   const { user, walletConnecting, userLoading } = useAppContext();
-  const walletAddress = user?.walletAddress || '';
-
-  // Referential equality needed here to avoid infinite loop in effect.
-  const decision = useMemo(
-    () => getDecisionFromLocalStorage(walletAddress),
-    [walletAddress],
+  const decision = useSelector<CoreDecisionsRecord, Decision | undefined>(
```
willm30, line 36: The decision can now be consumed by another part of the app. Note here that sharing component state would be very awkward because it would need to live in the highest ancestor of all the decision components, i.e. our `Routes` file, and be passed down many levels to many components.

src/components/common/ColonyDecisions/DraftDecisionItem/DraftDecisionActions.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const openDeleteDecisionDialog = useDialog(DeleteDecisionDialog);
+  const openDecisionDialog = useDialog(DecisionDialog);
+
+  return (
+    <div className={styles.buttonContainer}>
+      <Button
+        appearance={{ theme: 'blue' }}
+        text={{ id: 'button.delete' }}
+        onClick={(e) => {
+          e.stopPropagation();
```
willm30, line 25: Prevents the click event on the parent from firing (i.e. the redirect to preview)

src/components/common/ColonyDecisions/DraftDecisionItem/DraftDecisionItem.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const redirectToPreview = () =>
+    navigate(`${pathname}${PREVIEW_ROUTE_SUFFIX}`);
+
+  if (!decision) {
+    return null;
+  }
+
+  return (
+    <div className={styles.draftDecision}>
+      <ListItem
```
willm30, line 51: This `ListItem` component is an initial attempt to extract the view of the base "ListItem" from the accompanying logic associated with `ActionListItem`. I expect it will be modified should it become the base component of `ActionListItem` also (which could really do with a refactor), but this is a start.

src/components/shared/ListItem/ListItem.tsx
```diff
@@ -0,0 +1,117 @@
+import React, { ReactNode } from 'react';
```
willm30, line 1: As mentioned above, the intention is here that when the time comes, this component also becomes the basis of `ActionListItem` as well as `DraftDecisionItem`

src/redux/reducers/decisions.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { ActionTypes } from '../actionTypes';
+
+const coreDecisionsReducer: ReducerType<CoreDecisionsRecord> = (
+  state = CoreDecisions(),
+  action,
+) => {
+  switch (action.type) {
+    case ActionTypes.DECISION_DRAFT_CREATED:
+      return state.setIn(
```
willm30, line 14: This will set one decision per wallet address to a list of draft decisions being held by the store.

src/redux/reducers/decisions.ts
```diff
@@ -0,0 +1,25 @@
...
+        [CORE_DECISIONS_LIST, action.payload.walletAddress],
+        Decision(action.payload),
+      );
+    case ActionTypes.DECISION_DRAFT_REMOVED:
```
willm30, line 18: This will delete the decision at the given wallet address.

src/redux/sagas/setupUserContext.ts
```diff
@@ -61,6 +62,19 @@ export default function* setupUserContext() {
       type: ActionTypes.WALLET_OPEN_SUCCESS,
     });
 
+    /*
+     * Get saved decision from local storage and, if it exists, add to the store.
+     */
+
+    const decision = getDecisionFromLocalStorage(wallet?.address || '');
```
willm30, line 69: Whenever a new wallet is opened, the decision held in local storage will be saved to the store.

src/utils/colonyMotions.ts
```diff
@@ -21,133 +22,107 @@ export enum MotionState {
...
-  },
-  escalateTag: {
-    id: 'dashboard.ActionsPage.escalateTag',
-    defaultMessage: 'Escalate',
-  },
-  forcedTag: {
-    id: 'dashboard.ActionsPage.forcedTag',
-    defaultMessage: 'Forced',
-  },
+const { formatMessage } = intl({
```
willm30, line 32: Refactored these messages since they are not being defined inside a component.

src/components/common/ColonyDecisions/DecisionDialog/DecisionDialog.tsx
```diff
@@ -49,27 +49,27 @@ const validationSchema = object()
...
     setDecisionToLocalStorage(values, walletAddress);
     if (!pathname.includes(DECISIONS_PREVIEW)) {
       navigate(`${pathname}${DECISIONS_PREVIEW}`);
```
rdig, line 75: If we're setting the decision inside redux, shouldn't the local storage step moved inside a saga then ?

src/components/common/ColonyDecisions/DeleteDecisionDialog/DeleteDecisionDialog.tsx
```diff
@@ -37,11 +39,19 @@ const DeleteDecisionDialog = ({
 }: DeleteDecisionDialogProps) => {
   const navigate = useNavigate();
   const { pathname } = useLocation();
+  const dispatch = useDispatch();
 
   const handleDeleteDecision = () => {
     removeDecisionFromLocalStorage(decision.walletAddress);
```
rdig, line 45: Same as above, I think this step should be moved inside the saga

src/components/common/ColonyDecisions/DraftDecisionItem/DraftDecisionItem.tsx
rdig: Makes sense, could you please open a issue to track that ?

src/components/common/ColonyDecisions/RemoveExistingDecisionDialog/RemoveExistingDecisionDialog.tsx
```diff
@@ -41,6 +43,11 @@ const RemoveExistingDecisionDialog = ({
 
   const handleRemoveDecision = () => {
     removeDecisionFromLocalStorage(walletAddress);
```
rdig, line 45: Same here

src/components/shared/ListItem/ListItem.tsx
```diff
@@ -0,0 +1,117 @@
...
+  return (
+    <li>
+      <div
+        className={getMainClasses({}, styles, {
+          [ListItemStatus[status]]: !!status,
+        })}
+        onClick={onClick}
+        onKeyDown={onClick}
+        role="button"
+        tabIndex={0}
```
rdig, line 77: I don't think the tabIndex-es should be here at all, since we're basically hardcoding all to 0, meaning that keyboard nav is virtually impossible.

Then again, I don't really recall the original reason for why they were placed in here to being with

rdig (approved):
Looks good. I like the change to Redux for the draft.

I've left some comments that need to be addressed, but approving this anyway.

src/components/common/ColonyDecisions/DecisionDialog/DecisionDialog.tsx
jakubcolony: I noticed you are going away from the pattern of using action creators - is that intentional?

jakubcolony (approved):
LGTM, still works as expected 👍

What was the reasoning between migrating this to Redux vs. for example, using a context, which seemed like a pattern we were moving towards? I don't mean the technical differences of these, but specifically our use case and the impact of this change on other features - presumably app context, colony context and extensions should all be in Redux too

src/components/common/ColonyDecisions/DraftDecisionItem/DraftDecisionItem.tsx
willm30: Worked on this here #229

src/components/shared/ListItem/ListItem.tsx
willm30: I did try initially without, but the linter doesn't like it: 

> error  Elements with the 'button' interactive role must be focusable  jsx-a11y/interactive-supports-focus

src/components/common/ColonyDecisions/DecisionDialog/DecisionDialog.tsx
willm30: No, thank you for pointing out. I have created a couple for these

willm30:
@jakubcolony We were using redux in the dapp / cdapp before I incorporated the decisions functionality so wouldn't say this addition in itself would be responsible for rethinking how we manage global state. 

To use context in this case I would need to wrap the DialogProvider since the dialogs need access to the decision, but this is already wrapped by both the AppContextProvider and the ReduxProvider. Decisions felt too specific for `AppContext`, but not substantial enough to warrant its own context and an additional component wrapper at `Entry.tsx`. Redux felt like a natural choice in this case.

To your point about moving everything to redux, I don't have a strong opinion. I do like redux and think it's preferable to having lots of context providers all over the app, but doesn't seem too high a priority for right now. 

Happy to set up a call if you want to discuss more.

PR #215, opened 2023-01-24
Add index modal of the UAC

ArmandoGraterol:
Added initial index modal of the UAC. Access to it it's like always, just click the "New Action" button.

Also included some misc stuff that I noticed that needed to change.

FireShot Capture 785 - Colony CDapp - localhost

Contributes to #58


src/components/common/ColonyHome/ColonyHomeLayout.tsx
```diff
@@ -71,7 +69,10 @@ const ColonyHomeLayout = ({
                   />
                 </div>
                 {showActions && (
-                  <NewActionButton /* ethDomainId={ethDomainId} */ />
+                  <NewActionButton
+                    colony={colony}
```
jakubcolony, line 73: We're trying to go away from passing the current colony as a prop and tend to get it from the ColonyContext in the component itself:
```js
```

jakubcolony (approved):
Nice, the modal looks good:
image

I noticed you can click New Action even if you don't have a wallet connected / you aren't registered. Will this be covered in subsequent PRs? If so, thumbs up from me 👍

ArmandoGraterol:
> I noticed you can click New Action even if you don't have a wallet connected / you aren't registered. Will this be covered in subsequent PRs? If so, thumbs up from me +1

That's actually a good point I didn't notice before, I'll take a look at the logic as perhaps I commented it out at some point. Thank you!

src/components/common/NewActionButton/NewActionButton.tsx
```diff
@@ -281,12 +290,14 @@ const NewActionButton = (/* { colony, ethDomainId }: Props */) => {
           <Button
             appearance={{ theme: 'primary', size: 'large' }}
             text={MSG.newAction}
-            onClick={() => startWizardFlow('dashboard.ColonyActionsDialog')}
+            onClick={() =>
```
willm30, line 293: This view needs to be reused by `NewDecisionButton` for Simple Decisions, but naturally doesn't need most of the accompanying logic. 

I created [`DialogButton`](https://github.com/JoinColony/colonyCDapp/blob/port/188-remove-draft-decision-modal/src/components/shared/Button/DialogButton.tsx) to extract the view for reuse between both. 

Fancy copying that over and using it here?

src/components/common/NewActionButton/NewActionButton.tsx
```diff
@@ -87,172 +95,173 @@ const NewActionButton = (/* { colony, ethDomainId }: Props */) => {
...
-  //   {
-  //     component: ColonyTokenManagementDialog,
-  //     props: {
-  //       prevStep: 'dashboard.ManageFundsDialog',
-  //       colony,
-  //     },
-  //   },
-  // ]);
-  const startWizardFlow = (...args) => args;
+  const startWizardFlow = useNaiveBranchingDialogWizard([
```
willm30, line 98: Perhaps have this live in a separate config file? I wouldn't say it needs to live inside the component like the one in `ColonyActionsDialog` as it's only receiving a couple of props (colony and domainId) that can be passed as arguments to the wrapper function (e.g. getWizardFlow, which would return `startWizardFlow` for use in the click handler).

src/utils/colonyMotions.ts
```diff
@@ -30,47 +30,47 @@ export enum MotionVote {
 
 const MSG = defineMessage({
   stakedTag: {
-    id: 'dashboard.ActionsPage.stakedTag',
+    id: 'common.ColonyHome.ActionsPage.stakedTag',
```
willm30, line 33: I proposed the following change again in simple decisions, since these messages aren't being defined inside a component (and at least one of which is used in DraftDecisionItem and not Action Page), they could use the other style of message ids: https://github.com/JoinColony/colonyCDapp/pull/201/files#r1085082413

willm30 (approved):
Nice job @ArmandoGraterol! And thanks for splitting it all up into smaller prs 👍 

Only a couple of comments about coordinating changes between this branch and decisions.

src/components/common/NewActionButton/NewActionButton.tsx
ArmandoGraterol: Moved the flow array to its own config file inside the folder

src/utils/colonyMotions.ts
ArmandoGraterol: I'll revert this change then, and let the changes you did on Simple Decision take over eventually

src/components/common/Dialogs/index.ts
```diff
@@ -0,0 +1,23 @@
...
+// export { default as CreateDomainDialog } from './CreateDomainDialog';
+// export { default as EditDomainDialog } from './EditDomainDialog';
+// export { default as CreatePaymentDialog } from './CreatePaymentDialog';
+// export { default as ManageDomainsDialog } from './ManageDomainsDialog';
+// export { default as ManageFundsDialog } from './ManageFundsDialog';
+// export { default as UnlockTokenDialog } from './UnlockTokenDialog';
+// export { default as TransferFundsDialog } from './TransferFundsDialog';
+// export { default as AdvancedDialog } from './AdvancedDialog';
+// export { default as PermissionManagementDialog } from './PermissionManagementDialog';
+// export { default as RecoveryModeDialog } from './RecoveryModeDialog';
```
rdig, line 12: See my comment on #217 about not needing recovery mode for now

rdig (approved):
Good to go

PR #220, opened 2023-01-24
Port "Manage teams" related dialogs

ArmandoGraterol:
Added the add and edit teams dialogs. I'm excluding the manage whitelist dialog because that would make this PR much bigger than it already is.

FireShot Capture 793 - Colony CDapp - localhost
FireShot Capture 794 - Colony CDapp - localhost
FireShot Capture 795 - Colony CDapp - localhost

Contributes to #58


src/components/common/Dialogs/CreateDomainDialog/CreateDomainDialogForm.tsx
```diff
@@ -0,0 +1,240 @@
...
+} from '~hooks'; // useEnabledExtensions
+import { getAllUserRoles } from '~redux/transformers';
+import { canArchitect } from '~utils/checks';
+
+import { FormValues } from './CreateDomainDialog';
+
+import styles from './CreateDomainDialogForm.css';
+
+const displayName =
+  'common.ColonyHome.CreateDomainDialog.CreateDomainDialogForm';
```
rdig, line 35: Shouldn't this be under `Dialogs` rather than under `ColonyHome` ?

src/components/common/Dialogs/EditDomainDialog/EditDomainDialog.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+interface CustomWizardDialogProps extends ActionDialogProps {
+  filteredDomainId?: number;
+}
+
+type Props = DialogProps &
+  Partial<WizardDialogType<object>> &
+  CustomWizardDialogProps;
+
+const displayName = 'common.ColonyHome.EditDomainDialog';
```
rdig, line 35: Same here

src/components/common/Dialogs/EditDomainDialog/EditDomainDialogForm.tsx
```diff
@@ -0,0 +1,358 @@
...
+import { getAllUserRoles } from '~redux/transformers';
+import { canArchitect } from '~utils/checks';
+import { DomainColor } from '~gql';
+import { sortBy } from '~utils/lodash';
+
+import { FormValues } from './EditDomainDialog';
+
+import styles from './EditDomainDialogForm.css';
+
+const displayName = 'common.ColonyHome.EditDomainDialog.EditDomainDialogForm';
```
rdig, line 36: Same here

src/components/common/Dialogs/ManageDomainsDialog/ManageDomainsDialog.tsx
```diff
@@ -0,0 +1,144 @@
...
+import { FormattedMessage, defineMessages } from 'react-intl';
+
+import { DialogProps, ActionDialogProps } from '~shared/Dialog';
+import IndexModal from '~shared/IndexModal';
+
+import { WizardDialogType, useTransformer, useAppContext } from '~hooks'; // useEnabledExtensions
+import { getAllUserRoles } from '~redux/transformers';
+import { canArchitect, hasRoot } from '~utils/checks';
+
+const displayName = 'common.ColonyHome.ManageDomainsDialog';
```
rdig, line 11: yet another component name that needs to be fixed

rdig (approved):
Other than components name, all's good here

src/components/common/Dialogs/EditDomainDialog/EditDomainDialog.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+  const transform = pipe(
+    mapPayload((payload) => getEditDomainDialogPayload(colony, payload)),
+    withMeta({ navigate }),
+  );
+
+  return (
+    <Form<FormValues>
+      defaultValues={{
+        forceAction: false,
```
willm30, line 82: Does `forceAction` need to be included in the form's payload? It's only used to determine which saga to fire, I believe?

In which case, we could pass the `ForceToggle` component the setState function directly, meaning we wouldn't need to derive the `forced` state from the form state, which would mean we could remove lines 98-102

src/components/common/Dialogs/EditDomainDialog/EditDomainDialog.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+interface CustomWizardDialogProps extends ActionDialogProps {
+  filteredDomainId?: number;
+}
+
+type Props = DialogProps &
+  Partial<WizardDialogType<object>> &
+  CustomWizardDialogProps;
+
+const displayName = 'common.EditDomainDialog';
```
willm30, line 27: I can't remember from a previous review if these displayNames are supposed to include `common.Dialogs...`?

src/components/common/Dialogs/CreateDomainDialog/CreateDomainDialogForm.tsx
```diff
@@ -0,0 +1,156 @@
...
+        <DialogHeading title={MSG.titleCreate} />
+      </DialogSection>
+      {!userHasPermission && (
+        <DialogSection>
+          <PermissionRequiredInfo requiredRoles={[ColonyRole.Architecture]} />
+        </DialogSection>
+      )}
+      <DialogSection>
+        <div className={styles.nameAndColorContainer}>
+          <div className={styles.domainName}>
```
willm30, line 86: I see that this combination of domain and color select is also included in the EditDomainDialogForm, so I think could be moved to own component.

src/components/common/Dialogs/CreateDomainDialog/CreateDomainDialog.tsx
```diff
@@ -0,0 +1,97 @@
...
+    <Form<FormValues>
+      defaultValues={{
+        forceAction: false,
+        teamName: '',
+        domainColor: Color.LightPink,
+        domainPurpose: '',
+        annotationMessage: '',
+        // motionDomainId: Id.RootDomain,
+      }}
+      submit={getFormAction('SUBMIT')}
```
willm30, line 70: This `getFormAction` logic is being duplicated in every form. 

An alternative is to move the part concerned with the action type suffix into a util, which gets called directly inside the ActionForm

``` js
export const getFormAction = (
  action: ActionTypes,
  actionType: 'ERROR' | 'SUCCESS',
) => `${action}_${actionType}`;
```

then 

``` js
const ActionHookForm = <V extends Record<string, any>>({
  actionType,
  onSuccess,
  onError,
  onSubmitError,
  transform,
  ...props
}: Props<V>) => {
  const submit = actionType;
  const error = getFormAction(actionType, 'ERROR');
  const success = getFormAction(actionType, 'SUCCESS');
 ...
 }
```

Then you only need to pass in the actionType, the logic for which can be kept in the parent component. E.g.

``` tsx 
const actionType = !isForce /*&& isVotingExtensionEnabled*/
      ? ActionTypes.MOTION_DOMAIN_CREATE_EDIT
      : ActionTypes.ACTION_DOMAIN_CREATE

```

willm30 (commented):
Nice job @ArmandoGraterol!

When I open the Edit Team dialog from a colony I created manually (as opposed to via the script), I'm seeing the following: 

Screenshot (659)

Which in itself is fine, but drew my attention to the select team dropdown, which presumably should also be greyed out? Also the pointers are inconsistent in the disabled inputs: some show the disabled pointer, some don't. And finally, although I couldn't catch it in the screenshot, the tooltip "?" icon has a title, which overlaps the actual tooltip text and should probably be removed.

Just commenting for now as I left some over comments, but good stuff 🙂

rdig (approved):
Will has been very thorough in his review here, so I don't really have anything that I can point to.

I'm pre-approving this, but please address Will's issues, especially the ones that deal with extracting into separate components as that will help with readability and code reuse

src/components/common/Dialogs/EditDomainDialog/EditDomainDialog.tsx
ArmandoGraterol: I don't think so, that would mean that there's a `Dialogs` component when that's just a containing folder

src/components/common/Dialogs/EditDomainDialog/EditDomainDialog.tsx
ArmandoGraterol: It doesn't need to be included in the payload, but it's also not only used to determine which saga to fire as we have various conditionals inside the dialog forms that use this value, mainly revolving around permission stuff. Do we believe that in this case is it worth it to prop drill `setForceAction` and `forceAction`? 

FYI in case we prop drill those values instead of using the form context, currently, it would travel like this in almost every dialog that has an action that can be made into a motion:
```
Dialog > DialogForm > DialogHeading > ForceToggle
Dialog > DialogForm > DialogHeading > MotionDomainSelect (only forceAction)
Dialog > DialogForm > useDialogActionPermissions (only forceAction)
Dialog > DialogForm > DialogControls (only forceAction)
```

src/components/common/Dialogs/EditDomainDialog/EditDomainDialogForm.tsx
```diff
@@ -0,0 +1,258 @@
...
+  //   [currentFromDomain],
+  // );
+
+  // const handleMotionDomainChange = useCallback(
+  //   (motionDomainIdValue) =>
+  //     setFieldValue('motionDomainId', motionDomainIdValue),
+  //   [setFieldValue],
+  // );
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
ArmandoGraterol: Hmmm, I thought I removed this old code, perhaps I did it in another place and forgot this one :thinking: Thank you!

src/components/common/Dialogs/EditDomainDialog/EditDomainDialog.tsx
willm30: Ok, let's leave as it is then. If it were just prop drilling to the ForceToggle component then I might've thought it was worth it but passing to multiple components probably isn't preferable to what we have currently. I was trying to think of a way to make the forced state accessible to the form's contents and the form itself. The only way I can think of would be to wrap the dialog in a context that stores the forced state, but that's just me thinking out loud (have not tested). Also that would be a separate issue if it did work. But mentioning just so you know what I was thinking.
dialog

src/components/common/Dialogs/CreateDomainDialog/CreateDomainDialog.tsx
willm30: @ArmandoGraterol Adding to this for completeness, Raul mentioned in one of my prs that this should also include a way to pass in the states individually, but that they have default values assigned

https://github.com/JoinColony/colonyCDapp/pull/261#discussion_r1107359898

src/components/common/Dialogs/EditDomainDialog/EditDomainDialog.tsx
ArmandoGraterol: Yeah, I was also thinking of ways of including this state in a more general component like the `ActionHookForm`, but I haven't encountered a good, easy way of doing it. Thank you for your thoughts! I will think about this some more

src/components/common/Dialogs/CreateDomainDialog/CreateDomainDialog.tsx
ArmandoGraterol: Thank you for the update :+1:

src/components/common/Dialogs/CreatePaymentDialog/validation.ts
```diff
@@ -21,7 +21,7 @@ const MSG = defineMessages({
 
 const validationSchema = object()
   .shape({
-    domainId: string().required(),
+    fromDomain: number().required(),
```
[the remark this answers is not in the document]
ArmandoGraterol: I'll wait for a conclusion here to update the code if we decide to store the id as a number or string but yeah I think initially the domain ids are coming as numbers  (For example, `Id.RootDomain` is a number).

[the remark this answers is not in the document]
jakubcolony: Good spot @willm30, that was actually an oversight on my side, #263 changes the domain fields to `Int`

src/utils/domains.ts
```diff
@@ -1,10 +1,21 @@
...
   if (!domains || !domainId) {
     return undefined;
   }
-  return domains.filter(notNull).find(({ id }) => id === domainId);
+  return domains
+    .filter(notNull)
+    .find(({ id }) =>
+      skipRootDomain && (domainId === undefined || domainId === Id.RootDomain)
+        ? Number(id) !== Id.RootDomain
+        : Number(id) === domainId,
```
jakubcolony, line 19: This logic is actually wrong - it should be comparing the `domainId` against `nativeId`.

jakubcolony (changes requested):
Nice work @ArmandoGraterol!

I'm requesting changes mainly due to the issue with missing permissions UI:
I'm experiencing a similar issue to Will, but it's with the colonies created via the script. I see a message regarding missing permission but the team selector stays active and when I select one, the entire form becomes editable:
image
image

Just a small note, in other forms we're able to show a required error that doesn't include the camelCased field name - could that also be done here?
image

The `useEnabledExtensions` should be now available in `port/colony-actions` so there's no need to comment it out anymore

ArmandoGraterol:
@jakubcolony 
> The useEnabledExtensions should be now available in port/colony-actions so there's no need to comment it out anymore

I see that this new hook only provides whether the extension is enabled or not, but the old one also returned the version of the voting reputation extension. Would you know if that's obtained now from another place? Or do we want to update this new hook to also return that value?

jakubcolony:
@ArmandoGraterol

src/components/common/Dialogs/CreateDomainDialog/CreateDomainDialogForm.tsx
```diff
@@ -0,0 +1,123 @@
...
+    !values.forceAction;
+
+  return (
+    <>
+      <DialogSection appearance={{ theme: 'sidePadding' }}>
+        <DialogHeading title={MSG.titleCreate} />
+      </DialogSection>
+      {!userHasPermission && (
+        <DialogSection>
+          <PermissionRequiredInfo requiredRoles={[ColonyRole.Architecture]} />
```
willm30, line 69: The `requiredRoles` const can be reused here

src/components/common/Dialogs/CreateDomainDialog/CreateDomainDialogForm.tsx
```diff
@@ -0,0 +1,123 @@
...
+          label={MSG.annotation}
+          name="annotationMessage"
+          disabled={inputDisabled}
+          dataTest="createDomainAnnotation"
+        />
+      </DialogSection>
+      {!userHasPermission && (
+        <DialogSection appearance={{ theme: 'sidePadding' }}>
+          <NoPermissionMessage
+            requiredPermissions={[ColonyRole.Architecture]}
```
willm30, line 99: And here

src/components/common/Dialogs/CreateDomainDialog/CreateDomainDialogForm.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const [userHasPermission, onlyForceAction] = useDialogActionPermissions(
+    colony,
+    isVotingReputationEnabled,
+    requiredRoles,
+    [Id.RootDomain],
+  );
+
+  const inputDisabled = !userHasPermission || onlyForceAction || isSubmitting;
+
+  const cannotCreateMotion =
```
willm30, line 58: Could the logic above the view here be wrapped up into a hook that only exposes the data needed by the view? (eg `useCreateDomainDialogForm`)

src/components/common/Dialogs/CreatePaymentDialog/CreatePaymentDialogForm.tsx
```diff
@@ -221,46 +181,33 @@ Props) => {
...
         <DialogSection appearance={{ theme: 'sidePadding' }}>
-          <div className={styles.noPermissionFromMessage}>
-            <FormattedMessage
-              {...MSG.noPermissionFrom}
-              values={noPermissionFromMessageValues}
-            />
-          </div>
+          <NoPermissionMessage
+            requiredPermissions={[
+              ColonyRole.Administration,
```
willm30, line 186: There's a  required roles const above

src/components/common/Dialogs/CreatePaymentDialog/CreatePaymentDialogForm.tsx
```diff
@@ -137,17 +98,16 @@ Props) => {
...
-  //     VotingReputationExtensionVersion.FuchsiaLightweightSpaceship &&
-  //   !values.forceAction;
+  const cannotCreateMotion =
+    votingReputationVersion === noMotionsVotingReputationVersion &&
+    !values.forceAction;
 
-  const canMakePayment = userHasPermission; // && isOneTxPaymentExtensionEnabled;
+  const canMakePayment = userHasPermission && isOneTxPaymentEnabled;
 
   const inputDisabled = !canMakePayment || onlyForceAction || isSubmitting;
```
willm30, line 112: Similar thought about wrapping up as much logic as possible so we're only exposing what the view needs (and reducing the number of lines someone needs to read through before they get to the view itself)

src/components/common/Dialogs/CreatePaymentDialog/CreatePaymentDialogForm.tsx
```diff
@@ -221,46 +181,33 @@ Props) => {
...
         </DialogSection>
-      )} */}
+      )}
       <DialogSection appearance={{ align: 'right', theme: 'footer' }}>
         <DialogControls
           onSecondaryButtonClick={back}
           disabled={
-            // cannotCreateMotion ||
-            !isValid || !!customAmountError || inputDisabled
+            cannotCreateMotion ||
```
willm30, line 207: I think this should be moved to a const (e.g `areControlsDisabled`)

src/components/common/Dialogs/EditDomainDialog/EditDomainDialog.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+const validationSchema = object()
+  .shape({
+    forceAction: boolean().defined(),
+    domainName: string().max(20).required(),
```
willm30, line 30: Can these have the appropriate messages to avoid the following?

domainname

src/components/common/Dialogs/EditDomainDialog/EditDomainDialogForm.tsx
```diff
@@ -0,0 +1,195 @@
...
+    defaultMessage: `Cannot create motions using the Governance v{version} Extension. Please upgrade to a newer version (when available)`,
+  },
+});
+
+const EditDomainDialogForm = ({
+  back,
+  colony,
+  colony: { domains },
+}: ActionDialogProps) => {
+  const {
```
willm30, line 56: Same thought here re wrapping this logic into a hook (applies to all components really where there's more than a few lines of business logic above the view (so I don't repeat myself any more in the comments))

src/components/common/Dialogs/MintTokenDialog/MintTokenDialogForm.tsx
```diff
@@ -141,41 +131,19 @@ const MintTokenDialogForm = ({ colony, back }: ActionDialogProps) => {
...
-                    permission={ColonyRole.Root}
-                    name={{
-                      id: `role.${ColonyRole.Root}`,
-                    }}
-                  />
-                ),
-              }}
-            />
-          </div>
+          <NoPermissionMessage requiredPermissions={[ColonyRole.Root]} />
```
willm30, line 134: required roles const above

willm30 (commented):
UI working for me 🙂 Left some comments on tidying up some of the logic in the components

jakubcolony (commented):
All seems to be working well, great job! 👍 
I left a few comments about some minor improvements

src/hooks/useEnabledExtensions.tsx
```diff
@@ -1,9 +1,11 @@
 import { Extension } from '@colony/colony-js';
 
+import { Colony } from '~types';
+
 import useExtensionsData from './useExtensionsData';
 
-const useEnabledExtensions = () => {
-  const { installedExtensionsData, loading } = useExtensionsData();
+const useEnabledExtensions = (colony: Colony | undefined) => {
```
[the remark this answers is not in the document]
ArmandoGraterol: If we add `?`, this would mean that the param is optional no? We could call `useEnabledExtensions()` if that was the case, which is wrong because without a `colony` the hook will never return the data, as opposed to requiring a `colony` param that can be defined or not (We are assuming that eventually the `colony` will be defined, if not already).

src/components/common/Dialogs/CreateDomainDialog/CreateDomainDialog.tsx
```diff
@@ -0,0 +1,100 @@
...
+  callStep,
+  prevStep,
+  cancel,
+  close,
+  colony,
+}: Props) => {
+  const [isForce, setIsForce] = useState(false);
+  const navigate = useNavigate();
+
+  const { isVotingReputationEnabled } = useEnabledExtensions(colony);
```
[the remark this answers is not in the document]
ArmandoGraterol: It's possible because we could use that hook in the `NewActionButton` and pass the data from there to the dialog wizard, however, we have to take into consideration that we also need to pass `isOneTxPaymentEnabled` and `votingReputationVersion`. So, we would need to pass 3 new props or 1 with the 3 props inside it (something like `extensionData`?).

What do you think?

src/components/common/NewActionButton/NewActionButton.tsx
```diff
@@ -82,7 +79,7 @@ const NewActionButton = ({ filteredDomainId }: Props) => {
   // const mustUpgradeOneTx = oneTxMustBeUpgraded(oneTxPaymentExtension);
   const hasRegisteredProfile = !!user?.name && !!user.walletAddress;
   // const mustUpgrade = colonyMustBeUpgraded(colony, networkVersion as string);
-  // const isLoadingData = isLoadingExtensions || isLoadingUser;
+  const isLoadingData = loading; // || isLoadingUser;
```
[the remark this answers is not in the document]
ArmandoGraterol: This is a good question, thank you! I have noticed that there's a `walletConnecting` available on `useAppContext` that should be the equivalent of the `isLoadingUser` so I'm gonna implement that

src/components/shared/ColorSelect/ColorSelect.tsx
```diff
@@ -57,20 +57,17 @@ const ColorSelect = ({
     return <ColorTag color={activeOption} />;
   }, [activeOption]);
 
-  const options = useMemo<ComponentProps<typeof Select>['options']>(() => {
-    const colors = Object.keys(DomainColor).filter(
-      (val) => typeof DomainColor[val as any] === 'number',
-    );
-    return [
-      ...colors.map((color) => {
+  const options = useMemo<ComponentProps<typeof Select>['options']>(
```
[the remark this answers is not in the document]
ArmandoGraterol: Nice one! Thank you

src/components/shared/DomainFundSelectorSection/helpers.ts
```diff
@@ -1,11 +1,18 @@
+import { Id } from '@colony/colony-js';
+
 import { Domain } from '~types';
 import { sortBy } from '~utils/lodash';
 
-export const getDomainOptions = (colonyDomains: Domain[]) =>
+export const getDomainOptions = (
+  colonyDomains: Domain[],
+  skipRootDomain?: boolean,
```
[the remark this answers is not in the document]
ArmandoGraterol: Sounds good to me :ok_hand:  I'll make the changes

src/hooks/useEnabledExtensions.tsx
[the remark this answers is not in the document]
jakubcolony: Fair enough, thanks 👍

src/components/common/Dialogs/CreateDomainDialog/CreateDomainDialog.tsx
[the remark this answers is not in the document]
jakubcolony: I don't have a preference, whatever feels right for you. You could possibly introduce a type that we'd use for both the return value of `useEnabledExtensions` and the 1 props you mentioned

ArmandoGraterol:
@willm30 I've added a hook called `useActionDialogStatus` in which I tried to add as much logic in common between the dialog forms as I could, let me know what you think!

@jakubcolony I've reverted the changes to `useEnabledExtensions` and added the data as props to the action dialogs.

src/components/common/Dialogs/CreatePaymentDialog/helpers.ts
```diff
@@ -32,3 +36,33 @@ export const getCreatePaymentDialogPayload = (colony: Colony, payload: any) => {
...
   };
 };
+
+export const useCreatePaymentDialogStatus = (
+  colony: Colony,
+  requiredRoles: ColonyRole[],
+  enabledExtensionData: EnabledExtensionData,
+) => {
+  const { getValues } = useFormContext();
+  const values = getValues();
```
jakubcolony, line 46: So apparently, after talking to @willm30, I think you should use the `watch` function here. `getValues` doesn't guarantee that the component will get re-rendered if `values.fromDomain` changes
On a side note, you could destructure `fromDomain`

src/components/common/Dialogs/EditDomainDialog/EditDomainDialog.tsx
```diff
@@ -0,0 +1,124 @@
+import React, { useState } from 'react';
+import { useNavigate } from 'react-router-dom';
+import { string, object, number, boolean, InferType } from 'yup';
+import { defineMessages } from 'react-intl';
+
+import Dialog, { DialogProps, ActionDialogProps } from '~shared/Dialog';
+import { ActionHookForm as Form } from '~shared/Fields';
+
+import { Domain } from '~types';
```
jakubcolony, line 9: This line gives me a TS unused variable error

src/components/common/Dialogs/EditDomainDialog/EditDomainDialog.tsx
```diff
@@ -0,0 +1,124 @@
...
+import { ActionTypes } from '~redux';
+import { WizardDialogType } from '~hooks';
+import { pipe, withMeta, mapPayload } from '~utils/actions';
+import { DomainColor } from '~gql';
+import { findDomainByNativeId } from '~utils/domains';
+import { notNull } from '~utils/arrays';
+import { getDomainOptions } from '~shared/DomainFundSelectorSection/helpers';
+
+import EditDomainDialogForm from './EditDomainDialogForm';
+import { getEditDomainDialogPayload, notRootDomain } from './helpers';
```
jakubcolony, line 19: As above

src/components/shared/ColorSelect/ColorSelect.tsx
```diff
@@ -36,11 +38,13 @@ const displayName = 'ColorSelect';
 
 const ColorSelect = ({
   disabled,
-  activeOption,
   onColorChange,
   appearance,
   name = 'color',
 }: Props) => {
+  const { getValues } = useFormContext();
+  const activeOption = getValues(name);
```
jakubcolony, line 46: I think this should also be using `watch`

src/components/shared/Fields/Select/HookForm/HookFormSelect.tsx
```diff
@@ -53,11 +53,12 @@ const HookFormSelect = ({
 }: SelectProps) => {
   const [id] = useState<string>(idProp || nanoid());
   const {
-    formState: { errors },
+    formState: { errors, touchedFields },
     watch,
     setValue,
   } = useFormContext();
-  const error = errors[name] as Message;
+  const error = errors[name]?.message as Message;
```
jakubcolony, line 60: Should be `as Message | undefined`

src/components/shared/TokenAmountInput/TokenAmountInput.tsx
```diff
@@ -86,7 +89,7 @@ const TokenAmountInput = ({ colony, disabled }: Props) => {
           <div className={styles.tokenAmountUsd}>
             <EthUsd
               // appearance={{ theme: 'grey' }}
-              value={values.amount}
+              value={values.amount.replace(/,/g, '') || 0} // @NOTE: Get the amount without commas as BigNumber doesn't like them
```
jakubcolony, line 92: I think the problem should be fixed at the root (`FormattedInputComponent` should be storing the Cleave's raw value instead of formatted value which may contain the comma separators). I fixed it in #320 so do you want to add a TODO comment here to change it later?

jakubcolony (changes requested):
The UI works mint, but I had a few comments on the new changes. Otherwise nice work!

src/components/common/Dialogs/EditDomainDialog/EditDomainDialog.tsx
```diff
@@ -0,0 +1,124 @@
...
+  cancel,
+  close,
+  colony,
+  filteredDomainId,
+  enabledExtensionData,
+}: Props) => {
+  const colonyDomains =
+    colony.domains?.items.filter(
+      (domain) => notNull(domain) && notRootDomain(domain),
+    ) || [];
```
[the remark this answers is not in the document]
ArmandoGraterol: I also considered this and seeing that you concur with what I thought then will do :+1:

src/components/common/Dialogs/EditDomainDialog/EditDomainDialog.tsx
ArmandoGraterol: Do you mean it's appearing as if they are unused? It's not the case for me :thinking:

src/components/shared/TokenAmountInput/TokenAmountInput.tsx
ArmandoGraterol: For sure, I will add it to the other instances where we are removing commas like this as this ain't the only instance.

src/components/shared/TokenEditDialog/TokenEditDialog.tsx
willm30, line 1: Belongs in `common/Dialogs`?

src/components/shared/TokenAmountInput/TokenAmountInput.tsx
willm30, line 1: This probably also belongs in the dialogs folder since it appears to be an abstraction specific to a couple of dialogs only

src/components/shared/DomainFundSelectorSection/DomainBalance.tsx
willm30, line 1: Same thought for this one re moving somewhere inside `common/Dialogs`

willm30 (approved):
Nice one Armando. I think some of the components either introduced here or edited here actually belong inside a `common/Dialogs` as opposed to `shared`, since they are only and may only be expected to be used in some of the dialogs. 

Also, I think this pr would be easier to review if it were even smaller, e.g. one pr per dialog introduced. (That's definitely something I want to get better at also, having prs address the smallest logical / practical set of changes possible).  

But don't let me hold you up 🙂 UI's good and all will be revisited again in the not too distant future as part of the ui refresh.

src/components/common/Dialogs/CreatePaymentDialog/validation.ts
```diff
@@ -39,7 +39,7 @@ const validationSchema = object()
         'more-than-zero',
         () => MSG.amountZero,
         (value) => {
-          const numberWithoutCommas = (value || '0').replace(/,/g, '');
+          const numberWithoutCommas = (value || '0').replace(/,/g, ''); // @TODO: Remove this once the fix for FormattedInputComponent value is introduced.
```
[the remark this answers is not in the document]
ArmandoGraterol: Do you mean the downgrading of `cleave.js`? Nope, this is related to this fix that @jakubcolony implemented: https://github.com/JoinColony/colonyCDapp/pull/320/commits/dd6f97d886bb3207d2c29018a6ebdd85544e86ae

Basically, the cleave was using the formatted value of the input instead of the raw value like in the Dapp. So, we would get errors when dealing with the value stored by the form because they would have commas `,`. This is a solution I implemented before Jakub informed me of the issue with the cleave value.

I should remove this line when Jakub's fix makes its way into the branch.

src/components/shared/CannotCreateMotionMessage/CannotCreateMotionMessage.tsx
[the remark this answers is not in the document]
ArmandoGraterol: Sure, sounds good to me

jakubcolony (approved):
All good, nice work! 👍

src/components/common/Dialogs/CreateDomainDialog/CreateDomainDialog.tsx
```diff
@@ -0,0 +1,102 @@
...
+        actionType={actionType}
+        validationSchema={validationSchema}
+        transform={transform}
+        onSuccess={close}
+      >
+        {({ watch }) => {
+          const forceActionValue = watch('forceAction');
+          if (forceActionValue !== isForce) {
+            setIsForce(forceActionValue);
+          }
```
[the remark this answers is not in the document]
willm30: The way things are currently set up, it's needed in the `actionType` variable which lives outside the form. I did mention in another pr we could look into storing the dialog state in the dialog provider, but that would need looking into separately.

PR #221, opened 2023-01-24
Add Manage Whitelist dialog

ArmandoGraterol:
Added manage whitelist dialog and updated some of the related components like `CSVUploader`.

FireShot Capture 796 - Colony CDapp - localhost
FireShot Capture 797 - Colony CDapp - localhost
FireShot Capture 798 - Colony CDapp - localhost

Contributes to #58

### DIFF
--- a/src/components/common/Dialogs/ManageWhitelistDialog/ManageWhitelistActiveToggle.tsx
+++ b/src/components/common/Dialogs/ManageWhitelistDialog/ManageWhitelistActiveToggle.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { defineMessages, FormattedMessage } from 'react-intl';
 
-import { Toggle } from '~shared/Fields';
+import { HookFormToggle as Toggle } from '~shared/Fields';
 import { Heading4 } from '~shared/Heading';
 
 import styles from './ManageWhitelistDialogForm.css';
@@ -22,14 +22,14 @@ const MSG = defineMessages({
   },
   tooltipText: {
     id: `${displayName}.tooltipText`,
-    defaultMessage: `Address book is active by default once at least one address is added to the list. You can turn this feature “Off” to deactivate. Use with caution.`,
+    defaultMessage: `Address book is automatically activated when at least one address is added to the list. You can turn this feature “Off” to deactivate. Use with caution.`,
   },
   warningText: {
     id: `${displayName}.warningText`,
     defaultMessage: `<span>Warning.</span>
-      You are deactivating your colony’s address book.
-      This is a list of addresses your colony has flagged as “safe”.
-      Now, instead, all addresses watching your colony will
+      You are deactivating your colony’s address book
+      which is a list of addresses your colony has flagged as “safe”.
+      Once this is deactivated, all addresses watching your colony will
       appear in recipient selection menus. Be vigilant against spoof accounts.`,
   },
 });
--- a/src/components/common/Dialogs/ManageWhitelistDialog/ManageWhitelistActiveToggle.tsx
+++ b/src/components/common/Dialogs/ManageWhitelistDialog/ManageWhitelistActiveToggle.tsx
@@ -6,7 +6,7 @@ import { Heading4 } from '~shared/Heading';
 
 import styles from './ManageWhitelistDialogForm.css';
 
-const displayName = `common.ColonyHome.ManageWhitelistDialog.ManageWhitelistDialogForm.ManageWhitelistActiveToggle`;
+const displayName = `common.ManageWhitelistDialog.ManageWhitelistDialogForm.ManageWhitelistActiveToggle`;
 
 const MSG = defineMessages({
   toggleLabel: {
@@ -54,7 +54,6 @@ const ManageWhitelistActiveToggle = ({ isWhitelistActivated }: Props) => (
         labelValues={{ isWhitelistActivated }}
         name="isWhitelistActivated"
         tooltipText={MSG.tooltipText}
-        tooltipClassName={styles.tooltip}
       />
     </div>
     {!isWhitelistActivated && (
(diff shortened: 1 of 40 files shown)


src/components/common/Dialogs/ManageWhitelistDialog/ManageWhitelistDialog.tsx
```diff
@@ -0,0 +1,246 @@
...
+interface CustomWizardDialogProps {
+  colony: Colony;
+  userAddress?: string;
+}
+
+type Props = DialogProps &
+  Partial<WizardDialogType<object>> &
+  CustomWizardDialogProps;
+
+const displayName = 'common.ColonyHome.ManageWhitelistDialog';
```
rdig, line 44: This should be under `Dialogs` not `ColonyHome` here as well

src/components/common/Dialogs/ManageWhitelistDialog/NoWhitelistedAddressesState/NoWhitelistedAddressesState.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import { FormattedMessage, defineMessages } from 'react-intl';
+
+import styles from './NoWhitelistedAddressesState.css';
+
+const displayName =
+  'common.ColonyHome.ManageWhitelistDialog.NoWhitelistedAddressesState';
```
rdig, line 7: This should be in `Dialogs` too

src/components/common/Dialogs/ManageWhitelistDialog/WhitelistedAddresses/WhitelistedAddresses.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+import styles from './WhitelistedAddresses.css';
+
+interface Props {
+  colony: Colony;
+  whitelistedUsers: User[];
+}
+
+const displayName =
+  'common.ColonyHome.ManageWhitelistDialog.WhitelistedAddresses';
```
rdig, line 19: Same here about being in `Dialogs`

rdig (approved):
Besides component names, this is mostly good, however I'm still unsure how are we going to handle Whitelist with the database backend.

For now feel free to leave it in.

Julianahlk (changes requested):
I am seeing a few issues after uploading a .csv

- The confirm button is still disabled, I would either expect it to be enabled or display an error message explaining what is wrong
- The back button is displayed as "button.back"
- If you switch tabs to Address Book then return back to the Add Address tab, the state in the UI of the uploaded csv is not saved.

Screen Shot 2023-02-16 at 10 02 18 PM


- Uploading the template.csv that does not contain any whitelisted addresses triggers an uncaught exception

Screen Shot 2023-02-16 at 10 09 44 PM

jakubcolony (changes requested):
@ArmandoGraterol I'm seeing similar issues to Juliana with the back button, but more importantly, when I tried to drop the unmodified template, I got stuck in render loop:
image

When I tried to replicate it however, all I was seeing is the same error that Juliana already reported.

Does it make sense for the textarea to be shown when there are no addresses in the book?
image

I'll re-review once the above is addressed

src/components/common/Dialogs/ManageWhitelistDialog/ManageWhitelistActiveToggle.tsx
```diff
@@ -0,0 +1,76 @@
+import React from 'react';
+import { defineMessages, FormattedMessage } from 'react-intl';
+
+import { Toggle } from '~shared/Fields';
+import { Heading4 } from '~shared/Heading';
+
+import styles from './ManageWhitelistDialogForm.css';
+
+const displayName = `common.ManageWhitelistDialog.ManageWhitelistDialogForm.ManageWhitelistActiveToggle`;
```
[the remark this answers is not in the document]
ArmandoGraterol: If by path you mean the folder path, that's correct, but this component is only used inside `ManageWhitelistDialogForm`. It's like `NetworkFee` in that sense, it's not used on `CreatePaymentDialog` directly but in the form of the dialog.

src/components/common/Dialogs/ManageWhitelistDialog/ManageWhitelistActiveToggle.tsx
```diff
@@ -0,0 +1,76 @@
...
+    defaultMessage: 'Only show address book contacts as recipients',
+  },
+  tooltipText: {
+    id: `${displayName}.tooltipText`,
+    defaultMessage: `Address book is active by default once at least one address is added to the list. You can turn this feature “Off” to deactivate. Use with caution.`,
+  },
+  warningText: {
+    id: `${displayName}.warningText`,
+    defaultMessage: `<span>Warning.</span>
+      You are deactivating your colony’s address book.
```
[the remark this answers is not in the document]
ArmandoGraterol: Sounds good to me x2

ArmandoGraterol:
@Julianahlk @jakubcolony Added some changes to could solve the issues that you guys are having, please let me know if it works! Working on this CSV uploader is a bit of a hassle.
 
> * If you switch tabs to Address Book then return back to the Add Address tab, the state in the UI of the uploaded csv is not saved.

That's intentional per line 114 in `ManageWhitelistDialogForm`. Do you think that it would be best that we don't reset the form when we change tabs? I don't have a strong opinion on it :)

jakubcolony (approved):
LGTM, all seems to work 👍 Nice job!

Julianahlk (approved):
Working much better and approving for now.

Regarding saving the state when switching between the add address and address book tabs, in an ideal world it would be nice if the state saved because if you upload a csv then switch tabs, you need to upload it again.

PR #222, opened 2023-01-24
Add smite and award reputation dialogs

ArmandoGraterol:
FireShot Capture 799 - Colony CDapp - localhost
FireShot Capture 800 - Colony CDapp - localhost
FireShot Capture 801 - Colony CDapp - localhost

Contributes to #58


src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationContainer/ManageReputationContainer.tsx
```diff
@@ -0,0 +1,190 @@
...
+// import { getVerifiedUsers } from '~utils/verifiedRecipients';
+import { notNull } from '~utils/arrays';
+
+import DialogForm from '../ManageReputationDialogForm';
+import {
+  AwardAndSmiteDialogProps,
+  ManageReputationDialogFormValues,
+} from '../types';
+
+const displayName = 'common.ColonyHome.ManageReputationContainer';
```
rdig, line 26: This should be in `Dialogs` rather than `ColonyHome`

src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationDialogForm/TeamDropdownItem/TeamDropdownItem.tsx
```diff
@@ -0,0 +1,49 @@
...
+import styles from './TeamDropdownItem.css';
+
+interface Props {
+  domain: Omit<Domain, 'createdAt' | 'updatedAt'> | null;
+  user: User | null;
+  userReputation: string | undefined;
+  totalReputation: string | undefined;
+}
+
+const displayName = `common.ColonyHome.SmiteDialog.TeamDropdownItem`;
```
rdig, line 18: Should be in `Dialogs`

src/components/common/Dialogs/AwardAndSmiteDialogs/SmiteDialog/SmiteDialog.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+import { AwardAndSmiteDialogProps } from '../types';
+import ManageReputationContainer from '../ManageReputationContainer';
+
+const displayName = 'common.ColonyHome.SmiteDialog';
```
rdig, line 6: `Dialogs` again

rdig (approved):
Yep, all good here _(other than the names of the components)_

src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationDialogForm/ManageReputationDialogForm.tsx
```diff
@@ -0,0 +1,329 @@
...
+import ExternalLink from '~shared/ExternalLink';
+import SingleUserPicker, {
+  filterUserSelection,
+} from '~shared/SingleUserPicker';
+import UserAvatar from '~shared/UserAvatar';
+import NoPermissionMessage from '~shared/NoPermissionMessage';
+// import NotEnoughReputation from '~dashboard/NotEnoughReputation';
+import { REPUTATION_LEARN_MORE } from '~constants/externalUrls';
+
+import { ColonyWatcher, User } from '~types/index';
```
jakubcolony, line 28: Probably don't need to include `index`

src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationDialogForm/ReputationAmountInput/ReputationAmountInput.tsx
```diff
@@ -0,0 +1,122 @@
+import React from 'react';
+import { defineMessages, FormattedMessage } from 'react-intl';
+import Decimal from 'decimal.js';
+import { useFormContext } from 'react-hook-form';
+
+import { HookFormInput as Input } from '~shared/Fields';
+import Numeral from '~shared/Numeral';
+import { Colony } from '~types/index';
```
jakubcolony, line 8: Probably don't need to include `index`

src/components/shared/NoPermissionMessage/NoPermissionMessage.tsx
```diff
@@ -22,7 +22,7 @@ const MSG = defineMessages({
 
 interface Props {
   requiredPermissions: ColonyRole[];
-  domainName?: string;
+  domainName?: string | null;
```
jakubcolony, line 25: Can this ever be null? If `domain.metadata` is present, the name will be there
image

jakubcolony (changes requested):
Good start @ArmandoGraterol but I spotted a couple of issues with the dialog:

- There is something off with the reputation in the domain selector:

https://user-images.githubusercontent.com/112586815/224078322-6ff61feb-63f3-41c9-a0b8-f4b22ea1721d.mov

- Clicking on the Max button causes an error in the console:
image

- For some reason, there seems to be an upper limit of 999 reputation points (above which the button becomes inactive). I don't know if it's intentional but I can put a higher value in the Dapp without it complaining

willm30:
> * Clicking on the Max button causes an error in the console:

@Julianahlk  experienced the same in #288. The [advice here](https://github.com/nosir/cleave.js/issues/601) suggests downgrading to 1.4.10, which I believe worked (Juliana pls confirm 🙂)

src/components/shared/NoPermissionMessage/NoPermissionMessage.tsx
ArmandoGraterol: It can be, in this case inside `TransferFundsDialogForm`:
```
          <NoPermissionMessage
            requiredPermissions={[ColonyRole.Funding]}
            domainName={
              hasRoleInFromDomain
                ? fromDomain?.metadata?.name ||
                  `Domain #${fromDomain?.nativeId}`
                : toDomain?.metadata?.name || `Domain #${toDomain?.nativeId}`
            }
          />
```

`fromDomain` and `toDomain` come from here:
```
  const fromDomain = colonyDomains.find(
    (domain) => domain?.nativeId === fromDomainId,
  );
```

```
  const toDomain = colonyDomains.find(
    (domain) => domain?.nativeId === toDomainId,
  );
```

So unfortunately even if `.metadata` wasn't optional, the `.find` assumes that it could be `null` in the case that it doesn't find the domain.

ArmandoGraterol:
@jakubcolony The changes you suggested should be up! There are still some that will be added that are on the "Manage teams" PR, but I'm pinging you anyways in case you want to check that the current code is good with you.

@willm30 @Julianahlk That was also the same issue here, I had to downgrade `cleave.js`. Thank you all!

src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationContainer/ManageReputationContainer.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+  // const verifiedUsers = useMemo(() => {
+  //   return getVerifiedUsers(colony.whitelistedAddresses, colonyWatchers) || [];
+  // }, [colonyWatchers, colony]);
+
+  const updateReputationCallback = (userRepPercentage: number) => {
+    setUserReputation(userRepPercentage);
+  };
+
+  const { isVotingReputationEnabled } = useEnabledExtensions(colony);
```
jakubcolony, line 85: I thought you'd pass this as a prop instead of making `useEnabledExtensions` accept `colony` as its argument

src/components/shared/NoPermissionMessage/NoPermissionMessage.tsx
jakubcolony: I don't think that's the case (although I might be wrong). If `.find` doesn't find anything, it'll return `undefined` instead of `null`. Removing `null` from the prop type doesn't trigger any type errors

jakubcolony (changes requested):
Hey @ArmandoGraterol, I had another look and spotted a couple of issues:

The domain selector still seems to be buggy (showing all the domains' reputation as the currently selected domain's reputation):
https://user-images.githubusercontent.com/112586815/226297854-5c65f000-ab26-42e4-9956-4f82a40420b3.mov

I can award negative reputation, is that intentional?
image

I see some errors in the console when the input value is `.` or `-` but that will presumably go away when my fix for Cleave using raw value is merged - we'll be able to change the validation from using a custom test and `Decimal.js` to native yup methods

There's a problem with the smite reputation Max button, it seems like the validation doesn't re-run when it's clicked:

https://user-images.githubusercontent.com/112586815/226301188-9643b804-e377-4343-bc56-7819b3557cab.mov

src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationContainer/ManageReputationContainer.tsx
ArmandoGraterol: Yeah, this branch doesn't have those changes. Until now that I'm rebasing it

src/components/shared/NoPermissionMessage/NoPermissionMessage.tsx
ArmandoGraterol: You are right, it does return `undefined`. I do remember adding the `null` because at some point Typescript was yelling me about it :thinking: But that doesn't seem to be case for `TransferFundsDialogForm` anymore

ArmandoGraterol:
@jakubcolony Hey! I think I've addressed all of the issues you found except for the error when you input `.`, in case the fix you uploaded in the other PR deals with that.

Feel free to re-review and let me know if the fixes are working for you

src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationContainer/ManageReputationContainer.tsx
```diff
@@ -0,0 +1,187 @@
...
+            },
+          ),
+      })
+      .required();
+    smiteValidationSchema = defaultValidationSchema.concat(
+      amountValidationSchema,
+    );
+  }
+
+  const nativeTokenDecimals = nativeToken?.decimals || DEFAULT_TOKEN_DECIMALS;
```
jakubcolony, line 124: Could you use the `getTokenDecimalsWithFallback` method instead?

src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationContainer/ManageReputationContainer.tsx
```diff
@@ -0,0 +1,187 @@
...
+        amount: '',
+        annotation: '',
+      }}
+      actionType={actionType}
+      validationSchema={smiteValidationSchema || defaultValidationSchema}
+      onSuccess={close}
+      transform={transform}
+    >
+      {({ getValues }) => {
+        const values = getValues();
```
jakubcolony, line 159: Should that be `watch`?

jakubcolony (approved):
Nice work!

There's just a few `getValues` that you could probably change to `watch` but other issues have been fixed

>EDIT: Forgot to mention, also I haven't been able to fix the issue with the smite reputation max button, as for now, it seems like userReputation is not being updated fast enough and therefore when you select a new user or domain, the user reputation used by the schema is the old one. Still looking into a solution for this one, figured I let you know in case you have any ideas or suggestions to fix this.

Could you possibly move the logic for getting the user reputation to the schema test? Also I think you'll need to call `trigger` whenever the selected user/team changes to re-run the validation

src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationContainer/ManageReputationContainer.tsx
ArmandoGraterol: True, I forgot about this function

PR #228, opened 2023-01-25
Add Actions model

### LINKED ISSUE #203: Create Actions Model (Schema)
- No comments (will be handled separately)
- Keep simple for now, can be amended as we implement more actions (only handle Actions that we care about for now, no generics)
- Should be tied to a chain, transaction hash and a colony

### Actions Data Flow

cdapp-actions-data-flow

jakubcolony:
## Description

Resolves #203


jakubcolony:
This PR only focused on adding a basic action model, there have now been some subsequent changes to it in #232 and #245 so I'll close this PR

Issue #207, opened 2023-01-22
Port over Actions List component


rdig:
- Bring over, cleanup (imports) and if needed refactor the actions list component
- Needs to support pagination, but don't handle it here. Will be handled by #210 
- No comments (will be handled separately)
- Transactions log is not part of this
- Still has the sort functionality (newest/oldest)
- Don't handle the various tags (Passed, Forced, etc) as they will be handled by the Motions project
- Don't port over / handle the action notification logic (the one used for Recovery Mode)

### Actions Data Flow

Below is a rough outline of how the Actions Data Flow should work

![cdapp-actions-data-flow](https://user-images.githubusercontent.com/1193222/213923691-9f7f6512-861d-4e0a-97d3-4997f7e0d0cf.png)

willm30:
MERGED #231

PR #229, opened 2023-01-26
Refactor Actions List Item

willm30:
## Description

This PR refactors the `ActionsListItem` to:

a) Use a base `ListItem` component that can be shared with `DecisionsListItem`
b) Handle all the title-related logic outside the component.

# Testing

More of a straight-forward code review, just make sure the ActionListItem works the same as in #231.

**New stuff** ✨

* `ListItem`
* Mapping between action types and message descriptor values in `en-actions`

**Changes** 🏗

* Combined `SetUserRoles` in `en-actions` and updated `formatRolesTitle` utility
* Minor tidying of `InputLabel`
* Moved `ActionsListItem` to `ColonyActions` since it's no longer a shared component.

Contributes to #207


src/Entry.tsx
```diff
@@ -41,7 +40,6 @@ const Entry = ({ store }: Props) => {
         ...actionMessages,
         ...eventsMessages,
         ...systemMessages,
-        ...motionMessages,
```
willm30, line 44: We seem to have moved away from having two separate files for actions and motions messages, which makes sense since they share the same set of messages, so have removed.

src/components/common/ColonyActions/ActionsListItem/ActionsListItem.tsx
```diff
@@ -0,0 +1,109 @@
...
+//     motionState === MotionState.FailedNoFinalizable;
+
+//   return motionId && !isMotionFinished;
+// };
+
+interface Props {
+  item: FormattedAction;
+}
+
+const ActionsListItem = ({
```
willm30, line 44: There are two parts to this refactor. First is the abstraction of the view `ListItem`, which will also be shared by `DecisionListItem`.

The second is handling the logic, most of which was dedicated to formatting the title correctly. This is now handled by `getActionListItemTitleValues`.

The commented out sections refer to motions functionality.

src/components/common/ColonyActions/ActionsListItem/config/config.ts
```diff
@@ -0,0 +1,102 @@
...
+        ActionTitleMessageKeys.Recipient,
+      ];
+
+    default:
+      return [];
+  }
+};
+
+/* Returns the correct message values according to the action type. */
+const getActionListItemTitleValues = (
```
willm30, line 84: Before we were passing the [`FormattedMessage`](url) component a massive object with all possible title values. Now we fetch the relevant values, according to the action type, and only pass in those.

All this needs to be maintained is that the mapping above gets updated every time a new message is added at `en-actions`.

src/components/common/ColonyActions/ActionsListItem/config/helpers.tsx
```diff
@@ -0,0 +1,146 @@
+import React from 'react';
```
willm30, line 1: I can see this file being merged with `config.ts` in the event the data wrangling requirements are simplified (e.g. no async) and we don't have much to handle in the component. But for now I've separated as it makes it easier to see what's going on in `config`. 

This file is basically just responsible for ensuring that the items object has key value pairs that match what we're expecting in our message descriptors.

src/components/common/ColonyActions/ActionsListItem/config/helpers.tsx
```diff
@@ -0,0 +1,146 @@
...
+    ),
+    reputationChangeNumeral: <Numeral value={reputationChange} />,
+    reputationChange,
+    rolesChanged: formattedRolesTitle.roleTitle,
+    toDomain: findDomain(item.toDomain, colony)?.name,
+  };
+
+  /*
+
+Note that the following transformations also exist in the Dapp. Because they are async,
```
willm30, line 59: I've left this comment here as a reminder of outstanding transformations that hopefully will be handled by the cdapp equivalent of the actions resolver.

Can't do anything in this pr since we're using mock data.

src/components/shared/ListItem/ListItem.css
```diff
@@ -65,23 +60,16 @@
   background-color: var(--colony-blue);
 }
 
-.stateYellow {
+.stateDraft {
```
willm30, line 63: For decisions

src/components/shared/ListItem/ListItem.tsx
```diff
@@ -0,0 +1,124 @@
...
+  status?: ListItemStatus;
+  /** The tag to be displayed next to the title */
+  tag?: MotionState;
+  /** A title */
+  title: Message;
+  /** Values for the react-intl interpolation */
+  titleValues: UniversalMessageValues;
+}
+
+const ListItem = ({
```
willm30, line 59: This is the same as in `simple-decisions`

src/components/shared/ListItem/ListItem.tsx
```diff
@@ -0,0 +1,124 @@
...
+  return (
+    <li>
+      <div
+        className={getMainClasses({}, styles, {
+          [ListItemStatus[status]]: !!status,
+        })}
+        onClick={onClick}
+        onKeyDown={onClick}
+        role="button"
+        tabIndex={0}
```
willm30, line 80: Linter asks for this since we're making a non-interactive element interactive. I would've merged this div with the `<li>` wrapper but linter doesn't like that either (again, to do with making non-interactive elements interactive)

src/components/shared/ListItem/ListItem.tsx
```diff
@@ -0,0 +1,124 @@
...
+      >
+        <div
+          className={styles.avatar}
+          onClick={(e) => e.stopPropagation()}
```
willm30, line 84: So when you click on the avatar it doesn't fire the parent div's click event

src/hooks/useFormatRolesTitle.ts
```diff
@@ -1,80 +0,0 @@
-import { isEmpty } from 'lodash';
```
willm30, line 1: This doesn't need to be a hook any longer so moved to `~utils/colonyActions`

src/i18n/en-actions.ts
```diff
@@ -26,28 +26,35 @@ const actionsMessageDescriptors = {
       ${ColonyMotions.EmitDomainReputationPenaltyMotion} {Smite {recipient} with a {reputationChangeNumeral} {reputationChange, plural, one {pt} other {pts}} reputation penalty}
       ${ColonyActions.EmitDomainReputationReward} {Award {recipient} with a {reputationChangeNumeral} {reputationChange, plural, one {pt} other {pts}} reputation reward}
       ${ColonyMotions.EmitDomainReputationRewardMotion} {Award {recipient} with a {reputationChangeNumeral} {reputationChange, plural, one {pt} other {pts}} reputation reward}
+      ${ColonyActions.SetUserRoles} {{rolesChanged} in {fromDomain} {direction} {recipient}}
+      ${ColonyMotions.SetUserRolesMotion} {{rolesChanged} in {fromDomain} {direction} {recipient}}
       other {Generic action we don't have information about}
     }`,
-  [`action.${ColonyActions.SetUserRoles}.assign`]: `Assign the {roles} in {fromDomain} to {recipient}`,
```
willm30, line 31: Simplified these. Exact text is handled by `formatRolesTitle` in ~utils/colonyActions

src/i18n/en-actions.ts
```diff
@@ -26,28 +26,35 @@ const actionsMessageDescriptors = {
...
-  [`action.${ColonyMotions.SetUserRolesMotion}.assign`]: `Assign the {roles} in {fromDomain} to {recipient}`,
-  [`action.${ColonyActions.SetUserRoles}.remove`]: `Remove the {roles} in {fromDomain} from {recipient}`,
-  [`action.${ColonyMotions.SetUserRolesMotion}.remove`]: `Remove the {roles} in {fromDomain} from {recipient}`,
-  [`action.${ColonyActions.SetUserRoles}.assignAndRemove`]: `{roles} in {fromDomain} to/from {recipient}`,
-  [`action.${ColonyMotions.SetUserRolesMotion}.assignAndRemove`]: `{roles} in {fromDomain} to/from {recipient}`,
   'action.type': `{actionType, select,
       ${ColonyActions.WrongColony} {Not part of the Colony}
       ${ColonyActions.Payment} {Payment}
+      ${ColonyMotions.PaymentMotion} {Payment}
```
willm30, line 36: Added motions here since they've already been added to `action.title` (which I think is a good idea since actions and motions share the same messages)

src/utils/intl.ts
```diff
@@ -51,6 +50,26 @@ const isMessageDescriptor = (message?: Message): message is MessageDescriptor =>
 
 const { formatMessage: formatIntlMessage } = intl<ReactNode>();
 
+const addKeyToFormattedMessage = (
```
willm30, line 53: Required when complex message values are included in a component that is part of a list (e.g. actions list item)

src/components/common/ColonyActions/ActionsListItem/ActionsListItemMeta.tsx
```diff
@@ -0,0 +1,82 @@
...
+        title={commentCountTitle}
+      />
+      <Numeral value={commentCount} title={commentCountTitle} />
+    </span>
+  );
+};
+
+interface ActionsListItemMetaProps {
+  commentCount: any;
+  fromDomainId: any;
```
jakubcolony, line 61: Could better types be provided here?

src/components/common/ColonyActions/ActionsListItem/config/config.ts
```diff
@@ -0,0 +1,102 @@
...
+  item: FormattedAction,
+  colony?: Colony,
+) => {
+  const updatedItem = mapItemToExpectedFormat(item, colony);
+  const keys = getMessageDescriptorKeys(item.actionType);
+  const titleValues = keys.reduce<UniversalMessageValues>(
+    (values, key) => {
+      // eslint-disable-next-line no-param-reassign
+      values[key] = updatedItem[key];
+      return values;
```
jakubcolony, line 94: Could this be rewritten so that it doesn't mutate the `values` (meaning we can get rid of the rule disable)?
For example:
```js
return {
  ...values,
  [key]: updatedItem[key]
}
```

src/components/common/ColonyActions/ActionsListItem/config/helpers.tsx
jakubcolony: Local resolvers are [a deprecated feature](https://www.apollographql.com/docs/react/local-state/local-resolvers/), so custom hooks sound like the best next option

jakubcolony (approved):
LGTM, great idea with abstracting this!

src/components/common/ColonyActions/ActionsListItem/config/helpers.tsx
willm30: @jakubcolony Or have all/most of the transformation handled by the block ingestor? I.e. it picks up the on-chain event and adds to the db an actions item that's as close as possible (minus the jsx) to the format expected by the actions list item. (E.g. providing a User object instead of an address string). Then we query the db for that action to display in the list. 

Isn't that more or less the flow we're expecting to implement anyway? 
cdapp => on-chain event => block ingestor => db => cdapp

src/components/common/ColonyActions/ActionsListItem/ActionsListItemMeta.tsx
```diff
@@ -0,0 +1,82 @@
...
+    id: `${displayName}.domain`,
+    defaultMessage: 'Team {domainId}',
+  },
+  titleCommentCount: {
+    id: `${displayName}.titleCommentCount`,
+    defaultMessage: `{formattedCommentCount} {commentCount, plural,
+        one {comment}
+        other {comments}
+      }`,
+  },
```
rdig, line 26: We won't have comments enabled for the initial launch.

So either remove these, or just comment them out

src/components/common/ColonyActions/ActionsListItem/config/helpers.tsx
rdig: That's kinda the general idea of the block investor anyway, to handle as much data processing as possible, so we can offload it from the client.

Granted, this is not possible in all cases, but where that's the case, we should strive to achieve that.

src/components/shared/ListItem/ListItem.tsx
rdig, line 84: Shouldn't this be handled inside a callback ?

This is because, having it like this, will instantiate 2 new anonymous functions on every render and subsequent re-renders.

Having it as a callback instantiates it just once.

rdig (approved):
Very nice work again.

Left some comments, one of which requires attention,  but other than that, it's good to go.

src/types/actions.ts
```diff
@@ -129,6 +129,11 @@ export enum ColonyAndExtensionsEvents {
   AgreementSigned = 'AgreementSigned',
 }
 
+export type ActionUserRoles = {
+  id: ColonyRole;
+  setTo: boolean;
+};
+
```
jakubcolony, line 136: Could you clean up the imports in this file? They still contain the old `~core` or `~data`

PR #231, opened 2023-01-30
Port Actions List

willm30:
Note: Only about 500 actual diffs. 

The rest is:
300: cloning the Select component for use with `HookForm` (without changing anything else), 
200: mock data, 
200: Basic wiring of ActionListItem (which gets a full refactor in #229)

## Description

This PR wires in the Actions List using mock data. Consider it phase 1. Phase 2 will be adapting it where necessary to work with the `ActionItem` data model we end up using.

action

# Testing

Clicking the list items won't do anything since they're not real transactions. Transactions log isn't yet wired in.

So really it's a case of double checking sorting, (fake) pagination &  the user avatar popover, and that every else looks as expected. Opened #233 to track  mobile styles.

**New stuff** ✨

* Wired in Actions List using mocked data.

**Changes** 🏗

* Select component to HookForm
* Removed a couple of redundant props from User Avatar
* Light refactoring to `FriendlyName`, & extracted `ColonyActionsHeading` and `Sort Controls` to own components
* Added changes from `simple-decisions` to `ColonyHomeLayout`

**Deletions** ⚰️

* ActionsList.css not used.

## TODO
- [x] Bring over, cleanup (imports) and if needed refactor the actions list component
- [x] Needs to support pagination, but don't handle it here. (Have just copied the fake pagination like we had in the dapp)
- [x] Still has the sort functionality (newest/oldest)

Contributes to #207


rdig:
Why are there two PRs with the same name ? This and #229 ?

willm30:
I've split up one large pr into two smaller ones. The other one is focused on refactoring Actions Item List. Name now reflects that.

jakubcolony (approved):
LGTM 👍 My only request would be to keep the Test Mint Tokens button for now, until we can create actions using the dialog

src/components/common/ColonyActions/ColonyActions.tsx
```diff
@@ -1,33 +1,184 @@
...
 
-import { ActionTypes } from '~redux';
-import Button, { ActionButton } from '~shared/Button';
+import { SpinnerLoader } from '~shared/Preloaders';
+import LoadMoreButton from '~shared/LoadMoreButton';
+import { SortOptions } from '~shared/SortControls';
+import ActionsList from '~shared/ActionsList';
 
-const displayName = 'Common.ColonyActions';
+import { actionsSort, ActionsListHeading } from '.';
```
rdig, line 9: I like this way of importing, however it's really bad for readability.

Don't need to do anything about it, just wanted to throw my 5 cents in :)

src/components/common/ColonyHome/ColonyHomeLayout.tsx
```diff
@@ -17,6 +18,10 @@ import ColonyDomainDescription from './ColonyDomainDescription';
 
 import styles from './ColonyHomeLayout.css';
 
+const isExtensionsRoute = (params: Params) => {
+  return params['*'] === 'extensions';
+};
```
rdig, line 23: This is temporary, right ?

src/components/shared/UserAvatar/UserAvatar.tsx
```diff
@@ -25,33 +22,30 @@ interface Props extends Pick<AvatarProps, 'size' | 'className' | 'notSet'> {
...
   /** The corresponding user object if available */
   user?: User | null;
-
+  /** Use the user's thumbnail instead of full-size avatar */
   preferThumbnail?: boolean;
 }
 
 const displayName = 'UserAvatar';
 
 const UserAvatar = ({
```
rdig, line 31: I don't know if it makes sense anymore but I always wanted to get rid of the `notSet` prop since it was added as a crutch the first time around.

Maybe not necessarily now, but at least when we redo the UI we definitely have to get rid of it

src/types/actions.ts
```diff
@@ -133,12 +133,12 @@ export interface FormattedAction {
   id: string;
   status?: ItemStatus;
   actionType: ColonyActions | ColonyMotions;
-  initiator: Address;
-  recipient: Address;
+  initiator: User;
+  recipient: User;
```
rdig, line 137: Same with the recipient

rdig (approved):
Very nice work. I left some notes, but nothing that would prevent your from merging this.

I also appreciate the comments left to explain the various decisions taken :100:

src/components/common/ColonyHome/ColonyHomeLayout.tsx
willm30: Is that what you were asking?

PR #232, opened 2023-01-30
List actions query

### LINKED ISSUE #205: Create / adapt List Actions Query
- Please see the various "rules" and edge cases present in the current resolver as they will need to be accounted for: https://github.com/JoinColony/colonyDapp/blob/master/src/data/resolvers/colonyActions.ts
- Should be as "light" as possible to only load the required amount of data for the list
- Needs to support pagintation
- No comments for now (will be handled at a later date)

### Actions Data Flow

cdapp-actions-data-flow

jakubcolony:
## Description

This PR adds `GetColonyActions` query and wires it in with `ColonyActions` and related components.

image

Pagination will be added in #245.

## Testing
The ingestor image has been modified to run [block-ingestor#11](https://github.com/JoinColony/block-ingestor/pull/11), so you can start everything as usual.
1. Create a colony (you can't use the auto created ones to mint tokens) **NOTE**: The extension adding step will fail, just ignore it and go to the created colony
2. Click on the Test Mint Tokens button a few times.
3. Verify the actions appear in the list after reloading. The sorting should work, but it'll be reworked in #245.

Resolves #205


amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -378,3 +396,22 @@ type CurrentVersion @model {
...
+  id: ID!
+  colonyId: ID! @index(name: "byColony", queryField: "getActionsByColony")
+  colony: Colony! @belongsTo(fields: ["colonyId"])
+  type: ColonyActionType!
+  blockNumber: Int!
+  transactionHash: String!
+  initiatorAddress: ID
+  initiator: User @hasOne(fields: ["initiatorAddress"])
+  recipientAddress: ID
+  recipient: User @hasOne(fields: ["recipientAddress"])
```
jakubcolony, line 410: As discussed in a call with Raul and Will, this will be updated to support different types of initiator/recipient (e.g. Colony, Extension)

codegen.js
```diff
@@ -40,6 +40,11 @@ const codegen = async () => {
           ],
         },
       },
+      config: {
+        scalars: {
+          AWSDateTime: 'string',
```
jakubcolony, line 45: If we don't specify the config for a custom scalar, the generated type is any

willm30: I think we use others like AWSEmail that could also be added here?

src/components/common/ColonyActions/ActionsListItem/ActionsListItem.tsx
```diff
@@ -69,6 +69,8 @@ const ActionsListItem = ({
   //   motionState,
   // );
 
+  const status = ListItemStatus.Defused;
```
willm30, line 72: Is this a temporary solution? I'm not certain on how "NeedAction/NeedAttention" were used in the dapp but if we don't plan on using them they could be removed from the enum.

src/components/common/ColonyActions/ActionsListItem/ActionsListItem.tsx
```diff
@@ -5,7 +5,7 @@ import { Placement } from '@popperjs/core';
 import UserAvatar from '~shared/UserAvatar';
 import ListItem from '~shared/ListItem';
 import { ListItemStatus } from '~shared/ListItem/ListItem';
```
willm30, line 7: can drop the second `/ListItem`

willm30 (approved):
Nice one, working for me as per the pr description.

src/components/common/ColonyActions/ActionsListItem/ActionsListItem.tsx
jakubcolony: Yes, this is temporary (we don't have `status` field on `ColonyAction` yet)

rdig (approved):
Looks good to me

ArmandoGraterol (approved):
Nice work Jakub :ok_hand:

PR #238, opened 2023-02-01
Refactor ingestor image to depend on amplify 

### LINKED ISSUE #137: Refactor container orchestration to support block-ingestor data writing on startup
As described in [block-ingestor#1](https://github.com/JoinColony/block-ingestor/pull/1), the `ExtensionAddedToNetwork` event is emitted before Amplify container has started, therefore preventing the `SetCurrentVersion` mutation from running.
@rdig suggested it should be possible to modify the compose orchestration config in such a way that Amplify starts before the events are emitted.

jakubcolony:
## Description

This PR adds a healthcheck for amplify image, and makes block-ingestor dependent on amplify so that it can call the necessary mutations on startup (tracking extensions for now, but I expect this will also be needed for actions and motions)

It also fixes the failing colony creation by refactoring hardcoded OneTxPayment version in `colonyCreate` saga to programatically get it from the db (something I missed while bumping the colony network version), and removes the `setCurrentVersion` mutation calls from temp-create-data script.

Resolves #137

### DIFF
--- a/src/redux/sagas/colony/colonyCreate.ts
+++ b/src/redux/sagas/colony/colonyCreate.ts
@@ -24,10 +24,7 @@ import {
   GetTokenFromEverywhereQueryVariables,
 } from '~gql';
 import { ColonyManager, ContextModule, getContext } from '~context';
-import {
-  DEFAULT_TOKEN_DECIMALS,
-  LATEST_ONE_TX_PAYMENT_VERSION,
-} from '~constants';
+import { DEFAULT_TOKEN_DECIMALS } from '~constants';
 import { ActionTypes, Action, AllActions } from '~redux/index';
 import { createAddress } from '~utils/web3';
 import { TxConfig } from '~types';
@@ -47,6 +44,7 @@ import {
   getColonyManager,
 } from '../utils';
 import { createTransaction, createTransactionChannels } from '../transactions';
+import { getOneTxPaymentVersion } from '../utils/extensionVersion';
 
 interface ChannelDefinition {
   channel: Channel<any>;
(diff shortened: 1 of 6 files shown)


src/redux/sagas/utils/extensionVersion.ts
```diff
@@ -0,0 +1,24 @@
...
+    GetCurrentExtensionVersionQuery,
+    GetCurrentExtensionVersionQueryVariables
+  >({
+    query: GetCurrentExtensionVersionDocument,
+    variables: {
+      extensionHash: getExtensionHash(Extension.OneTxPayment),
+    },
+  });
+
+  return data.getCurrentVersionByKey?.items[0]?.version ?? null;
```
rdig, line 23: Why not default to `1` _(the first version that can exist)_ in the event of the query failing ?

```suggestion
  return data.getCurrentVersionByKey?.items[0]?.version ?? 1;
```

src/redux/sagas/colony/colonyCreate.ts
```diff
@@ -439,12 +437,10 @@ function* colonyCreate({
...
        */
+      const oneTxHash = getExtensionHash(Extension.OneTxPayment);
+      const oneTxVersion = yield call(getOneTxPaymentVersion);
       yield put(
-        transactionAddParams(deployOneTx.id, [
-          getExtensionHash(Extension.OneTxPayment),
-          /* @TODO: get latest version of OneTxPayment extn programatically */
-          LATEST_ONE_TX_PAYMENT_VERSION || 0,
-        ]),
+        transactionAddParams(deployOneTx.id, [oneTxHash, oneTxVersion || 0]),
```
rdig, line 443: And if you default `getOneTxPaymentVersion` to `1` if no data is available from the query, than you can remove the fallback from here

```suggestion
        transactionAddParams(deployOneTx.id, [oneTxHash, oneTxVersion]),
```

Alternatively, if you don't want to do this change, make sure to default to `1`, not `0`, as `1` is the first version that can exist, `0` cannot

rdig (approved):
Left some comments, but this is good otherwise

willm30 (approved):
Nice one, can confirm versions are being added to the db by the block ingestor on startup

Issue #209, opened 2023-01-22
Port over the Action page (individual)


rdig:
- Port over the action page components, but just the ones that deal with actions
- No motions
- No comments
- Route stays the same for now (although it will need to include the chain id at some point)
- redirect rules are still relevant (ie: redirect to 404 if the user tries to load an action not created by this colony, or if the transaction hash is not found)
- Just handle the "normal" actions, don't handle the Recovery Mode action for now

### Actions Data Flow

Below is a rough outline of how the Actions Data Flow should work

![cdapp-actions-data-flow](https://user-images.githubusercontent.com/1193222/213923691-9f7f6512-861d-4e0a-97d3-4997f7e0d0cf.png)

Issue #233, opened 2023-01-31
Port Colony Actions List mobile styles


willm30:
The ColonyHome/Layout mobile styles are handled here: https://github.com/JoinColony/colonyCDapp/pull/101

So when those changes are available in this branch, we need to make sure the Colony Actions List is also mobile responsive.

Also wire in ColonyHomeInfo from #239

PR #101, opened 2022-11-25
Port over members page components

chinins:
## Description

This PR ports over components for the members page and uses members query to pull the data.

Does not port:

- Dialogs connected to members actions
- Banned/whitelisted data
- Search bar
- Member Actions
- Filters

Screenshot 2022-12-13 at 20 40 58

## TO DO
- [x] Create issue to migrate members page to React Hook Form #157 
- [x]  Create issue to port Search and Filter functionality #181 
- [x] Create issue to port Member Actions #182

resolves #51

PR #239, opened 2023-02-01
Port Actions Page

willm30:
## Description

This PR ports the actions page. 

domainchange

For a reminder of how it looks in the dapp, see [here](https://xdai.colony.io/colony/meta/tx/0x4f0d5b7a62855ae8106214cafe4eb2b1f14fa323798aae7d21a81509ea68444f).

It does not port:
-- the feed
-- comments

## Testing

-> Click on an item in the actions list. I'm still using mocked data so we can test multiple screens easily. I have, though, commented out some sections that aren't yet supported by the `ColonyActions` model. I have tested these also, but leaving out to ensure the components remain consistent with the model.

-> Navigate to an invalid tx hash, should see:

er

 
**New stuff** ✨

* Actions Page and related components (ActionsPageEvent, DefaultAction etc)
* Generate the event description using a map and helper functions, much like generating the action title (`getEventTitleValues`).

**Changes** 🏗

* Tweaks to the DetailsWidget introduced in the previous pr so it works as expected in this one.
* Extracted some logic from `Numeral` so it can be reused in `~utils/reputation` (`formatReputationChange`)
* Tidy en-events

## TODO

- [x] Port over the action page components, but just the ones that deal with actions
- [x] No motions
- [x] No comments
- [x] No recovery mode
- [x] Route stays the same for now (although it will need to include the chain id at some point)
- [x] redirect rules are still relevant (ie: redirect to 404 if the user tries to load an action not created by this colony, or if the transaction hash is not found) -> Note, dapp is not redirecting to 404 but displaying something similar. I'm doing the same as in the dapp (`TransactionNotFound`).

Contributes to #209


src/utils/events.ts
```diff
@@ -0,0 +1,183 @@
...
+interface ColonyMetadataChanges {
+  nameChanged: boolean;
+  logoChanged: boolean;
+  tokensChanged: boolean;
+  verifiedAddressesChanged: boolean;
+}
+
+type DomainMetadata = Pick<Domain, 'color' | 'name' | 'description'>;
+
+export const getColonyMetadataMessageValues = (
```
willm30, line 15: Dynamically generating the message text according to the values that changed

src/utils/events.ts
```diff
@@ -0,0 +1,183 @@
...
+const getDomainMetadataValues = (
+  values: (string | Falsy)[],
+  direction: 'from' | 'to',
+) => {
+  const fallback = direction === 'from' ? ', but the values are the same' : '';
+  const vals = values.filter(Boolean) as string[];
+  const color = vals.find(isDomainColor);
+  let restValuesFormatted = fallback;
+
+  if (vals.length) {
```
willm30, line 115: This is a bit more complex because the colour is wrapped in a jsx color tag, so we can't just concatenate strings and return a string. It returns the string without the colour, and the colour to be formatted separately, and they get properly formatted by `getDomainMetadataValues` in `~common/ColonyActions/helpers`

src/utils/strings/index.ts
```diff
@@ -60,6 +60,18 @@ export const humanReadableFileSize = (size: number) => {
...
 
+/**
+ * Capitalize a word (converts the word to lower case, except for the first letter)
+ *
+ * @method convertToCapitalized
+ *
+ * @param {string} word The word / string to capitalize
+ * @return {string} The capitalized string
+ */
+export const convertToCapitalized = (word: string): string => {
```
willm30, line 71: Different from the function above in that it will first convert the word to lowercase before capitalising the first letter.

src/components/common/ColonyActions/ActionsPage/TransactionStatus/TransactionStatus.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import { defineMessages, FormattedMessage } from 'react-intl';
+
+import { Tooltip } from '~shared/Popover';
+import { getMainClasses } from '~utils/css';
+
+import { Status } from '../staticMaps';
```
jakubcolony, line 7: I think this naming lacks the word "transaction" in it, so that it's clear straightaway about which status we're talking

jakubcolony (changes requested):
Overall, this looks good!
image

image

I just spotted this issue (not sure if part of this PR, but still):
image

Note the address displayed in the detail row is different to the one in the popover. It seems like the popover simply displays the current user details, not one with the given address...

src/components/common/ColonyActions/ActionsPage/ActionsPage.tsx
```diff
@@ -0,0 +1,73 @@
...
+    id: `${displayName}.loading`,
+    defaultMessage: `Loading Transaction`,
+  },
+  unknownTransaction: {
+    id: `${displayName}.unknownTransaction`,
+    defaultMessage: `Unknown Transaction`,
+  },
+});
+
+type ActionsPageParams = Record<'colonyName' | 'transactionHash', string>;
```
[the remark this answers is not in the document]
willm30: More for the keys themselves than the types

src/components/common/ColonyActions/ActionsPage/ActionsPage.tsx
```diff
@@ -0,0 +1,73 @@
+import React from 'react';
```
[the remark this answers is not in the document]
willm30: Have renamed to ActionDetailsPage

src/components/common/ColonyActions/ActionsPage/TransactionStatus/TransactionStatus.tsx
willm30: Renamed to TransactionStatuses to avoid conflict with the component

willm30:
> Note the address displayed in the detail row is different to the one in the popover. 

@jakubcolony Good catch, fixed.

src/styles/variables.css
```diff
@@ -21,7 +21,9 @@
   --light-grey-pink: rgb(244, 238, 242); /* Used for user profile gradient start color */
   --light-ash: rgb(231, 242, 246); /* Used for user profile gradient end color */
   --light-purple: rgb(200, 214, 247); /* Used for some box shadows */
+  --light-purple-1: 200, 214, 247;
   --grey-purple: rgb(118, 116, 139); /* Used for the wallet connect button in the user navigation */
+  --grey-purple-1: 118, 116, 139; /* for color-mod replacement */
```
rdig, line 26: I really don't like the `-1` suffixed naming scheme.

Why not:
```suggestion
  --grey-purple-color-mod: 118, 116, 139;
```

Or yet another approach:
```css
--light-purple-raw: 200, 214, 247;
```

Which you can then pass both to `rgb` and `color-mod`:
```css
--light-purple: rgb(var(--light-purple));
...
background-color: rgba(var(--light-purple-raw), 0.33);
```

src/utils/events.ts
rdig: As I mentioned previously, we should store metadata history in an array format, so this will end up having to account for that

src/utils/events.ts
```diff
@@ -0,0 +1,182 @@
...
+    }
+  }
+
+  return domainMetadata;
+};
+
+const isDomainColor = (color: string): color is DomainColor =>
+  convertToCapitalized(color) in DomainColor;
+
+const getDomainMetadataValuesAndColor = (
```
rdig, line 106: Same here

src/utils/strings/index.ts
rdig: Yeah, but the naming (for both) is a bit confusing. 

How about `capitalizeFirstLetter` and `capitalizeWord` ?

rdig (approved):
Very nice work. Left some comments that need to be addressed, but overall quality work here! :+1: 

Regarding this:
> It does not port:
> -- the feed

I think the feed, as I originally implemented it, has run it's course. It's format and constraints is one of the biggest reasons for why we ended up with that spagetti monster of a transformer inside the Dapp. 

So since we can now get data better using clever queries, I think it might be time to remove it / refactor it, and just extract the components we need from it.

Julianahlk (approved):
Very nice work Will! Working well for me. 👍

src/styles/variables.css
willm30: I notice elsewhere in others' prs that color-mod was working just fine and have now confirmed that for myself in this branch. Don't know what I was doing wrong to make me think it'd been deprecated. I've reverted everything in this branch back to normal. 

Apologies to anyone I may have confused by suggesting color-mod was deprecated😬

src/components/common/ColonyActions/ActionDetailsPage/DefaultAction/DefaultActionContent.tsx
```diff
@@ -0,0 +1,45 @@
...
+      className={styles.heading}
+      data-test="actionHeading"
+      text={{ id: 'action.title' }}
+      textValues={getActionTitleValues(item, colony)}
+    />
+    <ActionsPageFeed actionData={item} />
+    {/*
+           *  @NOTE A user can comment only if he has a wallet connected
+           * and a registered user profile
+
```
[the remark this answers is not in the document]
willm30: Think it makes sense to remove 👍

src/components/common/ColonyActions/ActionDetailsPage/DefaultAction/DefaultAction.tsx
```diff
@@ -0,0 +1,45 @@
...
+  if (!colony) {
+    return null;
+  }
+
+  const isVotingExtensionEnabled = false;
+  const motionStyles = MOTION_TAG_MAP[MotionState.Forced];
+
+  return (
+    <div className={styles.main}>
+      {/* {isMobile && <ColonyHomeInfo showNavigation isMobile />} */}
```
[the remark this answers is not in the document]
willm30: Yes, is introduced in #101 and tracked in #233

src/components/common/ColonyActions/ActionDetailsPage/ActionDetailsPage.tsx
```diff
@@ -0,0 +1,77 @@
...
+    (item) => item.transactionHash === transactionHash,
+  );
+
+  const createdAt = action?.createdAt;
+  // const status = action?.transactionStatus;
+
+  const isValidTx = isTransactionFormat(transactionHash);
+  const events = ['event']; // to be taken from real data
+
+  if (!isValidTx || !events?.length || !action) {
```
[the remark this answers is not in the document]
willm30: Have updated

src/utils/events.ts
willm30: @jakubcolony I have left a comment in the code but just pinging you here as well since I imagine you'll be looking after this

src/components/common/ColonyActions/helpers/mapItemToMessageFormat.tsx
```diff
@@ -1,60 +1,125 @@
...
   };
+};
 
-  /*
+export const mapColonyEventToExpectedFormat = (
+  event: MockEvent & { eventName: ColonyAndExtensionsEvents },
+  item: FormattedAction,
+  colony?: Colony,
+) => {
+  const colonyMetadataChanges = { namedChanged: true, logoChanged: true };
```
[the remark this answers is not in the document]
jakubcolony: Thanks for pointing this out @willm30, I'll look into it

src/components/common/ColonyActions/helpers/mapItemToMessageFormat.tsx
```diff
@@ -28,16 +28,17 @@ export const mapItemToExpectedFormat = (
   //   item.decimals,
   //   item.reputationChange,
   // );
+
   return {
     ...item,
     amount: (
       <Numeral
         value={item.amount ?? 0} // @TODO: getAmount(item.actionType, item.amount)
```
jakubcolony, line 36: Could you open an issue to track it? This is for getting the amount after network fees, right?

src/graphql/queries/colony.graphql
```diff
@@ -1,3 +1,16 @@
+query CombinedColony($name: String!, $address: ID!) {
```
jakubcolony, line 1: I don't think this is the best approach. As far as I understand, there isn't a strong use case for it right now - we tend to know whether we want to get a colony by name or address, and all this is doing is making a redundant duplicated query. 

Why not add a separate, `GetFullyColonyByAddress` query and call the generated hook in `ActionDetailsPage`?

jakubcolony (changes requested):
Nice work! All looks good, but requesting changes since I'd like you to address the duplicated query comment

src/graphql/queries/colony.graphql
willm30: Sure

src/components/common/ColonyActions/helpers/mapItemToMessageFormat.tsx
willm30: Yes, left some more notes at the bottom of that file. https://github.com/JoinColony/colonyCDapp/issues/256

jakubcolony (approved):
LGTM! 👍

PR #242, opened 2023-02-03
Fix colony home layout at extensions route

willm30:
## Description

Tiny fix to use `pathname` instead of `params` to determine whether we're at `/extensions` or not in `ColonyHome`. Params wasn't including `extensions` when at `extensionsId`, so the wrong home layout was showing. 

## To Test

Go to `/extensions`. View should display correctly.
Click on an extension. View should display correctly.

That is all, it's a 5 line fix 🙂

**Changes** 🏗

* Pathname instead of params to determine whether we're at extensions in ColonyHomeLayout


jakubcolony (approved):
Good spot 👍

ArmandoGraterol:
@willm30 Yo Will, I think this PR needs a rebase

willm30:
@ArmandoGraterol Thanks, rebase complete

Julianahlk (approved):
Nice one! 👍

PR #245, opened 2023-02-05
Actions pagination

### LINKED ISSUE #210: Refactor / Implement better Actions List Pagination
This needs to account for the new data model and implement actual pagination, not just visual "trickery"

- Load at max 10 items at a time (configurable via a constant somewhere)
- Needs to support sorting (new/old)
- Not show duplicate items
- As light as possible

### Actions Data Flow

cdapp-actions-data-flow

jakubcolony:
## Description

This PR adds pagination to colony actions list

image

## Testing
1. Since you can only see actions of one type, apply the following patch to also display block numbers next to action items (it'll make it easier to verify sorting works as expected):
```
diff --git a/src/components/common/ColonyActions/ActionsListItem/ActionsListItem.tsx b/src/components/common/ColonyActions/ActionsListItem/ActionsListItem.tsx
index 6da2a0d3..67c0f38b 100644
--- a/src/components/common/ColonyActions/ActionsListItem/ActionsListItem.tsx
+++ b/src/components/common/ColonyActions/ActionsListItem/ActionsListItem.tsx
@@ -82,13 +82,14 @@ const ActionsListItem = ({
       }
       createdAt={createdAt}
       extra={
-        null // displayCountdownTimer && (
-        //   <CountdownTimer
-        //     state={motionState as MotionState}
-        //     motionId={Number(motionId)}
-        //     isFullyNayStaked={isFullyNayStaked(totalNayStake, requiredStake)}
-        //   />
-        // )
+        // null // displayCountdownTimer && (
+        // //   <CountdownTimer
+        // //     state={motionState as MotionState}
+        // //     motionId={Number(motionId)}
+        // //     isFullyNayStaked={isFullyNayStaked(totalNayStake, requiredStake)}
+        // //   />
+        // // )
+        <div>{item.blockNumber}</div>
       }
       meta={<ActionsListItemMeta fromDomainId={fromDomain ?? ''} />}
       onClick={handleActionRedirect}

```

2. Create a colony using the colony creation wizard (**NOTE**: Mint tokens will not work with a script-created colony)
3. Mint tokens at least 11 times.
5. Check that only 10 items are shown initially, in a correct order, the Load more buttons works and that you can change sorting options

Resolves #210


src/context/apolloClient.ts
```diff
@@ -14,13 +14,35 @@ const cache = new InMemoryCache({
...
       fields: {
         getActionsByColony: {
-          keyArgs: false,
-          merge(existing = {}, incoming) {
+          keyArgs: ['$colonyAddress'],
+          /**
+           * The following function takes care of merging already fetched actions
+           * with the next set of paginated actions
+           */
+          merge(existing = {}, incoming, { readField, args }) {
```
jakubcolony, line 22: This is more or less following the docs: https://www.apollographql.com/docs/react/pagination/cursor-based#keeping-cursors-separate-from-items

In our case, the `getActionsByColony` doesn't return just an array but an object with a few fields. We store them in a map (id <> action), then transform them back before returning the results in the `read` function

src/components/common/ColonyActions/ColonyActions.tsx
```diff
@@ -33,18 +33,31 @@ const MSG = defineMessages({
 //   ethDomainId?: number;
 // };
 
+const ITEMS_PER_PAGE = 10;
+
 const ColonyActions = (/* { ethDomainId }: Props */) => {
   const { colony } = useColonyContext();
 
-  const { data, loading: loadingActions } = useGetColonyActionsQuery({
+  const [sortDirection, setSortDirection] = useState<SortDirection>(
```
willm30, line 41: This all could perhaps live in a hook? Just to try and keep the component as neat as possible.

src/components/common/ColonyActions/ColonyActions.tsx
```diff
@@ -185,16 +196,25 @@ const ColonyActions = (/* { ethDomainId }: Props */) => {
...
-          {/* {loadMoreItems && (
+          <ActionsListHeading
+            sortDirection={sortDirection}
+            onSortChange={setSortDirection}
+          />
+          <ActionsList items={actions} />
+          {hasMoreItems && (
             <LoadMoreButton
-              onClick={showMoreItems}
+              onClick={() =>
```
willm30, line 208: If you were to move the above into a hook, this could be returned from it as, e.g. "handleLoadMore"

willm30 (approved):
Nice job Jakub, working for me!

src/components/common/ColonyActions/ColonyActions.tsx
rdig: I agree with this

src/context/apolloClient.ts
rdig: I like this, however I feel like the Cache file will keep growing, so I think it'll soon come time to separate it out in it's own module / multiple modules

rdig (approved):
I like this. Very nice! Left some notes, but nothing major.

src/context/apolloClient.ts
jakubcolony: I extracted cache into its own file - as it grows, we can then separate it further into more logical chunks

src/components/common/ColonyActions/ColonyActions.tsx
jakubcolony: I extracted the entire fetching/sorting/pagination logic into its own hook, `usePaginatedActions`. I'm not sure if this was your suggestions, or maybe you wanted me to extract just the sorting/pagination logic so it's reusable, but having thought about it, sorting and pagination will not always go hand in hand. I'll revisit it once there's another use case for sorting or pagination

PR #247, opened 2023-02-06
Single action query

### LINKED ISSUE #206: Create / Adapt Single Action Query
- Please see the various "rules" and edge cases present in the current resolver as they will need to be accounted for: https://github.com/JoinColony/colonyDapp/blob/master/src/data/resolvers/colonyActions.ts
- No comments for now (will be handled at a later date)
- Load the "full" action from the db

### Actions Data Flow

cdapp-actions-data-flow

jakubcolony:
## Description

This PR wires the `GetColonyAction` query with action details page UI. Further changes will follow as we handle more action types.

It also removes `decimals` and `tokenSymbol` fields from the action model since it can be retrieved via the relation with `Token` model.

image

## Testing

To test, create a colony, generate a few mint token actions (you can play with the values passed to `ActionButton` in `~common/ColonyActions` to modify the amount of tokens minted). Click on the actions you generated and verify the details match the action values.

Another thing to test is to modify the `colonyName` param in the URL so that it points to another colony, e.g.:
```
http://localhost:9091/colony/jakub/tx/0x0f2a0674e440a7e179d0d61063126bf14336d4c083deb391c270d80c601bfc2c
```
becomes
```
http://localhost:9091/colony/a/tx/0x0f2a0674e440a7e179d0d61063126bf14336d4c083deb391c270d80c601bfc2c
```
This should result in rendering the action not found view:
image

Resolves #206


docker/colony-cdapp-dev-env-block-ingestor
```diff
@@ -1,6 +1,6 @@
 FROM colony-cdapp-dev-env/base:latest
 
-ENV BLOCK_INGESTOR_HASH=bbad0279d810d2088c09c726fc8396f732ca1db5
+ENV BLOCK_INGESTOR_HASH=e082fe153e0f91a5df53d55bc01597a900f44760
```
rdig, line 3: Heads up if updating to the latest that colony network released `glwss3` which includes colony `v12`. I'm not sure the current release of colonyJS `v5` supports this yet

rdig (approved):
I'm all good with this however I just realized something _(which might be a bit off topic for this PR)_

The  current block ingestor setup doesn't work for Metatransaction, since they emit an event, with the same name _(ex: `TokensMinted`, `ColonyFundsClaimed`, etc)_, but with a different format altogether, meaning it won't be able to find data values.

We'll need to decide how to handle this different data format, or even if we want to handle it inside the same block ingestor

src/components/common/ColonyActions/ActionDetailsPage/ActionDetailsPage.tsx
```diff
@@ -33,38 +32,40 @@ const ActionDetailsPage = () => {
   const { colony } = useColonyContext();
   const { transactionHash, colonyName } = useParams<ActionDetailsPageParams>();
 
-  const action = mockActionData.find(
-    (item) => item.transactionHash === transactionHash,
-  );
-
-  const createdAt = action?.createdAt;
+  const { data, loading } = useGetColonyActionQuery({
```
willm30, line 35: I imagine the queries / hooks here being wrapped up into a single `useGetColonyAction` hook, such as https://github.com/JoinColony/colonyCDapp/blob/ddfba48a30746c2617878456cfeead0a567d4905/src/hooks/useGetColonyAction.ts

willm30 (approved):
Nice job, working as expected for me. Just a note to others that I needed to run the block ingestor separately in order for it to pick up the events. Don't know if this is just happening on my machine. 

This will probably fix the issue I was having testing your redirect pr as well so if you're changing how you handle polling in that I will wait before reviewing.

docker/colony-cdapp-dev-env-block-ingestor
jakubcolony: Thanks, this is only the ingestor image - as far as I'm aware we kept the network hash the same for now

src/components/common/ColonyActions/ActionDetailsPage/ActionDetailsPage.tsx
jakubcolony: Cool, I think it makes sense for your PR to bring those changes rather than replicating them here

Issue #253, opened 2023-02-07
Refactor `GasStation` to support arbitrary statuses


jakubcolony:
We want to be able to add arbitrary statuses (avoiding the word "messages", since it's already used in that context) to the Gas Station popover. Those should be controlled from within sagas. Currently, the Gas Station only shows wallet transactions / messages

![image](https://user-images.githubusercontent.com/112586815/217378270-802b4809-457c-4f08-ab1e-653cdc44b7b1.png)

rdig:
@jakubcolony can you refresh my memory on why we wanted this? As it feels like this is not relevant anymore, but it might just be me forgetting

PR #254, opened 2023-02-07
Implement mint tokens saga redirect

### LINKED ISSUE #211: Implement Mint Action Saga Redirect
Now that most of the Action Data Flow has been wired / created, we can handle saga redirect at the end of Action Creation.

- Note that his has to account for the race condition (time between tx sent out and the block ingestor / db data being available)

### Actions Data Flow

cdapp-actions-data-flow

jakubcolony:
## Description

This PR adds redirection at the end of the `mintTokens` saga, after waiting for the ingestor to handle the `MintTokens` event. There is infinite polling at 3s interval happening in the background, until there's an error or we receive the expected response.

https://user-images.githubusercontent.com/112586815/217384286-9031dc65-0830-420a-bb9a-2df57bd7dfda.mov

## Testing
Create a colony using the colony creation wizard.
Click on the Test Mint Tokens button.
Wait for a few seconds after the transaction has finished, you should be redirected to the corresponding actions page.

Resolves #211


src/redux/types/actions/index.ts
```diff
@@ -112,8 +114,13 @@ export type ActionTypeString = AllActions['type'];
...
 export type TakeFilter = (action: AllActions) => boolean;
 
+// To be replaced with MetaWithNavigate and deleted
 export type MetaWithHistory<M> = {
   history?: {
     push: <A>(route: A) => void;
   };
 } & M;
+
+export type MetaWithNavigate<M> = {
```
jakubcolony, line 124: The history object is no longer exposed by `react-router`, so all sagas with redirect will need to use this instead

src/redux/sagas/utils/pollAction.ts
```diff
@@ -0,0 +1,42 @@
...
+      const result = (yield apolloClient.query({
+        query: GetColonyActionDocument,
+        variables: {
+          transactionHash,
+        },
+        fetchPolicy: 'network-only',
+      })) as ApolloQueryResult<GetColonyActionQuery>;
+
+      if (result.data.getColonyAction) {
+        yield put({ type: ActionTypes.CANCEL_ACTION_POLLING });
```
rdig, line 34: I'd argue we would want to differentiate actions names if the poll action succeeds vs. it fails.

Just from a debug/logical standpoint _(since the actions don't actually do anything, other than being dispatched)_

But it would still be nice to have something like `SUCCESS_ACTION_POLLING` on success and `CANCEL_ACTION_POLLING` on fail / cancel

rdig (approved):
I like it, although I'd like us to introduce a separate action for polling success.

willm30 (commented):
Have tried to review a few times but unable to reproduce. (Building this branch completely fresh from a pc reboot). The saga completes fine but the block ingestor isn't picking up the event and so nothing is being written to the db, hence no redirect. 

This does however offer an opportunity to discuss the optimal way of handling problems in the user flow. Currently, this won't alert the user to a problem if you poll from the saga. 

However, if you redirect to the actions page, and then poll via an apollo query from there, you will at least get the actions page loading state which times out eventually, alerting the user to the fact there's something up. You also wouldn't need to duplicate the polling logic that apollo provides.

jakubcolony:
>Currently, this won't alert the user to a problem if you poll from the saga.

There's a plan of adding arbitrary messages to the Gas Station: #253 

>However, if you redirect to the actions page, and then poll via an apollo query from there, you will at least get the actions page loading state which times out eventually, alerting the user to the fact there's something up. You also wouldn't need to duplicate the polling logic that apollo provides.

That's an interesting suggestion. Thinking about it now, we also have the optimistic UI pattern that I used for extensions - maybe it could work here too. I'll look into it

willm30:
@jakubcolony I've implemented this pattern in wiring in the root motion saga: https://github.com/JoinColony/colonyCDapp/blob/ddfba48a30746c2617878456cfeead0a567d4905/src/hooks/useGetColonyAction.ts
https://github.com/JoinColony/colonyCDapp/pull/261/files#diff-a4b4b9f45aa0b72b3920e2f62a2eead1162722e6c77356053a3b1bb6c31b5157

ArmandoGraterol (approved):
LGTM :sunglasses:

Julianahlk (approved):
Very nice! Working well for me 👍

jakubcolony:
Following a conversation with Raul and Will, I'm removing the polling from inside the saga and simply pushing the user to the transaction page. The polling will be handled there and it's already implemented in #261

PR #261, opened 2023-02-09
Port root motion saga

### LINKED ISSUE #258: Wire in Root Motion
Using this issue as the first of the motion-wiring issues, and to handle some of the heavier lifting for subsequent motions sagas.

It comes in at least the following parts:

1. Wire up the root motion saga (needed for Mint Tokens, e.g.)
2. Handle the createMotion event in the block ingestor
3. Either adapt the ColonyAction model or create ColonyMotion model to host relevant motion data in the db (needs investigation / liaising with raul & jakub)
4. Send aforementioned data to db so it can be consumed by "DefaultMotion" component 
5. Pull in data from db and display at DefaultMotion

At this stage, all we're looking for is some evidence in the UI that the motion has been successfully created. 

The various stages of motion progression will be handled subsequently. #260

willm30:
## Description

This PR is responsible for wiring in the root motion saga and handling the first phase of the motion lifecycle, such that we have record of the motion in the db (via block-ingestor) and can access relevant data (such as action type) for display in the ui.

It does not handle any subsequent phases of the motion lifecycle (staking, voting etc.). This will be handled subsequently.

## Testing

Switch to this branch and run a fresh build (`npm run dev`). I've updated the block ingestor hash to point to block-ingestor#16.
 
Then go to https://github.com/JoinColony/block-ingestor/pull/16 and follow the instructions to get the reputation and the voting extension installed.

Finally, once all set up, click "Test mint tokens". It should redirect you to the relevant motion page:

Screenshot (656)

You can change the amount of tokens you mint in `ColonyActions.tsx`.

To be clear, we're just testing the UI is picking up the correct data in the mint tokens motion payload and displaying it correctly. (e.g. amount, symbol, domain id). 

We're not testing that anything else works yet in motions (voting, staking etc). The fact that some of the messages in the screenshot above aren't accurate (E.g. Steven did x) or that it goes straight to voting --> doesn't matter for the scope of this pr and will be addressed in subsequent issues.

**New stuff** ✨

* `useGetColonyAction` to handle queries at `ActionDetailsPage`

**Changes** 🏗

* Refactor action button
* Wire in root motion saga

## TODO

- [x] Wire in root motion saga
- [x] Adapt "mint token test" button to trigger it
- [x] Add necessary changes to block ingestor to pick up event (https://github.com/JoinColony/block-ingestor/pull/16)
- [x] Fire off db mutations  (https://github.com/JoinColony/block-ingestor/pull/16)
- [x] Pull data into ui and display motions page at generated tx hash

Resolves #258

### DIFF
--- a/amplify/backend/api/colonycdapp/schema.graphql
+++ b/amplify/backend/api/colonycdapp/schema.graphql
@@ -419,7 +419,7 @@ type ColonyAction @model {
   colony: Colony! @belongsTo(fields: ["colonyId"])
   type: ColonyActionType!
   blockNumber: Int!
-  transactionHash: String!
+  isMotion: Boolean
   createdAt: AWSDateTime!
   initiatorAddress: ID
   initiator: User @hasOne(fields: ["initiatorAddress"])
--- a/src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionCountdown/EscalateButton/EscalateButton.tsx
+++ b/src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionCountdown/EscalateButton/EscalateButton.tsx
@@ -46,9 +46,7 @@ const EscalateButton = ({ motionId, userAddress }: EscalateButtonProps) => {
     <div className={styles.escalation}>
       <ActionButton
         appearance={{ theme: 'blue', size: 'small' }}
-        submit={ActionTypes.MOTION_ESCALATE}
-        error={ActionTypes.MOTION_ESCALATE_ERROR}
-        success={ActionTypes.MOTION_ESCALATE_SUCCESS}
+        actionType={ActionTypes.MOTION_ESCALATE}
         transform={escalateTransform}
         text={MSG.escalate}
         /*
--- a/src/components/common/ColonyActions/ColonyActions.tsx
+++ b/src/components/common/ColonyActions/ColonyActions.tsx
@@ -1,13 +1,19 @@
 import React from 'react';
 import { defineMessages, FormattedMessage } from 'react-intl';
 import { BigNumber } from 'ethers';
+import { useNavigate } from 'react-router-dom';
 
 import { SpinnerLoader } from '~shared/Preloaders';
 import LoadMoreButton from '~shared/LoadMoreButton';
 import ActionsList from '~shared/ActionsList';
 import { ActionButton } from '~shared/Button';
-import { ActionTypes } from '~redux';
-import { useColonyContext, usePaginatedActions } from '~hooks';
+import { ActionTypes, RootMotionOperationNames } from '~redux';
+import { withMeta } from '~utils/actions';
+import {
+  useColonyContext,
+  useEnabledExtensions,
+  usePaginatedActions,
+} from '~hooks';
 
 import { ActionsListHeading } from '.';
 
@@ -32,6 +38,7 @@ const MSG = defineMessages({
 
 const ColonyActions = (/* { ethDomainId }: Props */) => {
   const { colony } = useColonyContext();
+  const navigate = useNavigate();
 
   const {
     loading: loadingActions,
@@ -41,6 +48,11 @@ const ColonyActions = (/* { ethDomainId }: Props */) => {
     hasMoreActions,
     loadMoreActions,
   } = usePaginatedActions();
+
+  // only to test root motion saga
+  const {
+    enabledExtensions: { isVotingReputationEnabled },
+  } = useEnabledExtensions();
 
   if (!colony) {
     return null;
@@ -165,18 +177,28 @@ const ColonyActions = (/* { ethDomainId }: Props */) => {
     );
   }
 
+  const isForce = false;
+  const isMotion = !!isVotingReputationEnabled && !isForce;
+  const actionType = isMotion
+    ? ActionTypes.ROOT_MOTION
+    : ActionTypes.ACTION_MINT_TOKENS;
+
+  const amount = BigNumber.from(1);
+  const transform = withMeta({ navigate });
+
   return (
     <div className={styles.main}>
       <ActionButton
-        submit={ActionTypes.ACTION_MINT_TOKENS}
-        error={ActionTypes.ACTION_MINT_TOKENS_ERROR}
-        success={ActionTypes.ACTION_MINT_TOKENS_SUCCESS}
+        actionType={actionType}
         values={{
+          operationName: RootMotionOperationNames.MINT_TOKENS,
           colonyAddress: colony.colonyAddress,
           colonyName: colony.name,
           nativeTokenAddress: colony.nativeToken.tokenAddress,
-          amount: BigNumber.from(1),
+          motionParams: [amount],
+          amount,
         }}
+        transform={transform}
         text="Test Mint Tokens"
       />
       {actions.length ? (
--- a/src/components/shared/Button/ActionButton.tsx
+++ b/src/components/shared/Button/ActionButton.tsx
@@ -4,32 +4,40 @@ import { MessageDescriptor } from 'react-intl';
 import { ActionTransformFnType } from '~utils/actions';
 import { useAsyncFunction, useMounted } from '~hooks';
 import DefaultButton from '~shared/Button';
+import { ActionTypes } from '~redux';
 import { Props as DefaultButtonProps } from './Button';
 
+const getFormAction = (
+  action: ActionTypes,
+  actionType: 'SUBMIT' | 'ERROR' | 'SUCCESS',
+) => {
+  const actionEnd = actionType === 'SUBMIT' ? '' : `_${actionType}`;
+  return `${action}${actionEnd}`;
+};
+
 interface Props extends DefaultButtonProps {
+  actionType: ActionTypes;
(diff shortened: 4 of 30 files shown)


amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -408,7 +409,7 @@ type ColonyAction @model {
   colony: Colony! @belongsTo(fields: ["colonyId"])
   type: ColonyActionType!
   blockNumber: Int!
-  transactionHash: String!
+  isMotion: Boolean
```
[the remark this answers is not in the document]
jakubcolony: @willm30 nice, I like that idea, but do you know what's the behaviour of the data field when calling the update mutation? 

The reason I'm asking is because for actions/motions that consist of more than event, you'll need to add some data fields on each of those events. So if the first event creates the action and adds two fields to the data, and then the second event wants to add another few by updating the action, will they be merged or overwritten?

src/components/shared/Button/ActionButton.tsx
```diff
@@ -4,32 +4,40 @@ import { MessageDescriptor } from 'react-intl';
...
-  error,
-  submit,
-  success,
   onSuccess,
   // @todo Remove `values` once async transform functions are supported
   values,
   transform,
   ...props
 }: Props) => {
+  const submit = getFormAction(actionType, 'SUBMIT');
```
willm30, line 38: These make more sense to live in here, right? Instead of duplicating this logic every time the ActionButton is used.

amplify/backend/api/colonycdapp/schema.graphql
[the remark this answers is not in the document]
willm30: Ah, would the problem be that we can't determine whether to merge / overwrite on an action type basis? We'd have to set the merge policy at the level of the model, which might not be appropriate for all action types (i.e. if some need merging and some need overwriting)? 

If we can't do this at a granular, action type level, and it is the case that some will need merging and some overwriting, then I guess we can't do it.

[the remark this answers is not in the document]
jakubcolony: I'll share the specific details of the case I'm dealing with right now (and you will most likely do as well):
The Payment action consist of 3 events: `PaymentAdded`, `PayoutClaimed`, `OneTxPaymentMade`
And the data associated with it is:
`
initiator
recipient
amount
tokenAddress
fundamentalChainId
fromDomain
`

`PaymentAdded` will create an action and fill out the `fromDomain` field, using a mutation with input such as:
```gql
  input: {
    id: ...,
    colony: ...,
    data: {
      fromDomain: ...
    }
  }
```

`PayoutClaimed` will update the existing action and provide info on tokenAddress and amount:
```gql
  input: {
    id: ...,
    data: {
      tokenAddress: ...,
      amount: ...
    }
  }
```

My concern is: will this update mutation override the existing `data` field, including the previously defined `fromDomain`? Or will the new fields, `tokenAddress` and `amount` get merged to create an object containing the 3 fields?

I think in any case we'd want it to behave like a standard update function, so if a field is not defined in the input, it should be left intact, and otherwise it should be overwritten.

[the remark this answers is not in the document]
willm30: Ok good to know. I'll open an issue for this in actions if that's alright with you, but the approach you outline above looks good to me.

src/hooks/useGetColonyAction.ts
```diff
@@ -0,0 +1,43 @@
...
+  const {
+    data: actionData,
+    loading: loadingAction,
+    stopPolling: stopPollingForAction,
+  } = useGetColonyActionQuery({
+    skip: skipQuery,
+    variables: {
+      transactionHash: transactionHash ?? '',
+    },
+    pollInterval: 1000,
```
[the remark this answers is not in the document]
jakubcolony: Currently, you do polling for every visit to the action page, even if it's not a redirect from a saga. I think there should be a flag stored somewhere (in query params, or route state) that would allow you to decide whether polling is necessary. 
Another thing I noticed is that the polling continues infinitely, even if the naked mole timeout is shown - shouldn't the polling stop at this point?

src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionCountdown/EscalateButton/EscalateButton.tsx
```diff
@@ -46,9 +46,7 @@ const EscalateButton = ({ motionId, userAddress }: EscalateButtonProps) => {
     <div className={styles.escalation}>
       <ActionButton
         appearance={{ theme: 'blue', size: 'small' }}
-        submit={ActionTypes.MOTION_ESCALATE}
-        error={ActionTypes.MOTION_ESCALATE_ERROR}
-        success={ActionTypes.MOTION_ESCALATE_SUCCESS}
+        actionType={ActionTypes.MOTION_ESCALATE}
```
rdig, line 49: In all fairness, for now, this shouldn't even be wired up to anything

src/components/common/ColonyActions/ColonyActions.tsx
```diff
@@ -165,18 +177,28 @@ const ColonyActions = (/* { ethDomainId }: Props */) => {
...
+
+  const amount = BigNumber.from(1);
+  const transform = withMeta({ navigate });
+
   return (
     <div className={styles.main}>
       <ActionButton
-        submit={ActionTypes.ACTION_MINT_TOKENS}
-        error={ActionTypes.ACTION_MINT_TOKENS_ERROR}
-        success={ActionTypes.ACTION_MINT_TOKENS_SUCCESS}
```
rdig, line 173: Why were the `error`/`success` action types removed from the Action Button ?

src/components/shared/Button/ActionButton.tsx
```diff
@@ -4,32 +4,40 @@ import { MessageDescriptor } from 'react-intl';
 import { ActionTransformFnType } from '~utils/actions';
 import { useAsyncFunction, useMounted } from '~hooks';
 import DefaultButton from '~shared/Button';
+import { ActionTypes } from '~redux';
 import { Props as DefaultButtonProps } from './Button';
 
+const getFormAction = (
```
rdig, line 10: Right. now I see that you've added a helper here for the success/error

I'm cool with this, however I'd still like to have the option to pass in an action name that will overwrite this.

Something like this:
```
const submit = submit || getFormAction(actionType, 'SUBMIT');
const error = error || getFormAction(actionType, 'ERROR');
const success = success || getFormAction(actionType, 'SUCCESS');
```

src/components/shared/Button/ActionButton.tsx
```diff
@@ -4,32 +4,40 @@ import { MessageDescriptor } from 'react-intl';
...
   transform?: ActionTransformFnType;
   values?: any | (() => any | Promise<any>);
 }
 
 const ActionButton = ({
+  actionType,
   button,
-  error,
-  submit,
-  success,
```
rdig, line 26: See above about my suggestion, which will require these props to come back

src/redux/sagas/motions/rootMotion.ts
```diff
@@ -139,11 +137,11 @@ function* createRootMotionSaga({
...
 
-      const ipfsHash = yield call(uploadIfpsAnnotation, annotationMessage);
+      //  const ipfsHash = yield call(uploadIfpsAnnotation, annotationMessage);
 
-      yield put(
-        transactionAddParams(annotateRootMotion.id, [txHash, ipfsHash]),
-      );
+      // yield put(
+      //   transactionAddParams(annotateRootMotion.id, [txHash, ipfsHash]),
+      // );
```
rdig, line 144: No more IPFS for now, so this can get removed entirely

rdig (approved):
Left some comments, but mostly good to go! Nice work!

src/hooks/useGetColonyAction.ts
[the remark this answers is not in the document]
willm30: Thanks, have addressed both. Also just fyi, I still need to add tracking motions functionality in the event the block ingestor goes down, so it populates the db on restarting with all the missed MotionCreated events. (issue: https://github.com/JoinColony/colonyCDapp/issues/294)

src/components/shared/Button/ActionButton.tsx
willm30: Have put back in

src/redux/sagas/motions/rootMotion.ts
willm30: @rdig Assuming we'll be uploading annotations to the db via block-ingestor, is it OK to use the `annotationMessage` itself as the second param to the `annotateTransaction` contract method (i.e. the metadata param)? My understanding is that generally storing strings on chain is expensive gas-wise, but since `annotateTransaction` just emits an event, is it OK to call `annotateTransaction` with the actual annotation message so it can be picked up by the block ingestor when it detects the event?

If not, we would presumably need to mutate the db from the saga, since I can't think of how else we'd get the actual annotation msg to the block ingestor.

jakubcolony (approved):
LGTM, great work with the polling! 👍

PR #263, opened 2023-02-10
Handle Payment action 

### LINKED ISSUE #214: Update Actions Model (Schema) to support Payment Action
- No comments (will be handled separately)
- Keep simple for now, can be amended as we implement more actions
- Should be tied to a chain, transaction hash and a colony
- A way to better load user/colony/extension data when they are the subject of the payment

### Actions Data Flow

cdapp-actions-data-flow

jakubcolony:
## Description

This PR adds support for the Payment action. It also contains a lot of wiring up that other actions will rely on.

image

## Testing
1. Create a colony using colony creation flow.
2. Click on the Test Mint Tokens button. This should mint 5 tokens of your colony. You will be redirected to the transaction not found page. This is expected and is re-implemented in #261.
3. Go back and click on the Test Create Payment button. This should trigger a payment from the colony to your wallet of 2 tokens. You will again see the same page as above, refresh it a few times and you should see the action details page for the Payment action. Verify all details are shown as expected (the amount will be ~1.98 due to network fees).

The ingestor image should already be set up so that it runs the correct version, but if it's not picking the action try deleting the container in Docker Desktop and restarting your dev env:
image

Resolves #214


amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -409,13 +410,17 @@ type ColonyAction @model {
   type: ColonyActionType!
   blockNumber: Int!
   createdAt: AWSDateTime!
-  initiatorAddress: ID
-  initiator: User @hasOne(fields: ["initiatorAddress"])
+  # Action type specific fields which might be null
+  initiatorAddress: ID # Amplify will automatically populate one of the following fields with related model if there is an initiator
+  initiatorUser: User @hasOne(fields: ["initiatorAddress"])
+  initiatorExtension: ColonyExtension @hasOne(fields: ["initiatorAddress"])
+  initiatorColony: Colony @hasOne(fields: ["initiatorAddress"])
```
jakubcolony, line 417: Unfortunately union connections are not supported by Amplify, so we can't do it cleaner like:
```gql
initiator: User | ColonyExtension | Colony @hasOne(fields: ["initiatorAddress"])
```

src/components/common/ColonyActions/ActionDetailsPage/ActionDetailsPage.tsx
```diff
@@ -85,7 +85,7 @@ const ActionDetailsPage = () => {
 
   return (
     <Layout>
-      <DefaultAction item={action} />
+      <DefaultAction actionData={action} />
```
jakubcolony, line 88: Renamed these in a few places so it's easier to understand what kind of data we're dealing with

src/utils/domains.ts
```diff
@@ -1,10 +1,15 @@
 import { Colony } from '~types';
 import { notNull } from './arrays';
 
-export const findDomain = (domainId?: string | null, colony?: Colony) => {
```
jakubcolony, line 4: We were looking up the domain using its database ID, whereas actions data references the native domain IDs

src/components/common/ColonyActions/ColonyActions.tsx
```diff
@@ -180,7 +180,7 @@ const ColonyActions = (/* { ethDomainId }: Props */) => {
             colonyAddress: colony.colonyAddress,
             colonyName: colony.name,
             nativeTokenAddress: colony.nativeToken.tokenAddress,
-            amount: BigNumber.from(1),
+            amount: BigNumber.from(5).mul(BigNumber.from(10).pow(18)), // this is in wei
```
rdig, line 183: Are these values hardcoded for testing purpouses, or it there something I don't understand ?

src/components/common/ColonyActions/ColonyActions.tsx
```diff
@@ -197,7 +197,7 @@ const ColonyActions = (/* { ethDomainId }: Props */) => {
             recipientAddress: user?.walletAddress,
             domainId: colony.domains?.items[0]?.nativeId,
             singlePayment: {
-              amount: BigNumber.from(1),
+              amount: BigNumber.from(2), // this is in ethers
```
rdig, line 200: Same question here

rdig (approved):
All changes here seem sensible, with the exception of the comments I left

src/components/common/ColonyActions/ColonyActions.tsx
jakubcolony: Correct, we don't have the UAC (I think that's what it's called) ported yet. They are here to trigger the relevant actions for testing

rdig: All good then :+1:

Julianahlk (approved):
Looking Good. Seems straightforward and working as expected. 👍

PR #264, opened 2023-02-10
Fix infinite loop when missing package.json in lambda function directory

### LINKED ISSUE #244: `Installing dependencies for Lambda Function "XXX"` infinitely repeats
On
```
npm install
```

`Installing dependencies for Lambda Function "XXX"`
infinitely repeats with XXX going through a loop.

jakubcolony:
## Description

Small fix for #244.

> I can confirm the issue, it happens whenever there's a folder in `/amplify/backend/function` that is missing `src/package.json` (it tends to happen when you switch between branches and have some leftover untracked files).
If package.json is missing, `npm i` will look in the parent directories until it finds the root package.json, which specifies the `postinstall` script, causing the infinite loop.

Resolves #244

### DIFF
--- a/scripts/lambda-functions-dependencies.sh
+++ b/scripts/lambda-functions-dependencies.sh
@@ -6,6 +6,9 @@ for d in amplify/backend/function/*/ ; do
     [ -L "${d%/}" ] && continue
     fnName=$(echo ${d} | cut -c 26- | rev | cut -c 2- | rev)
     cd "${pwd}/${d}src"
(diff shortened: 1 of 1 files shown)


Julianahlk (approved):
Nice fix! 👍

ArmandoGraterol (approved):
Working as expected! I've encountered this one in the wild also but I thought it was my env specifically being silly and I could not replicate it until now. Nice job!

rdig (approved):
Yup, pretty much what I thought the solution should be

PR #266, opened 2023-02-10
Port Member's search functionality

### LINKED ISSUE #181: Port Members Page: Search
- [x] Search Bar functionality

Dependent on #101

Julianahlk:
## Description

Port search bar functionality

Screen Shot 2023-02-10 at 4 39 50 PM

**New stuff** ✨

Adds filterMembers function
Adds filtering based on search text to `Members`

## TODO

- [x] Created issue #276 to track pagination for the members page

Resolves #181

### DIFF
--- a/src/components/common/Members/MembersTitle.tsx
+++ b/src/components/common/Members/MembersTitle.tsx
@@ -1,4 +1,4 @@
-import React, { useRef, MouseEvent, RefObject } from 'react';
+import React, { useRef, MouseEvent, ChangeEvent } from 'react';
 import { defineMessages } from 'react-intl';
 
 import Heading from '~shared/Heading';
@@ -36,16 +36,12 @@ const MSG = defineMessages({
   },
 });
 
-const handleSearchIconFocus = (searchInput: RefObject<HTMLInputElement>) => {
-  searchInput?.current?.focus();
-};
-
 interface Props {
   currentDomainId: number;
   domainSelectOptions: SelectOption[];
   handleDomainChange: (domainId: number) => void;
   searchValue: string;
-  handleSearch: (e: React.ChangeEvent | React.MouseEvent) => void;
+  handleSearch: (e: ChangeEvent | MouseEvent) => void;
 }
 
 const MembersTitle = ({
@@ -89,25 +85,12 @@ const MembersTitle = ({
             onChange={handleSearch}
             placeholder={formatText(MSG.searchPlaceholder)}
           />
-          {searchValue && (
-            <button
-              className={styles.clearButton}
-              onClick={handleSearch}
-              type="button"
-            >
-              <Icon
-                appearance={{ size: 'normal' }}
-                name="close"
-                title={{ id: 'button.close' }}
-              />
-            </button>
-          )}
+
           <Icon
             appearance={{ size: 'normal' }}
             className={styles.icon}
             name="search"
             title={MSG.search}
-            onClick={() => handleSearchIconFocus(searchInput)}
           />
         </div>
       )}
--- a/src/components/common/Members/MembersTitle.tsx
+++ b/src/components/common/Members/MembersTitle.tsx
@@ -1,4 +1,4 @@
-import React, { useRef, MouseEvent } from 'react';
+import React, { useRef, MouseEvent, RefObject } from 'react';
 import { defineMessages } from 'react-intl';
 
 import Heading from '~shared/Heading';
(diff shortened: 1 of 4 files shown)


src/components/common/Members/MembersTitle.tsx
```diff
@@ -36,6 +36,32 @@ const MSG = defineMessages({
...
+  }
+};
+
+const handleMouseEnter = (e: MouseEvent<HTMLInputElement>) => {
+  e.currentTarget.placeholder = formatText(MSG.searchPlaceholder) as string;
+};
+
+const handleMouseLeave = (e: MouseEvent<HTMLInputElement>) => {
+  e.currentTarget.placeholder = '';
+};
```
jakubcolony, line 63: Could that be possibly all handled in CSS with `:hover` and `::placeholder`?

src/components/common/Members/filterMembers.ts
```diff
@@ -0,0 +1,55 @@
...
+    // }
+
+    // if (filters?.verificationType === VerificationType.UNVERIFIED) {
+    //   return !isWhitelisted && textFilter;
+    // }
+
+    // return !isWhitelisted && textFilter;
+
+    return textFilter;
+  });
```
jakubcolony, line 54: I know you haven't came up with this initially, but why are we chaining things like that?
I think we could split the particular filters (search, verification type) into own small helper functions and then this entire long function could become something like:
```js
members.filter(member => filterMemberBySearchTerm(member, searchTerm) && filterMemberByVerificationStatus(member, verificationStatus))
```
This could be more sophisticated, if needed, but at least we won't be repeating so much code unnecessarily.

jakubcolony (commented):
I'm sorry I haven't raised this earlier, but I think we need to take a big step back here and look at a very crucial thing - pagination.

Metacolony currently has 11k+ members, and Amplify, by default, will only return 100 items. It looks like the entire `getMembersForColony` lambda function needs to be refactored to take that into account, and we need to figure some way of paginating the two lists it returns.

All the searching and filtering should happen on the server-side, so that we get all the data that possibly matches rather than just a subset of what we've already queried for.

Implementing server-side searching might be tricky though, since we are getting the reputation from the network for each query call, and it's gonna be expensive, so I'm wondering if we need to cache the user's reputation somewhere. Unfortunately I don't yet know the inner workings of that so it's hard for me to suggest a solution.

@rdig Tagging you here as I think the current setup is not production-viable and might require more planning taking the above into account.

rdig:
I agree that pagination is super important here. Also yes, search and filtering will be better served for us on the server side, due to their nature and how much that data can scale. However, it's more or less impossible _(in the practical sense)_ to cache the user's reputation, so we'll need to come up with a clever solution.

I think there's a couple of takeaways here:
- finish this PR without pagination (after it's review)
- open a new issue to track pagination for the members page
- have a call to figure out strategies on how we might achieve proper pagination while combining multiple sources inside a lambda function (as well as filtering and searching it)

jakubcolony:
@rdig thanks for sharing your thoughts, I'll wait for @Julianahlk to address my other comments before reviewing again

src/components/common/Members/MembersTitle.tsx
rdig: I agree with this. I'd rather have effects like these handled by CSS

_(If they don't work outright, feel free to look into adding post-css addons that handle that)_

rdig (commented):
I'm good with this in the context that we're going to refactor this whole lot to extract filtering, search and pagination onto the server side.

I will do another review pass once you've addressed the issues that Jakub pointed out.

src/components/common/Members/MembersTitle.tsx
Julianahlk: @jakubcolony I found a way to handle the hover and placeholder cases but I wasn't able to trigger a change of focus with only css. Do you think there is a way to do that as well?

src/components/common/Members/filterMembers.ts
Julianahlk: @jakubcolony split up the logic as suggested

src/components/common/Members/MembersTitle.tsx
jakubcolony: Let me make sure I understand what you're trying to achieve, since the naming seems a bit confusing.

You're trying to focus the searchbox on the search icon click (not focus, right?). In that case, you should be able to place both the input and the icon in the same wrapper with position relative, then position the icon so that it is on top of the input, and set its `pointer-events` property to `none`. So all the clicks on the icon will pass through and cause the input to focus.

If that doesn't work or doesn't make sense, feel free to ping me and I'll try to help

jakubcolony: Here's a demo: https://jsfiddle.net/0cvfkqtr/

Note you can click on the red box (representing the icon) and the input will focus

Julianahlk:
@jakubcolony ready for you to review again

src/components/common/Members/MembersTitle.tsx
Julianahlk: @jakubcolony updated with what you suggested for the search input focus

jakubcolony (approved):
I think it's good to go for now, but will still need some attention before going to production - thanks for opening an issue to track that @Julianahlk 👍

PR #217, opened 2023-01-24
Add Manage Funds related dialogs

ArmandoGraterol:
Examples of things to test:

- That the validation is working as expected for every field.
- That all of the states of the fields that are present on the Dapp are also present here. (Except for stuff not implemented yet like motions, of course).
- That the styles have not changed.

FireShot Capture 788 - Colony CDapp - localhost
FireShot Capture 789 - Colony CDapp - localhost
FireShot Capture 790 - Colony CDapp - localhost
FireShot Capture 791 - Colony CDapp - localhost
FireShot Capture 792 - Colony CDapp - localhost

Contributes to #58

PR #271, opened 2023-02-14
Cleanup and refactor Manage Permissions

### LINKED ISSUE #234: Cleanup Permissions Action Saga
- Clean up imports
- Remove old GraphQL queries and mutations
- Disable redirect for now, will be re-enabled once Actions supports it

Image

### LINKED ISSUE #235: Add User Permissions Model (Schema)
Needs to support:

- User Permissions
- Per Domain Permissions
- Per Colony Permissions
- Historic Permissions (ie: between block 100 - 200)
- Keep as simple as possible for now (other than the above)

Image

### LINKED ISSUE #236: Create / Adapt Permissions Query
GraphQL query (or subscription) to fetch permissions data for a certain user for a colony.

Be aware of the 1000 results limitation and account for that when fetching multiple, or historic permissions.

Image

### LINKED ISSUE #237: Wire up Permissions Action Modal with Saga and Permissions Query
Now that the query is available from #236 and the saga is available from #234 last thing to do is to wire everything up inside the Action Modal.

This will allow us to visualize current permissions, set, remove and update existing ones.

Currently blocked by #223, until it's merged

Image

rdig:
## Description

Cleans up imports and refactors the `managePermissions` saga, adds and refactors Schema Models, wires up the Actions list and page.

I've left comments where relevant, especially in the Schema file, as the changes to the models were less than obvious.

### Screenshots

Screenshot from 2023-04-24 07-52-24
Screenshot from 2023-04-24 07-52-30

### Testing

- Pull everything
- Start the dev enviroment
- create your user 
- create your colony _(I've noticed that the first ever colony created on the chain acts weird, and the ingestor doesn't always pick up it's events, so I'd recomment actually creating a second one, at which point it will work flawlessly)_
- Press `Test Permissions #1`
- Press `Test Permissions #2`
- Refresh the page 
- You should now see / click various actions entries for permissions _(Note that the entry that gives permissions to the extension, actually says that it gives permissions to the colony. This is visually "wrong", the data is all there, but the `FriendlyName` component doesn't know how to display it -- yet)_

Resolves #234
Resolves #235 
Resolves #236
Resolves #237


amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -322,6 +322,10 @@ type Colony @model {
   version: Int!
   actions: [ColonyAction] @hasMany
   metadata: ColonyMetadata @hasOne(fields: ["id"])
+  # @TODO This should not be fetched upfront
+  # It should be retrieved on demand at the earliest occasion it's needed
+  # ie: when opening a UAC modal
+  roles: [ColonyRole] @hasMany
```
jakubcolony, line 328: Would this comment be more appropriate in the CDapp where we actually fetch it?

src/graphql/fragments/actions.graphql
```diff
@@ -8,10 +8,28 @@ fragment ColonyAction on ColonyAction {
...
   }
+  initiatorColony {
+    ...Colony
+  }
+  initiatorExtension {
+    ...Extension
+  }
+  initiatorToken {
+    ...Token
+  }
```
jakubcolony, line 19: Do we really need to query all the fields of those, or should a more restricted fragments be created? Getting the `Colony` fragment after the `balances` were included already took a performance hit and I feel like this would slow down the action query quite a bit if any of these are not null

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -501,4 +513,99 @@ type ColonyAction @model {
...
+  # point of view, but due to the nature of the block ingestor, it works for our use case
+  type: String!
+    @index(
+      name: "getColonyHistoricRoleByDate"
+      queryField: "getColonyHistoricRoleByDate"
+      sortKeyFields: ["createdAt"]
+    )
+
+  domain: Domain! @hasOne
+  colony: Colony! @hasOne
```
jakubcolony, line 595: Nitpick, but I like to manually define the foreign key fields for these (e.g. add `colonyId` of type `ID` and pass it to the `@hasOne` decorator), as otherwise the generated names just end up a bit confusing:
image

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -501,4 +513,99 @@ type ColonyAction @model {
...
+  role_1: Boolean # root
+  role_2: Boolean # arbitration
+  role_3: Boolean # architecture
+  role_5: Boolean # funding
+  role_6: Boolean # administration
+}
+
+# it's a model, not a type since this is designed to work outside of the main
+# colony model, or colony role model, as, by design it should be only used
+# in very specific cases
```
jakubcolony, line 572: I don't think I know enough but why does it need to work outside of the colony role model?

jakubcolony (commented):
Great job navigating around all the obstacles! 

After starting the dev env, I'm seeing a bunch of Amplify mutation errors:
image

Then when I create a colony:
image

There are some actions in the list, so I don't know if the above breaks anything. The actions look duplicated however (I haven't clicked on either of the buttons, it's a fresh colony):
image

Looking at the Dapp, there seems to only have been one action related to permissions on colony creation.

Both buttons work and the actions they trigger show correct values:
image

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -501,4 +513,99 @@ type ColonyAction @model {
...
+  # As such, I've decided to simplify this, an just pass a stringified JSON around
+  # since this is for display purpouses only.
+  #
+  # If anyone has a better idea on how to handle this, I'll gladly change it!
+  #
+  # When / if you decide to use it, just be mindful of what format other people
+  # are using, and follow suit.
+  #
+  # Roles Types:
+  # [{ id: String, type: String, role: Number, setTo: boolean }]
```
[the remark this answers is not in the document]
jakubcolony: Circling back to it, I realised I actually didn't look into individual events as an actions concept at all. Should we come up with a solution that caters for all actions?
Otherwise maybe this should be renamed to be more roles-specific?

src/components/common/ColonyActions/ActionDetailsPage/ActionDetailsPageFeed/ActionDetailsPageFeed.tsx
```diff
@@ -17,14 +17,17 @@ const ActionDetailsPageFeed = ({
   actionData,
 }: // networkEvents,
 ActionsPageFeedProps) => {
-  const events = ACTIONS_EVENTS[actionData.type];
+  const events =
+    JSON.parse(actionData.individualEvents as string) ||
```
willm30, line 21: I would leave a comment here explaining why this is safe (i.e. how you know that actionData.individualEvents won't be undefined and cause the JSON.parse to error)

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -501,4 +513,99 @@ type ColonyAction @model {
...
+  individualEvents: String
```
willm30, line 536: I think I'm suggesting the same approach as Jakub above, but with motion messages all we do is fetch the current array, append to it, and then overwrite it. 

I would prefer that approach to this (type safety, consistency)

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -501,4 +513,99 @@ type ColonyAction @model {
...
+}
+
+type ColonyActionRoles {
+  role_0: Boolean # recovery
+  role_1: Boolean # root
+  role_2: Boolean # arbitration
+  role_3: Boolean # architecture
+  role_5: Boolean # funding
```
willm30, line 544: Perhaps leave a comment here too as to why 4 is missing

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -501,4 +513,99 @@ type ColonyAction @model {
...
+  role_6: Boolean # administration
+}
+
+# Roles have been set as role_<number> so we can more easily map them to their
+# actual contract counterpart on the client and block ingestor side(s)
+
+type ColonyRole @model {
+  id: ID! @index # colonyAddress_domainNativeId_userAddress_roles
+  domain: Domain! @hasOne
```
willm30, line 553: Nitpick: it's not clear which of these params are variable and constant in this comment. 

E.g. would be clearer:
``` graphql
# <colonyAddress>_<domainNativeId>_<contractAddress>_roles
```

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -501,4 +513,99 @@ type ColonyAction @model {
...
+
+type ColonyHistoricRole @model {
+  id: ID! # colonyAddress_domainNativeId_userAddress_blockNumber_roles
+
+  # I HATE AMPLIFY SORTING! MAY IT DIE A THOUSAND DEATHS!
+  #
+  # @NOTE Always set to `type` to "SortedHistoricRole" to enable Amplify sorting, it DOES NOT WORK with `id`
+  #
+  # Apparently, it can only filter non-primary fields, meaning that if we sort by `blockNumber` then that
+  # field is considered primary, so it won't allow us to filter by it.
```
willm30, line 584: What is the intended access pattern here? 

If you wanted to get permissions at an exact block number, you could also have block number as an index.

Or if you wanted to get all permissions for a given target address, index the address and then you can have the block number as the sort key. 

We _probably_ don't need to add an abritrary field just to enable the desired access pattern (although hard to comment for sure since I'm not sure what it is).

amplify/backend/api/colonycdapp/schema.graphql
willm30: Also, is there a reason for marking this as:

`id: ID! @index` ?

It's the primary key by default so I'm not sure what we gain from adding the `@index`

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -501,4 +513,99 @@ type ColonyAction @model {
   toDomain: Domain @hasOne(fields: ["toDomainId"])
   fundamentalChainId: Int
   newColonyVersion: Int
+  roles: ColonyActionRoles
```
willm30, line 516: Nitpick: couldn't this point to the ColonyHistoricRoles entry created at the same time? They contain the same info, no?

rdig: So these are different from the Historic Roles, this just has the exact roles that changed vs. historic role which have the exact snapshot at the time of them being set.

Eg: 
Action Role:
```
role_0: true
role_1: false
```
vs
```
role_0: true
role_1: false
role_2: null
role_3: null
role_5: null
role_6: null
```

src/components/common/ColonyActions/ActionDetailsPage/ActionDetailsPageFeed/ActionDetailsPageFeed.tsx
```diff
@@ -17,14 +17,17 @@ const ActionDetailsPageFeed = ({
...
+    ACTIONS_EVENTS[actionData.type];
```
willm30, line 22: I think this should be refactored so all action events are of the same type (instead of having some as just strings and others like this as objects). 

At the very least we want:

``` graphql

type ActionEvent {
 name: String!
}
```
And then we can add all the rest as optional.

That would mean you could pass the event down into ActionsPageEvent directly, so you wouldn't then have to "find" it again inside `getColonyRoleSetTitleValues`

willm30: Maybe not in this issue but it'd be nice to make this a bit more coherent

willm30 (commented):
I am also receiving many amplify/ingestor errors of the following nature after building the branch for the first time (before doing anything): 

errs

amplify/backend/api/colonycdapp/schema.graphql
rdig: I actually don't know why I left a todo here? But the comment should be also here, although I agree I need to move it to the Dapp as well

amplify/backend/api/colonycdapp/schema.graphql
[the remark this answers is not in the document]
rdig: Yes we should, I'm not particularly happy with my implementation here, I just did it since I wanted something fast to get it done already :)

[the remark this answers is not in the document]
rdig: That's why I asked @willm30 about his implementation on Motions, as to try to align either this to his, or come up with a middle ground.

amplify/backend/api/colonycdapp/schema.graphql
rdig: As I've mentioned above, I'm not particularly happy with this, and will gladly change it to something better.

Shall we set up a call ? @willm30 @jakubcolony ?

amplify/backend/api/colonycdapp/schema.graphql
rdig: Outside, as it not directly linked (via @hasOne @hasMany, etc) as this is envisioned to just be fetched on an instance by instance basis. 

This will become relevant for Recovery Mode.

amplify/backend/api/colonycdapp/schema.graphql
rdig: Good idea, didn't think about it.

src/components/common/ColonyActions/ActionDetailsPage/ActionDetailsPageFeed/ActionDetailsPageFeed.tsx
rdig: Sure, but I'm not going to bother with this since we're going to change this logic anyway.

src/graphql/fragments/actions.graphql
rdig: Good point. I think we need Fragments for these four (User, Colony, Extension, Token) in the "friendly name" format. Ie: just `name`, `displayName`, `address`

amplify/backend/api/colonycdapp/schema.graphql
willm30: If you indexed the targetAddress, and added the block number as a sortKeyField, then you'd be able to get all roles at the targetAddress and the block number without having to do any filtering. 

Then you'd only have to filter by colonyAddress to enable this pattern:

> get the permissions for user X at block Y within colony Z

This would also allow to you to get Roles for a target address from a  range block numbers since the sort key enables a variety of comparisons (gt, lt, gt & lt etc)

rdig:
This is ready for re-review/merge.

However do note, that in the interest of saving as much time as possible for the impeding launch, I've chosen to not implement _(currently)_ some of the more time consuming improvements that were pointed out.

They are still relevant, so I've extracted them in their own separate improvement issue #579 which should/will be tackled once we launch, in the subsequent fixes deployments

willm30:
I'm noticing a few issues that appear to be from the rebase:
1. `getRolesForUserAndDomain` - import is wrong
2. `getExtendedActionType` is missing
3. roles is commented out in `getDetailItems`
4. recipient is not defined in the ColonyActionFragment -> needs to be recipientAddress and references to recipient.walletAddress need to be replaced

I also needed to change my local appsync url in the block ingestor `.env` to  `http://localhost:20002/graphql` (previously it was just localhost:20002) to stop the block ingestor from erroring.
 
If I fix these manually, things appear to work as intended. I haven't noticed that there's anything I can't do as the creator of colony (i.e. all permissions enabled).  

I have noticed a few things I shouldn't be able to do. If I switch user to one without permissions, I can: 

- click "install" on the voting rep extn (which then fails). 
- submit the manage permissions form if I assign this user just the root permission (where you need permission 3 also). 

Will we leave these to be fixed in QA? 

The ActionDetailsPage looks as expected, aside from the message: "RecoveryRoleSet emitted by". Looks like it's missing an address.

permissions-screen

Happy to approve and fix the rest in qa once the rebase issues are fixed.

Julianahlk (commented):
I tried making all the fixes @willm30 mentioned but I am still getting an error every time I try to create a colony.

Screen Shot 2023-05-24 at 4 21 19 PM

rdig:
@willm30 @Julianahlk Fix everything you guys pointed out, and then some, as most of the dialogs were not wired with permissions at all

Some notes on testing:
- Due to QA changes you'll need to update your `.env` file permanently for the GraphQL link to include the `/graphql` part of the url _(as @willm30 correctly deduced)_ -- this is going to make it's way into `master` eventually and it's going to be a requirement from now on
- permissions preventing you to unlock token is correct as that's not wired in yet _(not in this PR at least -- I do remember it being available on the QA branch though)_

@willm30 

> submit the manage permissions form if I assign this user just the root permission (where you need permission 3 also).

This is actually correct the way it's currently implemented. You need Root in the Top Domain (Root Domain), or Architecture in the subdomains

willm30:
Unlock tokens should be available, dialog + saga were merged to master in these prs:
https://github.com/JoinColony/colonyCDapp/pull/217
https://github.com/JoinColony/colonyCDapp/pull/320

I'm also not able to mint tokens in a colony, despite having root permissions.

mint-tokens

According to this tooltip, if I have a permission in root, I should also have it in a subteam. I have created subteam `123` in Root, but the Payment dialog informs me I don't have the funding and administration permissions, despite having them in the root domain.

subteams

> This is actually correct the way it's currently implemented. You need Root in the Top Domain (Root Domain), or Architecture in the subdomains

I have a user with only root permission. When I go to assign that user more permissions in root, I get the following chain error:

managepermissions

Julianahlk (commented):
Overall, seeing the same issues as Will. One thing to add is in the Manage Permissions Dialog, when I view the permissions for a subteam with a user that has all the permissions in Root, the tooltip says I do not have the permission even though the permission is checked.

Screenshot 2023-05-31 at 1 57 27 PM

rdig:
@willm30 @Julianahlk Did another round of fixes based on your findings, so please give it another look

@willm30 

> Unlock tokens should be available, dialog + saga were merged to master in these prs:

Those PRs do not wire in the `colony.status.nativeToken` logic that the modals are basing their logic check on. 

I do remember however that there was some work towards that by @ArmandoGraterol so maybe he can shed some light here.

> I'm also not able to mint tokens in a colony, despite having root permissions.

Same deal as above, it checks for `colony.status.nativeToken.mintable` which is not available

> According to this tooltip, if I have a permission in root, I should also have it in a subteam. I have created subteam 123 in Root, but the Payment dialog informs me I don't have the funding and administration permissions, despite having them in the root domain.

Good spot, as well as correct. Fixed.

> I have a user with only root permission. When I go to assign that user more permissions in root, I get the following chain error:

> I would expect the form to be disabled with a message in the ui about missing permission 3.

Correct as well, seems I've misunderstood the first time you pointed this out. This was just a case of pointing the logic to the correct permissions. Fixed

---

@Julianahlk 

> When I view the permissions for a subteam with a user that has all the permissions in Root, the tooltip says I do not have the permission to set the role.

Fixed alongside the subteam logic I fixed in Will's comment above _(was a issue with `getUserRolesForDomain`'s fallbakc logic)_

src/components/common/ColonyActions/helpers/mapItemToMessageFormat.tsx
```diff
@@ -87,26 +94,28 @@ export const mapColonyActionToExpectedFormat = (
...
   return {
     amount: (
       <Numeral
         value={actionData.amount ?? 0} // @TODO: getAmount(item.actionType, item.amount)
         decimals={actionData.token?.decimals ?? DEFAULT_TOKEN_DECIMALS}
       />
     ),
-    // ...getColonyRoleSetTitleValues(role?.setTo),
+    ...getColonyRoleSetTitleValues(
+      actionData?.individualEvents as string,
```
jakubcolony, line 116: Optional chaining looks like it's not needed here, and could you provide a comment why the casting to string is safe here (otherwise this could be null or undefined as far as the type is concerned)?

src/graphql/fragments/actions.graphql
jakubcolony: This is still outstanding I believe and potentially more urgent than my other comments

src/utils/colonyActions.ts
```diff
@@ -487,6 +488,28 @@ export const getDetailsForAction = (
...
+  const extractedRoles = [
+    { id: 0, setTo: roles.role_0 },
+    { id: 1, setTo: roles.role_1 },
+    { id: 2, setTo: roles.role_2 },
+    { id: 3, setTo: roles.role_3 },
+    { id: 5, setTo: roles.role_5 },
+    { id: 6, setTo: roles.role_6 },
+  ];
+  return extractedRoles.filter(
+    ({ setTo }) => setTo !== null,
```
jakubcolony, line 506: Just making sure this is right, you're not excluding the case in which `setTo` is undefined (and looking at the type it could be)

jakubcolony (approved):
Left a few comments but otherwise LGTM, great effort!

PR #275, opened 2023-02-16
Refactor: migrate Members to react hook forms

Julianahlk:
## Description

- [] TODO write description

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

* Changes go here

**Deletions** ⚰️

* Deletions go here

## TODO

- [ ] Add TODOs


src/components/common/ColonyHome/ColonyDomainSelector/ColonyDomainSelector.tsx
```diff
@@ -2,7 +2,7 @@ import React, { ReactNode, useCallback } from 'react';
 // import { ColonyVersion, ROOT_DOMAIN_ID, Extension } from '@colony/colony-js';
```
willm30, line 2: The `ROOT_DOMAIN_ID` constant in this file can be replaced by `Id.RootDomain` from `colony-js`

src/components/common/ColonyMembers/MembersFilter/MembersFilter.tsx
```diff
@@ -46,18 +45,19 @@ const MembersFilter = ({ handleFiltersCallback, isRoot }: Props) => {
...
         onSubmit={() => {}}
-        enableReinitialize
       >
-        {({ resetForm, values }: FormikProps<FormValues>) => {
+        {({ formState: { isSubmitting, isValid }, getValues, reset }) => {
+          const values = getValues();
           const showReset =
             values.verificationType !== VerificationType.All ||
             values.memberType !== MemberType.All;
           handleFiltersCallback(values);
```
willm30, line 59: This line is causing a state change in the parent while this component is rendering: 

err

This function should really live in a change handler on the underlying select component, so that it updates the filter state in the parent when the filter state actually changes.

So you could change `checkFilters` in `ColonyMembers` to:

```js

const handleFilterChange = (name, value) => {
  if(filters[name] !== value) {
    setFilters((filters) => ({
         ...filters,
         [name]: value    
      }))
  }
}

```
Then in the `Filter` component call it like:

``` js
<Select
        onChange={(value) => handleFilterChange(name, value)}
        appearance={appearance}
        name={name}
        options={options}
        label={label}
      />

```

src/components/common/ColonyMembers/MembersFilter/MembersFilter.tsx
```diff
@@ -46,18 +45,19 @@ const MembersFilter = ({ handleFiltersCallback, isRoot }: Props) => {
...
           memberType: MemberType.All,
           verificationType: VerificationType.All,
         }}
         onSubmit={() => {}}
-        enableReinitialize
       >
-        {({ resetForm, values }: FormikProps<FormValues>) => {
+        {({ formState: { isSubmitting, isValid }, getValues, reset }) => {
+          const values = getValues();
           const showReset =
```
willm30, line 56: `showReset` could be replaced by the `isDirty` prop?

src/components/common/ColonyMembers/MembersFilter/MembersFilter.tsx
```diff
@@ -46,18 +45,19 @@ const MembersFilter = ({ handleFiltersCallback, isRoot }: Props) => {
...
-        initialValues={{
+        defaultValues={{
           memberType: MemberType.All,
           verificationType: VerificationType.All,
         }}
         onSubmit={() => {}}
-        enableReinitialize
       >
-        {({ resetForm, values }: FormikProps<FormValues>) => {
+        {({ formState: { isSubmitting, isValid }, getValues, reset }) => {
```
willm30, line 54: Not sure it makes sense to expose an `isSubmitting` prop here given form submission does nothing. Also there's no validation schema so the form will never be invalid.

src/components/common/ColonyMembers/MembersFilter/MembersFilter.tsx
```diff
@@ -66,9 +66,11 @@ const MembersFilter = ({ handleFiltersCallback, isRoot }: Props) => {
                 </span>
                 {showReset && (
                   <Button
+                    disabled={!isValid || isSubmitting}
```
willm30, line 69: See comment above. `disabled` is not needed here.

src/components/common/ColonyMembers/MembersFilter/MembersFilter.tsx
```diff
@@ -66,9 +66,11 @@ const MembersFilter = ({ handleFiltersCallback, isRoot }: Props) => {
...
                     text={MSG.reset}
+                    type="submit"
```
willm30, line 71: Don't think `type='submit'` is needed either given there's no submit handler on the form.

src/components/common/ColonyMembers/MembersFilter/MembersFilter.tsx
```diff
@@ -66,9 +66,11 @@ const MembersFilter = ({ handleFiltersCallback, isRoot }: Props) => {
...
                     appearance={{ theme: 'blue' }}
-                    onClick={() => resetForm()}
+                    onClick={() => reset()}
```
willm30, line 73: Could be `onClick={reset}`

willm30 (changes requested):
Nice work, requesting one change in particular to remove the bad set state call in `MembersFilter`.

rdig (changes requested):
Nothing really obvious stuck out, other than the stuff found by @willm30 and I'm trusting him on this since he's our resident React Hook Forms expert :)

But I'll give it another review once you fix those.

src/components/common/ColonyMembers/ColonyMembers.tsx
```diff
@@ -69,7 +79,7 @@ const ColonyMembers = () => {
           <TotalReputation selectedDomainId={selectedDomainId} />
           <MemberControls isRootDomain={isRootDomain} />
           <MembersFilter
-            handleFiltersCallback={setFilters}
+            handleFiltersCallback={checkFilters}
```
jakubcolony, line 82: The convention seems to be to name those `onSomeEvent` instead

jakubcolony (changes requested):
Pretty much piggybacking on Will's review here, but I also noticed that the search functionality doesn't work - is that expected in this PR?

src/components/common/ColonyMembers/MembersFilter/MembersFilter.tsx
Julianahlk: @willm30 After making this change the reset stops working, also Im getting this error, 
`Type 'UseFormReset<{ memberType: MemberType; verificationType: VerificationType; }>' is not assignable to type 'MouseEventHandler<HTMLButtonElement>'.
  Types of parameters 'values' and 'event' are incompatible.
    Type 'MouseEvent<HTMLButtonElement, MouseEvent>' is not assignable to type '{ memberType: MemberType; verificationType: VerificationType; } | { memberType?: MemberType | undefined; verificationType?: VerificationType | undefined; } | ResetAction<...> | undefined'.`

Julianahlk:
> Pretty much piggybacking on Will's review here, but I also noticed that the search functionality doesn't work - is that expected in this PR?

@jakubcolony I need to rebase from `feature/40-port-over-members-page`, the search functionality should work now.

src/components/common/ColonyMembers/MembersFilter/MembersFilter.tsx
Julianahlk: @willm30 this works however after this change, reset does not correctly reset the filter state. Is there a way to incorporate updating the filter state directly from the reset() function?

src/components/common/ColonyMembers/ColonyMembers.tsx
```diff
@@ -36,6 +36,16 @@ const ColonyMembers = () => {
     COLONY_TOTAL_BALANCE_DOMAIN_ID,
   );
 
+  const checkFilters = (state: FormValues) => {
```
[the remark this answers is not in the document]
Julianahlk: @willm30 I am a bit unsure of what you mean by all the logic above the view. Do you mean just the handleFilterChange function? and what would be the reason for doing this?

src/components/common/ColonyMembers/ColonyMembers.tsx
Julianahlk: @jakubcolony in this case the function which is now called `handleFilterChange` is passed to MembersFilter -> Filter -> Select and only once it reaches the Select is the function passed to onChange. So does it make sense to name the props that pass the function the same as the function `handleFilterChange` or change to the `onSomeEvent' pattern?

jakubcolony: I would still say so. I like the idea that `handleX` is what you define inside a particular component while `onX` is callback passed from somewhere else. It seem to be a very common pattern

src/components/common/ColonyMembers/ColonyMembers.tsx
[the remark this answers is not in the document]
willm30: By view I mean the JSX (i.e. what we want to show the user). 

So I'm suggesting we have a function, in this case lets call it `useColonyMembers`, that handles all the logic / data processing required for the view, and then have it return only the data needed. Personally find it neater to review components this way. An example of what I having in mind that I've been using in the motions branch: https://github.com/JoinColony/colonyCDapp/blob/2f329f725de0170b3bc391a7cc6e251035cad4a1/src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/StakingWidget/StakingInput/StakingControls/StakingControls.tsx

src/components/common/ColonyMembers/MembersFilter/MembersFilter.tsx
willm30: No, you will need to call `handleFilterChange` with the default values alongside calling reset, since we're holding said values outside the form state. Alternatively (and perhaps this is actually better), you wrap the whole thing in a form instead of just the filters aside, so the Members content can directly access the filter state instead of us duplicating it in component state in `ColonyMembers`

src/components/common/ColonyMembers/MembersFilter/MembersFilter.tsx
willm30: Ah yes, fair enough

willm30 (commented):
Left a couple of comments about how the Members component is currently managing state. There are currently multiple examples of duplicated state (e.g. the filter state is duplicated for accesss in `Members`, & inside `Members` the members query state is duplicated for filtering purposes). 

Thoughts on how to reduce this duplication:
- use a Members provider
-  wrap everything in a form (to share filter state)
- And / or, filter directly inside the members query.

src/components/common/ColonyMembers/ColonyMembers.tsx
```diff
@@ -25,22 +23,16 @@ const MSG = defineMessages({
...
-    COLONY_TOTAL_BALANCE_DOMAIN_ID,
-  );
-
-  const isRootDomain =
-    selectedDomainId === ROOT_DOMAIN_ID ||
-    selectedDomainId === COLONY_TOTAL_BALANCE_DOMAIN_ID;
-
-  const isMobile = useMobile();
+  const {
+    colony,
```
jakubcolony, line 27: What's the rationale between exporting `colony` from this hook? It only seems to introduce an extra dependency that could otherwise be avoided

src/components/common/Members/Members.tsx
```diff
@@ -81,19 +83,42 @@ const Members = ({ selectedDomain, handleDomainChange, filters }: Props) => {
...
-    data?.getMembersForColony?.contributors?.filter(notNull) ?? [];
-  const watchers = data?.getMembersForColony?.watchers?.filter(notNull) ?? [];
+  useEffect(() => {
+    setContributors(
+      data?.getMembersForColony?.contributors?.filter(notNull) ?? [],
+    );
+    setWatchers(data?.getMembersForColony?.watchers?.filter(notNull) ?? []);
+  }, [data]);
+
+  const filterContributorsAndWatchers = useCallback(() => {
```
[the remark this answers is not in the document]
jakubcolony: @willm30 The entire members query has to be reworked to support sorting, pagination, and filtering on the server-side. See https://github.com/JoinColony/colonyCDapp/issues/276

src/components/common/ColonyMembers/ColonyMembers.tsx
```diff
@@ -69,7 +61,7 @@ const ColonyMembers = () => {
           <TotalReputation selectedDomainId={selectedDomainId} />
           <MemberControls isRootDomain={isRootDomain} />
           <MembersFilter
-            handleFiltersCallback={setFilters}
+            handleFilterChange={handleFilterChange}
```
jakubcolony, line 64: I still think these should be called `onFilterChange`, to highlight they are handlers defined elsewhere, and being passed as props

src/components/common/Members/Members.tsx
```diff
@@ -81,19 +83,42 @@ const Members = ({ selectedDomain, handleDomainChange, filters }: Props) => {
...
     (event: React.ChangeEvent<HTMLInputElement>) => {
       const value = event.target?.value || '';
       setSearchValue(value);
+      filterContributorsAndWatchers();
     },
-    [setSearchValue],
+    [filterContributorsAndWatchers, setSearchValue],
   );
 
+  useEffect(() => {
```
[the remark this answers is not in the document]
jakubcolony: @willm30 I agree, especially having read the docs page you shared a few weeks ago: https://beta.reactjs.org/learn/you-might-not-need-an-effect

I don't think we need a state for members/contributors at all since they're derived from state (`data`, `searchValue`) and props (`filters`).

We could just call the `filterMembers` directly in the component body:
```js
const filteredContributors = filterMembers<Contributor>(
      data?.getMembersForColony?.contributors || [],
      searchValue,
      filters,
);
```

Or, if it proves to be an expensive computation, wrap it in `useMemo` with `data`, `searchValue` and `filters` as dependencies.

src/components/common/Members/MembersTitle.tsx
```diff
@@ -41,7 +41,7 @@ interface Props {
   domainSelectOptions: SelectOption[];
   handleDomainChange: (domainId: number) => void;
   searchValue: string;
-  handleSearch: (e: React.ChangeEvent | React.MouseEvent) => void;
+  handleSearch: (e: ChangeEvent | MouseEvent) => void;
```
jakubcolony, line 44: Is `MouseEvent` still needed?

jakubcolony (changes requested):
All works (including the search functionality) - nice work @Julianahlk! 

I left a few comments and will be happy to approve once they're addressed!

src/components/common/ColonyMembers/ColonyMembers.tsx
willm30: This was originally my idea, basically trying to keep the component logic as concise as possible. Raul has since suggested we keep hooks separate such as this one separate https://github.com/JoinColony/colonyCDapp/pull/275#discussion_r1133812909

src/hooks/useColonyMembers.ts
```diff
@@ -0,0 +1,52 @@
...
+    if (filters[name] !== value) {
+      setFilters((filter) => ({
+        ...filter,
+        [name]: value,
+      }));
+    }
+  };
+
+  const isRootDomain =
+    selectedDomainId === ROOT_DOMAIN_ID ||
```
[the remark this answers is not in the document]
Julianahlk: @willm30 should we just remove ROOT_DOMAIN_ID from ~constants to avoid this confusion?

src/components/common/Members/MembersTitle.tsx
Julianahlk: Nope, just removed it.

willm30 (approved):
LGTM 👍

src/hooks/useColonyMembers.ts
```diff
@@ -0,0 +1,49 @@
...
+import { COLONY_TOTAL_BALANCE_DOMAIN_ID } from '~constants';
+import { useMobile } from '~hooks';
+
+import { FormValues } from '~common/ColonyMembers/MembersFilter';
+import {
+  MemberType,
+  VerificationType,
+} from '~common/ColonyMembers/MembersFilter/filtersConfig';
+
+const useColonyMembers = () => {
```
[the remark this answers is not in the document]
willm30: I second this

src/components/common/ColonyMembers/useColonyMembers.ts
```diff
@@ -0,0 +1,49 @@
...
+      setFilters((filter) => ({
+        ...filter,
+        [name]: value,
+      }));
+    }
+  };
+
+  const isRootDomain =
+    selectedDomainId === Id.RootDomain ||
+    selectedDomainId === COLONY_TOTAL_BALANCE_DOMAIN_ID;
```
rdig, line 34: I mean this logic is not really correct. `COLONY_TOTAL_BALANCE_DOMAIN_ID` represents "all domains", root is just one of those domains.

If we're overloading this logic, then a comment explaining this is appropriate.

src/components/common/Members/MembersContent.tsx
```diff
@@ -27,7 +28,7 @@ const MembersContent = ({
   contributors,
 }: Props) => {
   const isRootDomain =
-    selectedDomain === ROOT_DOMAIN_ID ||
+    selectedDomain === Id.RootDomain ||
     selectedDomain === COLONY_TOTAL_BALANCE_DOMAIN_ID;
```
rdig, line 32: Same comment as I left above. It's either root, or all domains, can't really be both (unless we're overloading)

rdig (approved):
After all the review passes, I'm all good with this, however there's some logic related to the root which I don't really agree with.

I've left comments where relevant.

PR #289, opened 2023-02-19
Add support for CreateDomain action

### LINKED ISSUE #279: Add support for CreateDomain action
In both block-ingestor and the CDapp

image

jakubcolony:
## Description

This PR adds support for the CreateDomain action.

image

## Testing
1. Create a colony using the wizard.
2. Click on the Test Create Domain action.
3. You'll be redirected to a transaction not found page, wait a few seconds and refresh.
4. You should see the action details (including domain name, color, description). Please verify the domain is added to the domain list dropdown as well.

Resolves #279


amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -134,22 +134,22 @@ enum Network {
 }
 
 enum DomainColor {
-  LIGHTPINK
+  LIGHT_PINK
```
jakubcolony, line 137: Changes those to have properly generated PascalCase members in CDapp

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -297,14 +297,21 @@ type User @model {
...
+  colonyId: ID! @index(name: "byColony")
+  colony: Colony! @belongsTo(fields: ["colonyId"])
   nativeId: Int!
   nativeFundingPotId: Int!
   nativeSkillId: Int!
-  name: String
-  description: String
-  color: DomainColor
-  parent: Domain @hasOne
+  isRoot: Boolean!
```
jakubcolony, line 306: I thought this is an easier solution since we only support one level nesting (both in Dapp and on the contract level). Another thing that pushed me towards that change was that I couldn't get the parentId from the `DomainAdded` event or by calling `getDomain` on colony client - something to look at when we want to start supporting more complex nesting

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -297,14 +297,21 @@ type User @model {
...
+  metadata: DomainMetadata @hasOne(fields: ["id"])
+}
+
+type DomainMetadata @model {
+  id: ID! # This field is referenced by Domain id, so has to be in the same format: colonyAddress_nativeId
+  name: String!
+  description: String!
+  color: DomainColor!
```
jakubcolony, line 314: Separated this since it's a piece of information that's only available in the CDapp (and doesn't get emitted to the network, therefore inaccessible in the block-ingestor). This way, the saga is only responsible for creating the metadata, and the actual domain entity is created by action handler in block-ingestor

The advantage this gives is that block-ingestor will still pick up the events that originated from a contract call as opposed to a saga inside CDapp. Those will not have any metadata attached, but will still show correctly in the actions list.

We can follow the same pattern in the colony creation saga

package.json
```diff
@@ -33,7 +33,7 @@
   },
   "homepage": "https://colony.io",
   "devDependencies": {
-    "@aws-amplify/cli": "^10.6.1",
+    "@aws-amplify/cli": "10.4.1",
```
jakubcolony, line 36: https://github.com/aws-amplify/amplify-category-api/issues/1014 😠

src/components/common/ColonyActions/ColonyActions.tsx
```diff
@@ -206,6 +207,22 @@ const ColonyActions = (/* { ethDomainId }: Props */) => {
...
+            colonyAddress: colony.colonyAddress,
+            colonyName: colony.name,
+            domainName: 'Test Domain',
+            domainColor: DomainColor.BlueGrey,
+            domainPurpose: 'Test Domain Description',
+          }),
+          withMeta({ navigate }),
+        )}
+        text="Test Create Domain"
+      />
```
jakubcolony, line 225: Heads up: this is for testing only

src/types/domains.ts
```diff
@@ -1,57 +1,20 @@
 import { DomainColor } from '~gql';
 
-export enum Color {
```
jakubcolony, line 3: I think we don't need two separate enums (and a mapping between them) and we can simply use the GraphQL enum directly

amplify/backend/api/colonycdapp/schema.graphql
rdig: You mean the underscores ?

amplify/backend/api/colonycdapp/schema.graphql
rdig: Sure, I'm good with that.

Just keep in mind that at some point we'll need to deal with multiple nesting levels

src/components/shared/DomainDropdown/DomainDropdown.tsx
```diff
@@ -90,7 +89,16 @@ const DomainDropdown = ({
     const allDomainsOption: SelectOption = {
       children: (
         <DomainDropdownItem
-          domain={ALLDOMAINS_DOMAIN_SELECTION}
+          domain={{
+            id: '',
+            isRoot: true,
```
rdig, line 94: This is a bit misleading. 

"All teams" include all domains, including root. See it more of a grouping of domains, rather than a nested tree/list.

So maybe leave a comment explaining this so future devs won't be confused

src/types/domains.ts
rdig: Works for me.

rdig (approved):
Looks good. Left some comments, but otherwise stellar work

amplify/backend/api/colonycdapp/schema.graphql
jakubcolony: Yeah, if you don't include those, you get `Lightpink` instead of `LightPink`

jakubcolony:
@rdig A couple of extra changes since your approval:
- Modified temp-create-data and colonyCreate saga to call the domain metadata mutation and create the root domain in the database manually (since it doesn't trigger an event)
- Removed createUniqueDomain lambda function

Feel free to have another look when you have time

src/components/shared/DomainDropdown/DomainDropdown.tsx
jakubcolony: This is actually a dummy domain data that we use to display the "All Teams" option in the dropdown. `isRoot` has to be set to true so that it's not shown as a subdomain item like so:
image

But I agree it's confusing without further explanation so I added a relevant comment

Julianahlk (approved):
Working well for me and looks good on mobile 👍

src/components/shared/DomainDropdown/DomainDropdown.tsx
```diff
@@ -118,7 +130,8 @@ const DomainDropdown = ({
         .filter(notNull)
```
willm30, line 130: The `.slice()` and the comment re subgraph can presumably be removed now

src/components/shared/DomainDropdown/DomainDropdown.tsx
```diff
@@ -118,7 +130,8 @@ const DomainDropdown = ({
...
         .sort(sortByDomainId)
         .map((domain) => {
-          const { nativeId, name: domainName } = domain || {};
+          const { nativeId, metadata } = domain || {};
```
willm30, line 133: We expect `domain` to be defined here since we filter `notNull` above

willm30 (approved):
Nice one 👍

PR #291, opened 2023-02-23
Add support for MoveFunds action

### LINKED ISSUE #277: Add support for MoveFunds action
In both block-ingestor and the CDapp

image

jakubcolony:
## Description

This PR adds support for the MoveFunds action.
image

## Testing
1. Create a colony using the wizard.
2. Generate at least one each of Mint Tokens and Create Domain actions.
3. Click on the Test Move Funds button.
4. You'll be redirected to Transaction not found page. Refresh and verify the details shown are the same as the Dapp, e.g.: https://xdai.colony.io/colony/jakub/tx/0x38e7a37245fe91417f473b6150860db0f6e25596c12b0affa44796cef7b8936f.

## TODO

- [x] Clean up the saga
- [x] Wire data into the UI

Resolves #277


src/components/common/ColonyActions/ColonyActions.tsx
```diff
@@ -236,6 +236,23 @@ const ColonyActions = (/* { ethDomainId }: Props */) => {
...
+      <ActionButton
+        submit={ActionTypes.ACTION_MOVE_FUNDS}
+        error={ActionTypes.ACTION_MOVE_FUNDS_ERROR}
+        success={ActionTypes.ACTION_MOVE_FUNDS_SUCCESS}
+        transform={pipe(
+          mergePayload({
+            colonyAddress: colony.colonyAddress,
+            colonyName: colony.name,
+            fromDomain: colony.domains?.items.find((d) => d?.isRoot),
+            toDomain: colony.domains?.items.find((d) => !d?.isRoot),
```
jakubcolony, line 248: For testing, this will attempt to transfer from root to the first non-root domain it can find

src/redux/sagas/actions/moveFunds.ts
```diff
@@ -88,24 +53,24 @@ function* createMoveFundsAction({
       );
     }
 
-    const [{ fundingPotId: fromPot }, { fundingPotId: toPot }] = yield all([
-      call([colonyClient, colonyClient.getDomain], fromDomainId),
-      call([colonyClient, colonyClient.getDomain], toDomainId),
-    ]);
+    const { nativeFundingPotId: fromPot } = fromDomain;
+    const { nativeFundingPotId: toPot } = toDomain;
```
jakubcolony, line 57: Conveniently we have those in the database, so no need for a contract call

src/components/common/ColonyActions/ColonyActions.tsx
```diff
@@ -236,6 +236,23 @@ const ColonyActions = (/* { ethDomainId }: Props */) => {
...
+            amount: BigNumber.from(5).mul(BigNumber.from(10).pow(17)), // this is in wei
```
rdig, line 249: I know this is for testing, but why is this a 17 decimals token ?

If it's a WEI, and it's a common 18 decimals token _(like we create normally through the colony client)_, this should be `pow(18)`

rdig (approved):
Good to go as well.

src/components/common/ColonyActions/ColonyActions.tsx
jakubcolony: It is 18 decimals. I wanted 0.5 of a token which needed to be converted into wei, but I couldn't call `BigNumber.from(0.5)...` because that was throwing an error. So I start with `5`, and instead decreased the exponent by one

willm30 (approved):
Nice 1️⃣

PR #296, opened 2023-02-28
Extract colony metadata from colony model

### LINKED ISSUE #292: Extract colony metadata from the colony model
Similarly to the approach taken in handling domain creation, splitting the colony model into data which is available on chain and data only available on the client-side (e.g. name) allows us to handle colony creation in block-ingestor no matter if it came from the saga or contract call. 

This is a blocker for properly tracking the colonies in block-ingestor (https://github.com/JoinColony/block-ingestor/issues/6).

jakubcolony:
## Description

This PR extracts the colony metadata (the information that is not available on the chain) into its own model, allowing the client to write it separately from the "core" colony details.

To achieve this, I removed the Colony<>Profile relation. The display name, avatar, and thumbnail are now properties of Colony Metadata. This is to reduce complexity (you don't need either an extra mutation to create profile, or a custom lambda function like `createUniqueColony`) and allow for easier tracking of the changes (similarly to domain metadata's changelog).

I don't think it is yet possible to move the entire colony creation mutation to block-ingestor and only write the metadata in the saga (as we do with the domain). The problem I faced is that colony name isn't available on chain but at the same time it has to be unique and indexable (since we want to get colony by name reliably and fast). This needs further thought but one idea could be to create some mapping between given names and colony addresses (as well as chains possibly).

## Testing

Testing mostly focuses on double-checking I haven't broken anything. Please create some colonies using both the creation wizard and `temp-create-data` script. Inspect the CDapp to make sure none of the previous profile data (just display name for now) isn't missing/wrong.

Resolves #292


amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -27,7 +27,7 @@ input ProfileMetadataInput {
   emailPermissions: [String!]!
 }
 
-input MetadataInput {
+input ChainMetadataInput {
```
jakubcolony, line 30: Renamed to avoid confusion between all the different metadatas we now have

Julianahlk (approved):
Tested everything I could think of and couldn't find any issues. I do notice there are some merge conflicts. I can take another look once those are resolved.

jakubcolony:
@rdig Done

scripts/temp-create-data.js
```diff
@@ -253,7 +257,25 @@ const createMetacolony = async (singerOrWallet) => {
     if (metacolonyQuery?.errors) {
       console.log('METACOLONY COULD NOT BE CREATED.', metacolonyQuery.errors[0].message);
     } else {
-      console.log(`Creating metacolony { name: "meta", colonyAddress: "${utils.getAddress(metacolonyAddress)}", profile: { displayName: "Metacolony" }, nativeToken: "${utils.getAddress(metacolonyTokenAddress)}", version: "${metacolonyVersion.toString()}" }`);
+      console.log(`Creating metacolony { name: "meta", colonyAddress: "${utils.getAddress(metacolonyAddress)}", nativeToken: "${utils.getAddress(metacolonyTokenAddress)}", version: "${metacolonyVersion.toString()}" }`);
```
rdig, line 260: Is the temp script necessary anymore ?

We should decide what to do with it. Either:
- remove it and create everything manually
- keep it, but update it so that it creates proper colonies (with extensions, token authorities and everything)

rdig (approved):
Yup, sensible change.

No complaints from me. I just need to keep an eye out for when I merge #74 to make sure I update it to make use of the new naming scheme

scripts/temp-create-data.js
jakubcolony: Yeah, maintaining it is a bit of a pain. But it's nice to have the script that bootstraps everything for you.
One option could be to use some headless browser (cypress, puppeteer) to programmatically click through the user and then colony creation wizard, but I don't know how feasible that would be

rdig: We had that before, and it more or less work, however it provided quite a bit of overhead and ate alot of resources.

I guess once we stabilize the core features, maintaining this won't be as much of a problem anymore

PR #299, opened 2023-03-03
Port stakeMotion saga

### LINKED ISSUE #298: Port motion staking saga
So clicking the "Staking" button fires relevant actions on the chain

Then adapt block ingestor to update the db with latest motion state

willm30:
## Description

This PR ports the motion staking saga. This branch is currently a bit light on permissions and token activation, so we have to get into the truffle console to set things up correctly.

staked

### Setup instructions

1. Build Cdapp without block ingestor. Run webpack.
2. Checkout and run block ingestor separately on branch: https://github.com/JoinColony/block-ingestor/pull/18
3. Turn reputation monitor on: http://localhost:3001/reputation/monitor/toggle
4. Open truffle: `npm run truffle console`
5. Get your colony:  c = await IColony.at(<colony-address>)
6. Give yourself some reputation: c.emitDomainReputationReward(1, accounts[0], '1000000000000000000')
7. Then in the ui, install the voting reputation extension. Make the staking phase last a long time ((e.g. 500 hours) due to time in dev being weird)
8. Back in truffle, give voting rep extn the arbitration permission: `c.setArbitrationRole(1, "0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff", <voting-reputation-extn-address>, 1, true)` (voting rep address is its id in the db)

For staking to work, our user needs to have activated tokens. 

To do that, in truffle:

1. Mint some tokens: c.mintTokens('number in wei'), eg '10000000000000000000' or 10 eth
2. Then get the token address: ta = await c.getToken()
4. And claim the tokens you just minted: c.claimColonyFunds(ta)

At this point, we've just simulated the "mint tokens" action, which we can't do via the ui because we're testing a mint tokens motion.

Now we need to make a payment from the colony to the user.  This can be done in the ui. 

1. In `ColonyActions.tsx`, go to `ACTION_EXPENDITURE_PAYMENT` button and set the amount to some number less than or equal to the number you just minted. Then in the ui, click the create payment button. You have just paid yourself. 

Finally, we need to activate these tokens (again, in truffle console).

1. get the network address: n = await IColonyNetwork.at('0x5CC4a96B08e8C88f2c6FC5772496FeD9666e4D1F')
2. get tokenLocking address: tlAddress = await n.getTokenLocking()
3. instantiate tokenLocking contract: l = await TokenLocking.at(tlAddress)
5. Then, with the token address from earlier, instantiate a token contract : t = await Token.at(ta)
6.  Approve tokens: t.approve(l.address, 'number you paid yourself')
7. Activate tokens! l.deposit(t.address, 'number you approved' / 10) [I've noticed the amount you approve gets an "insufficient balance error"]

Now click "Test Mint Tokens". You should be directed to the motions ui.

Stake for the motion. The UI should update immediately, and persist after a refresh. Functionality should be same as: 
https://xdai.colony.io/colony/willsworld/tx/0xa5228de05deaa0f08dcbc4f8825ac66b00f2e564643de01c4b76a98f10dfed25

After you have reviewed this, it might be an efficient use of time to review this one also: https://github.com/JoinColony/colonyCDapp/pull/310

Since they share a setup.

**Changes** 🏗

* Wire in motion staking saga and associated components

## Note

Messages in the event feed are being handled separately. https://github.com/JoinColony/colonyCDapp/issues/312
Resolves  #298

### DIFF
--- a/src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/DefaultMotion.tsx
+++ b/src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/DefaultMotion.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 
 import DetailsWidget from '~shared/DetailsWidget';
 
-import { useColonyContext, useEnabledExtensions } from '~hooks';
+import { useDefaultMotion } from '~hooks';
 import { ColonyAction } from '~types';
 import { motionStateMap } from '~utils/colonyMotions';
 
@@ -19,30 +19,33 @@ interface DefaultMotionProps {
   actionData: ColonyAction;
 }
 
-const DefaultMotion = ({ actionData }: DefaultMotionProps) => {
-  const { colony } = useColonyContext();
+const DefaultMotion = ({
+  actionData,
+  actionData: { motionData },
+}: DefaultMotionProps) => {
+  const { colony, showBanner, setShowBanner, showMotionHeading } =
+    useDefaultMotion(motionData);
 
-  const { isVotingReputationEnabled } = useEnabledExtensions();
-
-  const motionState = actionData.motionData?.motionState;
-
-  if (!colony || !motionState) {
+  if (!colony || !motionData) {
     return null;
   }
 
-  const showBanner = true; // !shouldDisplayMotionInActionsList(currentStake, requiredStake);
+  const { motionState } = motionData;
 
   return (
     <div className={styles.main}>
       {/* {isMobile && <ColonyHomeInfo showNavigation isMobile />} */}
       {showBanner && <StakeRequiredBanner isDecision={false} />}
-      {isVotingReputationEnabled && (
+      {showMotionHeading && (
         <MotionHeading motionState={motionStateMap[motionState]} />
       )}
       <div className={styles.container}>
         <DefaultActionContent actionData={actionData} />
         <div className={styles.widgets}>
-          <MotionPhaseWidget actionData={actionData} />
+          <MotionPhaseWidget
+            actionData={actionData}
+            setShowStakeBanner={setShowBanner}
+          />
           {/* <div className={styles.details}>
         {motionState === MotionState.Voting && (
           <VoteWidget
--- a/src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/StakingWidget/helpers.ts
+++ b/src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/StakingWidget/helpers.ts
@@ -1,4 +1,6 @@
 import Decimal from 'decimal.js';
+import { MotionData } from '~gql';
+import { Address } from '~types';
 
 export const getStakeFromSlider = (
   sliderAmount: number,
@@ -11,11 +13,16 @@ export const getStakeFromSlider = (
     .plus(minUserStake);
 
 export const getFinalStake = (
-  amount: number,
+  sliderAmount: number,
   remainingToStake: Decimal,
   minUserStake: Decimal,
 ) => {
-  const stake = getStakeFromSlider(amount, remainingToStake, minUserStake);
+  const stake = getStakeFromSlider(
+    sliderAmount,
+    remainingToStake,
+    minUserStake,
+  );
+
   return stake.round().toString();
 };
 
@@ -24,7 +31,6 @@ export const convertStakeToPercentage = (
   requiredStake: Decimal,
 ) => {
   const divisibleRequiredStake = requiredStake.isZero() ? 1 : requiredStake;
-
   return new Decimal(stake).div(divisibleRequiredStake).mul(100).toDP(2);
 };
 
@@ -110,4 +116,12 @@ export const getUserStakeLimitPercentage = (
     remainingToStake,
     minUserStake,
   );
+};
+
+export const getUserStakes = (
+  usersStakes: MotionData['usersStakes'],
+  userAddress: Address,
+) => {
+  const userStake = usersStakes.find(({ address }) => address === userAddress);
+  return userStake?.stakes.raw ?? null;
 };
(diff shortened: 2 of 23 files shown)


src/hooks/useUserBalance.ts
```diff
@@ -0,0 +1,72 @@
...
+  getReputationOracleURL,
+  networkAddress,
+} from '~redux/sagas/utils/getNetworkClient';
+import { Address, Wallet } from '~types';
+
+import useAppContext from './useAppContext';
+import useColonyContext from './useColonyContext';
+
+// TODO: Get this data from db once we have token data: https://github.com/JoinColony/colonyCDapp/issues/286
+
```
willm30, line 15: This is a temp solution

src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/DefaultMotion.tsx
```diff
@@ -19,32 +19,33 @@ interface DefaultMotionProps {
...
 }
 
-const DefaultMotion = ({ actionData }: DefaultMotionProps) => {
-  const { colony } = useColonyContext();
+const DefaultMotion = ({
+  actionData,
+  actionData: { motionData },
+}: DefaultMotionProps) => {
+  const { colony, showBanner, setShowBanner, showMotionHeading } =
+    useDefaultMotion(motionData);
```
rdig, line 27: I feel iffy importing the colony from this hook.

I know it imports it properly under the hood, but from a eagle eye view it may looks like both `useDefaultMotion` and `useColonyContext` export the `colony` data object, without knowing for sure if there are differences between them.

I would stiill make use of the `colony` data internally in `useDefaultMotion`, but when used in components, I would import the colony data from the standard `useColonyContext`

rdig (approved):
Left a comment about a design decision I'm not really on the same page, otherwise this is all good.

src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/DefaultMotion.tsx
jakubcolony, line 27: I appreciate your efforts to keep the components lean, but I can't see a benefit in this particular hook. One of the reasons might be that it's primary purpose is to encapsulate some of this component dependencies and I don't think this is worthwhile.
Apart from what Raul mentioned regarding exporting the colony, I think exporting information on whether or not to show motion heading makes this hook more or less useless. It's not great in terms of reusability (maybe another component has got a different rules on when to show the heading) and it violates the separation of concerns between the data and the view that I think you originally wanted to improve. 
A good rule is that a piece of code should only have one reason to change, and while it's not always possible, I'd think a good reason for this hook to change would be some changes to the actions or extensions data. But right now, the hook also has to change whenever the underlying UI changes, meaning we will most likely be unable to use it for the refreshed UI "out of the box". 

Maybe this refactoring/optimisation happens a bit too early? This component was only ~90 lines long which I'd normally consider pretty short and there wasn't a ton of complex logic that you got rid of. I can imagine there's gonna be some more motions data mapping/manipulation down the line and it might be easier to see its value then

src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/StakingWidget/helpers.ts
```diff
@@ -111,3 +117,11 @@ export const getUserStakeLimitPercentage = (
     minUserStake,
   );
 };
+
+export const getUserStakes = (
+  usersStakes: MotionData['usersStakes'],
```
jakubcolony, line 122: You define this as the GraphQL type, which isn't necessarily what you query for (e.g. `UserStake` has `address` and `stakes` field, but your query might not ask the backend for both of those fields). It's safer to declare a GraphQL fragment and use the generated type for that

src/components/shared/Fields/Form/ActionHookForm.tsx
```diff
@@ -10,10 +10,15 @@ import HookForm, {
   CustomSubmitHandler,
   HookFormProps,
 } from './HookForm';
+import { UseFormReturn } from 'react-hook-form';
```
jakubcolony, line 13: should be higher up

src/hooks/motionWidgets/stakingWidget/helpers.ts
```diff
@@ -35,3 +45,60 @@ export const mapStakingSliderProps = ({
   },
   mutableRef,
 });
+
+export const getMotionSide = (vote: number) =>
+  vote === 0 ? MotionSide.NAY : MotionSide.YAY;
+
+const getUpdatedUserStakes = (
+  userStakes: MotionData['usersStakes'][0],
```
jakubcolony, line 53: Can you use the `UserStakes` defined above?

src/redux/sagas/transactions/createTransaction.ts
```diff
@@ -131,3 +132,18 @@ export function* waitForTxResult(channel: Channel<any>) {
     throw new Error('Transaction failed');
   }
 }
+
+export const createGroupTransaction = (
```
jakubcolony, line 136: I like this change but I think if you don't change all the sagas to use it (and pay attention to other PRs to make sure they're following this convention too), instead of improving things, it'll introduce inconsistencies that are really hard to resolve later (if I came across the two versions of `createGroupTransaction` in a couple months time, I'd need to dig through the code to understand if there's a difference or whether it's just refactoring-in-progress)
It might sound counterintuitive, but sometimes it's better to leave things as they are unless you're willing to make the change consistent

jakubcolony (commented):
The UI works and looks good! Left a few comments but otherwise top notch job 👍 

image

image

I initially struggled with this step of the instructions:
`Approve tokens: t.approve(l.address, 'number you paid yourself')`
I think it's worth adding the amount you paid yourself was in ethers, whereas the `t.approve` expects amount in wei (I paid myself `5` ethers and then was only activating `5` wei of that)

willm30:
Temporarily back to draft while prs down the chain get refactored

willm30:
Closing as covered here: https://github.com/JoinColony/colonyCDapp/pull/335

PR #311, opened 2023-03-08
Expose current VotingReputation version in `useEnabledExtensions` hook

### LINKED ISSUE #304: Expose the current version of extensions returned by `useEnabledExtensions` hook
The Dapp version of the hook returns the version of VotingReputation extension installed

This is a blocker for #220

jakubcolony:
## Description

This small PR refactors the `useEnabledExtensions` hook back into a shape similar to the Dapp's one.

I initially did some work to add relation between `CurrentVersion` and `ColonyExtension` models but then I realised we still need separate current version queries so I ditched all that.

## Testing

Log the value returned by the hook somewhere. Install, enable, deprecate and uninstall both extensions and see if the value returned matches your actions.

Resolves #304

### DIFF
--- a/src/hooks/useEnabledExtensions.tsx
+++ b/src/hooks/useEnabledExtensions.tsx
@@ -2,23 +2,21 @@ import { Extension } from '@colony/colony-js';
 
 import useExtensionsData from './useExtensionsData';
 
-type EnabledExtensionKey = `is${Extension}Enabled`;
-type EnabledExtensions = Partial<Record<EnabledExtensionKey, boolean>>;
-
 const useEnabledExtensions = () => {
   const { installedExtensionsData, loading } = useExtensionsData();
 
-  const enabledExtensions = installedExtensionsData.reduce<EnabledExtensions>(
-    (extensions, extension) => ({
-      ...extensions,
-      [`is${extension.extensionId}Enabled`]: extension.isEnabled,
-    }),
-    {},
+  const oneTxPaymentExtension = installedExtensionsData.find(
+    (extension) => extension.extensionId === Extension.OneTxPayment,
+  );
+  const votingReputationExtension = installedExtensionsData.find(
+    (extension) => extension.extensionId === Extension.VotingReputation,
   );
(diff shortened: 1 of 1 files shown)


src/hooks/useEnabledExtensions.tsx
```diff
@@ -2,23 +2,21 @@ import { Extension } from '@colony/colony-js';
...
-
 const useEnabledExtensions = () => {
   const { installedExtensionsData, loading } = useExtensionsData();
 
-  const enabledExtensions = installedExtensionsData.reduce<EnabledExtensions>(
-    (extensions, extension) => ({
-      ...extensions,
-      [`is${extension.extensionId}Enabled`]: extension.isEnabled,
-    }),
-    {},
```
jakubcolony, line 16: Keeping the dynamic reduce approach would mean I still have to search for the VotingReputation installation separately. Thinking about it, we don't add new extensions every day so I think it makes sense to be fixed for those extensions we care about

Julianahlk (approved):
I like the simplicity, makes it easier to use 👍

src/hooks/useEnabledExtensions.tsx
```diff
@@ -2,23 +2,21 @@ import { Extension } from '@colony/colony-js';
...
+    (extension) => extension.extensionId === Extension.OneTxPayment,
+  );
+  const votingReputationExtension = installedExtensionsData.find(
+    (extension) => extension.extensionId === Extension.VotingReputation,
   );
 
   return {
     loading,
-    enabledExtensions,
+    isOneTxPaymentEnabled: !!oneTxPaymentExtension?.isEnabled,
```
willm30, line 17: Just noting these two aren't exactly equivalent as enabledExtensions wasn't being spread so would have been:

``` js
enabledExtensions: {
   isOneTxPaymentEnabled: ...
   ....
}
```

willm30 (approved):

rdig (commented):
Same here. Good to go.

PR #316, opened 2023-03-13
Token activation sagas (& fetching user balance)

### LINKED ISSUE #286: Port Token Activation sagas
Related to [colonyCDapp#285](https://github.com/JoinColony/colonyCDapp/issues/285), porting token activation involves porting the `userDepositToken` and `userWithdrawToken` sagas. 

Setting these up may also involve payments (so there are tokens to activate) and permissions (for payments), so may need coordinating with Raul (handling permissions) and Jakub (handling actions). To be confirmed by the person who takes on the issue.

jakubcolony:
## Description

This PR will contain the token activation logic and, since it was its dependency, the lambda function for getting the user's current balance.

Resolves #286


src/components/frame/TokenActivation/TokensTab/TokensTab.tsx
```diff
@@ -161,23 +145,28 @@ const TokensTab = ({
...
-                <Numeral value={formattedActiveTokens} suffix={token.symbol} />
+                <Numeral
+                  value={activeBalance ?? 0}
+                  decimals={tokenDecimals}
+                  suffix={nativeToken.symbol}
+                />
               </span>
-              {formattedActiveTokens === SMALL_TOKEN_AMOUNT_FORMAT && (
-                <SmallTokenAmountMessage />
-              )}
```
jakubcolony, line 168: I felt like these are not relevant anymore. The `Numeral` component will display the small values using engineering notation. And I guess the UI refresh will bring some clarification in regards to number formatting rules

amplify/backend/function/getUserTokenBalance/src/index.js
```diff
@@ -0,0 +1,78 @@
...
+      // Get token balance
+      const tokenClient = await getTokenClient(tokenAddress, provider);
+      const {
+        etherRouterAddress: networkAddress,
+      } = require('../../../../mock-data/colonyNetworkArtifacts/etherrouter-address.json');
+      const networkClient = getColonyNetworkClient(Network.Custom, provider, {
+        networkAddress,
+      });
+      const tokenLockingClient = await networkClient.getTokenLockingClient();
+      const userLock = await tokenLockingClient.getUserLock(
```
Julianahlk, line 47: Do you know why this is called userLock?

Julianahlk (approved):
Looking good 👍 . It works as expected. I made some changes to the base PR today which may have caused some conflicts.

amplify/backend/function/getUserTokenBalance/src/index.js
jakubcolony: As we discussed in the call yesterday, this encapsulates the concept of user locking their tokens for the purpose of voting

amplify/backend/function/getUserTokenBalance/src/index.js
```diff
@@ -0,0 +1,78 @@
...
+        tokenAddress,
+        walletAddress,
+      );
```
willm30, line 50: nitpick: the lines from 39 down to here could be moved into utils to keep this lamda function body "on topic" (i.e just getting / calculating balances)

willm30 (approved):
Looking good to me (subject to your comment about validation). 

One question I have is why if I pay myself 2 eth, it only shows 1.98? I know there's a network fee involved somewhere along the way, but in the dapp this is hidden from the user in the sense that if I pay myself 1 token, it shows up as 1 token, and I can activate 1 token (not 0.98). I quite like that. (Not suggesting anything gets changed in this pr, just curious if you've noticed this also).

Also, have we any plans to implement optimistic ui with respect to token activation? Not in this pr, but it'd be nice to see instant feedback without refreshing the browser (as currently happens in the dapp).

amplify/backend/function/getUserTokenBalance/src/index.js
jakubcolony: Agree, I'm doing that in #336

jakubcolony:
@willm30 
>One question I have is why if I pay myself 2 eth, it only shows 1.98? I know there's a network fee involved somewhere along the way, but in the dapp this is hidden from the user in the sense that if I pay myself 1 token, it shows up as 1 token, and I can activate 1 token (not 0.98). I quite like that. (Not suggesting anything gets changed in this pr, just curious if you've noticed this also).

Could it be because on xDai the fees are negligibly small? Even for 200k tokens payment, the fee is less than 0.00001:
image

I don't really have another explanation...

>Also, have we any plans to implement optimistic ui with respect to token activation? Not in this pr, but it'd be nice to see instant feedback without refreshing the browser (as currently happens in the dapp).

We don't have that for colony balance yet so I kinda went along that. We know how to do it, it's just a matter of getting it done, presumably alongside the UI refresh.

ArmandoGraterol:
@willm30 That would be because we are not including the network inverse fee in the amount when we make a payment, for now.

When we submit the payment action payload in the Dapp, we add the fee to the amount that the user inputted so that the receiving user, action page, and events get the amount shown in the input and not the amount after fees.

This fee logic is something that I'm implementing here (https://github.com/JoinColony/colonyCDapp/pull/325/commits/91ed1ba845adfc86c85c9bd9ec7a122dbaed3622 ), in the `helpers` is where the magic happens :)

PR #320, opened 2023-03-15
Wire the Manage Funds dialog with action sagas

### LINKED ISSUE #306: Wire Manage Funds dialog with corresponding action sagas

jakubcolony:
## Description

What started as a simple ticket grew quite a bit... But we now have a working Manage Funds dialog!

I struggled with setting a custom error in `react-hook-form`. Apparently `setError` doesn't influence the `isValid` property of the form state (which I find counterintuitive, but that's how it is: https://react-hook-form.com/api/useformstate/). I had to resort to conditionally passing a custom error to the input component

## Testing
Test the Transfer Funds, Mint Tokens, and Unlock Tokens dialog. Ensure the validation works as intended and the correct action is created, and then picked up by block-ingestor.

Resolves #306

### DIFF
--- a/package.json
+++ b/package.json
@@ -186,6 +186,6 @@
     "redux-saga": "^1.1.3",
     "reselect": "^4.1.5",
     "scroll-into-view-if-needed": "^2.2.16",
-    "yup": "^0.29.0"
+    "yup": "^0.30.0"
   }
 }
--- a/src/components/common/Dialogs/DomainFundSelectorSection/DomainFundSelectorSection.tsx
+++ b/src/components/common/Dialogs/DomainFundSelectorSection/DomainFundSelectorSection.tsx
@@ -39,13 +39,16 @@ const DomainFundSelectorSection = ({
   disabled,
 }: Props) => {
   const { watch, setValue, trigger } = useFormContext();
-  const motionDomainId = watch('motionDomainId');
+  const toDomainId = watch('toDomainId');
 
   const handleFromDomainChange = (fromDomainId: string) => {
-    if (motionDomainId !== fromDomainId) {
-      setValue('motionDomainId', fromDomainId);
+    setValue('motionDomainId', fromDomainId);
+
+    if (transferBetweenDomains) {
+      // Touch the toDomainId field to show any error messages and trigger validation of the entire form (e.g. the amount)
+      setValue('toDomainId', toDomainId, { shouldTouch: true });
+      trigger();
     }
-    trigger();
   };
 
   return (
--- a/src/components/common/Dialogs/DomainFundSelectorSection/DomainFundSelectorSection.tsx
+++ b/src/components/common/Dialogs/DomainFundSelectorSection/DomainFundSelectorSection.tsx
@@ -41,9 +41,9 @@ const DomainFundSelectorSection = ({
   const { watch, setValue, trigger } = useFormContext();
   const { motionDomainId } = watch();
 
-  const handleFromDomainChange = (fromDomainValue) => {
-    if (motionDomainId !== fromDomainValue) {
-      setValue('motionDomainId', fromDomainValue);
+  const handleFromDomainChange = (fromDomainId: string) => {
+    if (motionDomainId !== fromDomainId) {
+      setValue('motionDomainId', fromDomainId);
     }
     trigger();
   };
--- a/src/components/common/Dialogs/DomainFundSelectorSection/DomainFundSelectorSection.tsx
+++ b/src/components/common/Dialogs/DomainFundSelectorSection/DomainFundSelectorSection.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect } from 'react';
+import React from 'react';
 import { defineMessages } from 'react-intl';
 import { useFormContext } from 'react-hook-form';
 import classNames from 'classnames';
@@ -38,29 +38,14 @@ const DomainFundSelectorSection = ({
   transferBetweenDomains,
   disabled,
 }: Props) => {
-  const {
-    watch,
-    setValue,
-    formState: { errors },
-    clearErrors,
-  } = useFormContext();
-  const { motionDomainId, fromDomain, toDomain } = watch();
+  const { watch, setValue } = useFormContext();
+  const { motionDomainId } = watch();
+
   const handleFromDomainChange = (fromDomainValue) => {
     if (motionDomainId !== fromDomainValue) {
       setValue('motionDomainId', fromDomainValue);
     }
   };
-
-  useEffect(() => {
-    if (fromDomain !== toDomain) {
-      if (errors.toDomain?.type === 'same-pot') {
-        clearErrors('toDomain');
-      }
-      if (errors.fromDomain?.type === 'same-pot') {
-        clearErrors('fromDomain');
-      }
-    }
-  }, [clearErrors, errors, fromDomain, toDomain]);
 
   return (
     <div
--- a/src/components/common/Dialogs/DomainFundSelectorSection/DomainFundSelectorSection.tsx
+++ b/src/components/common/Dialogs/DomainFundSelectorSection/DomainFundSelectorSection.tsx
@@ -54,7 +54,7 @@ const DomainFundSelectorSection = ({
       })}
     >
       <DomainFundSelector
-        name="fromDomain"
+        name="fromDomainId"
         label={MSG.from}
         colony={colony}
         disabled={disabled}
@@ -69,7 +69,7 @@ const DomainFundSelectorSection = ({
             appearance={{ size: 'medium' }}
           />
           <DomainFundSelector
-            name="toDomain"
+            name="toDomainId"
             label={MSG.to}
             colony={colony}
             disabled={disabled}
--- a/src/components/common/Dialogs/DomainFundSelectorSection/DomainFundSelectorSection.tsx
+++ b/src/components/common/Dialogs/DomainFundSelectorSection/DomainFundSelectorSection.tsx
@@ -38,13 +38,14 @@ const DomainFundSelectorSection = ({
   transferBetweenDomains,
   disabled,
 }: Props) => {
-  const { watch, setValue } = useFormContext();
+  const { watch, setValue, trigger } = useFormContext();
   const { motionDomainId } = watch();
 
   const handleFromDomainChange = (fromDomainValue) => {
     if (motionDomainId !== fromDomainValue) {
       setValue('motionDomainId', fromDomainValue);
     }
+    trigger();
   };
 
   return (
--- a/src/components/common/Dialogs/DomainFundSelectorSection/DomainFundSelectorSection.tsx
+++ b/src/components/common/Dialogs/DomainFundSelectorSection/DomainFundSelectorSection.tsx
@@ -60,7 +60,7 @@ const DomainFundSelectorSection = ({
         clearErrors('fromDomain');
       }
     }
-  }, [errors, fromDomain, toDomain]);
+  }, [clearErrors, errors, fromDomain, toDomain]);
 
   return (
     <div
--- a/src/components/common/Dialogs/DomainFundSelectorSection/DomainFundSelectorSection.tsx
+++ b/src/components/common/Dialogs/DomainFundSelectorSection/DomainFundSelectorSection.tsx
@@ -39,7 +39,7 @@ const DomainFundSelectorSection = ({
   disabled,
 }: Props) => {
   const { watch, setValue, trigger } = useFormContext();
-  const { motionDomainId } = watch();
+  const motionDomainId = watch('motionDomainId');
 
   const handleFromDomainChange = (fromDomainId: string) => {
     if (motionDomainId !== fromDomainId) {
--- a/src/components/common/Dialogs/MintTokenDialog/MintTokenDialog.tsx
+++ b/src/components/common/Dialogs/MintTokenDialog/MintTokenDialog.tsx
@@ -1,15 +1,14 @@
 import React, { useState } from 'react';
 import { defineMessages } from 'react-intl';
-import { string, object, bool, InferType } from 'yup';
+import { string, object, bool, InferType, number } from 'yup';
 import { useNavigate } from 'react-router-dom';
-import Decimal from 'decimal.js';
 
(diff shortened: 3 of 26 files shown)


willm30 (changes requested):
It appears that when you click in, then out, of the token input without change the value, the form state is changed from the default value of 0 to undefined, prompting this error:

undef

I had no trouble with minting or unlocking tokens.

src/graphql/fragments/colony.graphql
```diff
@@ -100,6 +98,12 @@ fragment ColonyMetadata on ColonyMetadata {
   }
 }
 
+fragment ColonyBalances on ColonyBalances {
+  items {
+    ...ColonyBalance
+  }
+}
+
```
rdig, line 106: If we're adopting this pattern, then I'd argue this should be done for all fragments (jn this file, at least).

 Ie: `fundsClaims` as well

rdig (approved):
All looks good from my end. Nice work.

jakubcolony (commented):
Just realised I didn't post my comments and they were showing as pending...

src/components/shared/Fields/Input/HookForm/FormattedInputComponent.tsx
```diff
@@ -89,6 +94,15 @@ const HookFormFormattedInputComponent = ({
     }
   };
 
+  /**
+   * A custom change handler must be provided to ensure the hook-form holds the raw value of the input
+   * E.g. 123456 instead of 123,456
+   */
+  const handleCleaveChange = (e: CleaveChangeEvent) => {
```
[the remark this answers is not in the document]
jakubcolony: I'll go by YAGNI until there's a use case to support

src/components/shared/TokenAmountInput/TokenAmountInput.tsx
```diff
@@ -68,6 +73,7 @@ const TokenAmountInput = ({ colony, disabled }: Props) => {
           formattingOptions={formattingOptions}
           disabled={disabled}
           dataTest="paymentAmountInput"
+          customError={notEnoughBalance ? MSG.notEnoughBalance : undefined}
```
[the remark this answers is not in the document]
jakubcolony: Great suggestion, thanks!

src/graphql/fragments/colony.graphql
jakubcolony: The reason for adopting this is that we used to have an incorrect casting here: https://github.com/JoinColony/colonyCDapp/pull/320/files#diff-7b7b3e96bf501571f1ccd3bfff0c483a4dd74ac8260dfd344b3e52010d1f134aL47
`ColonyBalances` is a GraphQL type, not what we query for.

To be honest, I don't see a point in wrapping the balance data in `items` at all, it feels like it could be just directly returned by the query as an array `[ColonyBalance]`. And we'd have one less type to worry about. Was your intention here to make it consistent with arrays `items` returned by the automatically generated queries?

src/components/common/Dialogs/TransferFundsDialog/TransferFundsDialogForm.tsx
```diff
@@ -54,10 +61,13 @@ const MSG = defineMessages({
...
   const {
-    getValues,
     formState: { isSubmitting, isValid },
+    watch,
   } = useFormContext();
-  const values = getValues();
+  const fromDomainId = watch('fromDomainId');
+  const toDomainId = watch('toDomainId');
+  const amount = watch('amount');
+  const tokenAddress = watch('tokenAddress');
```
ArmandoGraterol, line 70: You can also use `watch()` here and then `const { fromDomainId, toDomainId, amount, tokenAddress } = watch()`

ArmandoGraterol (changes requested):
Looking good my friend! It seems like the `.` input issue has been resolved in `TransferFundsDialog`, in the `MintTokenDialog` we still have it. I'm okay with changing it myself in other Dialog PRs when you merge this or if you want to, you can change it now.

Just noticed a thing, and it's that there are no errors when you select a `fromDomain` equal to the `toDomain`:
FireShot Capture 818 - Colony CDapp - localhost

The submit button stays disabled but there are no errors like when you select a `toDomain` equal to the `fromDomain`.

package.json
```diff
@@ -186,6 +186,6 @@
     "redux-saga": "^1.1.3",
     "reselect": "^4.1.5",
     "scroll-into-view-if-needed": "^2.2.16",
-    "yup": "^0.29.0"
+    "yup": "^0.30.0"
```
jakubcolony, line 189: @willm30 I found this version to have a working test context and not cause any issues with existing validation

src/components/shared/DomainFundSelectorSection/DomainFundSelectorSection.tsx
```diff
@@ -38,38 +38,23 @@ const DomainFundSelectorSection = ({
...
-    formState: { errors },
-    clearErrors,
-  } = useFormContext();
+  const { getValues, setValue, trigger } = useFormContext();
   const values = getValues();
   const handleFromDomainChange = (fromDomainValue) => {
     if (values.motionDomainId !== fromDomainValue) {
       setValue('motionDomainId', fromDomainValue);
     }
+    trigger();
```
jakubcolony, line 47: Changing from domain should trigger the validation of both toDomainId (to check if it's not the same domain) and amount (a previously valid amount might not be valid for the new domain anymore)

src/components/common/Dialogs/DomainFundSelectorSection/DomainFundSelectorSection.tsx
```diff
@@ -38,38 +38,24 @@ const DomainFundSelectorSection = ({
...
-    watch,
-    setValue,
-    formState: { errors },
-    clearErrors,
-  } = useFormContext();
-  const { motionDomainId, fromDomain, toDomain } = watch();
+  const { watch, setValue, trigger } = useFormContext();
+  const { motionDomainId } = watch();
+
   const handleFromDomainChange = (fromDomainValue) => {
```
willm30, line 44: nitpick: type `fromDomainValue`

src/components/common/Dialogs/DomainFundSelectorSection/DomainFundSelectorSection.tsx
```diff
@@ -38,38 +38,24 @@ const DomainFundSelectorSection = ({
...
     if (motionDomainId !== fromDomainValue) {
       setValue('motionDomainId', fromDomainValue);
     }
+    trigger();
```
willm30, line 48: Consider passing: `['fromDomainId', 'toDomainId']` as the argument to `trigger` so it only validates those fields, because currently (in a colony generated via the script) if you switch the `to` domain to the same domain as the `from` domain, but then change the `from` domain so they're different again, the validation message for the amount input is displayed, even though it hasn't been touched.

src/components/common/Dialogs/MintTokenDialog/MintTokenDialog.tsx
```diff
@@ -35,16 +34,10 @@ const validationSchema = object()
   .shape({
     forceAction: bool().defined(),
     annotation: string().max(4000).defined(),
-    mintAmount: string()
+    mintAmount: number()
```
willm30, line 37: nitpick: It starts a number since the default value is `0`, but if you type anything in the value is saved as a string.

src/components/common/Dialogs/TransferFundsDialog/TransferFundsDialog.tsx
```diff
@@ -98,18 +50,25 @@ const TransferFundsDialog = ({
...
+  const defaultFromDomainId = selectedDomainId || Id.RootDomain;
+  const defaultToDomainId =
+    Number(
+      domainOptions.find((option) => option.value !== defaultFromDomainId)
+        ?.value,
+    ) || Id.RootDomain;
+
+  const validationSchema = getValidationSchema(colony);
+
+  type FormValues = InferType<typeof validationSchema>;
```
willm30, line 62: You can still define this type outside the component with: `InferType<ReturnType<typeof getValidationSchema>>`

I think that would be preferable, if only for consistency

willm30 (approved):
Nice one! Just a couple of notes.

The validation message you added when transfering funds to the same domain doesn't show if you're in a colony with only the root domain. But the form is disabled so we may want to add that visual feedback in this case also.
samedomain

I'm getting an error from the `EthUsd` component when the input field is empty. It also displays the loading spinner indefinitiely if the input is empty. Not necessarily a fix for this pr, just bringing to your attention.
eths

src/components/common/Dialogs/TransferFundsDialog/validation.ts
```diff
@@ -0,0 +1,77 @@
...
+    id: `${displayName}.amountZero`,
+    defaultMessage: 'Amount must be greater than zero',
+  },
+  sameDomain: {
+    id: `${displayName}.sameDomain`,
+    defaultMessage: 'Cannot move to same team pot',
+  },
+  notEnoughBalance: {
+    id: `${displayName}.notEnoughBalance`,
+    defaultMessage: 'Insufficient balance in from team pot',
```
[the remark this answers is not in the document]
jakubcolony: Is it? That's what it is currently in the Dapp and while it does sound a bit off, my understanding is that it relates to "from team pot", as in "fromDomain", and as opposed to "to team pot"

src/components/common/Dialogs/DomainFundSelectorSection/DomainFundSelectorSection.tsx
jakubcolony: I thought about that, and I think it's necessary for the amount to get revalidated if the from domain changes

Say you have 2 domains: A with 5 tokens and B with 1 token
If you select domain A and enter an amount of 2 tokens, it works okay. But when you change the domain to B, unless you trigger the amount validation, you won't see an error (and there's not enough balance in B for 2 tokens)

Maybe the problem is that we show the error for fields that haven't been touched. I'll look into it

src/components/common/Dialogs/MintTokenDialog/MintTokenDialog.tsx
jakubcolony: Which is fine, isn't it? This just declares the validation expects a number (and we convert it using `toFinite` a few lines lower)
It was `number` in the Dapp, we changed it to string because we were storing a formatted value (e.g. `123,456`) instead of a raw one

src/components/common/Dialogs/TransferFundsDialog/validation.ts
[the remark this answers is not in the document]
willm30: hmm... if it's not a typo it's badly worded.

src/components/common/Dialogs/MintTokenDialog/MintTokenDialog.tsx
willm30: I would ask why the validation expects a number if we know it's going to be validating a string (albeit a string containing a number). I don't think it affects the actual validation at all (hence nitpick) but it is technically incorrect.

jakubcolony: I can't say it's incorrect - we want to validate a number, and therefore have access to `number` methods such as `moreThan`, `transform`.

willm30: > we want to validate a number, and therefore have access to number methods

That's true, I was thinking of it as if it were an interface.

ArmandoGraterol (changes requested):
Solid work! I've noticed something regarding the validation:

The error is not showing if you touch the first domain selector before the second domain selector:
[96d52575-b741-4e54-95a7-de788502c6ca.webm](https://user-images.githubusercontent.com/18473896/227383981-99333244-53ee-4a25-9ac5-fdf519e21037.webm)

src/components/common/Dialogs/TokenAmountInput/TokenAmountInput.tsx
```diff
@@ -83,13 +84,16 @@ const TokenAmountInput = ({ colony, disabled }: Props) => {
             elementOnly
             appearance={{ alignOptions: 'right', theme: 'grey' }}
             disabled={disabled}
+            onChange={() => {
+              trigger('amount');
+            }}
```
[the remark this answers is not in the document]
jakubcolony: Note this is the `Select` for the token symbol, not the amount. We have to trigger the `amount` field validation manually when the symbol changes

src/components/common/Dialogs/DomainFundSelectorSection/DomainFundSelectorSection.tsx
```diff
@@ -38,29 +38,15 @@ const DomainFundSelectorSection = ({
...
-      }
-      if (errors.fromDomain?.type === 'same-pot') {
-        clearErrors('fromDomain');
-      }
+  const handleFromDomainChange = (fromDomainId: string) => {
+    if (motionDomainId !== fromDomainId) {
+      setValue('motionDomainId', fromDomainId);
     }
-  }, [errors, fromDomain, toDomain]);
+    trigger();
```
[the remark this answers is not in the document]
jakubcolony: `HookFormSelect` already does the check to see if the newly selected value isn't the same:
image
So actually it's the `if (motionDomainId !== fromDomainId)` that may be redundant here

src/components/common/Dialogs/TokenAmountInput/TokenAmountInput.tsx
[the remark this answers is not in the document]
jakubcolony: To add to the above, `hook-form` triggers the validation on a field-level, unlike formik, that I believe was triggering the validation of the entire form

src/components/common/Dialogs/DomainFundSelectorSection/DomainFundSelectorSection.tsx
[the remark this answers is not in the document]
ArmandoGraterol: > HookFormSelect already does the check to see if the newly selected value isn't the same

Good to know! Thanks

ArmandoGraterol (approved):
If this ain't good code, I don't know what is

PR #321, opened 2023-03-16
Port finalize motion functionality

### LINKED ISSUE #314: Wire in motion finalized
We have motion staking. 

Now, once the motion times out, we should be able to see the motion finalized screen.

willm30:
Note on size: about 2,000 lines of the diff are boiler plate courtesy of adding a new lamda function.

## Description

This PR ports the motion finalized ui and saga. 

**New stuff** ✨

* Lamda function to fetch motion state from chain
* Ported Finalize Motion ui
* Wired in Finalize motion saga

**Changes** 🏗

* Split out `useMinAndRequiredStake` into two separate hooks since requiredStake needs fetching outside of `StakingWidgetProvider` (in `useDefaultMotion`)
* Modified `DetailItem` to accept a tooltip for use in `FinalizeMotion`

**Deletions** ⚰️

* `useMinAndRequiredStake` 

Resolves  #314


amplify/backend/function/updateMotionState/src/utils.js
```diff
@@ -0,0 +1,131 @@
...
+
+const getLatestMotionState = async (colonyAddress, motionId) => {
+  const votingReputationClient = await getVotingClient(colonyAddress);
+  return votingReputationClient.getMotionState(motionId);
+};
+
+const POLLING_FREQUENCY = 1000;
+const POLLING_TIMEOUT = 1000 * 30; // 30 seconds
+
+const pollForColonyAction = async (transactionHash, actionData) => {
```
willm30, line 100: I don't think polling can be done easily like on the front end when making post requests directly. Not sure if this is the correct doc, but functionality seems limited to the query and variables. https://www.apollographql.com/docs/apollo-server/v2/requests/

amplify/backend/function/updateMotionState/src/index.js
willm30, line 1: I've created a lamda because the motion state is independent of user-specific actions (and thus can't be reliably determined by the block-ingestor). It ensures the app always has access to the motion's latest state, as taken from the chain. Currently used by `DefaultMotion` to get the latest motion state, but I also imagine it being useful for the countdown timer (when timer hits 0, fetch latest state).

willm30:
Closing in favour of #350

PR #324, opened 2023-03-16
Port motion reward claim

willm30:
## Description

This PR implements the claim widget.

claim

## Testing

Testing for both functionality (does it all work as expected) and ui (does it look right).

Follow the [set up instructions](https://github.com/JoinColony/colonyCDapp/pull/338) to be able to stake. Be sure to set the staking duration to something longer than 72 hours, as occasionally this causes the motion to fail due to how time works in development. 720 would do.

Stake in favour of the motion 100%. Do not stake against. This will break it since we haven't set up voting/reveal yet.

Fast forward time by running the following in your terminal (fast forwards by 1000 hours):

```
curl -X POST --data '{"jsonrpc":"2.0","method":"evm_increaseTime","params":[3600000],"id":1}' localhost:8545

```

Click Finalize. You should then see it loading, before it displays the Claim Widget.
Click Claim. You should see the widget disabled, and you can verify that the rewards have been claimed by looking at the `stakerRewards` field of the motion in the database.

**Note**: It's possible that after hitting claim, your dev env will crash. It certainly has been for me ([this script specifically](https://github.com/JoinColony/colonyNetwork/blob/develop/scripts/start-blockchain-client.sh). While clearly a problem, I suggest we worry about it when we get to QA, if it's still occurring then). It doesn't fundamentally affect the testing of this branch. Just means you can't do anything after hitting claim (which is the end of the test anyway).

**New stuff** ✨

* Claim widget components
* Wired in claimReward saga

**Changes** 🏗

* Updated schema for claiming

Contributes to #322

### DIFF
--- a/amplify/backend/api/colonycdapp/schema.graphql
+++ b/amplify/backend/api/colonycdapp/schema.graphql
@@ -457,21 +457,22 @@ type UserStakes {
 type StakerRewards {
   address: String!
   rewards: MotionStakeValues!
+  isClaimed: Boolean!
 }
 
 type MotionData {
-  motionId: String!
+  motionId: String! # to ensure uniqueness, we format as: chainId-votingRepExtnAddress_nativeMotionId
+  nativeMotionId: String!
   usersStakes: [UserStakes!]!
   stakerRewards: [StakerRewards!]!
   motionStakes: MotionStakes!
   remainingStakes: [String!]! # tuple [nayRemaining, yayRemaining]
   userMinStake: String!
   requiredStake: String!
-  # For calculating user's max stake in client
-  rootHash: String!
-  # native domain id
-  motionDomainId: String!
+  rootHash: String! # For calculating user's max stake in client
+  motionDomainId: String! # native domain id
   isFinalized: Boolean!
+  createdBy: String! # voting rep extn address. Useful to check if we're viewing a "read-only" motion
 }
 
 # This will store the relevant events we care about for a particular colony
--- a/src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/MotionPhaseWidget.tsx
+++ b/src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/MotionPhaseWidget.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
+
 import { StakerRewards } from '~gql';
 import { useAppContext } from '~hooks';
-
 import { ColonyAction, ColonyActionType, Address } from '~types';
 import { MotionState } from '~utils/colonyMotions';
 
--- a/src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/MotionPhaseWidget.tsx
+++ b/src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/MotionPhaseWidget.tsx
@@ -1,13 +1,27 @@
 import React from 'react';
+import { StakerRewards } from '~gql';
+import { useAppContext } from '~hooks';
 
-import { ColonyAction, ColonyActionType } from '~types';
+import { ColonyAction, ColonyActionType, Address } from '~types';
 import { MotionState } from '~utils/colonyMotions';
 
+import ClaimMotionStakes from './ClaimMotionStakes';
 import FinalizeMotion from './FinalizeMotion';
 import StakingWidget, { StakingWidgetProvider } from './StakingWidget';
 
 const displayName =
   'common.ColonyActions.ActionDetailsPage.DefaultMotion.MotionPhaseWidget';
+
+const isMotionClaimed = (
+  stakerRewards: StakerRewards[],
+  walletAddress: Address,
+) => {
+  const userReward = stakerRewards.find(
+    ({ address }) => address === walletAddress,
+  );
+
+  return !!userReward?.isClaimed;
+};
 
 interface MotionPhaseWidgetProps {
   actionData: ColonyAction;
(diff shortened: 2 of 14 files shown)


src/redux/sagas/motions/claimMotionRewards.ts
```diff
@@ -39,83 +34,41 @@ function* claimMotionRewards({
   const apolloClient = getContext(ContextModule.ApolloClient);
 
   try {
-    const colonyManager = yield getColonyManager();
```
willm30, line 42: The following logic checks, of the motion ids provided to the saga, which can still be claimed by the user, and on which side there's a reward. 

The difference is that I've updated it to verify this with the db, instead of the chain. Reasoning being, querying the db is faster than the chain (I'm assuming).

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -406,21 +406,22 @@ type UserStakes {
...
   userMinStake: String!
   requiredStake: String!
-  # For calculating user's max stake in client
-  rootHash: String!
-  # native domain id
-  motionDomainId: String!
+  rootHash: String! # For calculating user's max stake in client
+  motionDomainId: String! # native domain id
   isFinalized: Boolean!
+  createdBy: String! # voting rep extn address. Useful to check if we're viewing a "read-only" motion
```
willm30, line 424: I anticipate this being useful when it comes to viewing motions that were created by a now-deleted voting reputation client. 

It would allow for a very simple check in the front-end for whether the motion is "current", and if not, render it "read-only" - i.e. can't be interacted with. It's essentially freezing the motion at the state it was when its voting client was uninstalled.

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -406,21 +406,22 @@ type UserStakes {
 type StakerRewards {
   address: String!
   rewards: MotionStakeValues!
+  isClaimed: Boolean!
 }
 
 type MotionData {
-  motionId: String!
+  motionId: String! # to ensure uniqueness, we format as: chainId-votingRepExtnAddress_nativeMotionId
```
willm30, line 413: Introducing this format for motion ids is useful in the block ingestor when fetching a motion from the db (since native motion ids are not unique across voting rep installations) and also in the claimReward saga, when we need to check which motions have outstanding claims.

willm30: https://github.com/JoinColony/block-ingestor/pull/42

src/redux/sagas/motions/claimMotionRewards.ts
```diff
@@ -209,6 +128,59 @@ function* claimMotionRewards({
...
+export default function* claimAllMotionRewardsSaga() {
   yield takeEvery(ActionTypes.MOTION_CLAIM, claimMotionRewards);
 }
+
+/**
+ * Client-side filtering the motions in the db for those
+ * with outstanding claims by the given user
+ */
+
+function getMotionsWithClaims(
```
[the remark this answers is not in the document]
rdig: I actually agree on both counts:
- the motion model has grown large and it now makes sense to split out and be independent
- I'm all for optimizing the speed at which we generate / send transactions, so this is a welcomed improvement

rdig (approved):
Looks good at a first initial pass.

I agree with both your improvement suggestions, however I think they should be done in separate issues.

src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/MotionPhaseWidget.tsx
```diff
@@ -1,14 +1,28 @@
 import React from 'react';
+import { StakerRewards } from '~gql';
```
ArmandoGraterol, line 2: There needs to be a line of separation between these imports

src/redux/sagas/motions/claimMotionRewards.ts
```diff
@@ -1,121 +1,85 @@
...
 } from '../transactions';
 
-import {
-  updateMotionValues,
-  putError,
-  takeFrom,
-  getColonyManager,
-} from '../utils';
+import { getColonyManager, putError, takeFrom } from '../utils';
+import { getMotionDatabaseId } from '~utils/colonyMotions';
```
ArmandoGraterol, line 24: This import should go above

src/redux/sagas/motions/index.ts
```diff
@@ -21,7 +21,8 @@ export default function* actionsSagas() {
     // call(voteMotionSaga),
     // call(revealVoteMotionSaga),
     call(finalizeMotionSaga),
-    // call(claimMotionRewardsSaga),
+    call(claimMotionRewardsSaga),
+    call(claimMotionRewardsSaga),
```
ArmandoGraterol, line 25: How come there are 2 calls for 1 saga? Typo?

ArmandoGraterol (approved):
Looks good to me! :ok_hand: Nice work

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -457,21 +457,22 @@ type UserStakes {
...
   motionStakes: MotionStakes!
   remainingStakes: [String!]! # tuple [nayRemaining, yayRemaining]
   userMinStake: String!
   requiredStake: String!
-  # For calculating user's max stake in client
-  rootHash: String!
-  # native domain id
-  motionDomainId: String!
+  rootHash: String! # For calculating user's max stake in client
+  motionDomainId: String! # native domain id
```
jakubcolony, line 473: Calling this `motionNativeDomainId` would probably remove the need for the comment (that's only visible in schema and not for the API consumers)

src/redux/sagas/motions/claimMotionRewards.ts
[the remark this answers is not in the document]
jakubcolony: Yep, as soon as there's a need to index on a specific field (motion ID in this case), it makes sense to split it out

jakubcolony (approved):
Nice work! 👍

PR #325, opened 2023-03-16
Wire up Create Payment dialog with action

### LINKED ISSUE #305: Wire Create Expenditure dialog with corresponding action sagas

ArmandoGraterol:
Integrated the create payment action with their related dialog.

I've added a new `useNetworkInverseFee` hook with a respective query to fetch the inverse fee needed by the payment action to calculate the correct amount to be paid. 

The `block-ingestor` should be able to pick up the initial network inverse fee (`100` in the dev env) with the help of `colony-js`, this fee is only updated by the Metacolony, and if you want to manually change the fee in the dev env or trigger the event associated with the change, you can do something like this:

1. `npm run truffle console`
2.  `n = await IColonyNetwork.at('0x5CC4a96B08e8C88f2c6FC5772496FeD9666e4D1F')`
3. `mca = await n.getMetaColony()`
4. `mc = await IMetaColony.at(mca)`
5. `mc.setNetworkFeeInverse(<Any number you would like to be the new fee, like 111>)`
6. Check if the new fee was picked up by going to the Payment dialog and seeing if the fee is different, or by directly querying for it in the Amplify API with `listCurrentNetworkInverseFees`

Resolves #305


amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -175,6 +175,8 @@ type Query {
     @function(name: "getReputationForTopDomains")
   getUserReputation(input: GetUserReputationInput): String
     @function(name: "getUserReputation")
+  getNetworkInverseFee: String
+    @function(name: "getNetworkInverseFee")
```
rdig, line 179: No no. This should not longer function as it did back in the Dapp.

This value is very rarely changed, even more so, when it gets changed an event gets emitted _(as far as I remember, check with @area)_

So this logic needs to be moved over in the block ingestor, fetch the value in there, stored in the database and retrieved properly via graphql queries.

rdig (changes requested):
I've stopped reviewing this after I noticed this fetches data via a lambda function. This needs to be moved over to the block ingestor.

I'll re-review this once that has been refactored.

src/components/common/Dialogs/CreatePaymentDialog/CreatePaymentDialog.tsx
```diff
@@ -45,18 +43,21 @@ const CreatePaymentDialog = ({
...
+  const { networkInverseFee } = useNetworkInverseFee();
 
   const actionType =
     !isForce && isVotingReputationEnabled
       ? ActionTypes.MOTION_EXPENDITURE_PAYMENT
       : ActionTypes.ACTION_EXPENDITURE_PAYMENT;
 
+  const validationSchema = getValidationSchema(colony);
+
+  type FormValues = InferType<typeof validationSchema>;
```
jakubcolony, line 55: As per Will's comment in #320, this can be moved outside the component and read:
```js
type FormValues = InferType<ReturnType<typeof getValidationSchema>>;
```

jakubcolony (changes requested):
@ArmandoGraterol I think this needs a rebase now that I also rebased #320 on master.

I can see the lambda is largely based off how I implemented the `setCurrentVersion` function, but as @willm30 pointed out, unless you call it directly from the CDapp, the entire logic of getting the existing fee entry and updating it can be moved to `block-ingestor`.

The enough balance check doesn't include network fee in the calculation, which leads to this form being valid (my balance is 10):
image
But clicking on Confirm triggers an underflow error:
image

Entering a decimal point (`.`) into the amount field crashes the entire app for me. I think we can now get rid of any conversion using decimal.js since you based this off #320?

Please ping me for another review once you've managed to rebase

src/components/common/Dialogs/CreatePaymentDialog/CreatePaymentDialogForm.tsx
```diff
@@ -63,7 +63,7 @@ const MSG = defineMessages({
 });
 
 interface Props extends ActionDialogProps {
-  verifiedUsers: ColonyWatcher['user'][];
+  verifiedUsers: User[];
```
jakubcolony, line 66: Shouldn't this be `MemberUser` instead? You get those off `watchers` and `contributors` which don't fetch the entire User fragment

jakubcolony (approved):
LGTM 👍

src/components/common/Dialogs/CreatePaymentDialog/CreatePaymentDialog.tsx
```diff
@@ -94,9 +116,9 @@ const CreatePaymentDialog = ({
       <Form<FormValues>
         defaultValues={{
           forceAction: false,
-          fromDomain: filteredDomainId || Id.RootDomain,
+          fromDomainId: filteredDomainId || Id.RootDomain,
           recipient: undefined,
-          amount: '',
+          amount: 0,
```
Julianahlk, line 121: Just curious of the reasoning behind changing this from empty string to 0? Also I noticed the validation message "amount is a required field" is no longer present if the amount field is empty.

Julianahlk (approved):
Nice work. The network fee display is working well for me. 

2 issues I am seeing compared to the dapp:
1. The network fee is being displayed at the same time with the error message. In the dapp, if there is an error, then only the error message is displayed.
2. The alignment of the network fee text should be similar to the error message text.

Screen Shot 2023-04-19 at 9 22 21 PM

From Dapp:
Screen Shot 2023-04-19 at 9 22 43 PM

src/components/common/Dialogs/CreatePaymentDialog/CreatePaymentDialog.tsx
ArmandoGraterol: Sure, we are validating the `amount` field as a number in the validation schema, and we infer the `FormValues` type from the schema, so the default value needs to be `0`, otherwise we get a typescript error.

Right now if you leave the field empty you should get an `Amount must be greater than 0`. I think that may be because `toFinite` interprets `''` as `0`.

PR #329, opened 2023-03-17
Port staking widget

### LINKED ISSUE #270: Port Staking Widget

willm30:
**Note:** all the prs that went into this one have already been reviewed, so we should be ok with giving this the green light.

## Description

This PR is the "mini"-feature branch for the staking widget. There is a lot to port so I have broken up into its various parts. 

Update: It's had a little bit more added to it as the staking widget validation was the last to be added, and came after merging finalize, claim and voting ui.

**New stuff** ✨

* Staking widget ui
* Staking widget saga
* Finalize widget
* Claim widget
* Voting ui

Resolves #270


ArmandoGraterol (approved):

rdig (approved):
Nothing jumps out, but then again, we've reviewed this as separate PRs prior

amplify/backend/function/fetchMotionState/src/utils.js
```diff
@@ -0,0 +1,36 @@
...
+const getVotingClient = async (colonyAddress) => {
+  const networkClient = getNetworkClient();
+  const colonyClient = await networkClient.getColonyClient(colonyAddress);
+  return colonyClient.getExtensionClient(Extension.VotingReputation);
+};
+
+const getLatestMotionState = async (colonyAddress, motionId) => {
+  const votingReputationClient = await getVotingClient(colonyAddress);
+  return votingReputationClient.getMotionState(motionId);
+};
```
jakubcolony, line 34: Is there no way (ie. event that gets emitted when the state changes) to cache that on the motion model directly?

src/graphql/fragments/actions.graphql
```diff
@@ -25,7 +25,58 @@ fragment ColonyAction on ColonyAction {
...
   isMotion
+  motionData {
+    ...MotionData
+  }
   colony {
     colonyAddress: id
+    nativeToken {
+      nativeTokenDecimals: decimals
+      nativeTokenSymbol: symbol
+      tokenAddress: id
```
jakubcolony, line 36: Do we need to get that here instead of using the colony data from the context?

src/utils/domains.ts
```diff
@@ -34,3 +34,20 @@ export const getDomainOptions = (colonyDomains: Domain[]) =>
...
+) => {
+  const { balance } =
+    balances?.items
+      ?.filter(notNull)
+      .find(
+        ({ domain: { nativeId } }) => nativeId === Number(nativeDomainId),
+      ) || {};
+
+  return balance;
+};
```
jakubcolony, line 53: I'm not sure but could this be potentially merged with `getBalanceForTokenAndDomain` from `utils/tokens`?

jakubcolony (approved):
Great work, a lot of effort went into this... 👍

amplify/backend/function/fetchMotionState/src/utils.js
willm30: The problem is when the timer runs out. The contracts don't have a concept of chron-jobs, therefore no event gets emitted when the timer runs out. Instead, you have to query the motion state on the contracts directly.

src/graphql/fragments/actions.graphql
willm30: Staking slider can't access context because it gets opened in the objection modal dialog

PR #340, opened 2023-03-23
Update staking widget ui after staking

willm30:
## Description

I initially intended this PR to "optimistically" update the ui, however, it transpires that it's much simpler just to poll the db and show a loading spinner while we wait. So the tradeoff is increased simplicity at the expense of waiting for the block ingestor to process the staking event before we show the user the updated values (where optimistic ui would be instant, but require every component to keep track of its own state, in addition to it already being held in the db). 

Therefore, this PR adds ensures the staking widget updates correctly after staking or objecting.

## Testing

Set up your env via [these instructions](https://github.com/JoinColony/colonyCDapp/pull/338)

Then stake both for and against the motion. You should see a loading state while we poll the db, and then updated values in the widget once the db updates (and, of course, the polling should stop).

Also, if you're testing this before testing #344 , #345, and #346, confirm that the objection dialog modal works, as does the grouped total stake view (and report comments in their respective prs).

updates

**New stuff** ✨

* `useStakingWidgetUpdate` to handle polling cancellation and loading state

**Changes** 🏗

* Add polling to stake widget
* Add loading state to stake widget
* Wire in handle success in both staking input and objection modal

Contributes to #270

### DIFF
--- a/src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/DefaultMotion.tsx
+++ b/src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/DefaultMotion.tsx
@@ -6,6 +6,7 @@ import DetailsWidget from '~shared/DetailsWidget';
 import { useColonyContext, useEnabledExtensions } from '~hooks';
 import { ColonyAction } from '~types';
 import { motionStateMap } from '~utils/colonyMotions';
+import { STAKING_THRESHOLD } from '~constants';
 
 import { DefaultActionContent } from '../DefaultAction';
 import MotionHeading from './MotionHeading';
@@ -32,17 +33,17 @@ const DefaultMotion = ({ actionData }: DefaultMotionProps) => {
   const motionState = MotionState.Staking;
   const {
     motionData: {
-      motionStakes: {
-        percentage: { nay, yay },
-      },
+      motionStakes: { percentage: percentageStaked },
     },
   } = actionData;
-  const stakeLessThan10Percent = Number(nay) + Number(yay) < 10;
+  const isUnderThreshold =
+    Number(percentageStaked.nay) + Number(percentageStaked.yay) <
+    STAKING_THRESHOLD;
 
   return (
     <div className={styles.main}>
       {/* {isMobile && <ColonyHomeInfo showNavigation isMobile />} */}
-      {stakeLessThan10Percent && <StakeRequiredBanner isDecision={false} />}
+      {isUnderThreshold && <StakeRequiredBanner isDecision={false} />}
       {isVotingReputationEnabled && (
         <MotionHeading motionState={motionStateMap[motionState]} />
       )}
--- a/src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/DefaultMotion.tsx
+++ b/src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/DefaultMotion.tsx
@@ -19,9 +19,11 @@ const displayName = 'common.ColonyActions.ActionDetailsPage.DefaultMotion';
 
 interface DefaultMotionProps {
   actionData: ColonyAction;
+  startPollingAction: (pollingInterval: number) => void;
+  stopPollingAction: () => void;
 }
 
-const DefaultMotion = ({ actionData }: DefaultMotionProps) => {
+const DefaultMotion = ({ actionData, ...rest }: DefaultMotionProps) => {
   const { colony } = useColonyContext();
 
   const { isVotingReputationEnabled } = useEnabledExtensions();
@@ -53,6 +55,7 @@ const DefaultMotion = ({ actionData }: DefaultMotionProps) => {
           <MotionPhaseWidget
             actionData={actionData}
             motionState={motionState}
+            {...rest}
           />
           {/* <div className={styles.details}>
         {motionState === MotionState.Voting && (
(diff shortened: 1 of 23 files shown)


src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/DefaultMotion.tsx
```diff
@@ -32,17 +39,17 @@ const DefaultMotion = ({ actionData }: DefaultMotionProps) => {
...
     motionData: {
-      motionStakes: {
-        percentage: { nay, yay },
-      },
+      motionStakes: { percentage: percentageStaked },
     },
   } = actionData;
-  const stakeLessThan10Percent = Number(nay) + Number(yay) < 10;
+  const isUnderThreshold =
+    Number(percentageStaked.nay) + Number(percentageStaked.yay) <
```
willm30, line 46: Changed for readability

src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/StakingWidget/StakingInput/StakingInput.tsx
```diff
@@ -21,58 +18,14 @@ const validationSchema = object({
...
-  const vote = isObjection ? MotionVote.Nay : MotionVote.Yay;
-  const transform = getStakingTransformFn(
-    remainingToStake,
-    userMinStake,
-    user?.walletAddress ?? '',
-    colony?.colonyAddress ?? '',
-    motionId,
-    vote,
-  );
+  const stakingInputRef = useRef<StakingInputRef>();
```
willm30, line 26: In my mind, this is the best way to pass data from `StakingSlider` up the component hierarchy (using `useImperativeHandler in `useStakingSlider`). It's reused in RaiseObjectionDialog, meaning we can't use lifted state or context.

src/hooks/useStakingWidgetUpdate.ts
```diff
@@ -0,0 +1,27 @@
+import { useState, useEffect } from 'react';
+
+import { MotionStakes } from '~gql';
+import { SetStateFn } from '~types';
+
+import { compareMotionStakes } from './helpers';
+
+const useStakingWidgetUpdate = (
```
willm30, line 8: When we stake successfully, we start polling the db again to get the updated staking values from the block ingestor. 

When the motion stakes object's content changes, we know to stop polling.

Julianahlk:
@willm30 When testing the staking widget UI, after doing all the steps and clicking the Test Mint Tokens button, I am able to get to the staking UI and successfully test but if I try to repeat the testing process I get this error when clicking the Test Mint Tokens button. Any ideas of why this could be happening?

Screen Shot 2023-03-29 at 9 05 10 PM

ArmandoGraterol (commented):
The code is looking good to me! However, I'm having trouble testing this branch in my env. Whenever I get to the part of minting tokens, `isVotingReputationEnabled` is `false` even though I've enabled it earlier, tried to force the action to be `ROOT_MOTION` but then the motion page doesn't load. Has any of you guys encountered this problem? Am I missing something? I thought I followed the instructions to the T

src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/StakingWidget/StakingInput/StakingSliderMessages/RequiredStakeMessage.tsx
```diff
@@ -4,7 +4,9 @@ import { defineMessages, FormattedMessage } from 'react-intl';
 import classNames from 'classnames';
 
 import Numeral from '~shared/Numeral';
-import { getStakeFromSlider, getStakePercentage } from '../..';
+import { STAKING_THRESHOLD } from '~constants';
```
rdig, line 7: I just made a comment in #344 about the staking threshold being a constant.

I guess this answers my concern :P

src/hooks/helpers/motionHookHelpers.ts
```diff
@@ -0,0 +1,51 @@
+import { BigNumber } from 'ethers';
+import { getStakeFromSlider } from '~common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/StakingWidget';
```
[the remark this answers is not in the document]
rdig: Yeah, I second that.

Even though this is super specific, I think that for "extreme" cases we can make exceptions and add an specific alias

rdig (approved):
All good.

This actually addresses stuff I pointed out in a diffrent PR's review :+1:

src/hooks/helpers/motionHookHelpers.ts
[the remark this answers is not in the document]
willm30: I will be moving these hooks into their component folders as over a few separate github comments with a few different people we seem to be converging on the idea that component-specific hooks should live with components and leave the top level hooks folder for hooks reused around the app. So that will solve the ugly import
#354

willm30:
@Julianahlk This looks like you've got no rep -- did you add it? If so, did the reputation monitor crash while you were testing?

willm30:
@ArmandoGraterol All I can think of here is that the event didn't get picked up in the block ingestor (or didn't handled quickly enough: for example, if you install the extn and then enable it before the install event has been processed, sometimes it causes the ui to become out of sync with the chain). Could you retry but refresh the browser after each stage of the installation to ensure it's been processed by the block ingestor?

ArmandoGraterol (approved):
Looks good to me :ok_hand: Thank you Will, that was it! I waited a bit and then reloaded before enabling the extension

PR #341, opened 2023-03-23
Wire up award/smite reputation dialogs

### LINKED ISSUE #308: Wire Manage Reputation dialog with corresponding action sagas

ArmandoGraterol:
Just like the Create/Edit Domain dialogs, the Award/Smite Dialogs didn't really need much work to connect them. 

As always, make sure the reputation monitor is on before testing. 

Resolves #308


src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationDialogForm/ManageReputationDialogForm.tsx
```diff
@@ -148,9 +150,7 @@ const ManageReputationDialogForm = ({
     ['value'],
   );
 
-  const domainName = colonyDomains.find(
-    (domain) => domain?.nativeId === domainId,
-  )?.metadata?.name;
+  const domainName = findDomainByNativeId(domainId, colony)?.metadata?.name;
```
rdig, line 153: I'm really not happy with this approach. 

Either destructure it, or use an intermediate variable

rdig (approved):
Mostly ok with this, however left a comment about a pattern used to fetch the domain's name

ArmandoGraterol:
> What's supposed to happen when I smite? I add 100 rep, shows up correctly, and then when I smite the max (100 rep), nothing happens. It still shows 100% after some time.

That shouldn't happen :thinking:. I just checked just to be sure, and first I tested it by generating reputation via payments and then smiting all of it and it worked, I had `0%`. Then, I got reputation by awarding myself 100pts and smiting it all as before and it showed `0%` once more. 

The only thing that I can think of right now that is happening is that maybe you forgot to have the reputation monitor ON or the monitor didn't mine the reputation in time by the time you refreshed the page and checked again.

willm30:
@ArmandoGraterol The ingestor error is saying that the query it's making is not correct. It's trying to write "skillId" to `ColonyAction`, which is not defined in this branch https://github.com/JoinColony/colonyCDapp/blob/c97ccb81ee62e219dfc6449a92c63c4dd01d97f4/amplify/backend/api/colonycdapp/schema.graphql#L435

The commit you're pointing it at (82bfd27bb9679e85fee476bc33df36e349ec3d57) is not the latest. Since you want to merge to master, we should ensure it's pointing to the latest block ingestor commit.

ArmandoGraterol:
@willm30 Oh I see, I think I misunderstood you. It's now something related to this PR (it's on `master` already) but the fix thankfully seems easy enough!

I've bumped the commit hash to the latest, and I haven't seen the error since. Let me know if it works for you too

willm30 (approved):
Second time lucky 🙂👍

PR #342, opened 2023-03-24
Wire up buttons of ColonyDomainSelector with domain dialogs

### LINKED ISSUE #187: Wire in domain logic
ColonyDomainSelector could do with some love. 

- Wire in the EditDomainDialog modal
- Convert to hook form

We also have the MotionDomainSelect component that needs bringing over.

I'm sure there is more that can be specced out. This is a start.

ArmandoGraterol:
[ebda8e2e-da32-4e8e-a717-fa414fa72002.webm](https://user-images.githubusercontent.com/18473896/227390916-03d8c7e3-4a59-4126-9026-d35948b947a3.webm)

Resolves #187

### DIFF
--- a/src/components/common/ColonyHome/ColonyDomainSelector/ColonyDomainSelector.tsx
+++ b/src/components/common/ColonyHome/ColonyDomainSelector/ColonyDomainSelector.tsx
@@ -1,13 +1,13 @@
 import React, { ReactNode, useCallback } from 'react';
 import { Id } from '@colony/colony-js';
 
+import { useDialog } from '~shared/Dialog';
 import ColorTag from '~shared/ColorTag';
 import { HookForm as Form, SelectOption } from '~shared/Fields';
 import DomainDropdown from '~shared/DomainDropdown';
-// import { useDialog } from '~shared/Dialog';
-// import EditDomainDialog from '~dialogs/EditDomainDialog';
+import { EditDomainDialog } from '~common/Dialogs';
 
-import { useColonyContext } from '~hooks';
+import { useColonyContext, useEnabledExtensions } from '~hooks';
 import { COLONY_TOTAL_BALANCE_DOMAIN_ID } from '~constants';
 // import { oneTxMustBeUpgraded } from '~modules/dashboard/checks';
 import { DomainColor } from '~types';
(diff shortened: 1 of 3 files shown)


src/components/common/ColonyHome/ColonyDomainSelector/ColonyDomainSelector.tsx
```diff
@@ -33,18 +33,16 @@ const ColonyDomainSelector = ({
...
-  //     }),
-  //   [openEditDialog, colony],
-  // );
-
-  const handleEditDomain = () => {};
+  const enabledExtensionData = useEnabledExtensions();
+
+  const openEditDialog = useDialog(EditDomainDialog);
+  const handleEditDomain = (selectedDomainId: number) =>
+    colony &&
```
rdig, line 40: This check is not needed.

The `useColonyContext` hook already does that

src/components/common/ColonyHome/ColonyDomainSelector/CreateDomainButton.tsx
```diff
@@ -18,42 +18,30 @@ const MSG = defineMessages({
...
-  //     });
-  //   },
-  //   [openCreateDomainDialog, colony],
-  // );
-
-  const handleClick = () => colony;
-
-  const text = formatMessage(MSG.buttonCreateNewDomain);
+  const handleClick = () =>
+    colony &&
```
rdig, line 26: Same here

src/components/common/Dialogs/CreateDomainDialog/CreateDomainDialog.tsx
```diff
@@ -86,7 +88,7 @@ const CreateDomainDialog = ({
           }
           return (
             <CreateDomainDialogForm
-              back={prevStep ? () => callStep(prevStep) : undefined}
+              back={prevStep && callStep ? () => callStep(prevStep) : undefined}
```
rdig, line 91: I think this might be a type issue here, as you shouldn't need to check `callStep` for existence

rdig (approved):
Left some comments that need to be addressed, but I'm pre-approving it

src/components/common/ColonyHome/ColonyDomainSelector/ColonyDomainSelector.tsx
willm30: The alternative pattern we've been using is:

``` js

if(!colony) {
 return null
}
```
which gets typescript off your case about it being maybe undefined.

@jakubcolony Maybe this is actually a legitimate use case for type casting the colony object to be defined if its consumed by the useColonyContext hook, given that we known if it's undefined, we render the 404?

willm30 (approved):
LGTM 👍

src/components/common/Dialogs/CreateDomainDialog/CreateDomainDialog.tsx
ArmandoGraterol: This is a result of me making `WizardDialogType<object>` Partial because now this Dialog can be called from inside the colony domain selector, which doesn't provide `prevStep` nor a `callStep` prop because it's not the wizard we use for the UAC.

src/components/common/ColonyHome/ColonyDomainSelector/ColonyDomainSelector.tsx
ArmandoGraterol: If the hook already handles `colony` not being defined, then yeah other than adding checks like this one or the one you suggested to reassure typescript, I think typecasting might be the way in this case.

Other than taking a deeper look into the hook and seeing if we can refactor it so we don't need to typecast that is.

PR #343, opened 2023-03-24
Add prettier config to cdapp

willm30:
## Description

I have wanted to do this for ages and just assumed there was some reason we didn't have one. But this came up in Monday's call and so now I'm proposing the following basic config (that can be extended over time as needed).

This will ensure we all commit according to the same presentation rules, eliminate purely presentation-based diffs, and get this file out of my source control!

**New stuff** ✨

* Prettier config

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -16,6 +16,7 @@ module.exports = {
     'eslint-config-airbnb/rules/react-a11y',
     'plugin:@typescript-eslint/recommended',
     '@colony/eslint-config-colony',
+    'prettier',
   ],
   plugins: ['@typescript-eslint', 'react', 'jsx-a11y', 'jsdoc', 'react-hooks'],
   overrides: [


.prettierrc
willm30, line 1: Note: I had a .prettierignore file that included `*.css` files to reduce diffs in the dapp, because in many css files we use singles and double quotes inconsistency. By omitting this, this will change all double quotes to single quotes in css files whenever you're working in a css file. I don't see this as a bad thing, but may result in some added diffs initially.

jakubcolony:
Good to see some movement on that! 

The reason I left #96 hanging around for so long and not just added this file is because it relies on people having the prettier extension installed and I don't think that's good enough (think of the eslint issues in block-ingestor that weren't picked up because you didn't have the right extension in place).

Another problem this poses is that some existing eslint rules may cause conflicts with how prettier formats files (I stumbled across this a few times, especially in block-ingestor)

Fortunately the solution doesn't seem too complex: https://prettier.io/docs/en/integrating-with-linters.html. Having read through it (and https://prettier.io/docs/en/precommit.html), it looks like it's not recommended to make linters check the formatting of code, so we should just disable the conflicting rules and then add prettier as a dev dependency and run its check in either pre-commit or pre-push (configured in `.huskyrc`)

jakubcolony:
Another thought I just had is whether this PR should apply the prettier rules to all the existing CDapp files, so that those purely stylistic diffs don't pollute future PRs

rdig (changes requested):
Fully agree with @jakubcolony here.

In the past we've had prettier integrated with eslint (without the need of code editor extensions -- although people still used them), so I'd like us to have that setup again, if we're to use prettier.

Second, I again agree with Jakub, after this PR adds the required rules and integrations, a subsequent PR with all the file changes needs to be added fast, so that the other PRs don't get polluted by non-relevant changes, making them a nightmare to review.

.eslintignore
```diff
@@ -2,12 +2,8 @@ node_modules
 scripts
 postcss.config.js
 webpack.*
-src/data/generated.ts
-src/lib/colonyNetwork
-src/lib/colonyServer
-src/lib/mock-oracle
-src/lib/subgraph
-src/lib/graph-node
```
willm30, line 10: These are all from dapp

.huskyrc
```diff
@@ -1,6 +1,6 @@
 {
-    "hooks": {
-        "pre-commit": "lint-staged",
-        "pre-push": "npm run typecheck && npm run lint && npm run stylelint"
```
willm30, line 4: There is no styelint command in the package.json, and there's no need to run the linter both before commit and push (right?).

.lintstagedrc
```diff
@@ -1,8 +1,11 @@
 {
   "src/**/*.{ts,tsx}": [
-    "eslint --fix"
+    "eslint --fix",
+    "prettier --write --ignore-unknown"
```
willm30, line 4: This ensures prettier runs after eslint, per the instructions:

> Note: If you use ESLint, make sure lint-staged runs it before Prettier, not after.

.prettierignore
willm30, line 1: Files to ignore

package.json
```diff
@@ -12,7 +12,6 @@
     "docker:build": "npm run docker:build:base && npm run docker:build:network && npm run docker:build:monitor && npm run docker:build:ingestor && npm run docker:build:amplify",
     "docker:compose": "docker compose --file docker/colony-cdapp-dev-env-orchestration.yaml",
     "dev": "npm run docker:build && npm run docker:compose up -- --force-recreate -V",
-    "lint": "eslint --fix 'src/**/*.{ts,tsx}'",
```
willm30, line 15: No need for own command, right? (I'm assuming no one uses this). I propose we use staged-lint as the single source of truth for linting commands.

package.json
```diff
@@ -82,7 +81,6 @@
     "eslint-plugin-jest": "^27.0.1",
     "eslint-plugin-jsdoc": "^39.3.6",
     "eslint-plugin-jsx-a11y": "^6.6.1",
-    "eslint-plugin-prettier": "^4.2.1",
```
willm30, line 85: Per Pretter this is:

> generally not recommended

And I don't think we fall under [the exceptions](https://prettier.io/docs/en/integrating-with-linters.html)

package.json
```diff
@@ -110,6 +108,7 @@
     "style-loader": "^3.3.1",
     "stylelint": "^13.3.3",
     "stylelint-config-css-modules": "^2.2.0",
+    "stylelint-config-prettier": "^9.0.5",
```
willm30, line 111: Ensures stylelint plays nicely with prettier

willm30:
@jakubcolony @rdig Please sense-check the proposed changes and re-review when you can.

.eslintrc.js
```diff
@@ -16,6 +16,7 @@ module.exports = {
     'eslint-config-airbnb/rules/react-a11y',
     'plugin:@typescript-eslint/recommended',
     '@colony/eslint-config-colony',
+    'prettier',
```
willm30, line 19: https://github.com/prettier/eslint-config-prettier

package.json
jakubcolony: You may want to run linting for the entire repo (probably something we'd want to do especially when trying to clean up all the errors). `lint-staged` only runs the linting for staged changes. I'd leave it there

jakubcolony (approved):
Looks sensible, other than the deleted `lint` script that I see some usefulness in 👍

package.json
ArmandoGraterol: I use this command in the Dapp. I think it could be useful in the beginning to catch stuff that may not be worked on, and therefore, no staging changes. After we are sure that we are lint-error-free (as we are currently in the Dapp for example), we could remove it

ArmandoGraterol (approved):
LGTM

willm30:
@rdig This could to go (aside from restoring the lint command as others have requested)?

rdig (approved):
Good from me as well

PR #348, opened 2023-03-27
Connect Edit Colony Details dialog with saga

ArmandoGraterol:
Connected EditColonyDetails with the related saga. Also updated sidebar and dialog `ColonyAvatar` to prefer avatar over the thumbnail so that the actual uploaded image is shown.

### Dialog and sidebar after uploading an avatar and reloading:
FireShot Capture 824 - Colony CDapp - localhost

Contributes to #309

### DIFF
--- a/src/components/common/Dialogs/EditColonyDetailsDialog/helpers.ts
+++ b/src/components/common/Dialogs/EditColonyDetailsDialog/helpers.ts
@@ -2,33 +2,16 @@ import { Colony } from '~types';
 
 export const getEditColonyDetailsDialogPayload = (
   colony: Colony,
-  {
-    colonyAvatarImage,
-    colonyDisplayName: payloadDisplayName,
-    annotationMessage,
-  },
-) => {
-  const colonyTokens = colony.tokens?.items || [];
-
-  return {
-    colonyAddress: colony.colonyAddress,
-    colonyName: colony.name,
-    colonyDisplayName: payloadDisplayName,
-    colonyAvatarImage:
-      typeof colonyAvatarImage === 'string' || colonyAvatarImage === null
-        ? colonyAvatarImage
-        : colony.metadata?.thumbnail,
-    colonyAvatarHash: colony.metadata?.avatar,
-    hasAvatarChanged: !!(
-      typeof colonyAvatarImage === 'string' || colonyAvatarImage === null
-    ),
-    colonyTokens: colonyTokens.filter(
-      (colonyToken) =>
-        colonyToken?.token.tokenAddress !== colony.nativeToken.tokenAddress,
-    ),
-    // verifiedAddresses: colonyData?.processedColony?.whitelistedAddresses,
-    annotationMessage,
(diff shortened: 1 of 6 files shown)


src/components/common/Dialogs/EditColonyDetailsDialog/helpers.ts
```diff
@@ -2,33 +2,16 @@ import { Colony } from '~types';
...
-    colonyTokens: colonyTokens.filter(
-      (colonyToken) =>
-        colonyToken?.token.tokenAddress !== colony.nativeToken.tokenAddress,
-    ),
-    // verifiedAddresses: colonyData?.processedColony?.whitelistedAddresses,
-    annotationMessage,
-    // isWhitelistActivated:
-    //   colonyData?.processedColony?.isWhitelistActivated,
-  };
-};
```
rdig, line 34: Is this logic that was brought over from the Dapp ?

src/redux/sagas/utils/metadataChangelog.ts
```diff
@@ -46,8 +47,10 @@ export const getUpdatedColonyMetadataChangelog = (
       transactionHash,
       newDisplayName: newDisplayName ?? metadata.displayName,
       oldDisplayName: metadata.displayName,
-      // @TODO: Actually check whether the avatar has changed
-      hasAvatarChanged: false,
+      hasAvatarChanged:
+        newAvatarImage === undefined
+          ? false
+          : newAvatarImage !== metadata.avatar,
```
rdig, line 53: So the logic in this util method stores the old/new values in an log entry, the returns the log. And it's up to whatever other function or component to handle the comparison logic (that's how I understand it anyway, feel free to check with @jakubcolony to confirm)

So the way you did it, breaks this pattern as just storing the comparison directly rather then the old/new values.

rdig (changes requested):
I've rejected this just because I don't think the logic for `getUpdatedColonyMetadataChangelog` was implemented properly.

Everything else, I'm ok with

src/components/common/Dialogs/EditColonyDetailsDialog/helpers.ts
ArmandoGraterol: Yep, I've updated this to accommodate for the new params of the saga

src/redux/sagas/utils/metadataChangelog.ts
ArmandoGraterol: Well as far as I have checked, this changelog is storing the old and new values of the colony name but not of the avatar, only if the avatar changed by the time this changelog was emitted (`hasAvatarChanged`). I'll wait for @jakubcolony to come back and throw some light into this tho

jakubcolony (approved):
All good here, nice work!

rdig (commented):
Turns out I was wrong about the avatar change logic. So this is all good to go then

rdig (approved):
Turns out I was wrong about the avatar change logic. So this is all good to go then

PR #313, opened 2023-03-09
Add support for Version Upgrade action

jakubcolony:
## Description

This PR adds support for the Version Upgrade action.

## Testing

Create a colony using the wizard. Click on the Test Version Upgrade button. It should fail since your colony already has the newest version.

You will need to add a new colony version to the chain. Run `npm run truffle console` and then the following commands one by one:
```js
// local dev ether router address
n = await IColonyNetwork.at('0x5CC4a96B08e8C88f2c6FC5772496FeD9666e4D1F')
// current version resolver 
vr = await n.getColonyVersionResolver(11)
// get metacolony
mca = await n.getMetaColony()
mc = await IMetaColony.at(mca)
// add new version (12) 
mc.addNetworkColonyVersion(12, vr)
```

Then create a colony in the CDapp and click on the testing button again. You should be redirected to the details page:
image

Resolves #280

PR #349, opened 2023-03-28
Wire up Upgrade Colony dialog

ArmandoGraterol:
In order to test, the process should be the same as the one for #313 but now use the dialog to upgrade the colony instead of the action button.

### No new version available:
FireShot Capture 825 - Colony CDapp - localhost

### New version available:
FireShot Capture 826 - Colony CDapp - localhost

### No new version available after upgrading:
FireShot Capture 827 - Colony CDapp - localhost

Contributes to #309


rdig (changes requested):
As with #325 I've stopped reviewing after noticing this adds an extra lambda function which is not really needed.

I'll re-review once this gets addressed

amplify/team-provider-info.json
```diff
@@ -26,4 +26,4 @@
       }
     }
   }
-}
+}
```
ArmandoGraterol, line 29: I don't understand what's changing here :thinking:

Julianahlk: Looks like the newline at the end of the file was removed. In general, we try to avoid having no newline at the end of files.

Julianahlk (approved):
LGTM

amplify/team-provider-info.json
willm30: That's only for these automatically generated files though? Otherwise I'm pretty sure we add a newline.

rdig: I think showed up because you removed the previous changes, then again, I've also seen this act up from time to time.

I believe this will be fixed once we manage to integrate prettier _(or at least I hope)_

src/components/common/Dialogs/NetworkContractUpgradeDialog/ContractVersionSection/ContractVersionSection.tsx
```diff
@@ -18,12 +20,17 @@ const MSG = defineMessages({
...
+  colony: Colony;
   currentVersion: number;
+  colonyContractVersion: number;
 }
 
-const LegacyPermissionWarning = ({ currentVersion }: Props) => {
+const LegacyPermissionWarning = ({
+  colony,
+  currentVersion,
+  colonyContractVersion,
```
rdig, line 31: Why not fetching them directly in here via the hook ?

src/components/common/Dialogs/NetworkContractUpgradeDialog/NetworkContractUpgradeDialogForm.tsx
```diff
@@ -86,7 +89,11 @@ const NetworkContractUpgradeDialogForm = ({
         </DialogSection>
       )}
       <DialogSection>
-        <ContractVersionSection currentVersion={version} />
+        <ContractVersionSection
+          colony={colony}
+          currentVersion={version}
+          colonyContractVersion={colonyContractVersion}
```
rdig, line 95: Same question as above? 

Is it just because of the loading logic ?

rdig (approved):
Left some comments, but mostly all good

src/components/common/Dialogs/NetworkContractUpgradeDialog/NetworkContractUpgradeDialogForm.tsx
ArmandoGraterol: Yep, it's because of the loading logic for the spinner and submit button of this dialog. Otherwise, the submit button would be displayed as available even though the colony version hasn't loaded in.

PR #353, opened 2023-03-30
Wire up ColonyUpgrade component

### LINKED ISSUE #59: Port over Colony Upgrade
Port over the ColonyUpgrade component rendered by `ColonyHomeLayout`.

ArmandoGraterol:
`colony-js` had to be updated to import the latest network release tag from it as the previous import was no longer available.

**When the user can update to a newer version:**
FireShot Capture 829 - Colony CDapp - localhost

In order to test, the process should be the same as the one for https://github.com/JoinColony/colonyCDapp/pull/313 but now a warning should appear in the colony's home.

**When the colony version is below v4:**
FireShot Capture 828 - Colony CDapp - localhost

In order to test, I just changed `colonyContractVersion` in line 60 of `ColonyUpgrade` to `3`.

Resolves #59

### DIFF
--- a/src/components/common/ColonyHome/ColonyUpgrade/ColonyUpgrade.tsx
+++ b/src/components/common/ColonyHome/ColonyUpgrade/ColonyUpgrade.tsx
@@ -1,20 +1,25 @@
-import React, { useCallback } from 'react';
+import React from 'react';
 import { defineMessages, FormattedMessage } from 'react-intl';
 
 import { useDialog } from '~shared/Dialog';
-import NetworkContractUpgradeDialog from '~dialogs/NetworkContractUpgradeDialog';
+import { NetworkContractUpgradeDialog } from '~common/Dialogs';
 import Alert from '~shared/Alert';
 import Button from '~shared/Button';
 import ExternalLink from '~shared/ExternalLink';
-import { useNetworkContracts } from '~data/index';
-import { useTransformer, useAppContext, useColonyContext } from '~hooks';
-import { getNetworkRelaseLink } from '~utils/external';
 import {
-  colonyMustBeUpgraded,
-  colonyShouldBeUpgraded,
-} from '~modules/dashboard/checks';
-import { hasRoot } from '~modules/users/checks';
-import { getAllUserRoles } from '~modules/transformers';
+  useTransformer,
+  useAppContext,
+  useColonyContext,
+  useColonyContractVersion,
+  useEnabledExtensions,
+} from '~hooks';
+import { getNetworkReleaseLink } from '~utils/external';
+import {
+  hasRoot,
+  mustColonyBeUpgraded,
+  canColonyBeUpgraded,
+} from '~utils/checks';
+import { getAllUserRoles } from '~redux/transformers';
 
 import styles from './ColonyUpgrade.css';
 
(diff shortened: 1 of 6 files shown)


src/components/common/ColonyHome/ColonyUpgrade/ColonyUpgrade.tsx
```diff
@@ -34,18 +39,16 @@ const MSG = defineMessages({
...
-  const handleUpgradeColony = useCallback(
-    () =>
-      openUpgradeVersionDialog({
-        colony,
-      }),
-    [colony, openUpgradeVersionDialog],
-  );
+  const enabledExtensionData = useEnabledExtensions();
+  const handleUpgradeColony = () =>
+    colony &&
```
rdig, line 47: You don't need to check for `colony` object existence, the `useColonyContext` already does that. You won't get to this point if `colony` is not set

src/utils/checks/canColonyBeUpgraded.ts
```diff
@@ -1,4 +1,16 @@
 import { Colony } from '~types';
 
-export const canColonyBeUpgraded = (colony: Colony, contractVersion: number) =>
-  contractVersion > colony.version;
+const minimumRequiredColonyVersion = 4;
+
+export const canColonyBeUpgraded = (
+  colony: Colony | undefined,
```
rdig, line 6: Colony will / should always be set here

src/utils/checks/canColonyBeUpgraded.ts
```diff
@@ -1,4 +1,16 @@
...
+  contractVersion: number,
+) => colony && contractVersion > colony.version;
+
+export const mustColonyBeUpgraded = (
+  colony: Colony | undefined,
```
rdig, line 11: Colony will / should always be set here

rdig (approved):
Mostly good, however some types need to be adjusted

src/components/common/ColonyHome/ColonyUpgrade/ColonyUpgrade.tsx
ArmandoGraterol: I agree but `useColonyContext` is returning `colony` as `ColonyFragment | undefined` anyways, so if I don't add these checks then TS just marks this as an error.

As we've mentioned in other PRs, I think we will need to refactor `useColonyContext` to actually return the correct type, or in the worst-case scenario, recast it inside the hook before returning it.

Julianahlk (approved):
Working well for me. 👍 Approving but just had a thought about the dismiss button.

Regarding the dismiss button in the upgrade banner, since clicking dismiss only closes the banner temporarily and the banner returns after a refresh, why not either:
1. Save the state of the banner so it remains closed until the next colony version is released, the upgrade dialog can still be accessed through the actions button.
2. Do not even show a dismiss button if the intent is that the upgrade banner should always be visible until a user upgrades. Clicking dismiss only to have the banner reappear could feel frustrating to users.

ArmandoGraterol:
@Julianahlk 
> Regarding the dismiss button in the upgrade banner, since clicking dismiss only closes the banner temporarily and the banner returns after a refresh, why not either:
> 
> 1. Save the state of the banner so it remains closed until the next colony version is released, the upgrade dialog can still be accessed through the actions button.
> 2. Do not even show a dismiss button if the intent is that the upgrade banner should always be visible until a user upgrades. Clicking dismiss only to have the banner reappear could feel frustrating to users.

I like the second option. I'll pass these improvements to Product and create an issue if they agree

PR #364, opened 2023-04-04
Port staking widget validation (I)

willm30:
## Description

This PR ports one part of the staking validation, namely, the view when the user has activated fewer tokens than the minimum needed to stake. Related prs are #365 and #368. 

Together, these add a fair amount of bloat and complexity to the staking widget. The main challenge is keeping the custom error state in sync with the StakingControls, as the validation messages live inside the StakingSlider (no access to colony or staking widget context), and the controls outside. I expect that the UI redesign will present an opportunity to rethink how we do this since it's currently fairly complex and fragile.

Now, back to the pr at hand.

activate-min-tokens

## Testing

In truffle console run:

``` 
c = await IColony.at('<colony-address>')
c.emitDomainReputationReward(1, accounts[0], '1000000000000000000')
c.setArbitrationRole(1, "0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff", '<voting-extn-address>', 1, true)

```

Then click "Test Mint Tokens".

Since you have not activated any tokens, you should see the invitation to activate some.
 
**New stuff** ✨

* `StakingValidationMessage`, `TokenErrorMessage`, `MinimumStakeMessage` components
* `useEnoughTokensForStaking` hook to get user balance data

**Changes** 🏗

* Removed `canBeStaked` prop as it's not a very useful abstraction

Contributes to #270

PR #365, opened 2023-04-04
Port activate button (II)

willm30:
## Description

This PR ports the activate tokens button as well as the corresponding error message when a user hits a staking limit caused by insufficient active tokens.

limit

## Testing

Following on from #364, you can activate tokens in the UI via TokenActivationPopover. Activate fewer than the required stake, and then refresh the browser. 

You should see a limit on the staking slider and the activate button/error msg appear if you hit that limit.

**New stuff** ✨

* Activate Button
* Limit exceeded message

**Changes** 🏗

* `getStakeFromSlider` to use Decimal for increased precision

Contributes to #270

### DIFF
--- a/src/hooks/useStakingSlider.ts
+++ b/src/hooks/useStakingSlider.ts
@@ -1,7 +1,9 @@
+import { BigNumber } from 'ethers';
 import { useLocation } from 'react-router-dom';
 import { useEnoughTokensForStaking } from '~common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/StakingWidget/StakingInput/useEnoughTokensForStaking';
 
-import { useGetColonyActionQuery } from '~gql';
+import { useGetColonyActionQuery, useGetUserReputationQuery } from '~gql';
+import { calculateStakeLimitDecimal } from './helpers';
 import useAppContext from './useAppContext';
 
 const getTransactionHashFromPathName = (pathname: string) =>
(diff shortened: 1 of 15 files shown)


src/hooks/useStakingSlider.ts
```diff
@@ -53,6 +57,29 @@ const useStakingSlider = (isObjection: boolean) => {
...
+        walletAddress: user?.walletAddress ?? '',
+        domainId: Number(motionDomainId),
+        rootHash,
+      },
+    },
+  });
+
+  const userReputation = data?.getUserReputation;
+  /* User cannot stake more than their reputation in tokens. */
+  const userMaxStake = BigNumber.from(userReputation ?? '0');
```
willm30, line 73: Cannot see a simple way of reusing the logic to get `userMaxStake`. I would wrap it up into a hook but I wouldn't be able to access colony or staking widget context inside, and it would therefore require an awkward number of props as args. 

I think duplication is warranted here.

rdig: I mean, it repeats like 2 times only. So you really shouldn't worry about. It's a good compromise

rdig (approved):
Good to go

ArmandoGraterol (commented):
Codewise this is looking good to me! The error is appearing correctly. I did notice something that may or may not be related to this PR:
[12d80f71-76b0-46c2-8dc9-15eb07d320ee.webm](https://user-images.githubusercontent.com/18473896/231004309-d78ff8dc-7f5b-43fe-a302-bad7822ebd92.webm)

When I reach the limit and try to go back, the slider doesn't move

willm30:
@ArmandoGraterol Thanks. Created #390 to investigate. This pr good to go then?

ArmandoGraterol (approved):
Yep, good to go

PR #367, opened 2023-04-04
Wire up Motion Timer components

### LINKED ISSUE #300: Wire in the motion timer
The motion timer needs wiring in.

ArmandoGraterol:
## Description

- [] TODO write description

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

* Changes go here

**Deletions** ⚰️

* Deletions go here

## TODO

- [ ] Add TODOs

Resolves #300


willm30:
@ArmandoGraterol I have been merging today so this can be rebased when you get a chance

amplify/backend/function/fetchMotionTimeoutPeriods/src/event.json
```diff
@@ -0,0 +1,5 @@
+{
+  "key1": "value1",
+  "key2": "value2",
+  "key3": "value3"
```
willm30, line 4: This file can be deleted

src/components/common/ColonyActions/CountDownTimer/CountDownTimer.tsx
```diff
@@ -55,139 +38,30 @@ interface Props {
 
 const CountDownTimer = ({
   state,
```
willm30: Any chance we could have all references to the MotionState as "motionState" (instead of "state")? It would maintain consistency and I find it easier to read

willm30: Also state could refer to all sorts of things

src/components/common/ColonyActions/CountDownTimer/helpers.ts
```diff
@@ -0,0 +1,171 @@
...
+   * Count it down
+   */
+  useEffect(() => {
+    const timer = setInterval(() => {
+      setTimeLeft(timeLeft - 1);
+    }, 1000);
+
+    if (timeLeft === 0) {
+      dispatch({
+        type: ActionTypes.MOTION_STATE_UPDATE,
```
willm30, line 139: If you're not wiring in `updateState` saga, I don't think you need use to redux and this (and the action type) can be removed. 

However, you should refetch the motion state from the lamda we have `fetchMotionState` when the timer hits 0. You should be able to pass through the `refetchMotionState` function from `useGetColonyAction`

willm30 (commented):
Looking good! 

We have a lamda to fetch motion state so this will need to be used instead of the saga (and that can be removed entirely, along with the action types since I believe they were only being used to trigger that saga). Let me know once you've rebased and changed it to use the `refetchMotionState` function and I'll take another look 🙂

src/components/common/ColonyActions/CountDownTimer/helpers.ts
```diff
@@ -60,14 +64,14 @@ const useMotionTimeoutPeriods = (colonyAddress = '', motionId: number) => {
...
-  motionId: number,
-  isFullyNayStaked: boolean,
-) => {
+export const useMotionCountdown = (state: MotionState, motionId: number) => {
   const { colony } = useColonyContext();
   const { user } = useAppContext();
   const dispatch = useDispatch();
+  const {
+    motionStakes: { percentage: percentageStaked },
+  } = useStakingWidgetContext();
```
willm30, line 73: How are you accessing Staking Widget context from outside the staking widget?

willm30:
Also as an FYI we need to show the countdown timer in the actions list. That can come separately. I have opened #416 to track

rdig:
This needs to be rebased right? as I've seen code that I've reviewed before.

I've had a initial look at it and have some thoughts about caching the timer between blocks, but please rebase this first first so I can read the code better. 

Let me know after

src/components/common/ColonyActions/CountDownTimer/helpers.ts
ArmandoGraterol: I've moved the context 1 level above in order to get the percentage of the NAY side staked

willm30 (changes requested):
~~Currently not seeing the timer at all~~ Sorry, needed to run `npm i` before building the branch

willm30:
Something is up in the lamda, which is causing it to return undefined. 

lamda

src/hooks/useGetColonyAction.ts
```diff
@@ -1,12 +1,29 @@
...
+  useGetColonyActionQuery,
+  useGetMotionStateQuery,
+  Exact,
+  GetMotionStateInput,
+  GetMotionStateQuery,
+} from '~gql';
 import { Colony } from '~types';
 import { isTransactionFormat } from '~utils/web3';
 
+export type RefetchMotionState = (
```
willm30, line 17:
``` ts
type Refetch = ReturnType<typeof useGetMotionStateQuery>['refetch']
```

src/components/common/ColonyActions/CountDownTimer/helpers.ts
```diff
@@ -0,0 +1,197 @@
...
+   * for the user to let them know what's going on
+   *
+   * So we are "faking" this by adding an extra 5 to the alloted time, so that the
+   * blockchain has time to process any transactions, detect the motion's timeout,
+   * and change the state so we can refresh it.
+   */
+  useEffect(() => {
+    if (
+      !loadingMotionTimeoutPeriods &&
+      /*  @NOTE: If we have loaded the motion timeout periods, we should only set/reset the timer:
```
willm30, line 117: The comment says:

> If we have loaded the motion timeout periods

But the check is:

`!loadingMotionTimeoutPeriods`

i.e. not loaded the motion timeout periods

Or am I misunderstanding?

willm30: I got here because I was trying to debug the following:

Let one motion hit zero, and fast forward time so it actually updates the motion state.

Create a new motion. The timeLeft state is zero.

src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionCountdown/MotionCountdown.tsx
```diff
@@ -43,37 +44,40 @@ const displayName = 'common.ColonyActions.ActionDetailsPage.MotionCountdown';
...
 }
 
-const MotionCountdown = ({ motionState, motionData }: MotionCountdownProps) => {
+const MotionCountdown = ({
+  motionState,
+  motionData,
+  refetchMotionState,
+}: MotionCountdownProps) => {
   // const motionDomain = 1;
   const motionId = 1;
```
willm30, line 57: This needs to come from the motion data...

src/components/common/ColonyActions/CountDownTimer/helpers.ts
willm30: This is the wrong solution for me. The context is staking widget specific, so moving it to encapsulate additional components feels like an ad-hoc refactor (naming wise for sure doesn't make sense). 

You have the motion data available already in the MotionCountdown component. I would prefer it to be passed down to the function, and then if we notice there is a lot of prop drilling going on in the countdown timer components, we can create a separate context for `MotionCountdown`, or extract the motion data into a context that is available for all DefaultMotion subcomponents (which then gets fed to the staking widget context).

src/components/common/ColonyActions/CountDownTimer/helpers.ts
```diff
@@ -0,0 +1,197 @@
...
+  return {
+    motionTimeoutPeriods,
+    loadingMotionTimeoutPeriods: loading,
+    refetchMotionTimeoutPeriods: refetch,
+  };
+};
+
+export const useMotionCountdown = (
+  state: MotionState,
+  motionId: number,
```
willm30, line 67: This should be typed as a string to avoid overflow in the event it's a very large number.

willm30 (changes requested):
Nice work on this. I needed to run `npm i` before building the branch (a note to other reviewers), but the timer is working nicely. 

One tiny but really important change that needs to happen is to not hardcode the motionId to `1`, or else the timer won't work for any other motions.

I also think it'd be good to show zeroes instead of nothing when the timer hits zero:

zeros

There will inevitably be a short delay before the motion state gets updated so that would be a nice visual addition.

And I'm not noticing the staking period reset when I stake 100%, although it works just fine when objecting 100%. I'm expecting it to happen with staking also, given the message says:

> Motion is fully staked; Staking period has reset.

For the other phases (voting, reveal), it reset perfectly.

amplify/backend/function/fetchMotionTimeoutPeriods/src/index.js
willm30, line 1: I'm getting a lot of prettier messages in this file re spacing/indentation

src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionCountdown/MotionCountdown.tsx
ArmandoGraterol: Yup, my bad. I missed this part when I was updating the component

src/components/common/ColonyActions/CountDownTimer/helpers.ts
ArmandoGraterol: The way I see it, when we get to this point in the flow we should have already requested the motion timeout periods via `useMotionTimeoutPeriods` above this. So, if `!loadingMotionTimeoutPeriods` is `true` then this would mean that either the query finished fetching the data or it tried to and there was an error, both ways the motion timeout periods have stopped loading and we should have some values to initialize the timer. (Even if the query fails, as we have fallbacks)

Julianahlk (approved):
Looks good. Working well for me.

src/components/common/ColonyActions/CountDownTimer/helpers.ts
willm30: Ah yes thanks, I was reading that wrong

src/components/common/ColonyActions/CountDownTimer/helpers.ts
```diff
@@ -0,0 +1,188 @@
...
+    loadingMotionTimeoutPeriods: loading,
+    refetchMotionTimeoutPeriods: refetch,
+  };
+};
+
+export const useMotionCountdown = (
+  state: MotionState,
+  motionId: string,
+  refetchMotionState: RefetchMotionState,
+  motionStakes: MotionData['motionStakes'],
```
willm30, line 63: Codegen generates a `MotionStakes` type for us, I think that's generally the way we're typing nested objects (As opposed to accessing as a property of the top level type)

willm30 (approved):
Thanks for making those changes, this looks good to go 👍

PR #368, opened 2023-04-04
Port staking error messages (III)

willm30:
## Description

This PR ports the following staking validation error messages:

### Amount left to stake is less than min user stake 
Screenshot (724)

### User does not have enough reputation to stake 100% of the motion
staking-limit

### User does not have enough reputation to stake the min user stake
min-rep

### Testing

1. Amount left to stake is less than min user stake 

Stake a motion 99%. You should find yourself in that awkward position where the amount left to stake is less than the min user stake. I have copied the logic over exactly from the dapp, but updated the copy as the previous message said: 
"The motion can't be staked more than the minimum stake", which is obviously false.

2.  User does not have enough reputation to stake 100% of the motion

For this it'll be easiest to manually hardcode the `userMaxStake` values in `useStakingSlider` and `useStakingControls`. It needs to be more than the min user stake, but less than the required stake. `10000000000000000` worked for me.

(Note that we have userMaxStake in two places due to the StakingSlider handling its own data. Because I can't access colony context or staking widget provider context inside the objection dialog, sharing logic between these two is very awkward).

As you stake, you should see the limit decrease since you're get closer and closer to your max stake.

3. User does not have enough reputation to stake the min user stake

Reset the max user stake to what it should be. Create a motion with a user with reputation, like normal. Then switch to a wallet without rep and you should see the error message.

**New stuff**  ✨

* The staking error messages seen above

Contributes to #270

### DIFF
--- a/src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/StakingWidget/StakingInput/StakingControls/StakingControls.tsx
+++ b/src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/StakingWidget/StakingInput/StakingControls/StakingControls.tsx
@@ -23,6 +23,7 @@ const StakingControls = ({ limitExceeded }: StakingControlsProps) => {
     showActivateButton,
     enoughTokensToStakeMinimum,
     isLoadingData,
+    cantStakeMore,
   } = useStakingControls(limitExceeded);
 
   return (
@@ -31,6 +32,7 @@ const StakingControls = ({ limitExceeded }: StakingControlsProps) => {
       <StakeButton
         isLoadingData={isLoadingData}
         enoughTokensToStakeMinimum={enoughTokensToStakeMinimum}
+        cantStakeMore={cantStakeMore}
       />
       {!showBackButton && (
         <ObjectButton
--- a/src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/StakingWidget/StakingInput/StakingControls/StakingControls.tsx
+++ b/src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/StakingWidget/StakingInput/StakingControls/StakingControls.tsx
@@ -24,7 +24,7 @@ const StakingControls = ({ limitExceeded }: StakingControlsProps) => {
     enoughTokensToStakeMinimum,
     enoughReputationToStakeMinimum,
     isLoadingData,
-    cantStakeMore,
+    canStakeMore,
     userNeedsMoreReputation,
     userMaxStake,
     userActivatedTokens,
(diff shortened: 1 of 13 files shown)


src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/StakingWidget/StakingInput/StakingControls/StakingControls.tsx
```diff
@@ -26,11 +36,17 @@ const StakingControls = ({ limitExceeded }: StakingControlsProps) => {
       <StakeButton
         isLoadingData={isLoadingData}
         enoughTokensToStakeMinimum={enoughTokensToStakeMinimum}
+        enoughReputationToStakeMinimum={enoughReputationToStakeMinimum}
```
willm30, line 39: The error messages introduced here don't have a lot to do with the form value, therefore we can't use the validation schema to handle error state. 

Without significant reengineering of this widget, I think passing a bunch of props is the simplest way to correctly handle the stake button's disabled stake.

src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/StakingWidget/StakingInput/StakingSlider.tsx
```diff
@@ -58,14 +61,19 @@ const StakingSlider = ({
...
+      {displayErrMsg && (
         <StakingValidationMessage
+          enoughReputationToStakeMinimum={enoughReputationToStakeMinimum}
           enoughTokensToStakeMinimum={enoughTokensToStakeMinimum}
-          userActivatedTokens={userActivatedTokens}
           userMinStake={userMinStake}
+          userActivatedTokens={userActivatedTokens}
+          userMaxStake={userMaxStake}
+          remainingToStake={remainingToStake}
+          nativeTokenDecimals={nativeTokenDecimals}
```
willm30, line 76: Again, without a significant redesign, it seems we are required to duplicate a lot of logic inside the Staking Slider so that it and the controls stay in sync.  (staking slider cannot access colony context nor staking widget provider because it's used inside the objection dialog).

src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/StakingWidget/StakingInput/StakingControls/StakingControls.tsx
rdig: Don't worry about, most likely this'll get changed along with the new UX. It's good enough for now.

rdig (approved):
Good work. Overall nice job across all three PRs :100:

Julianahlk (approved):
Looks good! All the tested cases you mentioned are working for me.

One thing I was wondering that is probably a separate issue to this PR. if its possible to get into the situation where the amount left to stake is less than the min user stake, couldn't we warn users that this will happen if they stake 99% for example? Of course it's an edge case but it feels like a bad experience for users to end up in this situation.

willm30:
@Julianahlk 
> if its possible to get into the situation where the amount left to stake is less than the min user stake, couldn't we warn users that this will happen if they stake 99% for example

Completely agree. I think it's best to put this on hold until we're working on the new UX, but either the ux can be improved here or we could just remove this problem entirely by ensuring the user doesn't "over stake" in the saga.

Julianahlk:
> Completely agree. I think it's best to put this on hold until we're working on the new UX, but either the ux can be improved here or we could just remove this problem entirely by ensuring the user doesn't "over stake" in the saga.

Either one sounds good to me! Would you be okay opening an issue to track this and fix it in the future?

willm30:
@Julianahlk Yep, when we start working on it 🙂

Issue #362, opened 2023-04-03
Wire up Staking Messages for Staking Widget


Julianahlk:
- [x] Add events field to MotionData
- [x] Save motion event values to events field in block ingestor for each staking message
- [x] Read events data from MotionData and display events in ActionDetailsPageFeed

Objections and voting are handled by separate issues.

PR #369, opened 2023-04-05
Wire up Staking Messages for Staking Widget

Julianahlk:
## Description

This PR wires up the motion created message and the staking messages when a user backs a motion or objection.

Screen Shot 2023-04-20 at 9 02 47 PM

### Testing

Follow the testing steps here, [port/270-update-ui-after-staking](https://github.com/JoinColony/colonyCDapp/pull/338)

Then stake and object the motion and the relevant messages should be displayed.

**New stuff** ✨

Motion created message
Staking message

(Resolves | Contributes to) #362


src/components/common/ColonyActions/ActionDetailsPage/ActionDetailsPageFeed/ActionDetailsPageFeed.tsx
```diff
@@ -10,21 +9,18 @@ const displayName =
...
     <>
-      {events?.map((event) => (
+      {motionEvents?.map((event) => (
         <ActionsPageEvent
           actionData={actionData}
-          eventName={event}
-          key={event}
+          motionEventData={event}
+          eventName={ColonyAndExtensionsEvents[event.name]}
+          key={`${event.name}-${event.initiatorAddress}-${event.amount}`}
```
Julianahlk, line 23: This key is temporary until a better key can be created when more of the messages are implemented.

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -460,6 +460,13 @@ type StakerRewards {
   isClaimed: Boolean!
 }
 
+type MotionEvent {
+  name: String!
+  initiatorAddress: String
```
rdig, line 465: Initiator address will always be set

src/components/common/ColonyActions/ActionDetailsPage/ActionDetailsPageFeed/ActionDetailsPageFeed.tsx
rdig: This is a really bad key, and you won't be able to solve with more values, and it will only get worse with time as we surface more and more events in that feed.

The only way to properly do this, is to also store the `transactionHash` of the transaction that emmited this event and the `logIndex` of the specific event.

Then do `${transactionHash}${logIndex}` which will give you a truly unique key.

rdig (changes requested):
I'm good with this, I only rejected this since I want you to refactor the way you store `MotionEvent`s and store the transaction hash and log index, so that you can properly create a unique id when listing the motion events in the feed

willm30 (changes requested):
@Julianahlk Nice start on this! 

We are missing some of the messages though, so they should be added in. Specifically:

1: The "Raised an objection" message the first time an objection is raised
2: The staking 100% and the objecting 100% messages

If an objection is fully staked first, this is what we should see:
 
obj100stake100

And if the motion is fully staked first, then fully objected:

stake100obj100

Naturally, the final voting message can come later.

Also, we're currently missing the user's reputation in the messages, but that can come in a separate issue if you'd prefer (in which case, please open in the motions project).

src/components/common/ColonyActions/ActionDetailsPage/ActionDetailsPageFeed/ActionDetailsPageFeed.tsx
willm30: Seconding my comments in the block ingestor, we can use the array index here. It would certainly be simpler and is not going to affect how [react renders the list](https://react.dev/learn/rendering-lists) because the list ordering does not change.

willm30:
@Julianahlk The block ingestor hash needs to be updated

jakubcolony:
@willm30 Could that be why the motion shows as failed straightaway for me? Or did I mess something up?
image

src/components/common/ColonyActions/ActionDetailsPage/ActionDetailsPageFeed/ActionDetailsPageEvent/ActionDetailsPageEvent.tsx
```diff
@@ -10,19 +15,25 @@ const displayName =
   'common.ColonyActions.ActionDetailsPage.ActionDetailsPageEvent';
 
 export interface ActionsPageEventProps {
-  eventName: ColonyAndExtensionsEvents;
+  eventName: ColonyAndExtensionsEvents | SystemMessagesName;
   actionData: ColonyAction;
+  motionMessageData: MotionMessage;
```
willm30, line 20: This is typed incorrectly. It will be a string if it's an action (not a motion)

src/components/common/ColonyActions/ActionDetailsPage/ActionDetailsPageFeed/ActionDetailsPageFeed.tsx
```diff
@@ -10,21 +14,23 @@ const displayName =
...
+    : ACTIONS_EVENTS[actionData.type];
   return (
     <>
-      {events?.map((event) => (
+      {messages?.map((message) => (
         <ActionsPageEvent
           actionData={actionData}
-          eventName={event}
-          key={event}
+          motionMessageData={message}
```
willm30, line 28: This prop doesn't need to be present for actions, which is one of the reasons this doesn't feel like an appriopriate abstraction for the motions use case.

src/components/common/ColonyActions/ActionDetailsPage/ActionDetailsPageFeed/ActionDetailsPageEvent/ActionEventData.tsx
```diff
@@ -14,20 +15,41 @@ const displayName = 'common.ColonyActions.ActionDetailsPage.ActionEventData';
...
+            actionData,
+            motionMessageData,
+            motionMessageInitiatorUser,
+            colony,
+          )}
         />
       </div>
       <div className={styles.details}>
-        <ActionRoles actionType={type} eventName={eventName} />
+        <ActionRoles
```
willm30, line 49: There's also no need for `ActionRoles` to be present when displaying motion messages, another reason I think we should compose `ActionEventData` and `MotionEventData` via a base `EventData` component that exposes "details" and "text" slots. 

(We could probably improve the name as it's not that clear that "Event Data" represents the list item / message a user will see)

src/components/common/ColonyActions/helpers/getEventTitleValues.ts
```diff
@@ -133,6 +138,17 @@ const EVENT_TYPE_MESSAGE_KEYS_MAP: {
     EventTitleMessageKeys.ReputationChangeNumeral,
     EventTitleMessageKeys.Recipient,
   ],
+  [SystemMessagesName.ObjectionFullyStaked]: [
```
willm30, line 141: These could live in their own map, `MOTION_MESSAGE_KEYS_MAP`, or similar

src/components/common/ColonyActions/helpers/mapItemToMessageFormat.tsx
```diff
@@ -98,8 +102,10 @@ export const mapColonyActionToExpectedFormat = (
 };
 
 export const mapColonyEventToExpectedFormat = (
-  eventName: ColonyAndExtensionsEvents,
+  eventName: ColonyAndExtensionsEvents | SystemMessagesName,
```
willm30, line 105: And we could create a motion-message specific version of this mapping also

willm30 (changes requested):
Good job on getting all the messages into the feed. That's all looking as expected 👍

Requesting changes specifically because this current iteration will break actions:

actions

But really, we should create a motions-specific set of message feed components and reuse as much from the actions set as we can (via styles and component composition). That way we have two clear abstractions, one for actions and one for motions, instead of one that isn't quite right for either. I have left a few notes but I imagine it'd be clearer to have a call to discuss.

src/components/common/ColonyActions/ActionDetailsPage/ActionDetailsPageFeed/ActionDetailsPageEvent/ActionEventData.tsx
```diff
@@ -14,20 +15,41 @@ const displayName = 'common.ColonyActions.ActionDetailsPage.ActionEventData';
...
   eventName,
-}: Pick<ActionsPageEventProps, 'actionData' | 'eventName'>) => {
+}: Pick<
+  ActionsPageEventProps,
+  'actionData' | 'eventName' | 'motionMessageData'
+>) => {
   const { colony } = useColonyContext();
+  const { user: motionMessageInitiatorUser } = useUserByNameOrAddress(
+    motionMessageData.initiatorAddress || '',
+  );
```
jakubcolony, line 27: We should really get rid of this hook, given there are actually no use cases for it. In majority of cases we just want to get user by their address, and the `UserProfile` component just wrongly seems to be allowing either address or name (in the Dapp it's only the name).
Here, we know we want to get user by its address, so there's no point in running a redundant username query. Please consider just using the `getUserByAddress` query directly

src/components/common/ColonyActions/ActionDetailsPage/ActionDetailsPageFeed/ActionDetailsPageEvent/ActionEventData.tsx
```diff
@@ -14,20 +15,41 @@ const displayName = 'common.ColonyActions.ActionDetailsPage.ActionEventData';
...
+  const messageId = Object.values<string>(ColonyAndExtensionsEvents).includes(
+    eventName,
+  )
+    ? 'event.title'
+    : 'systemMessage.title';
```
jakubcolony, line 32: Could this use `in` operator, as in `eventName in ColonyAndExtensionsEvents ? ... : ...`?

src/components/common/ColonyActions/ActionDetailsPage/ActionDetailsPageFeed/ActionDetailsPageEvent/ActionEventData.tsx
```diff
@@ -14,20 +15,41 @@ const displayName = 'common.ColonyActions.ActionDetailsPage.ActionEventData';
 const ActionEventData = ({
   actionData: { createdAt, transactionHash, type },
   actionData,
+  motionMessageData,
   eventName,
-}: Pick<ActionsPageEventProps, 'actionData' | 'eventName'>) => {
+}: Pick<
+  ActionsPageEventProps,
+  'actionData' | 'eventName' | 'motionMessageData'
+>) => {
```
jakubcolony, line 23: I feel like that should be extracted into its own line

jakubcolony (changes requested):
Looking good, but I'd like you to change the way we fetch user data for motion message initiator:

At the very least, we should get rid of the `useUserByNameOrAddress`, as it redundantly makes 2 queries when only one is needed. I put more details in a relevant comment.

Alternative approach I'm thinking about is making `MotionMessage` an amplify `@model`, which would allow to introduce a relation to the User model. That way, we could simply get the user data from the message without separate queries (similarly to `ColonyAction` model):
```gql
type MotionMessage @model {
    initiatorAddress: ID!
    initiatorUser: User @hasOne(fields: ["initiatorAddress"])
}

# in your query:
messages {
  initiatorUser {
    ...User
  }
}
```

While this would require some extra boilerplate in ingestor as we'd need to make 2 mutations instead of one, I think it's worthwhile considering since you only mutate once, but read the data multiple times (and this approach saves you an extra query each time).

@willm30 Can probably weigh in on this one

willm30:
@jakubcolony @Julianahlk I think that could work nicely, although I'd consider using the `getUserByAddress` query for now and then coming back to this in a separate issue as I see it as an optimisation. #423

src/components/common/ColonyActions/ActionDetailsPage/ActionDetailsPageFeed/ActionDetailsPageEvent/ActionEventData.tsx
rdig: This is actually correct.

In the dapp we could only fetch a user from a name due to limitations, but fetching it from an address is actually something I had introduced (work actually done by Olga) when we moved over to the CDapp.

I think it should still stand, unless you have a proper argument against it.

rdig (changes requested):
I'm going to do another review once you address all the issues the others pointed out.

willm30:
@jakubcolony If the motion fails straight away, it's probably because the staking duration is too low when you install voting rep extn. Try 720 hours instead of 72 (time in dev runs fast 😄)

arrenv:
> @jakubcolony If the motion fails straight away, it's probably because the staking duration is too low when you install voting rep extn. Try 720 hours instead of 72 (time in dev runs fast 😄)

You can also just disable the reputation monitor `http://127.0.0.1:3001/reputation/monitor/toggle` then turn it back on. I believe a cycle a progressed with every transaction when the monitor is active.

src/components/common/ColonyActions/ActionDetailsPage/ActionDetailsPageFeed/ActionDetailsPageFeed.tsx
Julianahlk: The key is now created in the block ingestor as messageKey

Julianahlk:
Created #449 to handle the user's reputation in the messages

src/components/common/ColonyActions/helpers/getEventTitleValues.ts
Julianahlk: @willm30 just to be clear you are thinking we should split EVENT_TYPE_MESSAGE_KEYS_MAP into separate maps for Actions and Motions? and the System messages will live in the motions map?

amplify/backend/api/colonycdapp/schema.graphql
Julianahlk: @rdig there are actually some messages that don't have an initiator/staker/agent etc. like MotionFinalized, ObjectionFullyStaked, MotionFullyStakedAfterObjection, MotionFullyStaked, MotionVotingPhase.
In these cases there is no value for initiatorAddress and also it would be not be needed.

src/components/common/ColonyActions/ActionDetailsPage/ActionDetailsPageFeed/ActionDetailsPageEvent/ActionEventData.tsx
```diff
@@ -1,39 +1,47 @@
...
+          values={getActionEventTitleValues(eventName, actionData, colony)}
         />
-      </div>
-    </div>
+      }
+      details={
+        <>
+          <ActionRoles
+            actionType={type}
+            eventName={eventName as ColonyAndExtensionsEvents}
```
willm30, line 36: Don't think this needs casting

src/components/common/ColonyActions/ActionDetailsPage/ActionDetailsPageFeed/MotionDetailsPageEvent/MotionEventData.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+const MotionEventData = ({
+  actionData: { createdAt, transactionHash },
+  actionData,
+  motionMessageData,
+  eventName,
+}: MotionEventDataProps) => {
+  const { user: motionMessageInitiatorUser } = useUserByNameOrAddress(
+    motionMessageData.initiatorAddress || '',
+  );
```
willm30, line 29: If you turn `getMotionEventTitleValues` into a hook, you can move this inside it and avoid passing an extra arg

willm30: I do think it makes sense to use the `getUserByAddress` query here as well, as Jakub mentioned, because we know we're getting an address.

src/i18n/en-system-messages.ts
```diff
@@ -1,21 +1,21 @@
...
+      ${SystemMessages.EnoughExitRecoveryApprovals} {Enough permission holders have now signed to exit recovery mode. As long as no further storage slots are updated, a recovery permission holder may now sign a transaction to reactivate the colony.}
+      ${SystemMessages.MotionHasPassed} {{motionTag} has {passedTag} and may be finalized.}
+      ${SystemMessages.MotionRevealPhase} {It's time to {revealTag} votes to the world!}
+      ${SystemMessages.MotionHasFailedNotFinalizable} {{motionTag} has failed.}
+      ${SystemMessages.MotionHasFailedFinalizable} {{motionTag} has {failedTag} and may be finalized.}
+      ${SystemMessages.MotionVotingPhase} {{votingTag} has started! Voting is secret 🤫, and weighted by Reputation.}
+      ${SystemMessages.MotionFullyStaked} {{motionTag} is fully staked; Staking period has reset. As long as there is no {objectionTag} , the motion will pass.}
+      ${SystemMessages.MotionFullyStakedAfterObjection} {{motionTag} is fully staked.}
+      ${SystemMessages.ObjectionFullyStaked} {{objectionTag} is fully staked; Staking period has reset. If the {motionTag} gets fully staked, a vote will start immediately; if not, the {motionTag} will fail.}
```
willm30, line 15: This one shows up even if the motion is already fully staked in favour, whereas in the dapp it doesn't show in this case, since it goes straight to voting and shows that message.

src/components/common/ColonyActions/helpers/mapItemToMessageFormat.tsx
```diff
@@ -111,17 +115,6 @@ export const mapColonyEventToExpectedFormat = (
...
-          value={actionData.amount ?? 0}
-          decimals={actionData.token?.decimals ?? undefined}
-          suffix={actionData.token?.symbol ?? ''}
-        />
-      </AmountTag>
-    ),
-    // backedSideTag:
-    //   event.vote === MotionVote.Yay ? <MotionTag /> : <ObjectionTag />,
     motionTag: <MotionTag />,
     objectionTag: <ObjectionTag />,
```
willm30, line 119: These two aren't needed here either anymore

src/components/common/ColonyActions/helpers/mapItemToMessageFormat.tsx
```diff
@@ -174,6 +162,43 @@ export const mapColonyEventToExpectedFormat = (
...
+  initiatorUser?: User | null,
+  motionMessageData?: MotionMessage,
+) => {
+  return {
+    eventNameDecorated: <b>{motionMessageData?.name}</b>,
+    amountTag: (
+      <AmountTag>
+        <Numeral
+          value={motionMessageData?.amount ?? 0}
+          decimals={actionData.token?.decimals ?? undefined}
```
willm30, line 176: The token might not be present on the actionData, as it's an optional field in the schema and will only get included if it's a token-related motion. 

However, it would be the same value as the token you can get from colony context, because we just take it from the colony client in the block ingestor. So here, to ensure you also have the decimals/symbol, you can just take it off the colony object, and you don't need to pass down `actionData`.

src/components/common/ColonyActions/helpers/getEventTitleValues.ts
```diff
@@ -171,4 +182,20 @@ const getEventTitleValues = (
...
 };
 
-export default getEventTitleValues;
+/* Returns the correct message values for Motions according to the event type. */
+export const getMotionEventTitleValues = (
+  eventName: ColonyAndExtensionsEvents | SystemMessages,
+  actionData: ColonyAction,
+  motionMessageData?: MotionMessage,
+  motionMessageInitiatorUser?: User | null,
+) => {
```
willm30, line 191: See my other comments but I think we can get away with just passing `eventName` and `motionMessageData` into this function

src/components/common/ColonyActions/helpers/getEventTitleValues.ts
```diff
@@ -171,4 +182,20 @@ const getEventTitleValues = (
   });
 };
 
-export default getEventTitleValues;
+/* Returns the correct message values for Motions according to the event type. */
+export const getMotionEventTitleValues = (
+  eventName: ColonyAndExtensionsEvents | SystemMessages,
+  actionData: ColonyAction,
+  motionMessageData?: MotionMessage,
```
willm30, line 189: This doesn't need to be optional

src/components/common/ColonyActions/helpers/getEventTitleValues.ts
```diff
@@ -154,13 +165,13 @@ export const generateMessageValues = (
     initialEntry,
   );
 
-/* Returns the correct message values according to the event type. */
-const getEventTitleValues = (
-  eventName: ColonyAndExtensionsEvents,
+/* Returns the correct message values for Actions according to the event type. */
+export const getActionEventTitleValues = (
+  eventName: ColonyAndExtensionsEvents | SystemMessages,
```
willm30, line 170: This doesn't need the `| SystemMessages` now?

willm30 (approved):
Thanks for making the suggested changes! 👍

src/components/common/ColonyActions/helpers/mapItemToMessageFormat.tsx
Julianahlk: Got it, Im assuming you are referring to the nativeToken

src/i18n/en-system-messages.ts
Julianahlk: Made a change in the block ingestor to only store the ObjectionFullyStaked message if the Nay vote is greater than the Yay vote when the block ingestor receives a fully staked nay vote.

src/hooks/useUserByAddress.ts
willm30, line 1: I've had a similar conversation previously with @jakubcolony and we concluded we don't need to wrap every query in a hook, even if it results in slightly fewer lines of code in the consuming component. 

Main reason being it's a pretty pointless abstraction, in the sense that all we're doing is unwrapping the data on line 11. 

So I would remove this and just call the `useGetUserByAddressQuery`.

src/hooks/useUserByAddress.ts
```diff
@@ -0,0 +1,19 @@
+import { useGetUserByAddressQuery } from '~gql';
+
+const useUserByAddress = (address: string) => {
+  const { data, error, loading } = useGetUserByAddressQuery({
+    variables: {
+      address,
+    },
+    fetchPolicy: 'cache-and-network',
```
willm30, line 8: Don't think we need to specify this here, either? The default of cache-first should be fine for the user.

src/hooks/useUserByAddress.ts
Julianahlk: Ok sounds good! Made the changes. Thanks for all the suggestions and help.

Julianahlk:
@jakubcolony @rdig All comments have been resolved and the PR is ready for re-review whenever you can. Thanks.

PR #379, opened 2023-04-06
Wire up Create/Edit Domain motion dialogs

### LINKED ISSUE #375: Wire up motion saga with Create/Edit domain dialogs

ArmandoGraterol:
Connected Create/Edit Domain dialogs with motion sagas. We should now be able to create motions to manage domains in the colony. I haven't touched the motion page as the issue scope is only connecting the dialogs, but I believe all the data for the motion page should be available already with this PR.

## Testing
In order to test, you are going to need reputation in root or/and a reputation in the domain that you want to edit. Also, the voting reputation extension is going to need the `Architecture` role in order to create the motion:
```
c.setArchitectureRole(1, "0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff", '<voting-reputation-id>', 1, true)
```

As for generating reputation in root and in a domain, you can do the following:
```
c = await IColony.at('<your-colony-address>')
c.emitDomainReputationReward(<id-of-the-target-domain>, accounts[0], '1000000000000000000')
```

If the team/colony doesn't have rep and you can only force the action, an error should be displayed in the dialog stating that.

## Help wanted!

Now, in this PR there's **1 know issue** that I haven't been able to squash yet and I was hoping you guys could chime in:

- The colony/team reputation is not being loaded fast enough when you first enter the dialog:
[d2100745-8843-4af3-a905-038a84214a3b.webm](https://user-images.githubusercontent.com/18473896/232564636-4a32ad49-3fe1-48c0-a9de-a829771ed241.webm)

What do you think could be a good alternative to solve this? Add a loading state?

Resolves #375


willm30:
> There's a warning being shown everytime you interact with the force toggle for the first time after a refresh:

I think the following is responsible:

``` js
const forceActionValue = watch('forceAction');
if (forceActionValue !== isForce) {
   setIsForce(forceActionValue);
}
```

I think the simplest solution is probably to move the logic into the child so you can wrap it in a `useEffect`.

ArmandoGraterol:
> @ArmandoGraterol
> 
> > There's a warning being shown everytime you interact with the force toggle for the first time after a refresh:
> 
> I think the following is responsible:
> 
> ```js
> const forceActionValue = watch('forceAction');
> if (forceActionValue !== isForce) {
>    setIsForce(forceActionValue);
> }
> ```
> 
> I think the simplest solution is probably to move the logic into the child so you can wrap it in a `useEffect`.

Nice one! Thank you, I was having problems finding the source

src/components/common/Dialogs/CreateDomainDialog/CreateDomainDialog.tsx
```diff
@@ -74,26 +75,22 @@ const CreateDomainDialog = ({
...
-          }
-          return (
-            <CreateDomainDialogForm
-              back={prevStep && callStep ? () => callStep(prevStep) : undefined}
-              colony={colony}
-              enabledExtensionData={enabledExtensionData}
-            />
-          );
-        }}
+        {() => (
```
willm30, line 85: You can just ditch the function and pass the component directly as the child

src/components/common/Dialogs/CreateDomainDialog/CreateDomainDialog.tsx
```diff
@@ -74,26 +75,22 @@ const CreateDomainDialog = ({
...
+          <CreateDomainDialogForm
+            back={prevStep && callStep ? () => callStep(prevStep) : undefined}
+            colony={colony}
```
willm30, line 88: We don't really need to pass colony down various level, it can be accessed via context where its consumed

willm30: E.g inside `useActionDialogStatus` and `DialogHeading`

src/components/common/Dialogs/EditDomainDialog/EditDomainDialog.tsx
```diff
@@ -103,20 +103,16 @@ const EditDomainDialog = ({
...
-            <EditDomainDialogForm
-              back={prevStep && callStep ? () => callStep(prevStep) : undefined}
-              colony={colony}
-              domainOptions={domainOptions}
-              enabledExtensionData={enabledExtensionData}
-            />
-          );
-        }}
+        {() => (
+          <EditDomainDialogForm
```
willm30, line 107: Same re func

src/components/common/Dialogs/Messages/NotEnoughReputation/NotEnoughReputation.css
```diff
@@ -0,0 +1,18 @@
+.text {
+  font-size: var(--size-small);
+  font-weight: var(--weight-bold);
+  color: var(--temp-grey-blue-7);
```
willm30, line 4: temp grey blue does not exist in cdapp

src/components/shared/Dialog/DialogHeading/MotionDomainSelect/MotionDomainSelect.tsx
```diff
@@ -0,0 +1,81 @@
...
+    label: string,
+  ) => {
+    /*
+     * @NOTE This is so that the active item is displayed as `Root/Current Domain`
+     * when a subdomain is selected
+     */
+    let displayLabel =
+      (option?.value || Id.RootDomain) === Id.RootDomain
+        ? `${formatMessage(dropdownLabel || MSG.createDomain)} ${label}`
+        : `${formatMessage(dropdownLabel || MSG.createDomain)} ${formatMessage({
```
willm30, line 50: Tiny note, you can also use `formatText` utility which saves you having to call `useIntl` above

src/redux/sagas/motions/createEditDomainMotion.ts
```diff
@@ -180,23 +193,67 @@ function* createEditDomainMotion({
...
+    //   yield takeFrom(annotateMotion.channel, ActionTypes.TRANSACTION_SUCCEEDED);
+    // }
     yield put<AllActions>({
       type: ActionTypes.MOTION_DOMAIN_CREATE_EDIT_SUCCESS,
       meta,
     });
 
     if (colonyName) {
-      yield routeRedirect(`/colony/${colonyName}/tx/${txHash}`, history);
+      yield routeRedirect(`/colony/${colonyName}/tx/${txHash}`, navigate);
```
willm30, line 256: This should follow the example of the rootMotion so that we get auto-redirected to the motion page instead of needing a browser refresh.

willm30 (changes requested):
Nice one, I didn't find the reputation issue you highlight in the pr to be much of a problem, i.e. the state changed pretty quickly. An alternative would be to disable it by default, but this wouldn't solve the fundamental problem of the state not being available at the time the component is mounted. In any case, I think any improvements in this regard could be discussed / handled in a separate issue, as there's potentially a bigger discussion surrounding the data flow in the UAC (e.g. fetching when a user hovers the item in the previous menu). I'm not a huge fan of adding another spinner, personally.

In other notes, there are a few things that need updating before we merge.

I have two subdomains, A and B. In the dapp, the motion domain selector is filtered by which team I've selected to edit, which makes sense to me, because I shouldn't be able to edit domain B from domain A. I think we should add this here. 

Currently, no filtering happens, so if you try editing B from A you get an error: 
 
err

domains

The details widget on the motions page should also display the full set of information for the motions. From the dapp: 

new-team-details-widget

Whereas it isn't currently on this branch. 

cdapp

Please cherry-pick [this commit](https://github.com/JoinColony/colonyCDapp/pull/429/commits/a7147924934fb2125d4624616abedfd63fe9201a) before doing so, as I've already updated the logic for the motion detail item in another branch.

And please update the redirect logic in the saga to follow the example in rootMotion so we get a redirect instead of seeing "unknown" and then needing to refresh.

src/components/common/Dialogs/CreateDomainDialog/CreateDomainDialog.tsx
ArmandoGraterol: These components are inside a `Dialog`, so they shouldn't be able to access the colony context

src/hooks/useColonyHasReputation.ts
```diff
@@ -0,0 +1,22 @@
...
+    variables: {
+      input: {
+        walletAddress: ADDRESS_ZERO,
+        colonyAddress,
+        domainId: reputationDomain,
+      },
+    },
+  });
+
+  return loading || (!!toFinite(data?.getUserReputation) && !error);
```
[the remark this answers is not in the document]
ArmandoGraterol: Brain fart

src/components/common/Dialogs/EditDomainDialog/EditDomainDialogForm.tsx
```diff
@@ -83,21 +98,33 @@ const EditDomainDialogForm = ({
           selectedDomain.metadata?.name || `Domain #${selectedDomain.nativeId}`,
         domainPurpose: selectedDomain.metadata?.description || '',
         forceAction,
+        motionDomainId:
+          motionDomainId !== Id.RootDomain &&
+          motionDomainId !== selectedDomainValue
+            ? selectedDomain.nativeId
+            : motionDomainId,
```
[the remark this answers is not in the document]
ArmandoGraterol: Let me know if the comment makes it clearer

ArmandoGraterol:
> Nice one, I didn't find the reputation issue you highlight in the pr to be much of a problem, i.e. the state changed pretty quickly. An alternative would be to disable it by default, but this wouldn't solve the fundamental problem of the state not being available at the time the component is mounted. In any case, I think any improvements in this regard could be discussed / handled in a separate issue, as there's potentially a bigger discussion surrounding the data flow in the UAC (e.g. fetching when a user hovers the item in the previous menu). I'm not a huge fan of adding another spinner, personally.

Agreed :+1: 

> The details widget on the motions page should also display the full set of information for the motions. From the dapp:
> 
> new-team-details-widget
> 
> Whereas it isn't currently on this branch.
> 
> cdapp
> 
> Please cherry-pick [this commit](https://github.com/JoinColony/colonyCDapp/pull/429/commits/67fe29e05759d908eeadff729c9c3c08926fff89) before doing so, as I've already updated the logic for the motion detail item in another branch.

I thought any update to the motion page would be out of the scope of this issue, nevertheless, the changes were easy to include :)

rdig (approved):
After your update yesterday, this seems all good. Nice work

src/components/common/Dialogs/EditDomainDialog/EditDomainDialogForm.tsx
```diff
@@ -57,18 +57,33 @@ const requiredRoles: ColonyRole[] = [ColonyRole.Architecture];
 
 interface Props extends ActionDialogProps {
   domainOptions: SelectOption[];
+  handleIsForceChange: Dispatch<SetStateAction<boolean>>;
```
willm30, line 60: We also have a `SetStateFn` type in ~types

src/components/shared/Dialog/DialogHeading/MotionDomainSelect/MotionDomainSelect.tsx
```diff
@@ -0,0 +1,74 @@
...
+      (option?.value || Id.RootDomain) === Id.RootDomain
+        ? baseDropdownLabel
+        : `${baseDropdownLabel}/${label}`;
+    return <div className={styles.activeItem}>{displayLabel}</div>;
+  };
+
+  return (
+    <div className={styles.main}>
+      <DomainDropdown
+        colony={colony}
```
willm30, line 59: This can access colony via context? Then no need to pass through

ArmandoGraterol: Unfortunately no, as this component is being used inside the dialogs

Issue #326, opened 2023-03-17
Port VotingWidget


willm30:
~~Remember to update Finalize/ClaimWidgets with relevant voting logic from FinalizeMotionAndClaimWidget in dapp.~~ Handled in #382 

(For example, the motion state mapping, but may be more so review each component's dapp logic to be sure)

PR #380, opened 2023-04-06
Wire in voting widget

willm30:
**Note: new lamda adds ~1600 of boilerplate to pr**

You might as well test #383 first and then come back to do code review since it's based off here.

## Description

This PR wires in the voting widget.

votingWidget

## Testing
Follow setup from https://github.com/JoinColony/colonyCDapp/pull/338. Make sure you set the voting phase to something long like 720 hours when installing the voting rep extn.
Stake both sides of a motion 100%.
You'll see the voting widget.
If your user has 100% of the rep in the domain, when you vote you'll be moved forward to the reveal phase.
If your user has under the threshold for ending the voting phase (70% by default), then you'll be able to switch users and test on multiple accounts (if you do try this, I'd be keen to know if you also experience #381)

Testing that everything works as expected. 

**New stuff** ✨

* getVoterRewards lamda

**Changes** 🏗

* Wire in motion voting saga
* Wire in voting components

Contributes to #326


amplify/backend/function/fetchVoterRewards/src/index.js
willm30, line 1: Voter reward range depends on the logged in user (not the users who staked), therefore can't be handled by block ingestor

src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/VotingWidget/useVotingWidgetUpdate.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  const [prevRecord, setPrevRecord] = useState(currentVotingRecord);
+  const [hasUserVoted, setHasUserVoted] = useState(!!currentVotingRecord);
+
+  if (currentVotingRecord && !hasUserVoted) {
+    setHasUserVoted(true);
+  }
+
+  // if user's vote count increased, db has been updated, stop polling
+  if (currentVotingRecord?.voteCount !== prevRecord?.voteCount) {
```
willm30, line 23: There are two cases here. One where the user votes and that changes the motion state to reveal because the reputation threshold has been reached. In this case, the voting widget will be dismounted and the effect below will cancel polling.

The other is that the user votes and it doesn't change the motion state. In this case, the component won't be dismounted, so to cancel polling, we check to see if the user's vote count has increased.

src/hooks/useGetColonyAction.ts
```diff
@@ -43,16 +43,24 @@ const useGetColonyAction = (colony?: Colony | null) => {
...
+      input: {
+        colonyAddress: colony?.colonyAddress ?? '',
+        motionId: Number(action?.motionData?.motionId),
       },
-    });
+    },
+  });
+
+  /* Ensures motion state is kept in sync with motion data */
+  useEffect(() => {
```
willm30, line 61: This ensures a smooth transition to the voting phase after both sides have been staked 100%

src/redux/sagas/motions/voteMotion.ts
```diff
@@ -71,19 +67,7 @@ function* voteMotion({
       'voteMotionTransaction',
```
willm30, line 67: Why do we add this message to metamask? (expand code above: "Sign this message to generate 'salt' entropy.")

It's currently too long for the gas stn so either I adjust the css there or we remove.

amplify/backend/function/fetchVoterRewards/src/index.js
willm30, line 1: Lamda needed here because a voter's reward depends on the current logged in user, not just who staked

Julianahlk (approved):
Overall works well for me. 

Occasionally, after I stake 100%, the loading staking values spinner does't go away until I refresh but I can't find a way to reproduce this consistently and there are no error messages.

I do see the same issue you mentioned when switching users when I follow the same steps you described in the PR.

src/redux/sagas/motions/voteMotion.ts
willm30: @rdig

rdig: Really, we remove it? :)

And how would the user know what he's signing then ?

The message is needed as it acts like the string the user is signing to prove their identity when voting and revealing. 

The problem here is the gas station, as it should be able to support any message length. However, since in it's current form it's not really long for this world anymore, we can just fix it for this instance, via CSS

willm30: Thanks. Will fix in the gas stn.

amplify/backend/function/fetchVoterRewards/src/utils.js
```diff
@@ -0,0 +1,92 @@
...
+  });
+
+  return networkClient;
+};
+
+const getVotingClient = async (colonyAddress) => {
+  const networkClient = getNetworkClient();
+  const colonyClient = await networkClient.getColonyClient(colonyAddress);
+  return colonyClient.getExtensionClient(Extension.VotingReputation);
+};
```
ArmandoGraterol, line 36: Is there no way to reuse these functions across the lambda functions? I think I've seen those in other functions and I'm curious if it's possible to reuse them

amplify/backend/function/fetchVoterRewards/src/utils.js
```diff
@@ -0,0 +1,92 @@
...
+    body = await response.json();
+    return body;
+  } catch (error) {
+    /*
+     * Something went wrong... obviously
+     */
+    console.error(error);
+    return null;
+  }
+};
```
ArmandoGraterol, line 87: Same with this function

docker/files/amplify/amplify-meta.json.base
```diff
@@ -80,6 +80,11 @@
       "build": true,
       "providerPlugin": "awscloudformation",
       "service": "Lambda"
+    },
+      "fetchVoterRewards": {
```
ArmandoGraterol, line 84: This line has 1 more indented space than what's needed

ArmandoGraterol (approved):
Looking good! I've been testing it indirectly by working on the motion timer and I haven't found an issue with it

amplify/backend/function/fetchVoterRewards/src/utils.js
willm30: @ArmandoGraterol There is something called Lamda layers which is intended to address this problem, but we aren't currently using it

PR #395, opened 2023-04-11
Fix `isClaimed` and `hasUserVoted` state in motions

willm30:
## Description

This PR keeps the `hasUserVoted` and `isClaimed` state in sync with the user, so that it's accurate after a user switches account or logs out.

I've also thrown in a small presentational fix by emboldening the "Claim your tokens" detail in the ClaimWidget.

**Changes** 🏗

*  sync `hasUserVoted` and `isClaimed` state with the user
*  embolden the "Claim your tokens" detail

### DIFF
--- a/src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/ClaimMotionStakes/ClaimMotionStakes.tsx
+++ b/src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/ClaimMotionStakes/ClaimMotionStakes.tsx
@@ -4,6 +4,8 @@ import DetailItem from '~shared/DetailsWidget/DetailItem';
 import { MotionData } from '~types';
 import { useClaimWidgetConfig } from '~hooks';
 
+import styles from './ClaimMotionStakes.css';
+
 const displayName = `common.ColonyActions.ActionDetailsPage.DefaultMotion.ClaimMotionStakes`;
 
 interface ClaimMotionStakesProps {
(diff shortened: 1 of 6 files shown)


rdig (approved):
Nice fix. Good to go from my side

src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/VotingWidget/useVotingWidgetUpdate.ts
```diff
@@ -19,6 +19,14 @@ export const useVotingWidgetUpdate = (
     setHasUserVoted(true);
   }
 
+  useEffect(() => {
```
willm30, line 22: Ideally we'd express this as:

``` js
useEffect(() => {
setHasUserVoted(!!currentVotingRecord);
}, [currentVotingRecord]
```

However, it doesn't satisfy the end case where a user votes for the first time, and then signs out before the db has been updated. In that case, the `currentVotingRecord` will start as undefined, and remain undefined, which is why I'm capturing the `!user` logic explicitly.

src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/ClaimMotionStakes/ClaimMotionStakes.tsx
```diff
@@ -4,23 +4,32 @@ import DetailItem from '~shared/DetailsWidget/DetailItem';
...
 }
 
+export type ClaimMotionStakesStyles = typeof styles;
+
 const ClaimMotionStakes = ({
   motionData,
   startPollingAction,
 }: ClaimMotionStakesProps) => {
-  const config = useClaimWidgetConfig(motionData, startPollingAction);
+  const config = useClaimWidgetConfig(motionData, startPollingAction, styles);
```
jakubcolony, line 22: Correct me if I'm wrong, but do they have to go through the hook? Looks like there's no logic involved in deciding which ones to apply and the `claimLabel` class could as well be passed to `DetailItem` directly

willm30: I only want to put the style on the first item (the claim detail), whereas applying it to the DetailItem will apply it to all of them

PR #405, opened 2023-04-12
Extend Colony Edit action handler to support avatar and whitelist changes

### LINKED ISSUE #370: Extend Colony Edit action handler to support avatar and whitelist changes
I initially left it out but makes sense to pick it up now that there's an easy way of testing these.

jakubcolony:
## Description

This PR adds the missing bits for the Edit Colony action:
- [x] Whitelist changes
- [x] Avatar updates

## Testing
Create a colony and play around with the Manage teams > Manage address book and Advanced > Edit colony details. Verify that they behave as expected and the actions titles/messages are correct.

image
image

Resolves #370


src/utils/lodash.ts
```diff
@@ -5,3 +5,4 @@ export { default as sortBy } from 'lodash/sortBy';
 export { default as isEmpty } from 'lodash/isEmpty';
 export { default as toFinite } from 'lodash/toFinite';
 export { default as isEqual } from 'lodash/isEqual';
+export { default as xor } from 'lodash/xor';
```
jakubcolony, line 8: I got rid of the code using it in this PR, but it's relevant for #434

ArmandoGraterol:
Hey @jakubcolony, is this ready for review? It's out of draft but there's no reviewers assigned

jakubcolony:
@ArmandoGraterol Yep it is, haven't realised I didn't click the button!

jakubcolony:
@willm30 Thanks for the thorough review!

All the issues you highlighted should be fixed (other than the text in details widget, which will be done in #434).

willm30 (commented):
@jakubcolony I'm noticing the address book is only letting me add users that are in the db. It will complete the action for other addresses but they don't show up, even after refreshing the browser. In the dapp I can add anyone to the address book, which I'm presuming is the intended functionality?

As an aside, we should probably refetch the colony object once the action completes so the user doesn't need to refresh the browser to see changes. This isn't that noticeable currently since you have to refresh anyway to update the ActionDetailsPage, but once the polling is wired in, we'll notice it (I noticed it in motions, actually, and [plan to refetch after finalizing the motion](https://github.com/JoinColony/colonyCDapp/pull/485/commits/cf12adfe497f1d0f21ae85a5189f0d77034c1142)).

src/redux/sagas/actions/manageVerifiedRecipients.ts
```diff
@@ -0,0 +1,213 @@
...
+
+    const {
+      payload: { hash: txHash },
+    } = yield takeFrom(
+      editColony.channel,
+      ActionTypes.TRANSACTION_HASH_RECEIVED,
+    );
+    yield takeFrom(editColony.channel, ActionTypes.TRANSACTION_SUCCEEDED);
+
+    // if (annotationMessage) {
```
rdig, line 135: Did we ever decide what we want to do with annotations ?

rdig (approved):
All good

jakubcolony:
@willm30 @ArmandoGraterol thanks for your input, I have now made the changes required to show non-registered users in the address book.

>As an aside, we should probably refetch the colony object once the action completes so the user doesn't need to refresh the browser to see changes. This isn't that noticeable currently since you have to refresh anyway to update the ActionDetailsPage, but once the polling is wired in, we'll notice it 

Even with polling wired in, won't there be a redirect which will cause the `ColonyContext` to refetch the colony object?

src/components/common/Dialogs/ManageWhitelistDialog/UserCheckbox/UserCheckbox.tsx
```diff
@@ -18,16 +17,17 @@ interface Props {
...
 
 const UserCheckbox = ({
   walletAddress,
   name,
   checkedTooltipText,
   unCheckedTooltipText,
+  user,
   showDisplayName = true,
 }: Props) => {
-  const { user } = useUserByNameOrAddress(walletAddress);
```
jakubcolony, line 30: This component is only used for the Whitelist dialog and by the time it's rendered we'd have already fetched the users info, no need for a separate query

src/graphql/queries/user.graphql
```diff
@@ -1,4 +1,4 @@
-query GetCurrentUser($address: ID!) {
+query GetUserByAddress($address: ID!) {
```
jakubcolony, line 1: Renamed as it better reflects what the query actually does. The same change is already in #257

src/hooks/index.ts
```diff
@@ -48,7 +48,6 @@ export {
   UserDomainReputation,
 } from './useUserReputationForTopDomains';
 export { default as useColonyContext } from './useColonyContext';
-export { default as useUserByNameOrAddress } from './useUserByNameOrAddress';
```
jakubcolony, line 51: I don't think there was a use case for this hook (or at least not one strong enough to justify its overheads) and I felt like it was misused a little.

Under the hood, it would call 2 queries (get user by address and by name), and then returned the results of one of them based on the arguments provided (which was already quite wasteful since we could tell if the argument is an address or username before querying).

Analysing all its usage, the majority of cases called only for the get user by address query rather than name, so I went ahead and replaced those with a hook doing specifically that. There was one case where it was truly getting the user by either name or address (the parameter in user profile route was accepting either one or another) but I've checked in the Dapp and we didn't allow the address in the profile route there, so I made CDapp consistent with that. Let me know if this was a new requirement and should be kept, and I'll go back and revert.

willm30:
> Even with polling wired in, won't there be a redirect which will cause the ColonyContext to refetch the colony object?

The redirect (i.e. the call to `navigate`) doesn't update the colony context.

src/hooks/index.ts
willm30: The question about the user profile route param would be one for @rdig

src/components/common/Dialogs/EditColonyDetailsDialog/EditColonyDetailsDialogForm.tsx
```diff
@@ -75,18 +80,25 @@ const EditColonyDetailsDialogForm = ({
       enabledExtensionData,
     );
 
-  const handleFileRead = async (file) => {
-    if (file) {
-      const base64image = file.data;
-      setValue('colonyAvatarImage', String(base64image), { shouldDirty: true });
+  const handleFileRead = async (file: FileReaderFile) => {
+    if (!file) {
```
willm30, line 84: I think we could update the file type if you believe it might be null/undefined

src/hooks/useUserByAddress.ts
```diff
@@ -0,0 +1,21 @@
+import { useGetUserByAddressQuery } from '~gql';
+import { Address } from '~types';
+
+const useUserByAddress = (address: Address) => {
```
willm30, line 4: We have discussed previously the pointlessness of wrapping queries in hooks. This is one of those cases, no?

src/hooks/useUserByName.ts
```diff
@@ -0,0 +1,20 @@
+import { useGetUserByNameQuery } from '~gql';
+
+const useUserByName = (username: string) => {
```
willm30, line 3: Same thought as above. From memory, your concern was that it just adds another layer of abstraction on top of codegen and isn't worth it given the only utility is destructuring the data.

willm30 (approved):
Address book now working as expected, and you mentioned the details widget will be fixed in the other pr so all good from me. I'll keep an eye on the refetching when this gets incorporated into the motions branch, as it's difficult to verify without having the polling available.

ArmandoGraterol (approved):
It seems like the issues were solved. Thank you, Jakub!

src/hooks/useUserByAddress.ts
jakubcolony: Yeah, that's right. I guess I just followed the existing pattern of this being a hook. However, I think it makes sense for the address query to stay wrapped in a hook, since it's being used in multiple places (which would otherwise lead to repetition of the boilerplate)

jakubcolony:
> The redirect (i.e. the call to navigate) doesn't update the colony context, or at least, I didn't notice this behaviour when on this branch but I can keep this in mind when re-reviewing.

@willm30 You're right, the colony doesn't update. Added the `refetchQueries` option to the mutation modifying the whitelist. Thanks!

Issue #392, opened 2023-04-11
Hitting "Sign out" doesn't update msg.sender without browser refresh.


willm30:
Go to motions and stake with a user.
Then sign out and sign in with another user. Stake again. 
Check the db record for user's stakes. You'll notice that the same entry has been updated twice (instead of two unique entries).
This is because when signing in/out, the msg.sender() is not being updated. 
It gets updated if you refresh the browser.

This will have something to do with the way we're handling user logins when the app first starts vs when merely signing-in/out.

I imagine this will resolve #381

PR #409, opened 2023-04-13
Fix inability to claim rewards when motion fails without vote

### LINKED ISSUE #406: Add claim to failed motion with objection
Currently, if we object a motion 100% and it ends, (and therefore fails), we aren't generating staking rewards which means the claim widget doesn't show.

Todo: think about how best to generate staking rewards in this case (as I don't think we can rely on an event being triggered when the motion ends). Lamba most likely (edit: can go in getMotionState lamda?)

willm30:
## Description

This PR fixes a problem with staker rewards not being present in the db in the event where a motion fails before it goes to a vote. Previously, I was exclusively populating staker rewards via the block ingestor (when a motion is finalized). 

However, it's also necessary in the event that the motion failed without going to a vote. If this happens, there is no specific event emitted via the contracts to let us know, so I have updated the `fetchMotionState` lamda we already have to check and update the db with staker rewards in the event the motion has failed without going to a vote. 

motionFailsNoVote

## To Test

Get yourself set up (see instructions in #380) and create a motion.
Stake / object, but don't: 
stake in favour 100% (so the motion doesn't pass)
stake in favour 100% and object 100% (so we don't trigger a vote).

Fast forward in time:

```
curl -X POST --data '{"jsonrpc":"2.0","method":"evm_increaseTime","params":[3600],"id":1}' localhost:8545

```

It's one hour in the example above, change that to however long you set up the voting rep extn to stake for. Refresh the browser. 

You should see:

1. The motion has failed
2. You are able to claim your rewards
3. Claiming your reward and refreshing doesn't change the information presented to you (i.e. stake, winnings, and total are all the same as before). 

If you test this with multiple users, keep #392 in mind and refresh the browser when switching

**Changes** 🏗

* Update `fetchMotionState` lamda to check if we need to update the staker rewards if the motion fails before going to a vote

Resolves #406

### DIFF
--- a/amplify/backend/api/colonycdapp/schema.graphql
+++ b/amplify/backend/api/colonycdapp/schema.graphql
@@ -116,6 +116,7 @@ input GetUserTokenBalanceInput {
 input GetMotionStateInput {
   motionId: Int!
   colonyAddress: String!
+  transactionHash: String!
 }
 
 input GetVoterRewardsInput {
@@ -455,9 +456,19 @@ input TokenInput {
   id: ID!
 }
 
+input MotionStakeValuesInput {
+  yay: String!
+  nay: String!
+}
+
 type MotionStakeValues {
   yay: String!
   nay: String!
+}
+
+input MotionStakesInput {
+  raw: MotionStakeValuesInput!
+  percentage: MotionStakeValuesInput!
 }
 
 type MotionStakes {
@@ -465,9 +476,20 @@ type MotionStakes {
   percentage: MotionStakeValues!
 }
 
+input UserStakesInput {
+  address: String!
+  stakes: MotionStakesInput!
+}
+
 type UserStakes {
   address: String!
   stakes: MotionStakes!
+}
+
+input StakerRewardsInput {
+  address: String!
+  rewards: MotionStakeValuesInput!
+  isClaimed: Boolean!
 }
 
 type StakerRewards {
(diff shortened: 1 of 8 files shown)


amplify/backend/api/colonycdapp/schema.graphql
willm30, line 1: All the changes in this file, aside from adding the transactionHash to `GetMotionStateInput` appear to be the necessary result of calling the `updateColonyAction` mutation from the lamda. 

I have had to duplicate all the types I use to query motion data, because they need to be specifically "inputs" instead of "types" when calling the mutation or else amplify throws an error.

I would love to be wrong about this so if anybody knows better please let me know.

amplify/backend/function/fetchMotionState/src/index.js
```diff
@@ -1,10 +1,42 @@
...
+  const { motionId, colonyAddress, transactionHash } =
+    event.arguments?.input || {};
   /* Get latest motion state from chain */
   const motionState = await getLatestMotionState(colonyAddress, motionId);
+  /*
+   * Check if we need to update staker rewards
+   * This ensures the rewards are present in the event a motion fails before going to a vote,
+   * so the side that has objected can reclaim their stake.
+   */
+  if (motionState === MotionState.Failed) {
```
willm30, line 21: This is clearly a side effect and makes it less obvious what the `fetchMotionState` lamda is doing...

However, the side effect is idempotent (it will only populate staker rewards once, for those that have not been populated previously), which mitigates the risk of hard-to-debug behaviour.

It's also simpler in the sense that it saves me calling a second lamda to perform the check every time we fetch the motion state.

So, I would argue this is ok to live here, despite the fact the lamda name now isn't perfect. I could, of course, change it, to something like "fetchMotionStateAndCheckStakerRewards"... 

But only if we think that's absolutely necessary as it's a mouthful.

amplify/backend/api/colonycdapp/schema.graphql
willm30, line 1: All the changes in this file, aside from adding the transactionHash to GetMotionStateInput are the result of calling the updateColonyAction mutation from the lamda.

I have had to duplicate all the types I use to query motion data, because they need to be specifically "inputs" instead of "types" when calling the mutation or else amplify throws an error.

I would love to be wrong about this so if anybody knows better please let me know.

willm30, line 1: All the changes in this file, aside from adding the transactionHash to GetMotionStateInput are the result of calling the updateColonyAction mutation from the lamda.

I have had to duplicate all the types I use to query motion data, because they need to be specifically "inputs" instead of "types" when calling the mutation or else amplify throws an error.

I would love to be wrong about this so if anybody knows better please let me know.

amplify/backend/function/fetchMotionState/src/utils.js
```diff
@@ -33,4 +38,129 @@ const getLatestMotionState = async (colonyAddress, motionId) => {
...
+  if (!actionData) {
+    console.error(
+      'Could not find motion in db. This is a bug and should be investigated.',
+    );
+  }
+
+  return actionData?.getColonyAction?.motionData;
+};
+
+const getStakerReward = async (motionId, userAddress, colonyAddress) => {
```
willm30, line 86: Logic copied straight from block ingestor.

ArmandoGraterol (approved):
Looks good to me! An alternative to fast-forwarding time through the console is also turning the monitor off before creating the motion, stake whatever you want to stake, and then turning it back on, and refreshing after the monitor stops processing all the blocks :)

rdig (approved):
Yup looks good, and I can't really poke any holes in the logic applied. So good to go :+1:

Issue #357, opened 2023-03-31
Handle motions that belong to an old voting rep extn


willm30:
What are we doing for motions that were installed by a voting rep that's now been uninstalled? 

Clearly we have to do something so they don't appear interactive to the user. 

Just disable the widget?

Or render something else in addition, e.g. some info for the user about why they can't interact with the motion.

Also handle the case in the ActionsList where if two motions have the same native id, they'll display the same state, even if one was created by an uninstalled voting rep version. I.e. "old" motions in the actions list should have their own tag.

willm30:
@rdig Opinion on this?

rdig:
Either render the widget disabled, or just don't render the any widget at all, whatever's easiest.

We'll do a proper solution in the the new UI/IX.

Whatever ends up being chosen, just make sure to leave a comment explaining the choice.

PR #410, opened 2023-04-14
Display motions in actions list

### LINKED ISSUE #302: Wire up motions in actions list
When you stake a motion above 10% it should show up in the actions list, and not before.

willm30:
## Description

This PR correctly displays motions in the actions list.

First, it filters out those that haven't been staked above the threshold (handled in https://github.com/JoinColony/block-ingestor/pull/53). It also filters out actions created by the voting rep extn (to avoid duplicates when finalizing a motion).

Second, it adds the correct tag to the action item.

tags

## Testing

You need to be able to stake a motion (see #380 for instructions)

1. Create a motion. It should not appear in the actions list.
2. Stake a motion. It should only appear in the actions list if you stake above 10% total. 
3. Finalize a motion. The corresponding action should not appear in the actions list (although you'll be able to see it added to the db).
4. Create an action normally. That should appear in the actions list. 

At all times, motion state should be correctly displayed as a tag on the ActionListItem. 

If you disable the voting rep extn, we remove the tags. #357 exists to handle the display of "old" motions, i.e. ones that belong to a voting rep extn that isn't currently installed. 

**Changes** 🏗

* Update schema to include "showInActionsList" flag
* Filter actions by this flag

Resolves #302, #414


src/components/common/ColonyActions/usePaginatedActions.ts
```diff
@@ -45,19 +50,23 @@ export const usePaginatedActions = (): UsePaginatedActionsReturn => {
     setSortDirection(newSortDirection);
   };
 
-  const loadMoreActions = () => {
+  if (hasMoreActions && filteredActions.length < actionsListSize) {
```
willm30, line 53: If we don't have any more actions to fetch: no problem. Just display what we have.

If we do have more actions to fetch, and we don't yet have enough actions to display, then fetch some more actions. Keep going until we have enough to display to the user.

src/components/common/ColonyActions/usePaginatedActions.ts
```diff
@@ -45,19 +50,23 @@ export const usePaginatedActions = (): UsePaginatedActionsReturn => {
...
     fetchMore({
       variables: {
         nextToken,
       },
     });
+  }
+
+  const loadMoreActions = () => {
+    setCount((ct) => ct + 1);
```
willm30, line 62: Instead of fetching directly from this call, we increase the amount of actions we want to display to the user (`actionListSize`). 

This will re-run the check on line 53. If we already have enough data in filtered actions, we don't need to fetch any more. Else, we will fetch more.

src/components/common/ColonyActions/ActionsListItem/helpers.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  if (motionState && motionData) {
+    return motionTags[getMotionState(motionState, motionData)];
+  }
+
+  // If no motion state / data, display as invalid
+  return motionTags.Invalid;
+
+  // TODO: handle case where motion is from now-uninstalled voting rep version.
```
willm30, line 36: #357

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -560,9 +560,11 @@ type ColonyAction @model {
   blockNumber: Int!
   isMotion: Boolean
   motionData: MotionData
+  showInActionsList: Boolean! # True for (forced) actions. True for motions if staked above 10%
   createdAt: AWSDateTime!
+  initiatorAddress: ID!
```
willm30, line 565: I am relying on this being present to determine whether to filter out an action from the actions list. If the initiatorAddress is a voting reputation extension, then it shouldn't appear. 

Therefore, making compulsory. Requires no downstream changes since it was already present on all db mutations, as far as I can see.

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -116,7 +116,7 @@ input GetUserTokenBalanceInput {
 input GetMotionStateInput {
   motionId: Int!
   colonyAddress: String!
-  transactionHash: String!
+  transactionHash: String
```
willm30, line 119: Fetching motion state from an ActionListItem does not need to update staker rewards, so making this optional

src/components/common/ColonyActions/ActionsListItem/ActionsListItem.tsx
```diff
@@ -58,19 +61,15 @@ const ActionsListItem = ({
...
-
   // const displayCountdownTimer = shouldDisplayCountDownTimer(
   //   motionId,
   //   motionState,
   // );
 
   const status = ListItemStatus.Defused;
+  const { isVotingReputationEnabled } = useEnabledExtensions();
+  const MotionTag = useMotionTag(isMotion, motionData);
+  const showMotionTag = isVotingReputationEnabled && !!MotionTag;
```
jakubcolony, line 72: Could that and the `useEnabledExtensions` call live in the hook above too?

jakubcolony (approved):
LGTM! 👍

src/graphql/fragments/actions.graphql
```diff
@@ -36,6 +36,7 @@ fragment ColonyAction on ColonyAction {
       tokenAddress: id
     }
   }
+  showInActionsList
```
rdig, line 39: The only thing I can't really wrap my head around is how this value is going to work when you have an voting rep extension installed, create a motion, it passes, then you uninstall the voting rep extension.

But that's maybe because I'm missing something

rdig (approved):
Looks all good

ArmandoGraterol (approved):
Nice work! I did notice a small details and it's that the motion is appearing when the staking is at 9%:
FireShot Capture 840 - Colony CDapp - localhost

The banner disappears and the motion appears in the action list. Surely it's some rounding error somewhere? Knowing how JS can be with math stuff. I'm still approving as this is something small that could be dealt with later if we want to

src/graphql/fragments/actions.graphql
willm30: I was imagining we'd leave it in but label it with a different a tag.

PR #429, opened 2023-04-18
Add UnlockToken and VersionUpgrade motion support

### LINKED ISSUE #428: Wire in VersionUpgrade and UnlockToken motions

willm30:
## Description

This PR makes a few small adjustments that enable testing of the other motions that use the rootMotion saga:

1. Unlock token
2. VersionUpgrade

It also updates the logic in the DetailsWidget so we can control whether the Motion item gets appended to the widget or replaces another item.

unlock-token

## Testing

Install voting rep extn
Give yourself reputation

1. Unlock Token: via the UAC, unlock your token
2. VersionUpgrade: via the testing button on ColonyHome, upgrade the version

Resolves  #428

### DIFF
--- a/src/components/shared/DetailsWidget/detailsWidgetConfig.tsx
+++ b/src/components/shared/DetailsWidget/detailsWidgetConfig.tsx
@@ -98,20 +98,6 @@ interface DetailItemConfig {
   item: ReactNode;
 }
 
-const getMotionDetailItem = (colony: Colony, motionDomainId?: number) => {
-  const motionDomain = findDomainByNativeId(
-    Number(motionDomainId ?? 1),
-    colony,
-  );
-  return {
-    label: MSG.motionDomain,
-    labelValues: undefined,
-    item: motionDomain?.metadata && (
-      <TeamDetail domainMetadata={motionDomain.metadata} />
-    ),
-  };
-};
-
 const getDetailItemsMap = (
   colony: Colony,
   {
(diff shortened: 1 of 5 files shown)


rdig (approved):
Seems reasonable

src/components/shared/DetailsWidget/detailsWidgetConfig.tsx
```diff
@@ -122,12 +108,18 @@ const getDetailItemsMap = (
...
     token,
+    motionData,
   }: // roles,
   ColonyAction,
 ): { [key in Exclude<ActionPageDetails, 'Permissions'>]: DetailItemConfig } => {
   const shortenedHash = getShortenedHash(transactionHash || '');
   const recipientWalletAddress = recipient?.walletAddress;
   const isSmiteAction = type === ColonyActionType.EmitDomainReputationPenalty;
+  const motionDomain = findDomainByNativeId(
+    Number(motionData?.motionDomainId ?? 1),
```
jakubcolony, line 119: Should the fallback value come from `Id.RootDomain`?

src/components/shared/DetailsWidget/detailsWidgetConfig.tsx
```diff
@@ -237,19 +236,12 @@ const getDetailItems = (
...
 
-  if (actionData.isMotion) {
-    detailItems.splice(1, 0, motionDetailItem);
-  }
+  const detailItems = detailItemKeys
+    .map<DetailItemConfig | undefined>((itemKey) => detailItemsMap[itemKey])
+    .filter((detail) => !!detail?.item);
 
-  return detailItems;
+  return detailItems as DetailItemConfig[];
```
jakubcolony, line 244: Could the casting be moved to a generic accepted by `filter`? 

```js
.filter<DetailItemConfig>()
```

jakubcolony (approved):
LGTM, nice work 👍

src/components/shared/DetailsWidget/detailsWidgetConfig.tsx
willm30: @jakubcolony 

This doesn't work:

``` ts
.filter<DetailItemConfig>((detail) => !!detail?.item)
```

Because the type paramter is telling the filter function that the return value of the predicate function will be `DetailItemConfig`, when in fact it's a boolean. 

Type guards, however, are compatible with filtering, so the following achieves the intended type inference:

``` ts
.filter((detail): detail is DetailItemConfig => !!detail?.item);
```

jakubcolony: My bad, ignore me then

PR #439, opened 2023-04-19
Wire Manage Tokens dialog with action saga

### LINKED ISSUE #434: Wire Manage Tokens dialog with saga
It's proving to be a bit difficult since the existing relation between colonies and tokens just doesn't seem to pick up the newly added token address. I span it from #405 so I can get that PR merged and work on this separately.

jakubcolony:
## Description

This PR wires the token management dialog with the existing data and ensures the saga stores the modified tokens properly.

## Testing
Test different cases in Token Management Dialog, e.g. removing a single/multiple tokens, adding, removing & adding at the same time. Verify the messages display correct value (note: you might have to wait until the action event message changes from "Colony details changed" to "Colony tokens were updated").

TODO:
- [x] Create colony/token relation in saga
- [x] Delete colony/token relation in saga when token is removed from list
- [x] Add validation to token management dialog checking if we could retrieve the token data 
- [x] Fix not being able to submit token management dialog when removing token
- [x] Display correct message in ActionDetailsPage

Resolves #434


ArmandoGraterol (approved):
Nice work! I think there are some files that were not there when I started reviewing this PR but that may be due to this needing a rebase. 

Nevertheless, tried several combinations and it's working as expected 👌

willm30 (commented):
I think the `TokenSelector` is using the wrong query. It's just querying the db, but we want it to query the chain also, right? I.e. it should call the `fetchTokenFromChain` lamda. I think that's also true for the ColonyCreationWizard, in which case we could keep using the same component for both (but we should move `TokenSelector` to `shared`).

Currently I can enter a token address that's not in the db or chain and it still lets me submit the form.

src/components/common/ColonyActions/helpers/getActionTitleMessageId.ts
```diff
@@ -5,13 +5,26 @@ const getActionTitleMessageId = (actionData: ColonyAction, colony?: Colony) => {
...
     return `action.${ColonyActionType.ColonyEdit}.verifiedAddresses`;
   }
 
+  if (
+    actionData.type === ColonyActionType.ColonyEdit &&
+    changelogItem?.haveTokensChanged
+  ) {
+    return `action.${ColonyActionType.ColonyEdit}.tokens`;
+  }
+
```
rdig, line 27: Yup, just noticed this. This is the thing I asked in https://github.com/JoinColony/colonyCDapp/pull/405#discussion_r1181887887

rdig (approved):
All good. but I agree with Will that we should be using the `fetchTokenFromChain` query every time we're dealing with Tokens

src/components/common/CreateColonyWizard/CreateColonyWizard.ts
```diff
@@ -19,9 +20,10 @@ const stepArray: StepType[] = [
 ];
 
 export type FormValues = {
-  tokenName: string;
-  tokenSymbol: string;
+  tokenName?: string;
+  tokenSymbol?: string;
   tokenAddress: string;
+  token?: Token | null;
```
jakubcolony, line 26: Either `tokenName`/`tokenSymbol` (when creating a new token) or `token` will be defined

jakubcolony: Storing `token` in hook-form state allows us to simplify the validation (we used to run the get token query twice, this allows us to simply check if token is null or not)

src/components/shared/Fields/Checkbox/HookForm/Checkbox.tsx
```diff
@@ -84,7 +83,10 @@ const HookFormCheckbox = ({
   );
 
   return (
-    <div className={classNames}>
+    <fieldset
+      className={classNames(mainClasses, className)}
+      disabled={disabled}
+    >
```
jakubcolony, line 89: When you set an input as disabled, the hook-form sets its value to undefined. Per docs (https://react-hook-form.com/api/useform/register/):
> disabled input will result in an undefined form value. If you want to prevent users from updating the input, you can use readOnly or disable the entire <fieldset />

readOnly doesn't work for checkboxes, so I went with wrapping the checkbox in a fieldset

src/components/common/CreateColonyWizard/StepSelectToken.tsx
```diff
@@ -74,16 +71,6 @@ export const switchTokenInputType = (
...
 
-const handleFetchSuccess = (
-  { getTokenByAddress }: GetTokenByAddressQuery,
-  setValue: UseFormSetValue<Step3>,
-) => {
-  const token = getTokenByAddress?.items[0];
-  const { name: tokenName, symbol: tokenSymbol } = token || {};
-  setValue('tokenName', tokenName || '');
-  setValue('tokenSymbol', tokenSymbol || '');
-};
```
jakubcolony, line 85: `StepSelectToken` doesn't need to know anything about the underlying queries

src/components/shared/Wizard/types.ts
```diff
@@ -1,8 +1,6 @@
 import { AppContextValues } from '~context/AppContext';
-import { Values } from '~types';
 
-type WizardValues<T> = Record<string, Values<T>>;
-export type StepValues<T> = Partial<WizardValues<T>>;
```
jakubcolony, line 5: @willm30 For some reason, this was interfering with introducing any fields that are not of primitive type (e.g. the `token` field). It would cause the types of all the other fields to become `undefined`. Removing it didn't throw any error and as far as I can tell the types still work. Would appreciate if you could check too

src/components/shared/TokenSelector/TokenSelector.tsx
```diff
@@ -61,51 +55,58 @@ const getStatusText = ({
...
 };
 
-const getTokenData = (tokenQuery?: GetTokenByAddressQuery) => {
-  const tokenData = tokenQuery?.getTokenByAddress?.items;
-  const { name, symbol } = tokenData?.[0] || {};
-  return { name, symbol };
-};
-
 const getLoadingState = (isLoading: boolean) =>
   // Don't show loading state if connection is 4G or unknown
```
jakubcolony, line 70: @willm30 I can't remember what was the reason behind this, do you remember why we're not showing loading state for 4g connections?

jakubcolony:
@willm30 Good shout, thanks, it should definitely use the `GetTokenFromEverywhereQuery`. I have updated that for both Token Management Dialog and Create Colony Wizard and moved the selector component to `shared`. It took me way longer than expected and I had to try a few approaches before coming up with storing the retrieved token object in hook-form. This allowed for simpler validation and to remove the second query one of the validation functions was calling, encapsulate as much token selector specific logic as possible and fix the console error I spotted on `master`:
image

Let me know if you have any thoughts

@ArmandoGraterol @rdig you're also welcome to re-review if you wish

src/components/shared/TokenSelector/TokenSelector.tsx
willm30: Showing the loading state for a query that returns quickly looks bad IMO, since it flashes momentarily then disappears. I'm not tied to this way of doing it but I do think it's an improvement over the flashing

src/components/shared/Wizard/types.ts
willm30: LGTM

src/components/common/CreateColonyWizard/StepSelectToken.tsx
willm30: Would removing these lines having something to do with this?

nativetoken

willm30 (commented):
Manage Token dialog working for me.

Did you say you were going to update the details widget in this pr? I.e. so it shows the update address book text instead of colony edit?
address-book

There's also a currently a problem with the create colony flow. The native token isn't showing if you enter it manually (I've left a comment with screenshot)

jakubcolony:
@willm30 

> There's also a currently a problem with the create colony flow. The native token isn't showing if you enter it manually

Thanks, should be fixed now

>  I.e. so it shows the update address book text instead of colony edit?

Isn't that what's happening already? If you see the colony edit text on the first try, can you refresh? It might be due to the colony metadata not being refetched which is fixed by #514

willm30:
> Isn't that what's happening already?

I mean in the details widget. It shows the Colony Edit detail instead of the Address Book detail. I was [under the impression](https://github.com/JoinColony/colonyCDapp/pull/405#issuecomment-1526550236) you would do in this PR. 

Just noticed a tiny bit of css misalignment in the Manage Tokens Dialog. The checkbox looks to be cut off on the left hand side:
manage-tokens-css

But these aside, all working for me so approving.

willm30 (approved):

jakubcolony:
@willm30 Right, I misunderstood you. The details are now showing specific types for updating tokens and address book. Thanks

PR #448, opened 2023-04-20
Wire up motion countdown timer in action list

### LINKED ISSUE #416: Wire MotionTimer with ActionItem in Actions list
We need to ensure the countdown timer is present in the actionItem in the actions list when it's a motion.

Base off #410

ArmandoGraterol:
To test just follow the same process for #367 or #410. 

Once you have rep, you can turn the monitor off and just create a bunch of motions with 2 users or more and advance the motion phases via staking/voting YAY and/or NAY like so:
FireShot Capture 841 - Colony CDapp - localhost

The timer should only show for the motion states above.

Resolves #416


willm30 (approved):
Nice!

jakubcolony (approved):
LGTM, nice one 👍

PR #452, opened 2023-04-21
Ensure actions list always displays correct number of actions

### LINKED ISSUE #441: Improve actions list query
Currently, the actions list query gets colony actions in created by order, up to the limit (10), and then applies the filter `showInActionsList`, meaning we'll get <= 10 results, whereas really we want exactly 10 results. 

This could be improved (in theory) if we added `showInActionsList` as a sort key, but you can't have boolean sort keys. You could, however, use a string with two options ("true" and "false"), which may get around this problem and ensure filtering is applied before the limit, not after.

willm30:
## Description

This PR adds some logic to ensure the actions list always shows the correct number of actions.

The "correct" number is: ITEMS_PER_PAGE * number of pages requested by user. 

The added logic ensures the query automatically refetches in the event we don't have enough actions to display to the user.

Note that I tried to implement this directly into the schema, but my idea of using a composite sort key on the colonyId index on the ColonyActions model, but:

a) it didn't work
b) composite sort keys are poorly documented, making it difficult to debug/fix.

In any case, this is not a lot of extra work for the client. We're just iterating through the query results once (via `slice`) to ensure the user always sees the correct number of actions.

## Testing

Easiest to test if you change the `ITEMS_PER_PAGE` constant in `usePaginatedActions` to something lower, like 3. Then set yourself up to be able to stake (#380), and create a few motions. 

Assuming you choose 3 as the constant, you'll notice that the actions list will always show the 3 most recently created actions that have `showInActionsList` set to `true` (verify this in the db). This is the case even if the first 3 results contain results with `showInActionsLis set to `false`, i.e. after creating a motion that hasn't been staked sufficiently, or finalizing a motion (which writes an additional action to the db).

This is the behaviour I thought I was introducing when using the built-in query filter, instead of writing the filtering logic in the client. But dynamodb applies the filter *after* the limit we supply, meaning we may receive fewer results. Hence the need for automatic refetching.

**Changes** 🏗

* Add aforementioned logic to `usePaginatedActions`

Resolves  #441


src/hooks/usePaginatedActions.ts
```diff
@@ -37,28 +37,35 @@ export const usePaginatedActions = (): UsePaginatedActionsReturn => {
...
+  const fetchMoreActions = nextToken && actions.length < visibleActions;
+  if (fetchMoreActions) {
     fetchMore({
       variables: {
         nextToken,
       },
     });
+  }
+
+  const loadMoreActions = () => {
```
willm30, line 60: Every time the user requests more actions, we increase the number of "visibleActions". 

The component will rerender and if it needs to fetch more actions from the db, it will.

src/hooks/usePaginatedActions.ts
```diff
@@ -37,28 +37,35 @@ export const usePaginatedActions = (): UsePaginatedActionsReturn => {
...
+    setVisibleActions((count) => count + ITEMS_PER_PAGE);
```
willm30, line 61: Every time the user requests more actions, we increase the number of "visibleActions". 

The component will rerender and if it needs to fetch more actions from the db, it will (line 52 above).

ArmandoGraterol (approved):
LGTM 👌

src/hooks/usePaginatedActions.ts
```diff
@@ -37,28 +37,35 @@ export const usePaginatedActions = (): UsePaginatedActionsReturn => {
     fetchPolicy: 'network-only',
   });
   const { items, nextToken } = data?.getActionsByColony || {};
+  const [visibleActions, setVisibleActions] = useState(ITEMS_PER_PAGE);
```
jakubcolony, line 40: How about `visibleActionsCount` to be specific it's a number, not an array of objects?

jakubcolony (approved):
Very clean solution, nice work 👍

PR #460, opened 2023-04-24
Hide Motion widget when voting rep uninstalled

willm30:
## Description

This PR ensures that we "disable" motions that do not belong to the current installation of the voting rep extn.

By this, I mean that:

- You can still view the details, at their most recent state
- You can't interact with the motion via the widget (because it's hidden)
- If a new voting rep exn is currently installed, you will see an "invalid" tag; else, you won't see any tag (because we don't display motion tags when the voting rep extn isn't installed).

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

* Changes go here

**Deletions** ⚰️

* Deletions go here

## TODO

- [ ] Add TODOs


willm30:
Closed in favour of https://github.com/JoinColony/colonyCDapp/pull/461

PR #461, opened 2023-04-24
Hide Motion widget when voting rep uninstalled

### LINKED ISSUE #357: Handle motions that belong to an old voting rep extn

willm30:
## Description

This PR ensures that we "disable" motions that do not belong to the current installation of the voting rep extn.

By this, I mean that:

- You can still view the details, at their most recent state
- You can't interact with the motion via the widget (because it's not present)
- If a new voting rep extn is currently enabled, you will see an "invalid" tag; else, you won't see any tag (because we don't display motion tags when the voting rep extn isn't enabled).

This applies in the following cases:
* The voting rep extn is uninstalled
* The voting rep extn is deprecated
* The voting rep extn is installed, but not the same installation that created the motion

Note that this hasn't been specced so I'm just relying on showing the "invalid" tag for motions that don't belong to the currently installed voting extn. Improvements in this respect should probably wait for the new UI.

## Testing

Install voting rep extn. Give yourself some rep.
Create a motion, stake it so it shows up in the action list.
Uninstall voting rep 
Check that you can no longer interact with the motion widget on the motion page
Reinstall voting rep
Check that the motion is now tagged with "invalid"

You can also try by deprecating the voting rep instead of uninstalling it. They should work the same way. We ultimately trying to ensure you cannot interact with a motion that was not created by the currently installed (and enabled) voting rep extn, and that there's some feedback in the ui to indicate why.

**Changes** 🏗

* Ensure motions from old voting rep versions can't be interacted with (are "read-only")
* Refactor the logic to display the motion tag in ActionsListItem

Resolves  #357

### DIFF
--- a/amplify/backend/api/colonycdapp/schema.graphql
+++ b/amplify/backend/api/colonycdapp/schema.graphql
@@ -114,9 +114,8 @@ input GetUserTokenBalanceInput {
 }
 
 input GetMotionStateInput {
-  motionId: Int!
   colonyAddress: String!
-  transactionHash: String
+  transactionHash: String!
 }
 
 input GetVoterRewardsInput {
@@ -150,9 +149,9 @@ type GetUserTokenBalanceReturn {
 }
 
 type GetMotionTimeoutPeriodsReturn {
-  timeLeftToStake: String!,
-  timeLeftToVote: String!,
-  timeLeftToReveal: String!,
+  timeLeftToStake: String!
+  timeLeftToVote: String!
+  timeLeftToReveal: String!
   timeLeftToEscalate: String!
 }
 
@@ -260,12 +259,13 @@ type Query {
   ): GetUserTokenBalanceReturn @function(name: "getUserTokenBalance")
   getMembersForColony(input: MembersForColonyInput): MembersForColonyReturn
     @function(name: "getMembersForColony")
-  getMotionState(input: GetMotionStateInput): Int
+  getMotionState(input: GetMotionStateInput): Int!
     @function(name: "fetchMotionState")
   getVoterRewards(input: GetVoterRewardsInput): VoterRewardsReturn
     @function(name: "fetchVoterRewards")
-  getMotionTimeoutPeriods(input: GetMotionTimeoutPeriodsInput): GetMotionTimeoutPeriodsReturn
-    @function(name: "fetchMotionTimeoutPeriods")
+  getMotionTimeoutPeriods(
+    input: GetMotionTimeoutPeriodsInput
+  ): GetMotionTimeoutPeriodsReturn @function(name: "fetchMotionTimeoutPeriods")
 }
 
 type Mutation {
@@ -686,5 +686,6 @@ type ColonyAction @model {
   fundamentalChainId: Int
   newColonyVersion: Int
   pendingDomainMetadataId: ID
-  pendingDomainMetadata: DomainMetadata @hasOne(fields: ["pendingDomainMetadataId"])
+  pendingDomainMetadata: DomainMetadata
+    @hasOne(fields: ["pendingDomainMetadataId"])
 }
--- a/amplify/backend/function/fetchMotionState/src/utils.js
+++ b/amplify/backend/function/fetchMotionState/src/utils.js
@@ -38,19 +38,17 @@ const getLatestMotionState = async (
   colonyAddress,
   { nativeMotionId, createdBy },
 ) => {
-  let motionState = MotionState.Null;
   try {
     const votingReputationClient = await getVotingClient(colonyAddress);
     const isDeprecated = await votingReputationClient.getDeprecated();
     /* Only fetch state for a motion that was created by the current (active) installation of the voting rep extn */
     if (!isDeprecated && votingReputationClient.address === createdBy) {
-      motionState = votingReputationClient.getMotionState(nativeMotionId);
+      return votingReputationClient.getMotionState(nativeMotionId);
     }
+    return MotionState.Null;
   } catch {
-    /* This will fail if the voting reputation extn is not installed, in which case we return 0. */
+    return MotionState.Null;
   }
-
-  return motionState;
 };
 
 const graphqlRequest = async (queryOrMutation, variables) => {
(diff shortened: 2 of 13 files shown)


amplify/backend/function/fetchMotionState/src/utils.js
```diff
@@ -33,9 +34,22 @@ const getVotingClient = async (colonyAddress) => {
...
-  const votingReputationClient = await getVotingClient(colonyAddress);
-  return votingReputationClient.getMotionState(motionId);
+const getLatestMotionState = async (
+  colonyAddress,
+  { nativeMotionId, createdBy },
+) => {
+  let motionState = MotionState.Null;
+  try {
+    const votingReputationClient = await getVotingClient(colonyAddress);
+    /* Only fetch state for a motion that was created by the current installation of the voting rep extn */
```
willm30, line 44: This ensures we don't get clashes whereby two motions of native Id 1 from different installations have the same status. 

One should be invalid (the uninstalled one), and the other whatever is the current state of the motion.

src/components/common/ColonyActions/ActionsListItem/helpers.ts
willm30, line 1: I have split `useMotionStatusDisplay` into the hooks it's comprimsed of as it makes it more obvious what's going on

src/components/common/ColonyActions/ActionsListItem/helpers.ts
```diff
@@ -1,64 +1,53 @@
...
+      skip: !isMotion || !motionData || !colony || !isVotingReputationEnabled,
+      variables: {
+        input: {
+          colonyAddress: colony?.colonyAddress ?? '',
+          transactionHash,
+        },
       },
-    },
-  });
+      fetchPolicy: 'cache-and-network',
```
willm30, line 24: We need it to fetch over the network to get the latest state in the event that the voting rep extn is uninstalled / reinstalled

src/components/common/ColonyActions/ActionsListItem/helpers.ts
```diff
@@ -1,64 +1,53 @@
...
+    });
   const networkMotionState = motionStateData?.getMotionState;
 
-  const showMotionTag = !loadingMotionState && isVotingReputationEnabled;
-
   const motionState =
-    networkMotionState && motionData
+    networkMotionState !== undefined && motionData
       ? getMotionState(networkMotionState, motionData)
-      : MotionState.Invalid;
```
willm30, line 51: We want a distinction between an invalid motion and a motion for which we don't currently have the motion state (else we'll see invalid when the motion state is loading).

src/components/common/ColonyActions/ActionsListItem/helpers.ts
```diff
@@ -1,64 +1,53 @@
...
+          transactionHash,
+        },
       },
-    },
-  });
+      fetchPolicy: 'cache-and-network',
+    });
   const networkMotionState = motionStateData?.getMotionState;
 
-  const showMotionTag = !loadingMotionState && isVotingReputationEnabled;
```
willm30, line 46: Since I've changed the fetch policy, hiding the tag if its loading essentially renders the cache ineffective since it'll hide the tag even if we have cached data available.

amplify/backend/function/fetchMotionState/src/utils.js
```diff
@@ -33,9 +34,23 @@ const getVotingClient = async (colonyAddress) => {
...
+const getLatestMotionState = async (
+  colonyAddress,
+  { nativeMotionId, createdBy },
+) => {
+  let motionState = MotionState.Null;
+  try {
+    const votingReputationClient = await getVotingClient(colonyAddress);
+    const isDeprecated = await votingReputationClient.getDeprecated();
+    /* Only fetch state for a motion that was created by the current (active) installation of the voting rep extn */
+    if (!isDeprecated && votingReputationClient.address === createdBy) {
```
willm30, line 46: This ensures we don't get clashes between two motions with the same native id but which were created by different voting rep extns. 

Also ensures the widget gets disabled in the event the voting rep extn is deprecated (in addition to uninstalled).

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -150,9 +149,9 @@ type GetUserTokenBalanceReturn {
 }
 
 type GetMotionTimeoutPeriodsReturn {
-  timeLeftToStake: String!,
-  timeLeftToVote: String!,
-  timeLeftToReveal: String!,
```
rdig, line 155: Wow, how did that get in there :)

src/components/common/ColonyActions/ActionsListItem/helpers.ts
rdig: Maybe leave this comment in code ?

rdig (approved):
Very nice work :+1:

ArmandoGraterol (approved):
LGTM 👌

amplify/backend/function/fetchMotionState/src/utils.js
```diff
@@ -33,9 +34,23 @@ const getVotingClient = async (colonyAddress) => {
...
 };
 
-const getLatestMotionState = async (colonyAddress, motionId) => {
-  const votingReputationClient = await getVotingClient(colonyAddress);
-  return votingReputationClient.getMotionState(motionId);
+const getLatestMotionState = async (
+  colonyAddress,
+  { nativeMotionId, createdBy },
+) => {
+  let motionState = MotionState.Null;
```
jakubcolony, line 41: Returning directly from the try/catch below would let you get rid of this variable (and you'd see straightaway what the catch returns without an extra comment)

jakubcolony (approved):
Nice work 👍

PR #469, opened 2023-04-25
fix carousel styles in storybook

krawczykkatarzyna:
## Description

- [] TODO write description

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

* Changes go here

**Deletions** ⚰️

* Deletions go here

## TODO

- [ ] Add TODOs


src/components/common/Extensions/ImageCarousel/ImageCarousel.css
```diff
@@ -0,0 +1,45 @@
+#storybook-root > div > div > div {
```
KatarzynaStefanowiczPagepro, line 1: We don't need this file (especially notice that these styles are related to the storybook but you put them in app component). Styles in storybook shouldn't be different than in the app.
If you need to add some extra containers helpers with some styles used only in a storybook, you can use a decorator.
https://storybook.js.org/docs/react/writing-stories/decorators

krawczykkatarzyna: it's not possible to overwrite library styles using decorators

KatarzynaStefanowiczPagepro (approved):

agnieszkajarosikloj (approved):

PR #471, opened 2023-04-25
Handle add/edit domain on motion finalized

### LINKED ISSUE #462: Connect pending domain metadata to real domain on motion finalized
Link pending domain metadata in motion with the real domain being created/edited when the motion has passed and has been finalized.

Image

ArmandoGraterol:
The bulk of this update is in the block-ingestor counterpart: https://github.com/JoinColony/block-ingestor/pull/64

- You can test this by following the same process needed to go through the whole motion ex. Get rep, get tokens to stake with, and assign the correct permissions. In the particular case of the add/edit domain motions, the voting extension needs the `Architecture` permission to create them:

  - `c.setArchitectureRole(1, "0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff", '<voting-reputation-id>', 1, true) `

  - and if you want to edit a domain and create the motion in the domain itself, you are going to need rep in that team, for example:
`c.emitDomainReputationReward(<id-of-subdomain>, accounts[0], '1000000000000000000')`

- The Add/Edit domain motions can be created through the UAC.
- We should make sure that every possible way of finalizing a motion with the YAY side winning (Through staking OR voting) creates/updates the targeted domain.
- If the NAY side wins or the motion failed but can't be finalized (Time ran out and nobody staked enough for example), nothing should happen.
- After finalizing the motion, and you refresh, you should see the new/updated domain

Resolves #462

### DIFF
--- a/docker/colony-cdapp-dev-env-block-ingestor
+++ b/docker/colony-cdapp-dev-env-block-ingestor
@@ -1,6 +1,6 @@
 FROM colony-cdapp-dev-env/base:latest
 
-ENV BLOCK_INGESTOR_HASH=a0b8868f3cb26f869711d7e92f4cf4b5560140fd
+ENV BLOCK_INGESTOR_HASH=013adb9cc1084573c87de9b785ec11c67c967537
 
 # Declare volumes to set up metadata
 VOLUME [ "/colonyCDapp/amplify/mock-data" ]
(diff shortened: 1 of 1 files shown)


willm30:
@ArmandoGraterol I rebased colony motions on master today

ArmandoGraterol:
> @ArmandoGraterol I rebased colony motions on master today

@willm30 Thank you! Updated

jakubcolony (approved):

willm30 (approved):
I'm noticing that the colony data isn't updating automatically so we need to refresh the browser to see the updates, so I've opened a pr for that https://github.com/JoinColony/colonyCDapp/pull/485. Everything else looks good 👍

rdig (approved):

Issue #388, opened 2023-04-10
Wire up Motion Revealed and Motion Finalized Messages for Staking Widget 


Julianahlk:
This is what it looks like when the motion went to a vote and passed:

![votepassed](https://user-images.githubusercontent.com/64402732/233966179-390a9c32-486b-4f4a-8ce9-e2375d42624f.png)

We're also looking to add in this issue:

- when it goes to a vote and fails
- when it doesn't go to a vote and passes
- when it doesn't go to a vote and fails

Don't forget you need to set the root role before finalizing:

c.setRootRole('<voting-rep-extn>', true)

PR #472, opened 2023-04-26
Add support for motion reveal system messages

Julianahlk:
## Description

Adds Support for the following Motion Messages with [block-ingestor#65](https://github.com/JoinColony/block-ingestor/pull/65)

- MotionRevealResultObjectionWon
- MotionHasFailedFinalizable
- MotionHasFailedNotFinalizable
- MotionRevealResultMotionWon
- MotionFinalized
- MotionRevealPhase
- MotionHasPassed

**Passed**

Screenshot 2023-04-28 at 3 02 42 PM

**Failed**

Screen Shot 2023-04-28 at 1 14 11 PM

Contributes to #388


amplify/backend/function/fetchMotionState/src/utils.js
```diff
@@ -34,10 +33,8 @@ const getVotingClient = async (colonyAddress) => {
   return colonyClient.getExtensionClient(Extension.VotingReputation);
 };
 
-const getLatestMotionState = async (
-  colonyAddress,
-  { nativeMotionId, createdBy },
```
willm30, line 39: Any reason for not destructuring the parameter?

amplify/backend/function/fetchMotionState/src/index.js
```diff
@@ -3,37 +3,73 @@ const {
...
+      'inRevealPhase',
+    );
+  }
+
+  if (
+    transactionHash &&
+    motionData &&
+    motionState === MotionState.Finalizable
+  ) {
+    const didPass = await didMotionPass(motionData);
```
willm30, line 55: `didMotionPass` isn't async

willm30 (changes requested):
All working fine for me with one user, but with two, it's telling me the motion passed when I click reveal with a user who voted for the motion, even though it actually failed. 

pass_fail

willm30:
I think we can also drop the last "Motion has Failed" message. There are only two in the dapp: 
dapp

amplify/backend/function/fetchMotionState/src/utils.js
Julianahlk: @willm30 When destructing directly in the parameter, I was getting issues with motionData being undefined.

Julianahlk:
> All working fine for me with one user, but with two, it's telling me the motion passed when I click reveal with a user who voted for the motion, even though it actually failed.
> 
@willm30
1. Not sure how this could happen unless the votingClient is returning different votes as different users reveal their votes and trigger the motionVoteRevealed in the block ingestor. Is the votingClient the correct place to retrieve the voting results? 
this is from the block ingestor 
`  const {
    votes: [nayVotes, yayVotes],
  } = await votingClient.getMotion(motionId);`
  
2. Do you think it makes sense to show the messages with the voting widget for every participant that reveals their vote? this is how it currently works.

willm30:
>Do you think it makes sense to show the messages with the voting widget for every participant that reveals their vote? This is how it currently works.

Not sure what you mean here. There's only one voting message ("Voting has started!"). There's also only one reveal message ("It's time to reveal votes to the world!"). 

Once everyone has revealed their vote, or the timer elapses, and the motion phases moves from Reveal into Finalizable, then we see the outcome of the vote and the motion in the message feed. 

> Not sure how this could happen 

It's because you're adding messages on every reveal event. We should sync it with the motion state instead (via the lamda). If it's the first time we've moved into the Finalizable state, add the relevant messages.

Julianahlk:
@willm30 when you say drop the last motion message, are you referring to? `${SystemMessages.MotionHasFailedNotFinalizable} {{motionTag} has failed.}`
It looks like this is still used in the dapp based on the code.

amplify/backend/function/fetchMotionState/src/utils.js
willm30: @Julianahlk  This isn't being caused by the destructuring, it's related to a refetch function that I have a fix for here:https://github.com/JoinColony/colonyCDapp/pull/505

amplify/backend/function/fetchMotionState/src/index.js
```diff
@@ -3,37 +3,91 @@ const {
...
 /**
  * @type {import('@types/aws-lambda').APIGatewayProxyHandler}
  */
 exports.handler = async (event) => {
   const { colonyAddress, transactionHash } = event.arguments?.input || {};
   /* Get latest motion state from chain */
   const motionData = await getMotionData(transactionHash);
-  const motionState = await getLatestMotionState(colonyAddress, motionData);
+  const motionState =
+    motionData && (await getLatestMotionState(colonyAddress, motionData));
```
willm30, line 17: With #505, you shouldn't need this guard as we shouldn't be calling this function with a tx hash that isn't a motion. In fact, motion data being undefined (and thus triggering a lamda error) is a clue this is being called incorrectly, and thus might be useful.

I'd say if you want to be cautious, wrap everything from line 16 down in:

``` js

if(motionData) {
} 
```

And we can also update the logging logic in `getMotionData` to log if `motionData` is undefined, instead of actionData, so we'll see in the console if motion data isn't found.

amplify/backend/function/fetchMotionState/src/index.js
```diff
@@ -3,37 +3,91 @@ const {
...
    * so the side that has objected can reclaim their stake.
    */
-  if (transactionHash && motionState === MotionState.Failed) {
-    if (motionData) {
-      const {
-        motionStakes: {
-          percentage: { yay: yayPercent },
-        },
-      } = motionData;
+  if (transactionHash && motionData && motionState === MotionState.Failed) {
```
willm30, line 23: The query should also be skipped if there's no transactionHash, so I'd say the same as above, if you want to be cautious, add the check for a tx hash in an `if` statement at the top of the function. But you'll get a lamda error if there's no tx hash or if it's an empty string:
> ValidationException: One or more parameter values are not valid. The AttributeValue for a key attribute cannot contain an empty string value. Key: id

So you'll know for sure the query is being called incorrectly.

amplify/backend/function/fetchMotionState/src/index.js
```diff
@@ -3,37 +3,91 @@ const {
...
+        motionData,
+        newMessages,
+        'hasPassed',
+      );
+    }
+
+    if (!didPass && !motionData.hasFailed) {
+      const newMessages = [];
+      newMessages.push('MotionRevealResultObjectionWon');
+      newMessages.push('MotionHasFailedFinalizable');
```
willm30, line 73: I think this can be:

``` js
const newMessages = ['MotionRevealResultObjectionWon', 'MotionHasFailedFinalizable']
```

amplify/backend/function/fetchMotionState/src/index.js
```diff
@@ -3,37 +3,91 @@ const {
...
+  if (
+    transactionHash &&
+    motionData &&
+    (motionState === MotionState.Finalizable ||
+      motionState === MotionState.Finalized)
+  ) {
+    if (didPass && !motionData.hasPassed) {
+      const newMessages = [];
+      newMessages.push('MotionRevealResultMotionWon');
+      newMessages.push('MotionHasPassed');
```
willm30, line 61: Same as below

amplify/backend/function/fetchMotionState/src/index.js
```diff
@@ -3,37 +3,91 @@ const {
...
+    );
+  }
+
+  if (
+    transactionHash &&
+    motionData &&
+    (motionState === MotionState.Finalizable ||
+      motionState === MotionState.Finalized)
+  ) {
+    if (didPass && !motionData.hasPassed) {
```
willm30, line 58: Some comments here would be handy for future us so we can quickly remember what each `if` really means

amplify/backend/function/fetchMotionState/src/index.js
willm30: Previously, because there was only one side effect in this lamda, adding the tx hash was effectively "putting it at the top" of the function. But now we have a few, I think it should be extracted

willm30 (changes requested):
We don't want to show this if the motion passes without going to a vote:
no-vote

amplify/backend/function/fetchMotionState/src/utils.js
Julianahlk: Got it, reverting 👍

amplify/backend/function/fetchMotionState/src/graphql.js
```diff
@@ -64,6 +75,14 @@ module.exports = {
           skillRep
           repSubmitted
           createdBy
+          hasObjection
+          hasPassed
+          hasFailed
+          hasFailedNotFinalizable
+          inRevealPhase
```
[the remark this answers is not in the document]
Julianahlk: added `motionStateHistory`

Julianahlk:
@willm30 Updated with all your suggestions

amplify/backend/function/fetchMotionState/src/index.js
```diff
@@ -34,8 +37,73 @@ exports.handler = async (event) => {
...
+          'hasFailed',
+        );
+      }
     }
+
+    return motionState;
   }
 
-  return motionState;
+  return undefined;
```
willm30, line 108: This lamda is typed to return a number so let's return MotionState.Null as the fallback

willm30 (approved):
@Julianahlk Nice job, this is working as expected 👍

amplify/backend/function/fetchMotionState/src/utils.js
```diff
@@ -139,6 +138,36 @@ const getStakerReward = async (motionId, userAddress, colonyAddress) => {
...
+    raw: { yay: yayStakes, nay: nayStakes },
+  },
+  requiredStake,
+  revealedVotes: {
+    raw: { yay: yayVotes, nay: nayVotes },
+  },
+}) => {
+  if (
+    BigNumber.from(nayStakes).gte(requiredStake) &&
+    BigNumber.from(yayStakes).gte(requiredStake)
```
jakubcolony, line 152: We could extract at least the `BigNumber.from(yayStakes)` to its own variable, it happens 3 times in this function

jakubcolony (approved):
Left a few minor comments, nice work overall 👍

src/components/common/ColonyActions/helpers/mapItemToMessageFormat.tsx
```diff
@@ -172,7 +180,9 @@ export const useMapMotionEventToExpectedFormat = (
       address: motionMessageData?.initiatorAddress ?? '',
     },
   });
-
+  if (!motionData) {
+    return {};
+  }
```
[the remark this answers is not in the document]
Julianahlk: added motionData to skip

PR #480, opened 2023-04-26
ui/15399707: specific side panel - base

krawczykkatarzyna:
## Description

- [] TODO write description

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

* Changes go here

**Deletions** ⚰️

* Deletions go here

## TODO

- [ ] Add TODOs


src/components/common/Extensions/SpecificSidePanel/SpecificSidePanel.tsx
```diff
@@ -0,0 +1,56 @@
...
+import ExtensionStatusBadge from '../ExtensionStatusBadge-new/ExtensionStatusBadge';
+import Permissions from './partials/Permissions';
+import DateInstalled from './partials/DateInstalled';
+import InstaledBy from './partials/InstaledBy';
+import Version from './partials/Version';
+import ContractAddress from './partials/ContractAddress';
+import Developer from './partials/Developer';
+
+const displayName = 'common.Extensions.SpecificSidePanel';
+
```
agnieszkajarosikloj, line 14: Why are you passing an array of types here? Maybe it will be better to pass current status?

src/components/common/Extensions/SpecificSidePanel/SpecificSidePanel.tsx
```diff
@@ -0,0 +1,56 @@
...
+            <div className={styles.panelRow}>
+              <div className={styles.panelTitle}>{statusType.title}</div>
+              <div className="w-[50%] justify-start flex flex-row">
+                {types?.map((type) => (
+                  <div className=" mr-1" key={type}>
+                    <ExtensionStatusBadge mode={type} text={type} />
+                  </div>
+                ))}
+              </div>
+            </div>
```
agnieszkajarosikloj, line 35: Maybe use enum to save status types?
And why are you comparing status to type[0]? This part of the code is hard to understand.

src/components/common/Extensions/SpecificSidePanel/SpecificSidePanel.tsx
krawczykkatarzyna: you are right, shouldn't be types but statuses. It should be array because there is a case when will be more statuses.

src/components/common/Extensions/SpecificSidePanel/SpecificSidePanel.tsx
krawczykkatarzyna: I use union of strings so enum is not necessary, fixed that: types?.[0]

agnieszkajarosikloj (approved):

src/components/common/Extensions/SpecificSidePanel/consts.tsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
+
+export const sidepanelData = [
```
KatarzynaStefanowiczPagepro, line 3: `sidePanelData`

src/components/common/Extensions/SpecificSidePanel/types.ts
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { ExtensionStatusBadgeMode } from '../ExtensionStatusBadge-new/types';
+import { sidepanelData } from './consts';
+
+export interface SpecificSidePanelProps {
+  statuses: SpecificSidePanelStatus[];
+  sidepanelData: SidepanelDataProps;
```
KatarzynaStefanowiczPagepro, line 7: `sidePanelData`

PR #486, opened 2023-04-26
Integrate motion components with Create Payment dialog

### LINKED ISSUE #475: Wire in Create Payment dialog with motion components and saga
- Motion domain selector.
- Force toggle.

Image

ArmandoGraterol:
FireShot Capture 016 - Colony CDapp - localhost

Some things that deserve an explicit mention:
- If the colony/team doesn't have rep, an error message should appear and the form should be disabled except for the domain selector.
- You should be able to create a motion in the parent domain (Root) even if the targeted domain is a subdomain.
- You can now get reputation by making a payment action/motion, so no need to get it via truffle console.
- To test creating motions in a subdomain, now you should be able to create domains through the UAC.
- To get funds in a team to make a payment, you can force the Transfer Funds action through the UAC by manually changing the saga to the non-motion action type.

Related block-ingestor PR: https://github.com/JoinColony/block-ingestor/pull/66

Resolves #475

### DIFF
--- a/src/components/common/Dialogs/CreatePaymentDialog/CreatePaymentDialogForm.tsx
+++ b/src/components/common/Dialogs/CreatePaymentDialog/CreatePaymentDialogForm.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import { defineMessages, FormattedMessage } from 'react-intl';
 import { ColonyRole } from '@colony/colony-js';
 import { isConfusing } from '@colony/unicode-confusables-noascii';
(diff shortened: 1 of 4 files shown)


willm30 (approved):
LGTM

ArmandoGraterol:
> P.S. How do I get me some AGT?

I know someone that knows someone 😎 I'm going to hook you in

src/components/common/Dialogs/CreatePaymentDialog/CreatePaymentDialogForm.tsx
```diff
@@ -65,6 +65,8 @@ const MSG = defineMessages({
 interface Props extends ActionDialogProps {
   verifiedUsers: MemberUser[];
   // showWhitelistWarning: boolean;
+  handleIsForceChange: SetStateFn;
```
jakubcolony, line 68: Small detail but I like the convention where you use the `handle` prefix for handlers that are internal (defined inside) for a particular component and the `on` prefix for any external handlers passed as props (or `set`, since this is a state setter)

jakubcolony (approved):
👍

PR #487, opened 2023-04-26
Integrate motion components with Mint Tokens dialog

### LINKED ISSUE #476: Wire in Mint Tokens dialog with motion components and saga
- Motion domain selector (**Disabled** as minting can only be done in the Root domain).
- Force toggle.

Image

ArmandoGraterol:
FireShot Capture 017 - Colony CDapp - localhost
FireShot Capture 018 - Colony CDapp - localhost

Some things that deserve an explicit mention:
- If the colony doesn't have rep, an error message should appear and the form should be disabled.
- You can now get reputation by making a payment action, so no need to get it via truffle console.

Resolves #476


willm30 (approved):
LGTM

jakubcolony (approved):
Very nice

PR #492, opened 2023-04-27
Add Motion Functionality to Manage Rep Dialog

### LINKED ISSUE #479: Wire in Award/Smite dialog with motion components and saga
With the voting reputation extension enabled, the dialogs should now have:
- Motion domain selector (**Disabled** in the case of the Award Reputation dialog as you can only create those motions in Root).
- Force toggle.

How these dialogs with the extension enabled look like in the Dapp:
Image
Image

willm30:
## Description

This PR makes it possible to create an Award / Smite Rep motion via the UAC.
It also makes some small refactors to the ManageReputation Dialog components, like moving validation logic to a separate file.

smite-rep

award-rep

## Testing

Install voting rep. 
c.setArchitectureRole(1, "0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff", '<voting-rep-extn-address>', 1, true)

You should:

- Not be able to submit the form without rep. (i.e. you see the NotEnoughRep and form is disabled).

After adding rep, you should be able to award/smite rep in any combination of domain and subdomain that is possible via the ui. (i.e from the subdomain in the subdomain, from root in root, or from root in the subdomain).

**Changes** 🏗

* Wire in motion components with Award/Smite
* Display reputation penalty as positive value in ActionTitle and DetailsWidget (like in Dapp)

Resolves #479

### DIFF
--- a/src/components/common/ColonyActions/helpers/mapItemToMessageFormat.tsx
+++ b/src/components/common/ColonyActions/helpers/mapItemToMessageFormat.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import Decimal from 'decimal.js';
 
 import Numeral from '~shared/Numeral';
 import FriendlyName from '~shared/FriendlyName';
@@ -87,7 +88,7 @@ export const mapColonyActionToExpectedFormat = (
     tokenSymbol: actionData.token?.symbol,
     reputationChangeNumeral: actionData.amount && (
       <Numeral
-        value={actionData.amount}
+        value={new Decimal(actionData.amount).abs()}
         decimals={getTokenDecimalsWithFallback(colony?.nativeToken.decimals)}
       />
     ),
--- a/src/components/common/ColonyHome/ColonyDomainSelector/ColonyDomainSelector.tsx
+++ b/src/components/common/ColonyHome/ColonyDomainSelector/ColonyDomainSelector.tsx
@@ -17,7 +17,7 @@ import CreateDomainButton from './CreateDomainButton';
 import styles from './ColonyDomainSelector.css';
 
 interface FormValues {
-  filteredDomainId: string;
+  filteredDomainId: number;
 }
 
 interface Props {
@@ -87,7 +87,7 @@ const ColonyDomainSelector = ({
   return (
     <Form<FormValues>
       defaultValues={{
-        filteredDomainId: String(filteredDomainId),
+        filteredDomainId,
       }}
       onSubmit={() => {}}
     >
--- a/src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationContainer/ManageReputationContainer.tsx
+++ b/src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationContainer/ManageReputationContainer.tsx
@@ -91,7 +91,6 @@ const ManageReputationContainer = ({
           domainId: filteredDomainId || Id.RootDomain,
           user: selectedUser,
           motionDomainId: (isSmiteAction && filteredDomainId) || Id.RootDomain,
-          amount: '',
           annotation: '',
         }}
         actionType={actionType}
--- a/src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationContainer/ManageReputationContainer.tsx
+++ b/src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationContainer/ManageReputationContainer.tsx
@@ -1,8 +1,6 @@
 import React, { useState } from 'react';
-import { string, object, number, boolean, InferType } from 'yup';
 import { Id } from '@colony/colony-js';
 import { useNavigate } from 'react-router-dom';
-import { defineMessages } from 'react-intl';
 
 import Decimal from 'decimal.js';
 import Dialog from '~shared/Dialog';
@@ -19,42 +17,9 @@ import { AwardAndSmiteDialogProps } from '../types';
 
 import { getManageReputationDialogPayload } from './helpers';
 
+import { FormValues, defaultValidationSchema } from './validation';
+
 const displayName = 'common.ManageReputationContainer';
-
-const MSG = defineMessages({
-  amountZero: {
-    id: `${displayName}.amountZero`,
-    defaultMessage: 'Amount must be greater than zero',
-  },
-  maxAmount: {
-    id: `${displayName}.maxAmount`,
-    defaultMessage: "Amount must be less than the user's reputation",
-  },
-});
-
-const defaultValidationSchema = object()
-  .shape({
-    domainId: number().required(),
-    user: object().shape({
-      walletAddress: string().address().required(),
-    }),
-    amount: string()
-      .required()
-      .test(
-        'more-than-zero',
-        () => MSG.amountZero,
-        (value) => {
-          const numberWithoutCommas = (value || '0').replace(/,/g, ''); // @TODO: Remove this once the fix for FormattedInputComponent value is introduced.
-          return !new Decimal(numberWithoutCommas).isZero();
-        },
-      ),
-    annotation: string().max(4000),
-    forceAction: boolean(),
-    motionDomainId: number(),
-  })
-  .defined();
-
-type FormValues = InferType<typeof defaultValidationSchema>;
 
 const ManageReputationContainer = ({
   colony: { nativeToken },
--- a/src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationContainer/ManageReputationContainer.tsx
+++ b/src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationContainer/ManageReputationContainer.tsx
@@ -83,7 +83,6 @@ const ManageReputationContainer = ({
   // const { isWhitelistActivated } = colony;
   const selectedUser = useSelectedUser(allColonyMembers);
   //   isWhitelistActivated ? verifiedUsers : colonyWatchers,
-
   return (
     <Dialog cancel={cancel}>
       <Form<FormValues>
@@ -91,7 +90,7 @@ const ManageReputationContainer = ({
           forceAction: false,
           domainId: filteredDomainId || Id.RootDomain,
           user: selectedUser,
-          motionDomainId: Id.RootDomain,
+          motionDomainId: (isSmiteAction && filteredDomainId) || Id.RootDomain,
           amount: '',
           annotation: '',
         }}
--- a/src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationContainer/ManageReputationContainer.tsx
+++ b/src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationContainer/ManageReputationContainer.tsx
@@ -2,22 +2,28 @@ import React, { useState } from 'react';
 import { Id } from '@colony/colony-js';
 import { useNavigate } from 'react-router-dom';
 
-import Decimal from 'decimal.js';
 import Dialog from '~shared/Dialog';
 import { ActionHookForm as Form } from '~shared/Fields';
 
 import { ActionTypes } from '~redux/index';
 import { pipe, withMeta, mapPayload } from '~utils/actions';
-// import { useSelectedUser } from '~hooks';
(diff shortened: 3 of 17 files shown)


src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationContainer/helpers.ts
```diff
@@ -17,10 +19,26 @@ export const getManageReputationDialogPayload = (
...
+    userAddress: user.walletAddress,
     annotationMessage: annotation,
     amount: reputationChangeAmount.toString(),
     motionDomainId,
     isSmitingReputation: isSmiteAction,
   };
 };
+
+export const useGetColonyMembers = (colonyAddress?: Address | null) => {
+  const { data } = useGetMembersForColonyQuery({
```
willm30, line 31: There's an implicit limit here of 100... 

Really we'd have a way of fetching more if you scroll to the bottom of the user picker (useful for colonies with more than 100 members).

But I think that's a "new UI" optimisation.

src/components/common/ColonyHome/ColonyDomainSelector/ColonyDomainSelector.tsx
```diff
@@ -87,7 +87,7 @@ const ColonyDomainSelector = ({
   return (
     <Form<FormValues>
       defaultValues={{
-        filteredDomainId: String(filteredDomainId),
```
willm30, line 90: This was a bug. 

We should be storing the domain id as a number, as the domain select components are using numbers and so a === comparison wasn't returning true when it should've.

src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationContainer/ManageReputationContainer.tsx
```diff
@@ -1,64 +1,36 @@
...
-    id: `${displayName}.amountZero`,
-    defaultMessage: 'Amount must be greater than zero',
-  },
-  maxAmount: {
-    id: `${displayName}.maxAmount`,
-    defaultMessage: "Amount must be less than the user's reputation",
-  },
-});
-
-const defaultValidationSchema = object()
```
willm30, line 36: Move to own file

src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationContainer/validation.ts
```diff
@@ -0,0 +1,49 @@
+import Decimal from 'decimal.js';
+import { boolean, number, object, string } from 'yup';
+
+import { intl } from '~utils/intl';
+
+const { formatMessage } = intl({
+  amountZero: 'Amount must be greater than zero',
+  maxAmount: "Amount must be less than the user's reputation",
+});
```
willm30, line 9: These messages don't actually show up so they're a bit pointless (maybe useful for devs revisiting this in the future? I actually found it confusing because it made me think the validation was broken). I am tempted to remove and just pass empty strings into the test function.

src/components/common/ColonyActions/helpers/mapItemToMessageFormat.tsx
```diff
@@ -87,7 +88,7 @@ export const mapColonyActionToExpectedFormat = (
     tokenSymbol: actionData.token?.symbol,
     reputationChangeNumeral: actionData.amount && (
       <Numeral
-        value={actionData.amount}
+        value={new Decimal(actionData.amount).abs()}
```
willm30, line 91: This is so a reputation penalty doesn't show up as a negative value

src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationContainer/validation.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export const getAmountValidationSchema = (schemaUserReputation: number) =>
+  object()
+    .shape({
+      amount: amountValidation.test(
+        'less-than-user-reputation',
+        () => formatMessage({ id: 'maxAmount' }),
+        (value) => {
+          const numberWithoutCommas = (value || '0').replace(/,/g, ''); // @TODO: Remove this once the fix for FormattedInputComponent value is introduced.
+          return !new Decimal(numberWithoutCommas).greaterThan(
```
willm30, line 43: Did this get fixed? @ArmandoGraterol

src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationDialogForm/ManageReputationDialogForm.tsx
```diff
@@ -105,24 +107,39 @@ const ManageReputationDialogForm = ({
   schemaUserReputation,
   updateSchemaUserReputation,
   nativeTokenDecimals,
-  verifiedUsers,
+  users,
```
willm30, line 110: Won't be verified if the whitelist is deactivated

src/index.ts
```diff
@@ -2,12 +2,15 @@ import { createElement } from 'react';
...
 import ReactModal from 'react-modal';
 import { Logger } from 'ethers/lib.esm/utils';
+import Decimal from 'decimal.js';
 
 import './styles/main.css';
 
 import Entry from './Entry';
 import store from '~redux/createReduxStore';
 
+Decimal.set({ toExpPos: 78 });
```
willm30, line 12: This is taken from the dapp and stops the Decimal class from trunating large numbers with exponent notiation (up to but not incl the power of 78) https://mikemcl.github.io/decimal.js/#toExpPos

src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationContainer/validation.ts
ArmandoGraterol: Yep, it should be fixed!

ArmandoGraterol (changes requested):
Nice work! I noticed some things:

The star in the domain selector has the wrong color:
FireShot Capture 027 - Colony CDapp - localhost

I think we can disable this domain selector in the Award Reputation dialog if there's no reputation in Root, as the motions can only be created there. Well, tbh I think we could disable the domain selectors in all dialogs if there's no rep in Root if my understanding of rep doesn't fail me but that's besides this PR:
FireShot Capture 028 - Colony CDapp - localhost

Even thought all of the form is disabled, I can still click the `Max` button:
FireShot Capture 029 - Colony CDapp - localhost

I think the reputation check is only taking into account the reputation in Root, we should pass the motion domain id to `useActionDialogStatus` in the case of the smite dialog:
FireShot Capture 030 - Colony CDapp - localhost

willm30:
@ArmandoGraterol Thanks, all fixed.

src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationDialogForm/ManageReputationDialogForm.tsx
```diff
@@ -105,24 +107,40 @@ const ManageReputationDialogForm = ({
...
+    disabledInput,
+    disabledSubmit,
+    canCreateMotion,
+    canOnlyForceAction,
+  } = useActionDialogStatus(
+    colony,
+    requiredRoles,
+    [domainId],
+    enabledExtensionData,
+    domainId,
```
ArmandoGraterol, line 142: I think this needs to be `motionDomainId` instead of `domainId`

src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationDialogForm/ManageReputationDialogForm.tsx
```diff
@@ -277,12 +302,15 @@ const ManageReputationDialogForm = ({
...
-          appearance={{ marginTop: 'negative' }}
-          domainId={Number(domainId)}
-        />
-      )} */}
+
+      {canOnlyForceAction && (
+        <DialogSection>
+          <NotEnoughReputation
+            appearance={{ marginTop: 'negative' }}
+            domainId={Number(domainId)}
```
ArmandoGraterol, line 310: Needs to be `motionDomainId` like above as we are validating if the domain where the motion will be created in has reputation, not if the _targeted_ domain has

ArmandoGraterol (commented):
Good one! Some of the things were fixed but now we have this happening:
FireShot Capture 031 - Colony CDapp - localhost
FireShot Capture 032 - Colony CDapp - localhost

- The reputation validation is still taking into account `domainId` when it should take `motionDomainId` instead, otherwise we can't create a motion in Root that targets a subdomain (like `Dogs` or `Dolphins` in this case) if the subdomain doesn't have rep.
- The domain selector is disabled when the team doesn't have rep, I think it should only be disabled when there's no rep in Root, otherwise, if you happen to select a team with no rep you can't change to another one that has because the selector is disabled.

src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationContainer/validation.ts
jakubcolony: I'm up for removing any dead code

src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationContainer/validation.ts
jakubcolony: @willm30 If you want to tick this TODO off in your PR, you can copy the validation from `TransferFundsDialog/validation.ts`

jakubcolony (commented):
Haven't spotted anything other than Armando pointed out, nice work overall!

Julianahlk (commented):
Overall, works well for me except that I am also seeing the subdomain issues that @ArmandoGraterol mentioned.

willm30:
@ArmandoGraterol Third time's a charm 😉 

As an aside, this is where automated testing would come in very handy, as it would centralise domain specific knowledge so that I, having not worked at all really on these dialogs, would know if I were introducing incorrect functionality before I asked for a review. (This is doubly true given how complex some of these forms are.) We should 100% be doing this when we start the new UI.

ArmandoGraterol (approved):
Working as expected now as far as I can notice 👌 Good job

willm30:
@Julianahlk @jakubcolony Can I get some green pls 🙂

src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationDialogForm/ManageReputationDialogForm.tsx
```diff
@@ -179,7 +195,7 @@ const ManageReputationDialogForm = ({
     }
   }, [schemaUserReputation, isSmiteAction, trigger]);
 
-  const formattedData = verifiedUsers.map((user) => ({
+  const formattedData = users.map((user) => ({
     ...user,
     id: user,
```
jakubcolony, line 200: Not your changes, but is that right? User object as an id instead of just the address?

jakubcolony (approved):
All good! 👍

Julianahlk (commented):
I am unable to finalize any motions for Award reputation, I keep getting the message below even though I clearly have enough funds:
Screenshot 2023-05-03 at 1 43 54 PM
Screenshot 2023-05-03 at 2 05 17 PM

@jakubcolony and @ArmandoGraterol were you able to finalize any award motions? I was able to complete a smite motion after giving myself reputation through a forced action.

src/components/common/Dialogs/AwardAndSmiteDialogs/ManageReputationDialogForm/ManageReputationDialogForm.tsx
willm30: One for @ArmandoGraterol

ArmandoGraterol: That part of the code ain't mine :P

Nevertheless, I think it's wrong, the `id` should be the address instead of the whole object

PR #499, opened 2023-05-01
Fix colony redirect

### LINKED ISSUE #498: Fix redirect to ColonyHome
Redirect to ColonyHome from Colony creation wizard currently broken due to a race condition and changing the fetch policy of the colony query inside useColonyContext.

willm30:
## Description

Fix to the redirect to ColonyHome from Colony creation wizard. Changing the network policy in #485 has surfaced a race condition after creating a colony via the wizard. 

`cache-and-network` appears to behave differently to `network-only`, in that now, we're not getting a loading state that persists  long enough for the colony to be populated in the db. Previously, it seems we were being shown a loading state that endured just enough for the redirect to be successful (margins are small here, for the record).

Anyway, to be absolutely sure we don't redirect to 404 when there is actually the colony in the db, I've implemented the same polling strategy as at the ActionDetailsPage. 

**Changes** 🏗

* Poll for colony inside colony context. 

Resolves #498

### DIFF
--- a/src/components/common/CreateColonyWizard/StepConfirmTransactions.tsx
+++ b/src/components/common/CreateColonyWizard/StepConfirmTransactions.tsx
@@ -123,7 +123,9 @@ const StepConfirmTransactions = ({ wizardValues: { colonyName } }: Props) => {
     getGroupStatus(newestGroup) === TRANSACTION_STATUSES.SUCCEEDED &&
     getGroupKey(newestGroup) === 'group.createColony'
   ) {
-    return <Navigate to={`/colony/${colonyName}`} />;
+    return (
+      <Navigate to={`/colony/${colonyName}`} state={{ isRedirect: true }} />
+    );
   }
 
   const createColonyTxGroup = findTransactionGroupByKey(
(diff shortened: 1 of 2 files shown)


Julianahlk (approved):
Very nice fix! Finally not being re directed to the 404 page 🎉

ArmandoGraterol (approved):
LGTM 👍

willm30:
@jakubcolony The problem I'm facing is that `cache-and-network` causes the loading state to be reactivated when refetching, which means you can't do a background refetch. I thought:

```js
fetchPolicy: 'network-only',
nextFetchPolicy: 'cache-first',
```
Was the solution, but this, for some reason, doesn't fetch the colony data that is already in the database when performing a redirect from the colony creation wizard to colony home (it just returns undefined). Refetching via a mutation inside the saga doesn't solve this problem (which makes sense since at this point, we haven't run the query yet at all because we're not inside a colony route). 

Network only is a possibility, but then you don't get any caching.

So this solution is an attempt to:
1. enable refetching
2. maintain the utility of the cache
3. ensure we don't redirect to 404 when we have the available data in the db.

PR #501, opened 2023-05-01
Fix inability to claim both sides of a motion

### LINKED ISSUE #500: Fix inability to claim two sides of a motion
To reproduce: 

1. Ensure you have two users with roughly similar amounts of reputation in your colony.
2. Create a motion.
3. Stake the motion both sides with just one of the users (User A). 
4. User A votes Yay, User B votes Nay.
5. Reveal, finalize.
6. User A should be able to claim their rewards on both sides, but currently I'm getting an error.

willm30:
## Description

This pr fixes a bug I introduced when porting across the claim motion saga. I was getting an error when claiming both sides of a motion, and it was caused by passing an undefined motion id to the claimRewardWithProofs function.

## Testing

See the reproduction steps in the issue. 

**Changes** 🏗

* Fix undefined motion id

Resolves #500

### DIFF
--- a/src/redux/sagas/motions/claimMotionRewards.ts
+++ b/src/redux/sagas/motions/claimMotionRewards.ts
@@ -70,16 +70,15 @@ function* claimMotionRewards({
       userAddress,
     );
 
-    const allMotionClaimsCount =
-      motionsWithYayClaim.length + motionsWithNayClaim.length;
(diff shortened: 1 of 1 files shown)


Julianahlk (approved):
All good! I was able to claim for both sides without any errors popping. 👍

jakubcolony (approved):
Works well now 👍

ArmandoGraterol (approved):
Working as expected 👍

PR #503, opened 2023-05-01
Fix incorrect domain balance in finalize motion

### LINKED ISSUE #502: Fix incorrect domain balance in Finalize Motion
Currently, the getDomainBalance function isn't returning the correct domain balance.

willm30:
## Description

This pr fixes a bug with fetching the domain balance when we finalize a motion. I wasn't discriminating by the token address, so was getting the wrong token balance. 

**Changes** 🏗

* Find domain balance by domain **and** token address

Resolves #502


Julianahlk (approved):
Nice fix Will 😉

ArmandoGraterol (approved):
It looks to be working as expected! Tested with the Root domain and a subdomain.

Also, I found this which is unrelated to the PR but something that might be worth creating an issue for:
Screenshot 2023-05-03 at 15 33 37

This error appeared in a transfer funds motion (it crashed the page) and I had many reputation points due to me minting and paying large amounts. The culprit seems to be `BigNumber` getting a decimal number somewhere in the code which is something that it doesn't like.

jakubcolony (commented):
@willm30 There's a `getBalanceForTokenAndDomain` util (at least in `master`) which I think might be doing the same thing

willm30:
@ArmandoGraterol That should be fixed by setting the Decimal exp pos to a large number (done in another pr).

PR #505, opened 2023-05-02
Fix motion refetching without data

willm30:
## Description

I was seeing the following lamda error caused by an effect refetching motion state without checking if there's any motion data. This PR fixes it.

lamda-err

**Changes** 🏗

* Don't refetch motion state without motion data


Julianahlk (approved):
👍

ArmandoGraterol (approved):
LGTM 👍 Finalized a motion without getting errors the whole flow

PR #510, opened 2023-05-03
Update user logout to allow for user switching

### LINKED ISSUE #392: Hitting "Sign out" doesn't update msg.sender without browser refresh.

ArmandoGraterol:
Removing the colony manager's context on logout is something that we do in the Dapp and it seems like it would be required for us to do here as well in order to avoid using a manager instance with an outdated signer.

You should be able to go through the whole motion flow without having to reload the page when you switch users. I haven't noticed any drawbacks/side effects from us removing this context on logout.

Resolves #392

### DIFF
--- a/src/redux/sagas/users/index.ts
+++ b/src/redux/sagas/users/index.ts
@@ -150,6 +150,7 @@ function* usernameCreate({
 
 function* userLogout() {
   try {
+    removeContext(ContextModule.ColonyManager);
     removeContext(ContextModule.Wallet);
     clearLastWallet();
 


willm30 (approved):
Very nice, the Colony Manager takes the signer from the wallet that instantiated it so makes sense to create a new one on each login

jakubcolony (approved):
thumbs up from me

PR #514, opened 2023-05-04
Refetch colony in `editDomain` and `createDomain` sagas

### LINKED ISSUE #511: Refetch colony in sagas with mutations
If you complete an action, do we need to refetch the colony from the action details page or does the polling update the colony object?

jakubcolony:
## Description

Following Will's suggestion, I have branched this PR off motions. It adds refetching for Edit Domain, Create Domain and Manage Whitelist action so that the updated data is visible in the UI without manual refresh.

## Changes:
- Refetch colony query in `editDomain` saga
- Add colony query refetch in `useGetColonyAction` hook when the action data loads - this is relevant for the Create Domain action, since the creation happens in block-ingestor and we won't be able to retrieve the newly created domain until the action loads
- Added `isRedirect` state config to `createDomain` and `manageVerifiedRecipients` sagas to enable action polling for these

## Testing
Verify that the created/edited domain is visible in the UI without refreshing the page (but wait until the action has finished loading).

Resolves #511

### DIFF
--- a/src/redux/sagas/actions/createDomain.ts
+++ b/src/redux/sagas/actions/createDomain.ts
@@ -157,7 +157,9 @@ function* createDomainAction({
     });
 
     if (navigate) {
-      navigate(`/colony/${colony.name}/tx/${transactionHash}`);
+      navigate(`/colony/${colony.name}/tx/${transactionHash}`, {
+        state: { isRedirect: true },
+      });
     }
   } catch (error) {
     return yield putError(ActionTypes.ACTION_DOMAIN_CREATE_ERROR, error, meta);
--- a/src/redux/sagas/actions/manageVerifiedRecipients.ts
+++ b/src/redux/sagas/actions/manageVerifiedRecipients.ts
@@ -194,7 +194,11 @@ function* manageVerifiedRecipients({
     });
 
     if (colony.name && navigate) {
-      yield navigate(`/colony/${colony.name}/tx/${txHash}`);
+      yield navigate(`/colony/${colony.name}/tx/${txHash}`, {
+        state: {
+          isRedirect: true,
+        },
+      });
     }
   } catch (error) {
     return yield putError(


willm30:
> I realised the createDomain refetching only works (ie. the newly created domain shows up without refreshing) if the race condition

We should therefore handle the refetch on the action details page in this case?

jakubcolony:
Alternatively we could go down the route of optimistic UI update, similarly to extensions UI. It only needs to be done for the Create Domain action. All of the others update rather than create data so they could stick with the neat `refetchQueries` option. Thoughts?

jakubcolony:
Converting to draft for now as I need to implement one of the strategies discussed above

willm30:
@jakubcolony Not sure how straightforward optimistic ui would be in this case? You'd need to get the data of the newly created domain to all the domain selectors across the app, which are currently accessing domain data directly from the db. 

My first thought is that calling refetch via colony context is simpler.

jakubcolony:
@willm30 Updating cache manually proved rather tricky so I went ahead with refetching the colony on action details page. The problem is that it won't work until action polling is on the same branch. I added a relevant comment and will revisit after the merge

src/components/common/ColonyActions/ActionDetailsPage/ActionDetailsPage.tsx
```diff
@@ -46,14 +46,17 @@ const ActionDetailsPage = () => {
...
   const events = ['event']; // to be taken from real data
 
-  const { data: colonyData, loading: loadingColony } =
-    useGetFullColonyByAddressQuery({
-      variables: {
-        address: action?.colonyAddress ?? '',
-      },
-      skip: !action?.colonyAddress,
-    });
-  const txColony = colonyData?.getColonyByAddress?.items[0];
```
jakubcolony, line 56: I think this query was redundant, we can check whether the action belongs to the current colony using the object from colony context. It made the details page load much faster

src/context/ColonyContext.tsx
```diff
@@ -1,6 +1,7 @@
-import React, { createContext, useMemo, ReactNode } from 'react';
```
jakubcolony, line 1: Bringing @willm30's changes from #257

Julianahlk (approved):
Overall, looks good. Just one question about the expected behavior. After editing a domain, when I open the domain selector dropdown in ColonyHome, I sometimes see the changed domain redrawing immediately after opening the dropdown. Is this expected?

jakubcolony:
@Julianahlk What do you mean by redrawing? Are you able to post a video by any chance?

Julianahlk:
@jakubcolony  Here is how it looks, this is may not be very important since it usually happens when viewing the dropdown quickly after editing the domain.

edit domain

jakubcolony:
Thanks @Julianahlk, I would say this is expected yeah, there's some delay before we get the updated domain data. We could make the refetch blocking but I'd say the probability of people jumping from the action details page straightaway is pretty low

ArmandoGraterol (commented):
Working great 👍 

I've been trying to create a colony with this branch and I haven't been able to due to this error:
Screenshot 2023-05-16 at 20 11 51

I was able to test the domain sagas because I used the colonies created by the script, but otherwise, I can't create a colony the normal way

willm30:
If this depends on motions, shouldn't we branch it off motions?

jakubcolony:
@willm30 Good suggestion, done 

@ArmandoGraterol @Julianahlk Would you mind re-reviewing when you get a chance please?

Julianahlk (approved):
Still works well after changing the base to port/colony-motions 👍

ArmandoGraterol (approved):
Solid work! I can create colonies and the domains are being updated correctly 👌

Issue #423, opened 2023-04-18
Refactor motion messages to be a model


willm30:
See comment: https://github.com/JoinColony/colonyCDapp/pull/369#pullrequestreview-1388873476

We'd like to make the Motion Messages model a `@model` in the schema so that amplify will attach it to the user model for us.

PR #520, opened 2023-05-06
Convert MotionMessages to a Model

Julianahlk:
## Description

- Makes MotionMessage a mode
- Makes initiatorAddress required
- Removes the extra query, useGetUserByAddressQuery as the initiator user is already fetched with the model

## Test

Use [block-ingestor#72](https://github.com/JoinColony/block-ingestor/pull/72) for testing as this adds the required values for initiatorAddress

Contributes to #423

### DIFF
--- a/amplify/backend/function/fetchMotionState/src/utils.js
+++ b/amplify/backend/function/fetchMotionState/src/utils.js
@@ -1,4 +1,4 @@
-const { providers, BigNumber } = require('ethers');
+const { providers, BigNumber, constants } = require('ethers');
 const {
   getColonyNetworkClient,
   Network,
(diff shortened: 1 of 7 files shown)


amplify/backend/function/fetchMotionState/src/utils.js
```diff
@@ -218,6 +218,7 @@ const updateMotionMessagesInDB = async (
 
   motionMessages.forEach((message) => {
     updatedMessages.push({
+      initiatorAddress: constants.AddressZero,
```
Julianahlk, line 221: Since SystemMessages do not have an initiator when saved to the db during fetchMotionState, AddressZero is used as a placeholder for initiatorAddress.

willm30 (changes requested):
The change to the type of `initiatorAddress` is triggering errors in the block ingestor:

err-bi

Which I believe is the source of something I'm noticing on this branch, which is that the loading state is hanging after staking both sides of a motion 100%. The voting message also doesn't appear after refreshing the browser.

It's probably worth syncing with @ArmandoGraterol also, since you're both working on the motion schema atm. Perhaps base this branch off of his or vice versa?

ArmandoGraterol (approved):
LGTM 👍

PR #525, opened 2023-05-08
Fix gas station message

### LINKED ISSUE #515: Fix gas station "salt entropy" msg
When revealing and voting, we include the following message in the gas station: 

`Sign this message to generate 'salt' entropy. Extension Address...`

See the respective sagas for full message.

In the gas station, it doesn't display correctly. It's too long. This issue is to wrap this text or whatever is required so that it displays correctly.

willm30:
## Description

This PR fixes a problem with the gas station message content not being wrapped correctly. It also fixes a mistake with the message title for the vote motion message, which was using the reveal vote message.

In the gas station, you should now see when voting / revealing:
entropy-msg

reveal-motion

Resolves #515


ArmandoGraterol (approved):
👌

Julianahlk (approved):
👍 Looks good!

PR #526, opened 2023-05-08
Refactor claim motion saga

### LINKED ISSUE #374: Separate claimMotionReward saga into two
Split the claimMotionReward saga into
- claimMotionReward that only claims one motion
- and claimAllMotionRewards that claims all motions passed to it (like how it functions currently).

So really it's just a case of implementing the first one.

## Original Comment
 I also think there's a claim to be made for splitting this saga into two: one for claiming multiple motion rewards at once, and one for just claiming one. This latter use case is far more common, and the logic to implement would be simpler since we'd have the transaction hash at the time of claiming, and could get the exact motion from the db instead of filtering. I will create an issue to handle separately if no objections.

_Originally posted by @willm30 in https://github.com/JoinColony/colonyCDapp/pull/324#discussion_r1155786672_

willm30:
## Description

This PR creates a separate saga to claim motion rewards in the event we're only claiming for a single motion.

## Testing

Go through the motion flow and ensure you can still claim motion rewards correctly.  I have tested with one user voting, and again with two users.

Cannot test "claim all" in the stakes tab since it's yet to be wired in (but, this hasn't changed -- only it's now pointing to the `claimAllMotionRewards` saga, which is the same as before).

**New stuff** ✨

* `claimMotionRewards` saga, as distinct from `claimAllMotionRewards` saga

Resolves #374


src/redux/sagas/motions/claimAllMotionRewards.ts
willm30, line 1: This has just been renamed

ArmandoGraterol (approved):
Working as expected 👍

Julianahlk (approved):
Nice job! Works as expected for me.

PR #527, opened 2023-05-08
Add ColonyMotion model

### LINKED ISSUE #356: Separate Colony Motion into own model
This is so we can query a motion id directly instead of requesting all motions from the db and client-side filtering (in claim motion rewards saga, for instance, and we also do this in the block ingestor).

While you are at it, rename motionDomainId to motionDomainNativeId

ArmandoGraterol:
Separated motion data from `ColonyAction` by creating a new model that is linked to the action.

I've tested the motion flow with several motions and updated the places where we were accessing motion data indirectly (by querying the action to get the data). Make sure to double-check that the flow is working as expected and that every place that needed updating, has been. 

Apart from moving `motionData` to their own model, other notable changes are the renaming of `motionDomainId` to `nativeMotionDomainId` in the DB, and the update to the `MotionMessage` model to allow them to get linked to the corresponding motion (These changes were done because you can't create/update motion messages from within the create/update motion mutations due to both data structures being `models` now).

`block-ingestor`'s counterpart PR: https://github.com/JoinColony/block-ingestor/pull/70

Resolves #356


amplify/backend/function/fetchMotionState/src/index.js
willm30, line 1: Are the changes in this lamda out of necessity or to make consistent how we fetch motion data? Just curious

src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/VoteOutcome/VoteResults/VoteResults.tsx
```diff
@@ -27,8 +27,8 @@ const MSG = defineMessages({
 });
 
 interface VoteResultsProps {
-  revealedVotes: MotionData['revealedVotes'];
-  voterRecord: MotionData['voterRecord'];
+  revealedVotes: ColonyMotion['revealedVotes'];
```
willm30, line 30: Could use the underlying types here also

amplify/backend/function/fetchMotionState/src/index.js
```diff
@@ -10,11 +10,12 @@ const {
...
 exports.handler = async (event) => {
-  const { colonyAddress, transactionHash } = event.arguments?.input || {};
+  const { colonyAddress, databaseMotionId, transactionHash } =
+    event.arguments?.input || {};
   /* Get latest motion state from chain */
-  const motionData = await getMotionData(transactionHash);
+  const motionData = await getMotionData(databaseMotionId);
 
-  if (transactionHash && motionData) {
+  if (databaseMotionId && transactionHash && motionData) {
```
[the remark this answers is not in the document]
ArmandoGraterol: True, thank you for catching this

amplify/backend/function/fetchMotionState/src/index.js
```diff
@@ -10,11 +10,12 @@ const {
  * @type {import('@types/aws-lambda').APIGatewayProxyHandler}
  */
 exports.handler = async (event) => {
-  const { colonyAddress, transactionHash } = event.arguments?.input || {};
+  const { colonyAddress, databaseMotionId, transactionHash } =
```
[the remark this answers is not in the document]
ArmandoGraterol: If we use the motion id instead of the transaction hash for the messages then sure it's no longer needed

amplify/backend/function/fetchMotionState/src/index.js
ArmandoGraterol: Mostly for consistency, it seems weird to me that we would update the colony motion through the colony action (and that we fetch the even though we only need the motion data of course) now that we have a motion model like we were doing in this function

amplify/backend/function/fetchMotionState/src/utils.js
```diff
@@ -223,13 +216,10 @@ const updateMotionMessagesInDB = async (
     });
```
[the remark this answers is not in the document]
willm30: I was thinking databse

ArmandoGraterol:
@willm30 Updated the claim reward saga as you suggested! Let me know if that's what you had in mind

willm30 (approved):
LGTM 👍

src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/VoteOutcome/VoteResults/VoteResults.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import { defineMessages } from 'react-intl';
 
-import { MotionData } from '~gql';
+import { MotionStakes, VoterRecord } from '~gql';
```
jakubcolony, line 4: Could these be either derived from `MotionData` or have fragments defined? Right now they describe the object type as defined in the schema rather than what we're querying for. If there's a new field in the schema, it will be included in this type regardless of whether it's returned in the query.

src/components/common/ColonyActions/CountDownTimer/CountDownTimer.tsx
```diff
@@ -36,7 +36,7 @@ interface Props {
   motionState: MotionState;
   refetchMotionState: RefetchMotionState;
   motionId: string;
-  motionStakes: MotionData['motionStakes'];
```
jakubcolony, line 39: Related to my previous comment, by keeping these dependant on `MotionData` fragment we were sure we're working with the data available from the query

jakubcolony (approved):
Looks good! 👍

src/components/common/ColonyActions/CountDownTimer/CountDownTimer.tsx
willm30: This was on my suggestion based on the general pattern we're following of not accessing nested types.

src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/VoteOutcome/VoteResults/VoteResults.tsx
jakubcolony: For ongoing development, yes. However, I can see the potential this has for giving us headache when refactoring. This was one of the reasons for moving away from Amplify OOTB codegen which would generate types for all the fields as opposed to what we query for

ArmandoGraterol: I'll see if we can use fragments instead then

ArmandoGraterol:
@jakubcolony @willm30 Can you guys re-review this?

I've introduced some changes to the `MotionMessage` model as according to amplify's docs and errors I was getting, you can't create/update a model and at the same time create/update the "parent" model like we were doing before

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -546,9 +547,16 @@ type MotionMessage @model {
   initiatorAddress: ID!
   name: String!
   messageKey: String!
+  motionId: ID!
```
jakubcolony, line 550: Is that the database or native motion id?

src/redux/sagas/motions/claimAllMotionRewards.ts
```diff
@@ -41,32 +40,36 @@ function* claimAllMotionRewards({
...
     const databaseMotionIds = motionIds.map((nativeMotionId) =>
       getMotionDatabaseId(chainId, votingClient.address, nativeMotionId),
     );
 
+    const motions: ColonyMotion[] = [];
+
+    for (const databaseMotionId of databaseMotionIds) {
+      const {
+        data: { getColonyMotion },
+      }: ApolloQueryResult<GetColonyMotionQuery> = yield apolloClient.query<
```
jakubcolony, line 52: I don't think `ApolloQueryResult` is necessary here - what you put as the first generic should already provide typing for the response

jakubcolony (approved):
👍

Julianahlk (approved):
LGTM 👍

src/redux/sagas/motions/claimAllMotionRewards.ts
willm30: We can't infer the type of a value injected into the generator. Although we trust redux-saga, it could be anything. That's why it won't infer the return value of the expression that gets evaluated by `yield`.

jakubcolony: Good point 👍

willm30 (commented):
My claims aren't showing if I object to a motion and then let it fail without going to a vote. 

In https://github.com/JoinColony/colonyCDapp/pull/591 I add a fix so that they show up without needing to refresh the browser a second time, but I'm not seeing them at all on this branch.

Screenshot (835)

~~Also, not sure if it's this branch but I'm not seeing the team in the actions list for forced mint tokens (and potentially all root?) actions~~ Not this branch. I think this was actually fixed on qa recently?

amplify/backend/function/fetchMotionState/src/utils.js
```diff
@@ -194,41 +194,49 @@ const updateStakerRewardsInDB = async (
...
     [flag]: true,
   };
 
   motionMessages.forEach((message) => {
-    updatedMessages.push({
-      initiatorAddress: constants.AddressZero,
-      name: message,
-      messageKey: `${transactionHash}_${message}`,
+    const messageKey = `${motionData.id}_${message}`;
+    if (
```
willm30, line 215: A set would be more efficient here:

``` js
const messageKeys = new Set(messages.items.map((m) => m.messageKey));

motionMessages.forEach((message) => {
  const messageKey = `${motionData.id}_${message}`;
  if(!messageKeys.has(messageKey)) {
  ...
  }
}

```

amplify/backend/function/fetchMotionState/src/utils.js
```diff
@@ -194,41 +194,49 @@ const updateStakerRewardsInDB = async (
...
+      updatedMessages.push({
+        initiatorAddress: constants.AddressZero,
+        name: message,
+        messageKey,
+        motionId: motionData.id,
+      });
+    }
+  });
+
+  const messagePromises = updatedMessages.map((message) => {
```
willm30, line 229: You could probably combine this with the loop above if you use `reduce` on motion messages instead for each. Then await all the promises after.

``` js
const updatedMessages = motionMessages.reduce(() => {})
await Promise.all(updatedMessages)

```

ArmandoGraterol: Sounds good to me!

amplify/backend/api/colonycdapp/schema.graphql
ArmandoGraterol: It's the database motion id

ArmandoGraterol:
@willm30

> My claims aren't showing if I object to a motion and then let it fail without going to a vote.

 Fixed

> I think this was actually fixed on qa recently?

That's correct! That was a fix inside `block-ingestor` and it should be on `master` and only deployed in QA

willm30:
LGTM 👍

PR #553, opened 2023-05-15
Wire Manage Token Motions Dialog & Edit Colony Saga

### LINKED ISSUE #484: Wire in Manage Tokens dialog with motion components and saga
- Motion domain selector (**Disabled** as you can only create this motion in Root).
- Force toggle.

Image

willm30:
## Description

This PR wires in the Manage Tokens dialog with motions, so you can create and finalize a Manage Tokens motion.
It also fixes a bug in the dialog, which previously permitted you to add a token that already existed in the colony.

tokensmotio

## Testing

See dialog specific requirements in #484.

As for functionality, you should be able to initiate a motion to add and/or remove tokens, and have the colony state updated once it is successfully finalized. 

You should also be able to make multiple motions that touch the same state (e.g. two motions that add or remove the same token), and expect it to behave logically (i.e. you can't add the same token twice and removing a removed token does not mean it gets added again).  I've left additional comments in the block ingestor pr on this: https://github.com/JoinColony/block-ingestor/pull/74

**Changes** 🏗

* Wire up ManageTokens dialog with motion components
* Wire up `editColonyMotion` saga

Resolves #484


src/components/common/Dialogs/TokenManagementDialog/TokenManagementDialog.tsx
```diff
@@ -120,20 +123,14 @@ const TokenManagementDialog = ({
         onSuccess={handleSuccess}
         onError={handleError}
       >
-        {({ watch }) => {
-          const forceActionValue = watch('forceAction');
-          if (forceActionValue !== isForce) {
-            setIsForce(forceActionValue);
```
willm30, line 126: Moving this inside the form component gets rid of a set state error

src/redux/sagas/motions/editColonyMotion.ts
```diff
@@ -0,0 +1,265 @@
...
+
+    const haveTokensChanged = !!(
+      modifiedTokenAddresses.added.length ||
+      modifiedTokenAddresses.removed.length
+    );
+
+    /*
+     * Save the pending colony metadata in the database
+     */
+    if (colony.metadata) {
```
willm30, line 193: Here we upload the intented changes to the metadata state ("pending" state) to the db, so it can be accessed in the block ingestor when the motion is finalized.

src/redux/sagas/utils/updateColonyTokens.ts
```diff
@@ -39,6 +49,49 @@ export const getModifiedTokenAddresses = (
   return modifiedTokenAddress;
 };
 
+export const getPendingModifiedTokenAddresses = (
```
willm30, line 52: The requirements here are different from the action.

Because the action completes synchronously, you can infer whether an address was added or removed by looking at whether it's currently present among the colony's tokens.

Because motions complete asynchrously, we need to be explicit about this.

src/components/common/Dialogs/TokenManagementDialog/validation.ts
```diff
@@ -0,0 +1,76 @@
...
+    forceAction: boolean().defined(),
+    tokenAddress: string()
+      .notRequired()
+      .test(
+        'is-address',
+        () => MSG.invalidAddress,
+        (value) => !value || isAddress(value),
+      )
+      .test(
+        'is-duplicate',
```
willm30, line 41: I've added this test for duplicate token

src/utils/colonyActions.ts
```diff
@@ -550,17 +551,31 @@ export const formatRolesTitle = (roles: ActionUserRoles[]) => {
...
 
+const getChangelogItem = (
+  {
+    isMotion: actionIsMotion,
+    transactionHash,
+    pendingColonyMetadata,
+  }: ColonyAction,
+  colonyMetadata?: ColonyMetadata | null,
+) => {
+  const metadataObject = actionIsMotion
```
willm30, line 562: colonyMetadata won't yet reflect the change if it's a motion, so we look at the pending metadata instead

ArmandoGraterol (approved):
Looks good 👍 I've noticed that the voting widget shows the action as "Colony Edit" instead of "Update Tokens":
FireShot Capture 054 - Colony CDapp - localhost
FireShot Capture 055 - Colony CDapp - localhost

Also, this may not be related to this PR but it's the first time I've encountered it:
FireShot Capture 056 - Colony CDapp - localhost

I've reproduced this in 2 motions that passed and the messages persisted after reloading. Afterward, I passed more motions but I didn't encounter this again

willm30:
> I've noticed that the voting widget shows the action as "Colony Edit" instead of "Update Tokens":

Thanks, I've updated across all voting widget components.

fixed
fixed2

>Also, this may not be related to this PR but it's the first time I've encountered it:

I have also not seen this before since that feature was merged onto the main branch. If we can persistly reproduce it, I'll take a deeper look.

Julianahlk (approved):
Overall, looks good and is working well. Two things I noticed.

1. The tokens shown in ColonyHome are not updated until a page refresh. In this example, I removed token B but it still appears here until the page is refreshed.
Screenshot 2023-05-17 at 2 05 20 PM

2. This isn't really about this PR but just an observation that if a user was actually voting on a real motion, there is no information displayed about which tokens are being added or removed just "Colony tokens were updated", so it would be difficult to make a decision for how to vote. For example, If I create a motion to remove Token B and add Token A, there is no way to know what is changing. Of course, in the future annotations could provide this information, however, a user could forget to add any details about what is changing.

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -711,4 +717,7 @@ type ColonyAction @model {
   pendingDomainMetadataId: ID
   pendingDomainMetadata: DomainMetadata
     @hasOne(fields: ["pendingDomainMetadataId"])
+  pendingColonyMetadataId: ID
+  pendingColonyMetadata: ColonyMetadata
+    @hasOne(fields: ["pendingColonyMetadataId"])
```
jakubcolony, line 722: Will this not lead to the same problem as we're currently facing with pending domain metadata, where the values can get overwritten?

src/redux/sagas/utils/updateColonyTokens.ts
```diff
@@ -12,8 +12,18 @@ import {
   GetTokenFromEverywhereQueryVariables,
 } from '~gql';
 import { Colony } from '~types';
+import { notNull } from '~utils/arrays';
 import { xor } from '~utils/lodash';
 
+/**
+ 
+ */
```
jakubcolony, line 20: Empty comment

jakubcolony (approved):
Great work 👍

willm30:
@Julianahlk 
> The tokens shown in ColonyHome are not updated until a page refresh. In this example, I removed token B but it still appears here until the page is refreshed.

This might be because you're leaving the action details page before the db has been updated? If you wait until staker rewards come through, everything's up to date.

> This isn't really about this PR

One to think about for new ui.

ArmandoGraterol:
> Was there anything in particular about these two motions I should be aware of? E.g. you were switching between users or anything like that?

Unfortunately no, I didn't do anything special.

Both motions were created, staked, voted on, and finalized by 1 and the same user so no switching between wallets. Something that also happened when the votes were revealed was that the voting widgets were not displaying the vote of the user (it showed 0% for both sides) along with the conflicting messages above, but after a refresh the voting widget it showed the correct vote result (can't say the same for the messages tho).

PR #563, opened 2023-05-17
Integrate Colony Edit dialog with motion components

### LINKED ISSUE #433: Wire in Edit Colony motion

willm30:
## Description

This PR wires up the colony edit form with motion components.

colonyedit

## Testing

Create a motion to edit your colony. Avatar / displayname should be updated upon finalizing the motion. Try removing the avatar as well.

Also note that finalizing a motion to change the avatar only won't affect the colony's display name, even if the display name changed after the motion was created (and vice versa). 

For dialog specs see #433

**Changes** 🏗

* Adding motion components to colony edit form.

Resolves #433

### DIFF
--- a/src/components/common/Dialogs/EditColonyDetailsDialog/ColonyAvatarUploader.tsx
+++ b/src/components/common/Dialogs/EditColonyDetailsDialog/ColonyAvatarUploader.tsx
@@ -4,7 +4,7 @@ import { useFormContext } from 'react-hook-form';
 import { defineMessages } from 'react-intl';
 
 import {
-  getOptimisedAvatar,
+  getOptimisedAvatarUnder300KB,
   getOptimisedThumbnail,
 } from '~images/optimisation';
 import Avatar from '~shared/Avatar';
@@ -54,7 +54,7 @@ const ColonyAvatarUploader = ({
     setAvatarFileError(undefined);
 
     try {
-      const optimisedImage = await getOptimisedAvatar(file.file);
+      const optimisedImage = await getOptimisedAvatarUnder300KB(file.file);
       const optimisedThumbnail = await getOptimisedThumbnail(file.file);
       setValue('colonyAvatarImage', optimisedImage, { shouldDirty: true });
       setValue('colonyThumbnail', optimisedThumbnail);
--- a/src/components/common/Dialogs/EditColonyDetailsDialog/ColonyAvatarUploader.tsx
+++ b/src/components/common/Dialogs/EditColonyDetailsDialog/ColonyAvatarUploader.tsx
@@ -1,0 +1,95 @@
+import React from 'react';
+import { FileRejection } from 'react-dropzone';
+import { useFormContext } from 'react-hook-form';
+import { defineMessages } from 'react-intl';
+
+import {
+  getOptimisedAvatar,
+  getOptimisedThumbnail,
+} from '~images/optimisation';
+import Avatar from '~shared/Avatar';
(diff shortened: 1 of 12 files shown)


ArmandoGraterol (approved):
Working as expected as fas as I can see 👍 

Also, good news! I've managed to get the bug I mentioned in the other PR on video:
[690f3343-1c55-42d3-8794-3000d110e47a.webm](https://github.com/JoinColony/colonyCDapp/assets/18473896/088abc31-5f8e-4457-a164-6446b3ebdb99)

I've replicated like 4 times while testing this PR, doing the same things that I did in the video and I think the key to getting it is going as fast as you can through the whole flow, and even then, it doesn't always happen

willm30:
@ArmandoGraterol Good spot, I have opened an issue: https://github.com/JoinColony/colonyCDapp/issues/569

Julianahlk (changes requested):
Overall working well except for an issue with the image sizes.

I attempted to change the colony profile picture to a png with a size of 766 KB and after pressing the create motion button I received the following error in the console and meanwhile the loading spinner remained spinning in the button.

So far what I noticed with the set of images I tried is that it works with pngs larger than 1 MB and fails with pngs smaller than 1 MB.

Screenshot 2023-05-22 at 10 07 16 PM

willm30:
@Julianahlk I can look at how we handle the error but I can't get round our db's memory limit.

willm30:
@Julianahlk Can you try again? You now won't able to upload files greater than 1mb, which was causing the saga to fail.

jakubcolony (approved):
Nice work, the only thing that could be improved is adding the ability to clear the avatar uploader after an error has been shown (say you wanted to no longer upload an avatar after it turned out to be too large). Feel free to ignore this though, I can't imagine the impact being huge

image

Julianahlk:
> @Julianahlk Can you try again? You now won't able to upload files greater than 1mb, which was causing the saga to fail.

I tried your latest changes for, https://github.com/JoinColony/colonyCDapp/pull/563 and I am still getting the error in the console: 'item size has exceeded the maximum allowed size'
When I use the image below which is less than a MB, it is accepted but then the error happens when pressing the button to confirm the change.

willm30:
@Julianahlk Interesting. So what's happening here is our compression function is producing an avatar of size ~330kb, and the dynamodb size limit for a single item is 400kb. So, when combined with the other metadata data, we're going over that 400kb limit.

I've mitigated this by ensuring the size of the compressed avatar is under 300kb, which should leave enough room for other data.  

That I think is good enough for this pr. 

I've created an issue for your broader point, however, which is that the UAC dialogs don't currently implement `onError`, so we have no way of responding to errors occuring inside the saga. See: https://github.com/JoinColony/colonyCDapp/issues/580

Julianahlk:
@willm30 The image upload for the colony avatar works great now! 

I noticed one other potential issue. Its possible to enter only spaces when changing the colony name so that the colony name appears as in this image. This could be confusing, also when creating a colony its not possible to enter a name with only spaces.

Screenshot 2023-05-24 at 12 07 12 PM

willm30:
@Julianahlk 

> I noticed one other potential issue. Its possible to enter only spaces when changing the colony name so that the colony name appears as in this image.

I think this is true for all forms (i.e. ability to enter whitespace and have the form validate) so can we handle this separately?

Julianahlk (approved):
Sure all good then 🙌

PR #570, opened 2023-05-19
Mitigate reveal state race condition (motions)

### LINKED ISSUE #569: Fix duplicate voting message bug in motions
The voting passed / failed messages is getting duplicated sometimes, and it appears to have something to do with how quickly you zip through the ui. This issue is to investigate and fix.

## Original Comment

Also, good news! I've managed to get the bug I mentioned in the other PR on video:
[690f3343-1c55-42d3-8794-3000d110e47a.webm](https://github.com/JoinColony/colonyCDapp/assets/18473896/088abc31-5f8e-4457-a164-6446b3ebdb99)

I've replicated it like 4 times while testing this PR, doing the same things that I did in the video and I think the key to getting it is going as fast as you can through the whole flow, and even then, it doesn't always happen

_Originally posted by @ArmandoGraterol in https://github.com/JoinColony/colonyCDapp/pull/563#pullrequestreview-1433231227_
            

## Notes

The logic in the lamda currently permits vote outcomes motion messages to be posted twice, as in the video above. We should change this so we only ever post the vote outcome once.

willm30:
## Description

Per #569, very occasionally the fetchMotionState lamda gets called after the motion state becomes finalizable, but before the revealed votes have been populated to the db. This was having the effect of the wrong vote outcome being shown.

I have mitigated this by checking that the votes are indeed present as we expect before adding messages to the motion message feed. 

The root cause is still not certain. I have not been able to reproduce at all. I thought it might be due to the refetch occuring in the motion timer, but @ArmandoGraterol has not been able to verify this. In any case, if it happens again now, the user will be able to reload the browser, which will fetch the latest data and display everything as expected. 

## Test

To test, just run through the motion lifecycle and confirm the reveal to finalize phase works as expected. 

**Changes** 🏗

* Add checks for revealed votes

Resolves #569

### DIFF
--- a/amplify/backend/function/fetchMotionState/src/index.js
+++ b/amplify/backend/function/fetchMotionState/src/index.js
@@ -29,7 +29,7 @@ exports.handler = async (event) => {
         },
       } = motionData;
 
-      // Motion did not go to a vote
+      // Motion failed without going to a vote
       if (yayPercent !== '100') {
         await updateStakerRewardsInDB(
           colonyAddress,
@@ -48,8 +48,6 @@ exports.handler = async (event) => {
       }
     }
 
-    const didPass = didMotionPass(motionData);
-
     if (
       motionState === MotionState.Reveal &&
       !motionStateHistory.inRevealPhase
@@ -66,6 +64,18 @@ exports.handler = async (event) => {
       motionState === MotionState.Finalizable ||
       motionState === MotionState.Finalized
     ) {
+      if (
+        // If the motion is finalizable, and we have voted, we expect the revealed votes to be populated to the db.
+        // If they have not yet been added, don't update the messages, else we'll show an incorrect vote outcome.
+        motionStateHistory.hasVoted &&
+        motionData.revealedVotes.raw.yay === '0' &&
+        motionData.revealedVotes.raw.nay === '0'
+      ) {
+        return motionState;
+      }
+
+      const didPass = didMotionPass(motionData);
+
       // Check if the motion passed and the messages have not already been stored in the db
       if (didPass && !motionStateHistory.hasPassed) {
         const newMessages = [];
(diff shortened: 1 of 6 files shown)


ArmandoGraterol (commented):
I've tested with several motions to create/edit domains and mint tokens and so far the only thing I've noticed is that sometimes the phase gets stuck like this:
FireShot Capture 063 - Colony CDapp - localhost

This is after revealing my vote, nothing changes. It's not happening every time or anything like that, but it's worth noting.

willm30:
@ArmandoGraterol 

Follow-up questions:

1. When that happens, are we still polling? (you can see in network tab)
2. If you refresh the browser, does it move to finalized?
3. Are you sure the `refetch` in the motion timer doesn't have any role to play here? I notice in the screenshot that the timer value is 0, which would cause that refetch to run independently of the widget, which may well explain why it's being called before we have the revealed votes data available but after the on-chain state has changed.

If the answer to 1 is "No", can you remove this line (34) from `useRevealWidgetUpdate` and see if you can still reproduce?

```
  useEffect(() => stopPollingAction, [stopPollingAction]);
```

This stops the polling when the reveal widget dismounts. I want to know if this problem is resolved by keeping the polling going until at least the revealed vote data is in the db. 

If the answer to 2 is "Yes", I feel like this is a reasonable compromise for the problem in the event we make no further progress. It's certainly preferable to the previous behaviour.

willm30:
@ArmandoGraterol 

If it's the case that removing that `useEffect` improves the situation, it implies that the RevealWidget is being dismounted and then remounted again (which is why, when it's present, we stop polling but don't move to the FinalizedWidget). It'd be great to know what's causing that remounting.

That said, I feel like this is a good enough solution. Worst case, user needs to refresh. I've removed the effect which should ensure the polling runs until the db gets updated. And we've mitigated the issue with duplicate messages. 

So yeah, I'll ask for a team review if you can approve this

ArmandoGraterol (approved):
This will do 👍

Julianahlk (approved):
LGTM 👍

PR #577, opened 2023-05-22
Add `useUserByNameOrAddress` hook

### LINKED ISSUE #512: Allow accessing the user profile by address
This feature got removed as part of optimisations in #405, the issue is to bring it back while only calling one query.

Julianahlk:
## Description

* Adds userUserByNameOrAddress hook that only makes one query depending if the input is a username or an address

## Testing

Make sure a user can be loaded by both username or address in the route:

`/user/0x9dF24e73f40b2a911Eb254A8825103723E13209C`
`/user/a`

Resolves #512

### DIFF
--- a/src/components/common/UserProfile/UserProfile.tsx
+++ b/src/components/common/UserProfile/UserProfile.tsx
@@ -3,7 +3,7 @@ import { useParams } from 'react-router-dom';
 
 import ProfileTemplate from '~frame/ProfileTemplate';
 import NotFoundRoute from '~routes/NotFoundRoute';
-import { useUserByName } from '~hooks';
+import { useUserByNameOrAddress } from '~hooks';
 
 import UserMeta from './UserMeta';
 import UserProfileSpinner from './UserProfileSpinner';
@@ -15,7 +15,7 @@ const displayName = 'common.UserProfile';
 
 const UserProfile = () => {
   const { username } = useParams();
-  const { user, loading, error } = useUserByName(username || '');
+  const { user, loading, error } = useUserByNameOrAddress(username || '');
 
   if (loading) {
     return <UserProfileSpinner />;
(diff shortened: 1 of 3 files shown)


src/hooks/useUserByNameOrAddress.ts
```diff
@@ -0,0 +1,82 @@
...
+  });
+
+  const user = data.getUserByAddress?.items[0];
+  return { user, loading, error };
+};
+
+/*
+ * Can't conditionally call the codegen query hooks so querying GraphQL directly
+ */
+const useUserByNameOrAddress = (username: string) => {
```
ArmandoGraterol, line 57: Same here

src/components/common/UserProfile/UserProfile.tsx
```diff
@@ -15,7 +15,7 @@ const displayName = 'common.UserProfile';
 
 const UserProfile = () => {
   const { username } = useParams();
-  const { user, loading, error } = useUserByName(username || '');
+  const { user, loading, error } = useUserByNameOrAddress(username || '');
```
ArmandoGraterol, line 18: I would also change the name of this param to something more general as with this PR now this could not be just a username.

jakubcolony (changes requested):
Nice work but I think we can avoid all the overheads of calling the queries "manually" by using the generated hooks with `skip` option

```ts
const isAddress = isAddress(username);
const { } = useUserByAddressQuery({
  skip: !isAddress,
});
const { } = useUserByNameQuery({
  skip: isAddress,
});
```

Julianahlk:
> Nice work but I think we can avoid all the overheads of calling the queries "manually" by using the generated hooks with `skip` option

@jakubcolony Thanks for the suggestion, updated.

jakubcolony (approved):
LGTM, thanks for making the changes 👍

PR #585, opened 2023-05-25
Fix available funds

willm30:
## Description

This PR fixes the disappearance of "Available Funds" in ColonyHome.

[A change to the colony fragment](https://github.com/JoinColony/colonyCDapp/commit/a2087b97fc203e75a7f8c6891c6eb7dc41edff10) made the token we're adding manually in the cache the wrong format, and so apparently invalidated our manual changes to the Colony field in the cache.

**Changes** 🏗

* Update token format in cache and add typing so we notice more quickly and easily in the future. 

## Testing

On master, you should see your available funds.

Once I get some approvals I will push the changes directly to qa also so no need to rebase to see the fix.


src/graphql/fragments/colony.graphql
```diff
@@ -47,6 +42,15 @@ fragment Colony on Colony {
   }
 }
 
+fragment ColonyTokensConnection on ModelColonyTokensConnection {
```
willm30, line 45: Making this a fragment lets it be used as a type which will make it obvious if something like this happens again in the future

src/cache/colony/tokens.ts
```diff
@@ -1,29 +1,33 @@
...
   tokens: {
-    read: (baseTokens) => ({
-      ...baseTokens,
-      items: [
-        ...baseTokens.items,
+    read: (baseTokens) => {
+      const updatedTokens: UnaliasedColonyTokensItem[] = [
         {
           __typename: 'ColonyTokens',
+          id: 'DEFAULT_TOKEN_ID',
```
willm30, line 14: This was the offender!

jakubcolony (approved):
nice one, thanks for fixing this!

ArmandoGraterol (approved):
LGTM 👌

PR #591, opened 2023-05-29
Wire unclaimed motions in stakes tab

### LINKED ISSUE #588: Wire in unclaimed motions in stakes tab

willm30:
## Description

This PR wires in the Stakes Tab of the Token Activation Popover. 

The tab now shows all motions in the given colony with unclaimed stakes for the logged in user.

**Changes** 🏗

* Add `unclaimedMotionsWithStakes` field to `Colony` object
* Stakes tab wired in

claims

## Testing

There are a number of routes to test.

1. Object to a motion so that it fails without going to a vote.
2. Stake a motion so that it passes without going to a vote.
3. Stake and object so that it goes to a vote. 
4. Stake and object with multiple users and vote with multiple users.

In all of the above cases, the stakes tab should correctly reflect the number of unclaimed stakes for the logged in user. Try switching users -> should still show correct number of unclaimed stakes for that user.

You can also check that the db also correctly reflects the state of unclaimed motions in the colony, and that we don't hang on to motions that no longer have any unclaimed stakes. (This is to make the client-side user-specific filtering of the array faster).

Then, you can claim either via the motion widget directly, or via the Claim All button in the stakes tab.

If you claim via the motion, it will automatically update the stakes tab after a moment.
If you claim via the stakes tab, it will also automatically update after a moment (via a short poll on the colony object in the db).

Let me know if you find any edge cases. 

Resolves #588


amplify/backend/function/fetchMotionState/src/utils.js
```diff
@@ -170,12 +175,49 @@ const didMotionPass = ({
...
+
+  const motionWithUnclaimedStake = motionsWithUnclaimedStakes?.find(
+    ({ transactionHash: txHash }) => txHash === transactionHash,
+  );
+
+  const unclaimedRewards = updatedStakerRewards.filter(
+    ({ isClaimed }) => !isClaimed,
+  );
+
+  if (!motionWithUnclaimedStake && unclaimedRewards.length) {
```
willm30, line 199: If we have an unfinalizable motion with stakes (e.g a motion that fails before going to a vote), we check to see if there's already an entry in the `motionsWithUnclaimedStakes` array. 

If not, we check to see if there are any unclaimed stakes, and if so, we add the motion to the array. 

These checks as described are necessary because we remove the motion from the array in the block ingestor once it no longer has any unclaimed rewards. These checks stop us from re-adding it with an empty unclaimed rewards array.

src/components/shared/Popover/Popover.tsx
```diff
@@ -45,6 +45,8 @@ interface Props {
   renderContentValues?: SimpleMessageValues;
   /** Set the open state from outside */
   isOpen?: boolean;
+  /** State setter to sync external state with internal state */
+  setIsOpen?: SetStateFn<boolean>;
```
willm30, line 49: The changes in this file fix an issue with the external state of the popover (passed in via TokenActivationPopover) not being updated to reflect the internal state of the popover component

src/context/ColonyContext.tsx
```diff
@@ -14,12 +14,16 @@ interface ColonyContextValue {
...
+  startPolling: (pollInterval: number) => void;
+  stopPolling: () => void;
 }
 
 const ColonyContext = createContext<ColonyContextValue>({
   loading: false,
   canInteractWithColony: false,
   refetchColony: () => null,
+  startPolling: () => undefined,
+  stopPolling: () => undefined,
```
willm30, line 26: Exposing the polling functions allows us to poll after clicking Claim All, which is the only way to ensure the Stakes Tab updates automatically without needing a browser refresh.

Julianahlk (commented):
I am having issues with the state of an already claimed motion getting reset and being able to claim it again after a different motion is finalized.

**Test Steps:**
1. Create a Mint Token motion with 1 user staking 100% and 1 other user objecting 100% and have it pass
2. Create a Mint token motion with a single user and have it fail
3. Create a Mint Token motion with 1 user staking 100% and 1 other user objecting 100% and have it fail
4. Finalize the 1st and 3rd motions then use "claim all" for user 1. Everything seems correct at this point
5. Finalize the 2nd motion that only involved user 1, now "claim all" will appear for all 3 motions, the 2 motions that were already claimed and the one that was just finalized. In the motions that were already claimed, the text "claimed their stake" will be displayed and the claim button will incorrectly be enabled.

Screenshot 2023-06-01 at 3 09 41 PM

If I claim again, the claim their stake message will appear twice:

Screenshot 2023-06-01 at 3 12 23 PM

willm30:
@Julianahlk I'm finding this tricky to follow and reproduce myself. 

For example, this step:

> Create a Mint token motion with a single user and have it fail

You mean, create a motion with a single user, object with that same user, and then fast-forward time so it fails? In which case, it can't be finalized per this step:

> Finalize the 2nd motion that only involved user 1

> If I claim again, the claim their stake message will appear twice:

Your screenshot shows two different users, so having two separate claim messages seems correct to me, no?

Maybe you could reproduce with a simpler example?

In any case, I have updated the claim widget so that it refetches the action data when it notices that the colony's `motionsWithUnclaimedStakes` array changes. Let me know if that improves it.

ArmandoGraterol (approved):
Working as expected for me 👍 Nice job

Julianahlk:
@willm30 Should a claim of 0 tokens show up in the list for the stakes tab?
 
Screenshot 2023-06-05 at 12 47 57 PM

Screenshot 2023-06-05 at 12 47 44 PM

Julianahlk (approved):
After retesting, I wasn't able to reproduce the issues I had with claiming a motion twice. Other than the issue with the 0 tokens claim everything looks good now 👍

src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/MotionPhaseWidget.tsx
```diff
@@ -12,6 +12,7 @@ import { VotingWidget } from './VotingWidget';
 import { RevealWidget } from './RevealWidget';
 import { VoteOutcome } from './VoteOutcome';
 import { MotionAction } from '~types/motions';
```
jakubcolony, line 14: that should go higher up

jakubcolony (approved):
Nice work! 👍

willm30:
> After retesting, I wasn't able to reproduce the issues I had with claiming a motion twice. Other than the issue with the 0 tokens claim everything looks good now 👍

Opened an issue for this https://github.com/JoinColony/colonyCDapp/issues/614

PR #609, opened 2023-06-02
feat: added dark theme, fixed few components on dark theme

CzarekDryl:
Implemented dark mode, fixed components that were broken on dark mode.


KatarzynaStefanowiczPagepro (approved):

src/components/frame/Extensions/themes/enum.ts
```diff
@@ -0,0 +1,4 @@
+export enum Themes {
```
krawczykkatarzyna, line 1: Theme

src/components/frame/Extensions/themes/index.ts
```diff
@@ -0,0 +1,7 @@
+import { light, dark } from './consts';
+import { IThemes } from './types';
```
krawczykkatarzyna, line 2: ITheme

src/components/frame/Extensions/themes/index.ts
```diff
@@ -0,0 +1,7 @@
...
+
+export const themes: IThemes = {
```
krawczykkatarzyna, line 4: theme

src/components/frame/Extensions/themes/enum.ts
CzarekDryl: Changed

src/components/frame/Extensions/themes/index.ts
CzarekDryl: Changed to `themeList`, I think it's more precise

src/components/frame/Extensions/themes/index.ts
CzarekDryl: Same as comment under

krawczykkatarzyna (approved):

PR #624, opened 2023-06-07
Set extension permissions on initialisation

### LINKED ISSUE #619: Set VotingReputation permissions on installation
Currenlty VotingReputation Extn does not have correct permissions when it gets installed.

willm30:
## Description

This pr adds the logic to set extension permissions when an extension is enabled.

**Changes** 🏗

* Add permissions logic to enable extn saga

## Testing

1. Create a colony.
2. Install the voting rep extension. Ensure the rep extn address is picked up in the extension details widget (you will need to refresh the browser). (@jakubcolony We need to fix this before launching to production, since we can't expect users to know to refresh the browser after installing the extension. If you don't, the wrong extension address ("0x000...") gets passed to the block ingestor.)

3. Enable the voting rep extension. All permissions should be enabled, and thus you should be able to freely create, stake, and finalize motions.

Resolves #619

### DIFF
--- a/src/constants/index.ts
+++ b/src/constants/index.ts
@@ -151,6 +151,6 @@ export const ADDRESS_ZERO = ethersContants.AddressZero;
 
 export const GANACHE_LOCAL_RPC_URL = 'http://localhost:8545';
 
-export const isDev = process.env.NODE_ENV === 'development';
+export const isDev = process.env.NETWORK === 'ganache';
 
 export const STAKING_THRESHOLD = 10;
--- a/src/redux/sagas/extensions/extensionEnable.ts
+++ b/src/redux/sagas/extensions/extensionEnable.ts
@@ -1,9 +1,12 @@
 import { all, call, put, takeEvery } from 'redux-saga/effects';
+import { ClientType, Id } from '@colony/colony-js';
 
+import { intArrayToBytes32 } from '~utils/web3';
 import { ActionTypes } from '../../actionTypes';
 import { Action } from '../../types/actions';
 import { createGroupTransaction, getTxChannel } from '../transactions';
 import {
+  Channel,
   modifyParams,
   putError,
   refreshEnabledExtension,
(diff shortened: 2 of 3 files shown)


src/redux/sagas/extensions/extensionEnable.ts
```diff
@@ -30,33 +40,56 @@ function* extensionEnable({
...
+          {
+            identifier: colonyAddress,
+            methodName: channelName,
+            ...channels[channelName],
+          },
         ),
       );
 
+      /* Delay action creation in development, else block ingestor doesn't detect all on-chain events */
+      if (isDev) {
```
willm30, line 82: For context, firing all actions concurrently was causing the block ingestor to "drop" or miss events, and therefore was  setting permissions but not initialising the extension in the db. 

Nothing wrong with colonyjs or interacting with chain itself, this is purely a case of the block ingestor not hearing the event correctly. 

Adding the delay appears to solve the problem, at least on my machine. 

I expect this will cease to be a problem once we change the processing to a block by block basis. ( @jakubcolony )

src/constants/index.ts
```diff
@@ -151,6 +151,6 @@ export const ADDRESS_ZERO = ethersContants.AddressZero;
 
 export const GANACHE_LOCAL_RPC_URL = 'http://localhost:8545';
 
-export const isDev = process.env.NODE_ENV === 'development';
```
willm30, line 154: NODE_ENV does not exist in my .env nor my .env.example so switching to something that does

ArmandoGraterol (changes requested):
Nice one! 1 step closer to not needing the truffle console to test stuff in the dev env :)

I did notice something, it's that the "Assign permissions" item in the gas station doesn't appear until that action is completed so a normal user would not notice that there are more transactions to sign and execute after the initialize one.

Julianahlk (approved):
Works well for me. One thing I notice is that sometimes its difficult to get the extension to actually enable on the first colony I create but generally works fine on the 2nd or 3rd colonies that are created.

ArmandoGraterol (approved):
It seems like the issue I found is due to the delay we've introduced to account for the block-ingestor on the dev env. Therefore, I think this should be good to merge as it shouldn't be an issue on QA or prod 👍

PR #630, opened 2023-06-08
Fix/address for contributor

Julianahlk:
## Description

This reverts a previous change and makes name a required field again on the User. The original issue is that its possible for a Contributor to exist as just an address without having a valid user.

This attempts to solve this by adding an address field to Contributors and Watchers so that a Contributor can just be an address.


src/components/common/ColonyHome/ColonyMembersWidget/MembersSubsection.tsx
```diff
@@ -143,12 +143,13 @@ const MembersSubsection = ({
       {setHeading(true)}
       <ul className={styles.userAvatars}>
         {members.slice(0, avatarsDisplaySplitRules).map((member) => (
```
jakubcolony, line 145: On a side note, this should probably use `filter(notNull)`

jakubcolony (approved):
👍

PR #633, opened 2023-06-08
Replace link history if traveling between transactions

### LINKED ISSUE #602: Go back button from motions doesn't go back to colony home
To reproduce, ensure you have at least two different motions in the staked claims tab. Click back and forth between them a few times. 

Then click "go back" in the top left hand corner. It should "go back" to the colony, but it will instead replicate the native browser back button and take you back to the previous url, which was the other motion.

Julianahlk:
## Description

This solves the issue that when linking from one motion to the next using go to motion from the stakes tab, each motion is added to the history stack and the back button goes back to the previous motion.

if linking to a motion when already on a motion, the history stack will be replaced instead of pushed and now the back button will go to the previous location in the stack before navigating between motions.

Resolves #602

### DIFF
--- a/src/components/frame/TokenActivation/StakesTab/StakesListItem.tsx
+++ b/src/components/frame/TokenActivation/StakesTab/StakesListItem.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { defineMessages, FormattedMessage } from 'react-intl';
+import { useLocation } from 'react-router-dom';
 
 import Link from '~shared/Link';
 import Numeral from '~shared/Numeral';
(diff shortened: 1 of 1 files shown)


src/components/frame/TokenActivation/StakesTab/StakesListItem.tsx
```diff
@@ -38,9 +39,15 @@ const StakesListItem = ({
   });
 
   const txHash = data?.getColonyActionByMotionId?.items[0]?.id ?? '';
+  const location = useLocation();
+  const replace = location.pathname.indexOf('/tx') !== -1;
```
jakubcolony, line 43: How about a more descriptive name for this variable, e.g. `isMotionPage`? I feel like it'd make it easier to understand at first glance

jakubcolony (approved):
Nice one 👍

ArmandoGraterol (approved):
Nice fix 👍

PR #653, opened 2023-06-14
Update Create Domain action payload

### LINKED ISSUE #647: Fix Create Domain Motion Payload
The Create Domain Motion payload is not consistent with the Edit Domain Motion payload. Both use the same saga, hence an error is being thrown when trying to create a domain via a motion.

To reproduce:

Create a domain motion. You should see an error in the console.

ArmandoGraterol:
Fixed the create domain payload having a different structure than the one we use in the edit domain saga and create/edit domain motion saga.

You should be able to create "create/edit domain" actions and motions without problems.

Resolves #647


Julianahlk (approved):
Tested and create domain is still working as expected 👍

jakubcolony (approved):
👍

PR #654, opened 2023-06-14
Add @index to extension hash field

jakubcolony:
## Description

In block-ingestor#88, there's a need for fetching extensions by their hash and in such scenarios @index should help speed things up.


Julianahlk (approved):
Looks good!

PR #659, opened 2023-06-15
Update error message display conditional on staking slider

### LINKED ISSUE #601: Not enough tokens error message hidden after motion staked 100%
To reproduce:

Have two users. User 2 has 0 activated tokens.

User 1 stakes a motion 100%. 
Then, switch to User 2. 
The staking widget input is correctly disabled, but, there's no message to inform the user to activate their tokens.

The reason for this is this in `StakingSlider.tsx`:

```
  const displayErrMsg = !!user && !isLoadingData && remainingToStake !== '0';

```

I think the "remainingToStake !== 0" was put in to stop the error message showing incorrectly after staking 100%, but obviously it needs to be refined a little more in light of the above.

ArmandoGraterol:
The error messages in the staking slider should work as always but now this shouldn't be reproducible:

```
Have two users. User 2 has 0 activated tokens.

User 1 stakes a motion 100%.
Then, switch to User 2.
The staking widget input is correctly disabled, but, there's no message to inform the user to activate their tokens.
```

Resolves #601

### DIFF
--- a/src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/StakingWidget/StakingInput/StakingSliderLabel.tsx
+++ b/src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/StakingWidget/StakingInput/StakingSliderLabel.tsx
@@ -65,7 +65,7 @@ const StakingSliderLabel = ({
           symbol={nativeTokenSymbol}
         />
       )}
-      {remainingToStake !== '0' && (
+      {enoughTokensToStakeMinimum && remainingToStake !== '0' && (
         <RequiredStakeMessage {...requiredStakeMessageProps} />
       )}
     </Tooltip>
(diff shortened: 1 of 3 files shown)


Julianahlk (approved):
LGTM 👍

src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/StakingWidget/StakingInput/StakingSliderLabel.tsx
```diff
@@ -57,13 +58,14 @@ const StakingSliderLabel = ({
...
-      {!enoughTokensToStakeMinimum ? (
+      {!enoughTokensToStakeMinimum && (
         <MinimumStakeMessage
           userMinStake={userMinStake}
           decimals={nativeTokenDecimals}
           symbol={nativeTokenSymbol}
         />
-      ) : (
+      )}
+      {remainingToStake !== '0' && (
```
willm30, line 68: Could we have this condition be `enoughTokensToStakeMinimum && remainingToStake !== '0'`, to avoid the following:

Screenshot (846)

I.e. they're showing up at the same time, which I originally wanted to avoid.

willm30 (approved):
LGTM

PR #676, opened 2023-06-21
fix button copy

krawczykkatarzyna:
fix button copy


CzarekDryl (approved):

src/components/frame/Extensions/pages/ExtensionDetailsPage/partials/ExtensionDetails/ExtensionDetails.tsx
```diff
@@ -112,7 +112,7 @@ const ExtensionDetails: FC<ExtensionDetailsProps> = ({ extensionData }) => {
                 id: 'extensionDetailsPage.confirmUninstall',
               })}
               closeMessage={formatMessage({
-                id: 'button.cancelUninstall',
```
KatarzynaStefanowiczPagepro, line 115: The `closeMessage` prop is a text in button so I think we should add a `button.cancelUninstall` to en.json and remove `extensionDetailsPage.cancelUninstall` from there

PR #697, opened 2023-06-26
Update permission management read only permission logic

### LINKED ISSUE #696: Unable to create ManagePermissions motion if user has no permissions
## To Reproduce 

On a colony with voting rep extn enabled, and with reputation in the colony,
And a user with no permissions in that colony,
Open the Manage Permissions dialog.

The inputs are disabled and you're unable to create a motion:
perms

ArmandoGraterol:
This is not a perfect solution. I think `canRoleBeSet` or the permission validation, and the UI in this dialog needs to be reworked to accommodate for the user actually needing `Root` or `Architecture` roles depending on the domain and permissions being modified. 

That would require more time and effort which is why I'm just uploading this fix, which should be enough to fix a user with no permissions not being able to create a motion.

FireShot Capture 083 - Colony CDapp - localhost

Resolves #696

### DIFF
--- a/src/components/common/Dialogs/PermissionManagementDialog/PermissionManagementForm.tsx
+++ b/src/components/common/Dialogs/PermissionManagementDialog/PermissionManagementForm.tsx
@@ -207,7 +207,10 @@ const PermissionManagementForm = ({
             return (
               <PermissionManagementCheckbox
                 key={role}
-                readOnly={(inheritedRole && !directRole) || !canRoleBeSet(role)}
+                readOnly={
+                  (inheritedRole && !directRole) ||
+                  (!canRoleBeSet(role) && showPermissionErrors)
+                }
                 disabled={disabledInput || !selectedUser}
                 role={role}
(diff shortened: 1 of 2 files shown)


Julianahlk:
@ArmandoGraterol When testing this PR I am getting the error "colony-network-extension-bad-version" right after starting the dev env when I try to create a colony. Any ideas why this could be happening? I haven't seen this error in awhile.

Screenshot 2023-06-26 at 9 19 40 PM

willm30:
@Julianahlk I have opened prs to fix what you're seeing, but you can also manually fix for the time being by adding:

> NODE_ENV=development

to `docker/files/block-ingestor/env.base`, which is the .env for the block-ingestor when run from Docker. We're now using that env var to detect whether the block ingestor is running locally, and if so, seed the database with data like extension versions.

willm30 (approved):
LGTM

Julianahlk (approved):
Working well for me 👍

PR #700, opened 2023-06-27
Update NODE_ENV env var in block ingestor dockerfile

willm30:
## Description

Small PR to update the block ingestor env vars to use NODE_ENV when running the block ingestor from docker.


Julianahlk (approved):
LGTM 👍

PR #704, opened 2023-06-28
Update staked tokens after staking a motion

### LINKED ISSUE #675: Update active tokens after staking
After staking a motion, the number of active tokens that you have available decreases (because you just staked some). This should be reflected instantly in the Token Activation Popover, since this data forms the basis of the staking limits in the Staking widget.

Currently it doesn't update till you refresh the browser, which means the staking limits won't reflect the actual number of active tokens you have available.

To reproduce:

Activate enough tokens to be able to stake, but not enough that you can stake 100%. You will see the staking limit in the staking widget. Then stake. Make a mental note of where the staking limit is once the staking widget has reloaded.

Now refresh the browser. Note that the staking limit has changed position, because the active tokens data now takes into account the stake you just made.

willm30:
## Description

This PR updates the Staked Tokens item in the Token Activation Popover after staking a motion, and also after claiming token rewards (via widget and via stakes tab in Token Activation Popover).

## To Test

We're testing two things. The first is that the "Staked" item in the Token Activation Popover correctly updates whenever we stake a motion, or claim our rewards (either via the widget, or via the Token Activation Popover).

This is fairly simple to test: ensure that the value updates after staking / claiming.

The second is that the Staking limit updates after staking. To test this, you need to activate enough tokens to stake but not enough to stake 100%. Then stake. Once the widget updates, the staking limit should have updated to reflect the fact that you have fewer active tokens now.

**New stuff** ✨

* moved `useGetUserTokenBalanceQuery` to new context provider
* add polling after staking and claiming tokens

**Changes** 🏗

* Removed prop drilling of token balance data

Resolves #675

### DIFF
--- a/src/context/UserTokenBalanceContext.tsx
+++ b/src/context/UserTokenBalanceContext.tsx
@@ -1,15 +1,19 @@
-import React, { ReactNode, createContext, useContext, useMemo } from 'react';
+import React, {
+  ReactNode,
+  createContext,
+  useCallback,
+  useContext,
+  useMemo,
+} from 'react';
 import { GetUserTokenBalanceReturn, useGetUserTokenBalanceQuery } from '~gql';
 import { useAppContext, useColonyContext } from '~hooks';
 
 export const UserTokenBalanceContext = createContext<{
   tokenBalanceData: GetUserTokenBalanceReturn | null | undefined;
-  startPollingTokenBalance: (interval: number) => void;
-  stopPollingTokenBalance: () => void;
+  pollTokenBalance: () => void;
 }>({
   tokenBalanceData: null,
-  startPollingTokenBalance: () => {},
-  stopPollingTokenBalance: () => {},
+  pollTokenBalance: () => {},
 });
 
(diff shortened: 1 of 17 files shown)


src/context/UserTokenBalanceContext.tsx
```diff
@@ -0,0 +1,97 @@
...
+      }
+    };
+
+    interval = setInterval(fetchActiveBalance, 3_000);
+  }, [tokenBalanceData, refetch]);
+
+  const pollLockedTokenBalance = useCallback(async () => {
+    let interval;
+
+    const fetchLockedBalance = async () => {
```
willm30, line 60: Distinguishing between active and locked.

When staking / claiming, we can stop polling once the locked tokens value has changed.

When activating / deactivating tokens via the popover, the locked tokens value won't change, and therefore we won't know to stop polling if we're only looking for the locked value to change.

ArmandoGraterol (approved):
Working as expected for me 👍 Nice job!

rdig (approved):

PR #709, opened 2023-06-29
Fix reputation in motion msg

willm30:
## Description

Currently, the reputation for the motion messages is using the reputation in the `fromDomain`, as opposed to the reputation in the `motionDomain`, which means that in the example below, it's showing that I have no rep, when I do. 

rep-calc

We are interested in displaying the rep in the `motionDomain` because that's what is used in the voting calculation. 

**Changes** 🏗

* Use the motion domain, not from domain, to calculate rep in the motion messages

### DIFF
--- a/src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/VotingWidget/useVoteDetailsConfig.tsx
+++ b/src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/VotingWidget/useVoteDetailsConfig.tsx
@@ -42,6 +42,7 @@ export const useVoteDetailsConfig = ({
     colony?.colonyAddress ?? '',
     user?.walletAddress ?? '',
     Number(nativeMotionDomainId),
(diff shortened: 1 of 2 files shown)


src/components/common/ColonyActions/ActionDetailsPage/DefaultMotion/MotionPhaseWidget/VotingWidget/useVoteDetailsConfig.tsx
```diff
@@ -42,6 +42,7 @@ export const useVoteDetailsConfig = ({
     colony?.colonyAddress ?? '',
     user?.walletAddress ?? '',
     Number(nativeMotionDomainId),
+    rootHash,
```
willm30, line 45: This is a different but related fix so I'm including it here. The reputation displayed in the voting widget should be that of the time at which the motion was created, just like with staking.

ArmandoGraterol (approved):
Nice catch! It's working as expected for me 👍

rdig (approved):

PR #712, opened 2023-06-29
Fix error in getMembersForColony Lambda

willm30:
## Description

This PR prevents the contributors sort function in the `getMembersForColony` lambda from erroring in the event that a contributor's reputation percentage is undefined. This is the source of a recent error in fetching colony members on Qa: https://qa-cdapp.colony.io/colony/leaningtowerofpisa

## Testing

The problem is the we're calling the Decimal constructor with values that might be undefined (I can see this in the AWS Lambda console). So, this should fix that error. Can't easily be tested locally, therefore just code review.

**Changes** 🏗

* Guard Decimal constructor from undefined values

### DIFF
--- a/amplify/backend/function/getMembersForColony/src/index.js
+++ b/amplify/backend/function/getMembersForColony/src/index.js
@@ -244,33 +244,54 @@ exports.handler = async (event) => {
     }),
   );
 
-  const sortedContributors = (() => {
-    return contributorsWithReputation.sort((contributor1, contributor2) => {
-      if (sortingMethod === SortingMethod.BY_HIGHEST_REP) {
-        return new Decimal(contributor2?.reputationPercentage)
-          .sub(contributor1.reputationPercentage)
-          .toNumber();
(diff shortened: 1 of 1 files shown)


ArmandoGraterol (commented):
How should we go about reproducing this in the local env? 

I've checked out your colony with this branch using the QA env but the members are not loading. (with or without this branch) 

In the dev env I tried to reproduce this with 2 members and 1 watcher and I also couldn't.

willm30:
@ArmandoGraterol Sorry, I should've clarified. I don't expect this to be tested locally, since it's a bug I'm only noticing in QA. However, the problem is the we're calling the Decimal constructor with values that might be undefined. So, this should fix that error. Then I can deploy to QA and see if the lambda starts working correctly again.

ArmandoGraterol (approved):
Thank you for the clarification! The code looks good to me 👍

rdig (approved):

PR #713, opened 2023-06-29
Ensure staking slider has correct max value

willm30:
## Description

I noticed a bug (difficult to reproduce) in which the staking slider would only stake 99% of the remaining stake, even if I had enough tokens. 

This is a rounding error, and so I've updated the logic to capture the state in which a user wants to stake 100% of the remaining stake amount.

**Changes** 🏗

* Add case specifically if slider is at 100


ArmandoGraterol (approved):
I couldn't reproduce the bug but the slider is working as expected so I guess it's good to go 👍

rdig (approved):

PR #714, opened 2023-07-03
Fix `ConfirmAllInput` step

willm30:
## Description

When selecting an existing token in the Colony Creation wizard, we're missing its name and symbol:
bigben

This PR fixes that. 

**Changes** 🏗

* Update form values to include name and symbol

Other (misc):

* Add focus state to token select options for feedback while tabbing
* Checksum token address when querying db
* Fix loading state in Token Selector


ArmandoGraterol (approved):
LGTM 👍

thomassdk (approved):
Works well for me

jakubcolony (approved):
Hm, the fix for that must have gotten lost in all the rebases - LGTM, thanks 👍

PR #715, opened 2023-07-03
Fix user name in action message

willm30:
## Description

Currently, the logic for generating recipient / initiator info in the action message doesn't properly account for the fact that it could be a Colony, Colony Extension or Token as well as a user. This is resulting in the following when paying a Colony:

It's incorrectly and inconsistently displaying the recipient
payment

This PR fixes it. 

bigbenafter

## Testing

Create an action such as paying another colony, and the correct data should be displayed in the message. 

**New stuff** ✨

*Add logic to detect the type of the initiator / recipient


ArmandoGraterol (approved):
Great work 🥃

thomassdk (approved):
LGTM!

jakubcolony (approved):
awesome 👍

PR #718, opened 2023-07-04
Update ingestor hash for block-ingestor#98

willm30:
Need to update block ingestor hash to point to: https://github.com/JoinColony/block-ingestor/pull/98


willm30:
No longer necessary, master is pointing to latest block ingestor hash after merging other prs

PR #719, opened 2023-07-04
Update funds after claim

willm30:
## Description

This PR updates the colony object after claiming an incoming funds item, meaning the ui will update automatically without need for a browser refresh.

## Testing
Transfer a colony some token so there's an incoming funds claim. Make the claim, either from the funding widget or the funds page. You should see the ui update automatically some seconds later.

**Changes** 🏗

* Poll colony object after claiming

### DIFF
--- a/src/components/common/ColonyHome/ColonyUnclaimedTransfers/ColonyUnclaimedTransfers.tsx
+++ b/src/components/common/ColonyHome/ColonyUnclaimedTransfers/ColonyUnclaimedTransfers.tsx
@@ -41,7 +41,12 @@ const MSG = defineMessages({
 });
 
 const ColonyUnclaimedTransfers = () => {
-  const { colony, canInteractWithColony } = useColonyContext();
+  const {
+    colony,
+    canInteractWithColony,
+    startPolling: startPollingColony,
+    stopPolling: stopPollingColony,
+  } = useColonyContext();
   const claims = useColonyFundsClaims();
 
   const firstItem = claims[0];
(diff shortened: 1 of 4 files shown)


src/context/ColonyContext.tsx
```diff
@@ -46,9 +52,19 @@ export const ColonyContextProvider = ({
   children: ReactNode;
 }) => {
   const { colonyName } = useParams<{ colonyName: string }>();
+  const [prevColonyName, setPrevColonyName] = useState<string>();
+
   const { state: locationState } = useLocation();
   const [isPolling, setIsPolling] = useState(!!colonyName);
 
+  /* Update polling state when routing between colonies */
```
willm30, line 60: This ensures we don't inadvertently restart polling when switching between colonies.

src/components/common/ColonyHome/ColonyUnclaimedTransfers/ColonyUnclaimedTransfers.tsx
```diff
@@ -59,6 +64,11 @@ const ColonyUnclaimedTransfers = () => {
    */
   const token = firstItem?.token;
 
+  const handleClaimSuccess = () => {
+    startPollingColony(1_000);
+    setTimeout(stopPollingColony, 10_000);
```
willm30, line 69: These are currently "magic" values. There's definitely a good first issue in moving all such "magic" values to our constants file.

thomassdk (approved):
LGTM

The only thing I noticed when testing it was that when I was transferring a native token it didnt appear in the `Claim Token` of the `funds` page and the Colony amount never increased.

Screenshot from 2023-07-07 12-06-09
Screenshot from 2023-07-07 12-15-25

willm30:
@thomassdk Thanks, I will add that to the qa list

ArmandoGraterol (approved):
Working as expected 👍

PR #725, opened 2023-07-05
Expand note on lambda dependencies

thomassdk:
-------

## Description

This PR expands the documentation in the README in relation to setting up lambda dependencies, including a note on possible error messages

**New stuff** ✨

* A script was added to the package.json to allow users to easily run the command to install lambda dependencies using npm

**Changes** ✨

* `lambda-functions-dependencies.sh`'s method to find the bash interpreter has been changed in order to improve portability (IE it doesn't work on some distros)

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,8 @@
     "amplify": "amplify",
     "DEPRECATED:webpack:prod": "NODE_OPTIONS=\"--max-old-space-size=4096\" NODE_ENV=production webpack-cli --config webpack.prod.js",
     "voyager": "ts-node temp-voyager/voyager",
-    "codegen": "node codegen"
+    "codegen": "node codegen",
+    "postinstall": "./scripts/lambda-functions-dependencies.sh"
   },
   "repository": {
     "type": "git",
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,6 +7,7 @@
     "": {
       "name": "colony-cdapp",
       "version": "5.0.0-alpha.0",
+      "hasInstallScript": true,
       "dependencies": {
         "@apollo/client": "^3.6.9",
         "@colony/colony-js": "^6.3.6",
--- a/scripts/lambda-functions-dependencies.sh
+++ b/scripts/lambda-functions-dependencies.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 pwd=$(pwd)
 
--- a/scripts/lambda-functions-dependencies.sh
+++ b/scripts/lambda-functions-dependencies.sh
@@ -1,15 +1,21 @@
 #!/usr/bin/env bash
 
 pwd=$(pwd)
+directory="amplify/backend/function"
 
-for d in amplify/backend/function/*/ ; do
-    [ -L "${d%/}" ] && continue
-    fnName=$(echo ${d} | cut -c 26- | rev | cut -c 2- | rev)
-    cd "${pwd}/${d}src"
-    # Skip if the folder doesn't contain package.json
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -20,7 +20,8 @@
     "amplify": "amplify",
     "DEPRECATED:webpack:prod": "NODE_OPTIONS=\"--max-old-space-size=4096\" NODE_ENV=production webpack-cli --config webpack.prod.js",
     "voyager": "ts-node temp-voyager/voyager",
-    "codegen": "node codegen"
+    "codegen": "node codegen",
+    "lambda:install": "./scripts/lambda-functions-dependencies.sh"
```
willm30, line 24: Re-adding the post install script should be as simple as adding a `postinstall` command to the package.json: https://docs.npmjs.com/cli/v9/using-npm/scripts#npm-install

I think it makes sense to do this in this pr (especially seeing as it's mentioned in the doc you've edited above).

It may also be worth adding something to that bash script to check in the `.env` for a local configuration (e.g. `NETWORK=ganache`), and only run if it's present. @rdig can confirm if this is necessary, i.e. if the postinstall command was causing problems for qa and by extension production.

willm30: Ok so the reason it was a pain for qa was because it results in a infinite loop if the lambda functions folder doesn't exist (which it only does locally).

So we could modify the bash script like so to short-circuit early if the dir doesn't exist:

``` sh
directory="amplify/backend/function"

if [ -d "$directory" ]; then 
    for d in amplify/backend/function/*/ ; do
        [ -L "${d%/}" ] && continue
        fnName=$(echo ${d} | cut -c 26- | rev | cut -c 2- | rev)
        # Check if src directory exists
        if [ -d "${pwd}/${d}src" ]; then 
            cd "${pwd}/${d}src"
            # Skip if the folder doesn't contain package.json
            # Otherwise npm will look for the root package.json and cause an infinite loop
            !([ -f "package.json" ]) && continue
            echo "Installing dependencies for Lambda Function \"${fnName}\""
            npm i
            echo
        fi
    done
fi

```

thomassdk: So it sounds that this PR needs to change from updating the Readme to modifying this bash script. Your updated script looks good, but why are we checking if the src directory exists? 
```bash
        # Check if src directory exists
        if [ -d "${pwd}/${d}src" ]; then 
```
Don't we want it to fail if the lambda doesn't have a src directory, as it suggests something has gone wrong somewhere locally?

willm30: Sorry that's fair enough, can definitely be handled separately. I was caught by:

> At the end of the install there's a post-install script that will recursively install dependencies for all the currently declared lambda functions.

Which currently isn't the case. And my thinking was if we're updating documentation, it should be accurate, hence suggesting we add it back in. But I see no problem in doing it separately either.

willm30: > Don't we want it to fail if the lambda doesn't have a src directory, as it suggests something has gone wrong somewhere locally?

The layer we're using doesn't have a src dir, and that's expected. So basically, it's: "skip the layer"

willm30 (approved):
LGTM but let's make sure we do something about the postinstall script so the docs remain uptodate 🙂

package.json
thomassdk: I was more suggesting that we remove the changes to the documentation and fix the `postinstall` script, as otherwise I think my suggested documentation changes are a bit too verbose. Otherwise, if we start documenting everything that might go wrong we will never get anything done :)

thomassdk: Also had a more in depth look at the lambda functions and realised that `colonycdappSSMAccess` which is why we need that bash guard.

Out of curiosity, why is that one structured differently to the others with `lib/nodejs` setup?

PR #728, opened 2023-07-05
Remove ability to deprecate one tx payment

willm30:
## Description

This PR prevents you from being able to deprecate the One Tx Payment extension.

### DIFF
--- a/src/components/common/Extensions/ExtensionDetails/ExtensionDetails.tsx
+++ b/src/components/common/Extensions/ExtensionDetails/ExtensionDetails.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { Route, Routes, useLocation, useParams } from 'react-router-dom';
 import { defineMessages, FormattedMessage } from 'react-intl';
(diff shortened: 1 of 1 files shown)


ArmandoGraterol (commented):
If the contracts allow it, shouldn't we also allow it?

jakubcolony (approved):
Don't know an answer to Armando's question but the change looks & works good

willm30:
@jakubcolony I was actually going to ask you... this was from a note you'd made in the notion qa doc. I.e. you said we should not be able to deprecate the extn. Any chance you remember the rationale?

jakubcolony:
@willm30 This was based on the current Dapp behaviour

willm30:
@ArmandoGraterol Thoughts? In the dapp you can't deprecate OneTxPayment extn.

jakubcolony:
@willm30 I came across this in #950 again and realised there's a `uninstallable` property on the extension config in `~constants/extensions`. I think we can close this PR.

PR #731, opened 2023-07-05
Consistent max colony name

### LINKED ISSUE #711: Make Colony name max length consistent
When creating a Colony, there's no character limit on the colony name.

But when editing a Colony, there's a 20 character limit on the name. 

Having a limit makes sense in principle (since there's an AWS max record size), so I'd suggest adding the limit to the Colony name input when creating a colony.

thomassdk:
## Description

This PR makes the colony (display) name max length consistent

Screenshot from 2023-07-05 12-56-49

image

**New stuff** ✨

* A new constant has been created in `MAX_DISPLAY_NAME` 

**Changes** 🏗

* `MAX_DISPLAY_NAME` is now used in colony creation validation and colony editing input limiting

Resolves #711

### DIFF
--- a/src/components/common/CreateColonyWizard/validation.ts
+++ b/src/components/common/CreateColonyWizard/validation.ts
@@ -1,10 +1,6 @@
 import { string, object } from 'yup';
 
-import {
-  ADDRESS_ZERO,
-  DEFAULT_NETWORK_TOKEN,
-  MAX_COLONY_DISPLAY_NAME,
-} from '~constants';
+import { ADDRESS_ZERO, DEFAULT_NETWORK_TOKEN } from '~constants';
 import { GetFullColonyByNameDocument } from '~gql';
 import { intl } from '~utils/intl';
 import { createYupTestFromQuery } from '~utils/yup/tests';
@@ -48,7 +44,6 @@ const { formatMessage } = intl({
 export const colonyNameValidationSchema = object({
   displayName: string()
     .trim()
-    .max(MAX_COLONY_DISPLAY_NAME)
     .required(formatMessage({ id: 'error.colonyNameRequired' })),
   colonyName: string()
     .required(formatMessage({ id: 'error.colonyURLRequired' }))
--- a/src/components/common/CreateColonyWizard/validation.ts
+++ b/src/components/common/CreateColonyWizard/validation.ts
@@ -42,9 +42,10 @@ const { formatMessage } = intl({
 });
 
 export const colonyNameValidationSchema = object({
-  displayName: string().required(
-    formatMessage({ id: 'error.colonyNameRequired' }),
-  ),
+  displayName: string()
+    .trim()
+    .max(20)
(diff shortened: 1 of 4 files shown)


src/components/common/CreateColonyWizard/validation.ts
```diff
@@ -42,9 +46,10 @@ const { formatMessage } = intl({
 });
 
 export const colonyNameValidationSchema = object({
-  displayName: string().required(
-    formatMessage({ id: 'error.colonyNameRequired' }),
-  ),
+  displayName: string()
+    .trim()
+    .max(MAX_DISPLAY_NAME)
```
willm30, line 51: We need to pass a custom message as the second param, since the automatically generated message uses the field's key in the validation schema (`displayName`), which is ugly.

https://github.com/jquense/yup#stringmaxlimit-number--ref-message-string--function-schema

src/constants/index.ts
```diff
@@ -158,3 +158,5 @@ export const isDev = process.env.NETWORK === 'ganache';
 export const CDAPP_VERSION = version;
 
 export const STAKING_THRESHOLD = 10;
+
+export const MAX_DISPLAY_NAME = 20;
```
willm30, line 162: Could we specify this is the max Colony display name? User display doesn't have a max, for instance.

src/components/common/Dialogs/EditColonyDetailsDialog/EditColonyDetailsDialogForm.tsx
```diff
@@ -123,7 +124,7 @@ const EditColonyDetailsDialogForm = ({
           name="colonyDisplayName"
           appearance={{ colorSchema: 'grey', theme: 'fat' }}
           disabled={disabledInput}
-          maxLength={20}
+          maxLength={MAX_DISPLAY_NAME}
```
willm30, line 127: You can pass this through to the display name input in `StepColonyNameInputs` also, so the user has some visual prompt wrt the character limit

willm30 (commented):
Nice one! Left a few comments, but happy to approve once addressed.

src/components/common/CreateColonyWizard/validation.ts
thomassdk: I haven't done this, as I implemented your other suggestion of setting the `maxLength` of the Input to have consistency across the app. As this is the case, I have just removed this validation as this change makes it obsolete.

src/components/common/Dialogs/EditColonyDetailsDialog/EditColonyDetailsDialogForm.tsx
thomassdk: Done

image

src/constants/index.ts
thomassdk: The variable is now named `MAX_COLONY_DISPLAY_NAME`

willm30 (approved):
LGTM!

ArmandoGraterol (approved):
This is it!

Issue #276, opened 2023-02-16
Update getMembersForColony lambda function to support pagination


Julianahlk:
- Searching and filtering should happen on the server
- Create solution to handle caching of the user reputation

See comments: https://github.com/JoinColony/colonyCDapp/issues/266#issuecomment-1428818532

PR #733, opened 2023-07-05
Wire in whitelist

willm30:
## Description

This PR wires in the whitelist functionality. It also adds some missing features from the members page, namely, the address book / permissions controls, and the verified filter.

Verified filter:

whitelist

Whitelist activated, users filtered out if not verified:

whitelistactivated

Whitelist activated, warning if trying to pay an address not on it:

whitelist warnign

Add to address book:
addressBook

**Changes** 🏗

* Wire in whitelist with create payment and manage reputation dialogs
* Wire in members controls items (whitelist and permissions)
* Add verified filter

## Testing

Dialogs:
- Payment: 
  - warning should appear if you try and pay someone who's not verified when the whitelist is activated
  - The user picker should also filter out non verified users when the whitelist is activated
 - Manage reputation
   - User picker should filter out non verified users when the whitelist is activated
- Manage addresses:
  - You should not be able to add the same address twice (e.g. one lower case version and one normal)
  - The Address Book tab should show all your verified users. If you switch the whitelist off, it should disable the UserPicker filtering in the other dialogs (i.e. you see all watchers, not just verified ones).

Members page:
- If your wallet has the root permission, you should be able to edit permissions and manage the address book from the MembersControls.
- The verified filter should work with all filter permutations
- You should see a blue tick next to a verified member in the members list
- Clicking "Add to address book" in the 3 dots menu who pre-populate the address field with the user's address

The intention is that all the unwired verified recipients logic is now wired, so let me know if you notice that I've missed anything.

Resolves https://github.com/JoinColony/colonyCDapp/issues/561

### DIFF
--- a/src/components/common/ColonyMembers/ColonyMembers.tsx
+++ b/src/components/common/ColonyMembers/ColonyMembers.tsx
@@ -69,7 +69,7 @@ const ColonyMembers = () => {
           <TotalReputation selectedDomainId={selectedDomainId} />
           <MemberControls isRootOrAllDomains={isRootOrAllDomains} />
           <MembersFilter
-            handleFilterChange={handleFilterChange}
+            onFilterChange={handleFilterChange}
             isRootOrAllDomains={isRootOrAllDomains}
           />
         </aside>
--- a/src/components/common/ColonyMembers/ColonyMembers.tsx
+++ b/src/components/common/ColonyMembers/ColonyMembers.tsx
@@ -69,7 +69,7 @@ const ColonyMembers = () => {
           <TotalReputation selectedDomainId={selectedDomainId} />
           <MemberControls isRootOrAllDomains={isRootOrAllDomains} />
           <MembersFilter
-            onFilterChange={handleFilterChange}
+            handleFilterChange={handleFilterChange}
             isRootOrAllDomains={isRootOrAllDomains}
           />
         </aside>
--- a/src/components/common/ColonyMembers/MemberControls/MemberControls.tsx
+++ b/src/components/common/ColonyMembers/MemberControls/MemberControls.tsx
@@ -1,72 +1,86 @@
 import React from 'react';
-// import { defineMessages } from 'react-intl';
 import { Extension } from '@colony/colony-js';
+import { defineMessages } from 'react-intl';
 
 import InviteLinkButton from '~shared/Button/InviteLinkButton';
-// import Button from '~shared/Button';
-// import { useDialog } from '~shared/Dialog';
-// import PermissionManagementDialog from '~dialogs/PermissionManagementDialog';
-// import ManageWhitelistDialog from '~dialogs/ManageWhitelistDialog';
 import { isInstalledExtensionData } from '~utils/extensions';
-import { useColonyContext, useExtensionData } from '~hooks';
+import {
+  useAppContext,
+  useCanInteractWithNetwork,
+  useColonyContext,
+  useEnabledExtensions,
+  useExtensionData,
+} from '~hooks';
+
+import { canArchitect, hasRoot } from '~utils/checks';
+import { getAllUserRoles } from '~transformers';
+import Button from '~shared/Button/Button';
+import { useDialog } from '~shared/Dialog';
+import {
+  ManageWhitelistDialog,
+  PermissionManagementDialog,
+} from '~common/Dialogs';
 
 import styles from './MemberControls.css';
 
 const displayName = 'common.ColonyMembers.MemberControls';
 
-// const MSG = defineMessages({
-//   editPermissions: {
-//     id: `${displayName}.editPermissions`,
-//     defaultMessage: 'Edit permissions',
-//   },
-//   manageWhitelist: {
-//     id: `${displayName}.manageWhitelist`,
-//     defaultMessage: 'Manage address book',
-//   },
-// });
+const MSG = defineMessages({
+  editPermissions: {
+    id: `${displayName}.editPermissions`,
+    defaultMessage: 'Edit permissions',
+  },
+  manageWhitelist: {
+    id: `${displayName}.manageWhitelist`,
+    defaultMessage: 'Manage address book',
+  },
+});
 
 type Props = {
   isRootOrAllDomains: boolean;
 };
 
 const MemberControls = ({ isRootOrAllDomains }: Props) => {
-  const { colony } = useColonyContext();
-
-  // const openPermissionManagementDialog =  useDialog(PermissionManagementDialog);
-  // const handlePermissionManagementDialog = useCallback(() => {
-  //   openPermissionManagementDialog({
-  //     colony,
-  //   });
-  // }, []);
-
-  // const openToggleManageWhitelistDialog = useDialog(ManageWhitelistDialog);
-  // const handleToggleWhitelistDialog = useCallback(() => {
-  //   return openToggleManageWhitelistDialog({
-  //     colony,
-  //   });
-  // }, []);
-
+  const { colony, isSupportedColonyVersion } = useColonyContext();
+  const canInteractWithNetwork = useCanInteractWithNetwork();
+  const { user } = useAppContext();
+  const currentUserRoles = getAllUserRoles(colony, user?.walletAddress);
+  const enabledExtensionData = useEnabledExtensions();
   const { extensionData } = useExtensionData(Extension.OneTxPayment);
+
+  const canManangePermissions =
+    (user && canArchitect(currentUserRoles)) || hasRoot(currentUserRoles);
+  const canManageWhitelist = user && hasRoot(currentUserRoles);
+
+  const openPermissionManagementDialog = useDialog(PermissionManagementDialog);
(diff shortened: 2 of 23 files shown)


src/components/common/Dialogs/ManageWhitelistDialog/ManageWhitelistDialogForm.tsx
```diff
@@ -180,7 +184,11 @@ const ManageWhitelistDialogForm = ({
         <DialogControls
           secondaryButtonText={backButtonText}
           onSecondaryButtonClick={back}
-          disabled={!isDirty || disabledSubmit}
+          disabled={
+            (!isDirty && !whitelistAddress) ||
```
willm30, line 188: This could probably be expressed better, but basically we want to disable the form if it isn't dirty, except for when there's an address already populated in the address field.

src/components/common/Dialogs/ManageWhitelistDialog/helpers.ts
```diff
@@ -30,7 +31,12 @@ export const getManageWhitelistDialogPayload = (
   } else {
     verifiedAddresses =
       whitelistAddress !== undefined
-        ? [...new Set([...whitelistedAddresses, whitelistAddress])]
+        ? [
+            ...new Set([
+              ...whitelistedAddresses,
+              createAddress(whitelistAddress),
```
willm30, line 37: Normalise addresses so you can't add the same address twice (but one normal and one lowercase)

src/components/common/Members/filterMembers.ts
```diff
@@ -1,55 +1,63 @@
...
   searchTerm?: string,
   filters?: FormValues,
-): M[] => {
-  /* No filters */
-  if (!searchTerm && filters?.verificationType === VerificationType.All) {
-    return members;
+): Member[] => {
+  let filtered = members;
+
+  if (filters?.verificationType !== VerificationType.All) {
```
willm30, line 46: There's [an issue](https://github.com/JoinColony/colonyCDapp/issues/276) to move this to the members lambda. Have not addressed here.

src/redux/sagas/colony/colonyCreate.ts
```diff
@@ -347,7 +347,7 @@ function* colonyCreate({
           input: {
             id: colonyAddress,
             displayName,
-            isWhitelistActivated: true,
+            isWhitelistActivated: false,
```
willm30, line 350: Should be deactivated by default, since there's no one in it to begin with

src/components/common/ColonyMembers/ColonyMembers.tsx
```diff
@@ -69,7 +69,7 @@ const ColonyMembers = () => {
           <TotalReputation selectedDomainId={selectedDomainId} />
           <MemberControls isRootOrAllDomains={isRootOrAllDomains} />
           <MembersFilter
-            onFilterChange={handleFilterChange}
+            handleFilterChange={handleFilterChange}
```
jakubcolony, line 72: Any reason for this change? I proposed this convention to easier distinguish between internal and external handlers:
https://github.com/JoinColony/ColonyFrontEndLivingStandard/commit/4a76e00551428ec638b02efd342b24512bd541d7

src/components/common/ColonyMembers/MemberControls/MemberControls.tsx
```diff
@@ -80,24 +94,24 @@ const MemberControls = ({ isRootOrAllDomains }: Props) => {
...
-          {/* {canMamangePermissions && (
-          <li>
-            <Button
-              appearance={{ theme: 'blue' }}
-              text={MSG.editPermissions}
-              onClick={handlePermissionManagementDialog}
-            />
-          </li>
-        )}
+          {canManangePermissions && (
```
jakubcolony, line 97: typo

src/components/common/Dialogs/CreatePaymentDialog/CreatePaymentDialog.tsx
```diff
@@ -7,16 +7,11 @@ import Dialog, { DialogProps, ActionDialogProps } from '~shared/Dialog';
...
-// } from '~data/index';
 import { pipe, withMeta, mapPayload } from '~utils/actions';
-// import { getVerifiedUsers } from '~utils/verifiedRecipients';
+import { getVerifiedUsers } from '~utils/verifiedUsers';
 import { WizardDialogType, useNetworkInverseFee } from '~hooks';
-import { useGetMembersForColonyQuery } from '~gql';
-
-import { extractUsersFromColonyMemberData } from '../helpers';
 
+import { useGetColonyMembers } from '../AwardAndSmiteDialogs/ManageReputationContainer/helpers';
```
jakubcolony, line 14: Is this a good enough reason to move to a shared hooks folder?

src/components/common/Dialogs/ManageWhitelistDialog/ManageWhitelistDialogForm.tsx
jakubcolony: Could this comment live in-code?

src/components/common/Members/filterMembers.ts
```diff
@@ -1,55 +1,63 @@
...
+    return !isVerified;
   }
 
   return true;
-  // return !member.isWhitelisted;
 };
 
-export const filterMembers = <M extends Contributor | Watcher>(
-  members: M[],
+export const filterMembers = <Member extends Contributor | Watcher>(
```
jakubcolony, line 38: There's a `Member` type exported from `~types/graphql` that could replace the `Contributor | Watcher` part (in a few places in this file). On a side note, my impression was that a single letter is more or less a convention for generics, to avoid confusion with regular types

jakubcolony (approved):
Left a few comments here and there, all good otherwise! 👍

src/components/common/ColonyMembers/ColonyMembers.tsx
willm30: Interesting. I had interpreted the rule as: "use 'handle' as the prefix for event handlers", as opposed to: "use 'handle' as the prefix for event handlers _and_ use 'on' as the prefix for all props that pass event handlers". 

That's perhaps worth clarifying in the entry in the living standard. 

We should also add this to help enforce it: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-handler-names.md

And create an issue to go and fix it everywhere it's not being adhered to, since I'm pretty sure it's not being adhered to generally.

willm30: Also, wouldn't this only apply to native events? `onFilterChange` is not an actual event, it wraps the `onChange` prop of the `Select` component (which also not a native event). 

That's how I've interpret your distinction between "internal" and "external" event handlers but I might be misreading that.

src/components/common/ColonyMembers/MembersFilter/MembersFilter.tsx
```diff
@@ -86,7 +86,7 @@ const MembersFilter = ({ onFilterChange, isRootOrAllDomains }: Props) => {
                         name={name}
                         options={options}
                         label={label}
-                        handleFilterChange={onFilterChange}
```
willm30, line 89: Like here for example @jakubcolony. What would be the correct behaviour?

The component wraps our Select component, which exposes an `onChange` prop.

willm30: It seems a bit odd that currently the handle is being defined as `handle`, then passed as `on`, then being passed again as `handle`

src/components/common/Members/filterMembers.ts
willm30: Yep, changed it back. Agree that it is also the convention (which can be frustrating as it's harder to read).

src/components/common/ColonyMembers/ColonyMembers.tsx
jakubcolony: The idea behind the rule in living standard was the same as the rule you linked, that is to enforce all event handlers parsed as prop to start with `on`, and the handlers defined within the component file to start with `handle`. The benefits I see are that it's instantly clear what type of handler you're dealing with, and whenever you need to write code like this:
```js
const handleChange = () => {
  // call handler passed as prop
  onChange();
};
```
If that rule wasn't followed, you'd most likely need to find some alternative naming, which isn't great.

I tried playing around with enabling the eslint rule but it seems like it catches too many cases, e.g.:
```js
  const { updateWallet } = useAppContext();

   <ActionButton
     onSuccess={updateWallet}
   />
```
Yields an error because it expects `updateWallet` to start with `handle` which doesn't feel justified for such a trivial case. The crucial thing is that the `onSuccess` prop is named appropriately.

Thanks for highlighting the potential confusion, I modified the section in living standards: https://github.com/JoinColony/ColonyFrontEndLivingStandard/pull/16

src/components/common/ColonyMembers/MembersFilter/MembersFilter.tsx
jakubcolony:
```js
onFilterChange={onFilterChange}
```

Since both are props

PR #756, opened 2023-07-06
Fix duplicate contributor in members query

willm30:
## Description

dupl


amplify/backend/function/getMembersForColony/src/index.js
```diff
@@ -121,177 +121,142 @@ exports.handler = async (event) => {
...
-      watchers.push({
-        address: item.user.id,
-        user: item.user,
-      });
-    }
-  });
-
-  // There will be no Watchers outside of the root domain
-  if (domainId > Id.RootDomain) {
-    watchers.length = 0;
```
willm30, line 162: This is just weird.

amplify/backend/function/getMembersForColony/src/index.js
```diff
@@ -121,177 +121,142 @@ exports.handler = async (event) => {
...
+        address: user.id,
+        user,
+      })),
     };
   }
 
-  // Identify watchers & contributors
-  data?.getColonyByAddress?.items[0]?.watchers.items.forEach((item) => {
-    if (
-      addressesWithReputation?.some(
```
willm30, line 144: This double loop is slower than using a hash set. m * n versus m + n

amplify/backend/function/getMembersForColony/src/index.js
```diff
@@ -121,177 +121,142 @@ exports.handler = async (event) => {
...
+    // Or it's a subdomain, in which case we can return two empty arrays since you can't watch a subdomain
+    return {
+      contributors: [],
+      watchers: [],
+    };
   }
 
-  // get reputation for each address
-  const contributorsWithReputation = await Promise.all(
-    contributors.map(async (contributor) => {
```
willm30, line 196: This logic was hard to follow, but basically it handles calculating the user's reputation in the domain. I have simplified.

amplify/backend/function/getMembersForColony/src/index.js
```diff
@@ -121,177 +121,142 @@ exports.handler = async (event) => {
...
-  });
-
-  // There will be no Watchers outside of the root domain
-  if (domainId > Id.RootDomain) {
-    watchers.length = 0;
-  }
-
-  // now catch addresses that have reputation but are not in the watchers list
-  // i.e. address that was awarded reputation but has not joined the colony
-  if (addressesWithReputation?.length !== contributors?.length) {
```
willm30, line 167: I presume this is responsible for my duplicate contributor, probably combined with some specific user flow like being awarded rep while not joined to a colony and then joining... or something like that. It's hard to tell. 

I have changed so that we generate contributors straight from the source of truth provided by the chain, instead of trying to splice together two separate lists. This should solve the problem.

amplify/backend/function/getMembersForColony/src/index.js
```diff
@@ -68,209 +67,196 @@ exports.handler = async (event) => {
...
+
   if (errors || !data) {
     const [error] = errors;
     throw new Error(
       error?.message || 'Could not fetch colony data from DynamoDB',
     );
   }
 
-  // Incase there are members but no reputation on the colony,
-  // we can stop here and return the list of watchers
```
willm30, line 125: Contributors can also have no rep if they have permissions, therefore we can't terminate early here

jakubcolony (approved):
Looks good, nice fix! 💪

ArmandoGraterol (approved):
Excellent work! 

Do we have an issue to fix the members not displaying the permission labels in the list? This is how the list looks in the Dapp in prod:
FireShot Capture 121 - Members - Colony - Metacolony - xdai colony io

Also, toggling the permission sorter doesn't do anything for me

willm30:
@ArmandoGraterol Yep, keeping track of both things you mention in the QA project 👌

PR #762, opened 2023-07-07
Store advanced user settings in DB

### LINKED ISSUE #584: Store user's advanced settings in the DB
Currently they're stored in local storage, meaning they won't persist between different machines and will get wiped when site data is cleared

thomassdk:
## Description

This PR saves the users advanced settings in the DB along with the rest of their settings.

**Changes** 🏗

* The schema has been extended to add these user settings to the profile model
* `metatransactions` has been renamed to `metatransactionsEnabled` to reference that it is a boolean value and match the naming strategy used for `decentralizedModeEnabled`
* `getCanUserSendMetatransactions` has been refactored to use the apollo context to fetch the user settings

**Deletions** ⚰️

* `userSettings` hook & context have been removed

Resolves #584


willm30 (approved):
LGTM

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -63,6 +63,12 @@ input CreateUniqueColonyInput {
   version: Int!
 }
 
+input ProfileAdvancedInput {
```
jakubcolony, line 66: Would be great if we could be more specific that those are advanced _settings_

src/components/common/UserProfileEdit/UserAdvancedSettings.tsx
```diff
@@ -83,23 +73,39 @@ const headingTextValues = {
   ),
 };
 
-const UserAdvancedSettings = () => {
+interface Props {
+  user: User;
+}
```
jakubcolony, line 78: Alternatively the current user could be retrieved from the `useAppContext` hook, especially since you already call it

jakubcolony (approved):
Looks good, a few minor comments left 👍

src/components/common/UserProfileEdit/UserAdvancedSettings.tsx
```diff
@@ -83,23 +73,39 @@ const headingTextValues = {
...
-      customRpc,
-    },
-    setSettingsKey,
-  } = useUserSettings();
+  const { updateUser } = useAppContext();
+  const [editUser, { error }] = useUpdateUserProfileMutation();
 
   const metatransasctionsDefault = metatransactionsAvailable
-    ? metatransactionsSetting
+    ? profile?.advanced?.metatransactionsEnabled || false
```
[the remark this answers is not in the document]
thomassdk: Personally, I find the double negation to cast a Boolean not very legible how does the team feel about the syntax
`Boolean(profile?.advanced?.metatransactionsEnabled)` @jakubcolony @willm30 ???

Also, (not that I really think it matters) but it seems the `Boolean` syntax is faster!

https://www.measurethat.net/Benchmarks/Show/11127/0/boolean-vs

src/components/common/UserProfileEdit/UserAdvancedSettings.tsx
thomassdk: This follows the pattern in `UserMainSettings` of passing down the User object as a prop.

I believe this is so that the possibility of the object being `null` or `undefined` is handled by the parent component:

(https://github.com/JoinColony/colonyCDapp/blob/3d206b51e22e22e016c1ea7a08203951ebf1b686/src/components/common/UserProfileEdit/UserProfileEdit.tsx#L36C1-L41)

otherwise it gets really annoying to constantly be dealing with the null / undefined possibility

src/components/common/UserProfileEdit/UserAdvancedSettings.tsx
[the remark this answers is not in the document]
willm30: @thomassdk `Boolean` also fine 👍

PR #771, opened 2023-07-07
Fix nodejs CI

### LINKED ISSUE #752: Fix Node.js CI typecheck
The Node.js CI fails everytime we push to the repo. There are only a couple of errors, one of which is that it can't find the colony network address, which I'm pretty sure we're good to hardcode? I.e. it doesn't change, even when developing locally? Confirm with Raul.

thomassdk:
## Description

This PR gives the Node.js CI GitHub workflow some love.

Failing typecheck tests are now passing, and removed scripts have been added back in.

Actions have also been updated and now the CI is passing and is 3 times are fast

image

Resolves #752

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -21,11 +21,6 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
-    - name: Get npm cache directory
-      id: npm-cache
-      run: |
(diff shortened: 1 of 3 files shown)


willm30 (approved):
Spot on!

ArmandoGraterol (approved):
Go Thomas go!

PR #789, opened 2023-07-11
Port control safe dialog

thomassdk:
## Description

- [] TODO write description

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

* Changes go here

**Deletions** ⚰️

* Deletions go here

## TODO

- [ ] Add TODOs


thomassdk:
The scope of this PR was expanded a bit by also bringing over the transaction types and transaction validation so that there is a common base for each transaction issue to work from

src/components/common/Dialogs/ControlSafeDialog/ControlSafeDialog.tsx
```diff
@@ -0,0 +1,82 @@
+import React from 'react';
+import { InferType } from 'yup';
+import { useNavigate } from 'react-router-dom';
+import { pipe, withMeta, mapPayload } from '~utils/actions';
+
+import Dialog, { DialogProps, ActionDialogProps } from '~shared/Dialog';
+import { ActionHookForm as Form } from '~shared/Fields';
+import { ActionTypes } from '~redux/index';
```
alicjakujawa, line 8: import { ActionTypes } from '~redux';

src/components/common/Dialogs/ControlSafeDialog/ControlSafeForm.tsx
```diff
@@ -0,0 +1,281 @@
...
+    colony,
+    requiredRoles,
+    [Id.RootDomain],
+    enabledExtensionData,
+  );
+
+  const submitButtonText = (() => {
+    return { id: 'button.confirm' };
+  })();
+
```
alicjakujawa, line 137: why we wrote this as a function and call this function immediately?

src/components/common/Dialogs/ControlSafeDialog/TransactionTypesSection/ContractInteractionSection.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+import { DialogSection } from '~shared/Dialog';
+
+const displayName = `dashboard.ControlSafeDialog.ContractInteractionSection`;
+
```
alicjakujawa, line 6: probably require update

src/components/common/Dialogs/ControlSafeDialog/TransactionTypesSection/TransferFundsSection.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+import { DialogSection } from '~shared/Dialog';
+
+const displayName = `dashboard.ControlSafeDialog.TransferFundsSection`;
+
```
alicjakujawa, line 6: we need to update this as well

src/components/common/Dialogs/ControlSafeDialog/TransactionTypesSection/TransferNFTSection.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+import { DialogSection } from '~shared/Dialog';
+
+const displayName = `dashboard.ControlSafeDialog.TransferNFTSection`;
+
```
alicjakujawa, line 6: here as well

alicjakujawa (changes requested):
Small comments and addiitonally i think we miss translation:

Screenshot 2023-07-14 at 12 04 17

src/components/common/Dialogs/ControlSafeDialog/ControlSafeForm.tsx
thomassdk: there is a bit more logic that needs to be added later that changes this text so this is a placeholder until we get around to adding that extra logic back in

ArmandoGraterol:
Can you rebase this, please? :)

alicjakujawa:
@thomassdk Would you mind rebase it before i will review/test it again?

thomassdk:
rebase complete @ArmandoGraterol and @alicjakujawa

alicjakujawa (changes requested):
The one final thing im seeing is that, when you open dialog, and you have preselected transaction type, there is no content for this section by default. So if you have by default selected RawTransaction, you need one more time select this to see the content

alicjakujawa (approved):

thomassdk:
> The one final thing im seeing is that, when you open dialog, and you have preselected transaction type, there is no content for this section by default. So if you have by default selected RawTransaction, you need one more time select this to see the content

Hmmm not sure why this is happening but it might be due the fact that this is all a temporary measure until we port the multiple transactions functionality. I'm of the opinion that we leave it for now and if it is still an issue at the end then we fix it

ArmandoGraterol (changes requested):
Good job, I've left some comments but considering the size, things are looking good!

I've noticed 2 things about the dialog:

- The `Select a safe from the menu or add a new one via Safe Control` error message is appearing in the initial render when this message should only appear if you select a transaction type without first selecting a safe. On that note, initially, there shouldn't be any transaction type selected, maybe that's why the error message appears prematurely.
- The "Wrong colony version" warning is showing, even thought my colony should be have the latest version.

Examples:
- Dapp:
FireShot Capture 124 - Actions - Colony - armando - qa-xdai colony io

- This branch:
FireShot Capture 125 - Colony CDapp - localhost

src/components/common/Dialogs/ControlSafeDialog/constants.ts
```diff
@@ -0,0 +1,47 @@
...
+  },
+  {
+    value: TransactionTypes.CONTRACT_INTERACTION,
+    label: MSG[TransactionTypes.CONTRACT_INTERACTION],
+  },
+  {
+    value: TransactionTypes.RAW_TRANSACTION,
+    label: MSG[TransactionTypes.RAW_TRANSACTION],
+  },
+];
```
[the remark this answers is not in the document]
thomassdk: good to know :) should everything go into the helpers file or should the enum stay in the constants file?

thomassdk:
Thanks for the review @ArmandoGraterol :raised_hands: 

To respond to your points:
1. This is the same issue that Alicija has noticed and think that it is due to just including the bare minimum in this PR to unblock the transaction type ports. I think it will be resolved when we port over the multi transaction functionality and so I am reluctant to put some time into fixing this now when it will just be a bodge job. Does that sound reasonable to you?
2. It seems in QA the Colony version is 11 and it is not possible to upgrade it to the minimum version of 12 that is defined in the requirements for this feature. Are your QA colonies a different version / is it possible for you to upgrade your colonies?

Screenshot from 2023-07-18 11-24-33

src/components/common/Dialogs/ControlSafeDialog/constants.ts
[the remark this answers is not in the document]
ArmandoGraterol: Yep, we even include enum in the helper files

ArmandoGraterol:
> 1. This is the same issue that Alicija has noticed and think that it is due to just including the bare minimum in this PR to unblock the transaction type ports. I think it will be resolved when we port over the multi transaction functionality and so I am reluctant to put some time into fixing this now when it will just be a bodge job. Does that sound reasonable to you?

Sounds good to me, although we could still include a fix for this bug:
> On that note, initially, there shouldn't be any transaction type selected, maybe that's why the error message appears prematurely.

I took a deeper look, and that it's not causing the error message to appear but it's still a bug we can solve if we include this fix that we did in the Dapp inside the `Select` component: https://github.com/JoinColony/colonyDapp/commit/d2670474d8d8b0ce36ef2a8ca7f097dfdc8bbff9

With this fix, the placeholder of the `Select` should show when no transaction type is selected :)

> 2. It seems in QA the Colony version is 11 and it is not possible to upgrade it to the minimum version of 12 that is defined in the requirements for this feature. Are your QA colonies a different version / is it possible for you to upgrade your colonies?

Oh interesting, I didn't consider that the CDapp is behind in the colony version. You can disregard my feedback then as this is something that should be addressed in another issue/PR.

ArmandoGraterol (approved):
Nice work porting over this huge chunk comrade! 

If we include the fix for `Select` then I think this should be good to go, so I'm pre-approving 👌

PR #814, opened 2023-07-14
fix: added selecting member to manage modal, fixed width on mobile

CzarekDryl:
Added selecting member to manage modal, member context, fixed width on mobile


src/components/v5/shared/MembersSelect/types.ts
```diff
@@ -1,5 +1,5 @@
 import { User } from '~types';
 
 export interface MemberSelectProps {
-  user?: User;
+  user?: User | null;
```
krawczykkatarzyna, line 4: why null?

src/context/MemberContext.tsx
```diff
@@ -0,0 +1,39 @@
...
+}>({
+  isMemberModalOpen: false,
+  setIsMemberModalOpen: noop,
+  setUser: noop,
+  user: undefined,
+});
+
+export const MemberContextProvider: FC<PropsWithChildren> = ({ children }) => {
+  const [isMemberModalOpen, setIsMemberModalOpen] = useState(false);
+  const [user, setUser] = useState<User | undefined | null>(undefined);
```
krawczykkatarzyna, line 27: <User | undefined | null> - why :P ?

src/components/v5/shared/MembersSelect/types.ts
CzarekDryl: It's null because this type comes from graphql and it's adding null by default

krawczykkatarzyna (approved):

PR #825, opened 2023-07-17
Add raw transaction section

### LINKED ISSUE #735: Port "Raw Transaction" transaction type section
Clarifications:
- This section contains a user picker that, as usual, should only contain the verified (or not if there's no whitelist) users.
- The Value field should only allow WEI values.  (It can be 0)
- The Data field should only allow hex strings.

Image

Image

alicjakujawa:
Port over the raw transaction section. The validation should work as we expect (only allow WEI and hex strings in their respective fields, for example.)

FireShot Capture 133 - Colony CDapp - localhost

Resolves #735


ArmandoGraterol (approved):
Wow! What an incredible PR.

Truly sublime.

thomassdk (approved):
LGTM

PR #840, opened 2023-07-19
Implement support for "Manage Safes" actions in action page and list

### LINKED ISSUE #743: Add support for adding/removing safes actions in the action page and action list
Clarifications:
- The event to look for in the case of adding/removing a safe is `ColonyEdit`.

Image

Image

Image

ArmandoGraterol:
Add support to the action page and action list for the actions related to managing safes.

To test this PR, try to add and remove safes through the UAC and ensure everything on the action page and the list is wired up and makes sense.

- Action page when you remove 1 safe:
FireShot Capture 131 - Colony CDapp - localhost

- Action page when you remove more than 1 safe:
FireShot Capture 128 - Colony CDapp - localhost

About this action page, I had to do some compromises to show all of the removed safes in the details widget. It's not like in the design where each removed safe is 1 detail item with its own label (Like the screenshot above this one), that's because by the way we implemented the details widget this time around in the CDapp, it's not really straightforward (the solution is either hacky or a rework of the section.) to make 1 detail item repeat for each removed safe.

- Action page when you add 1 safe:
FireShot Capture 130 - Colony CDapp - localhost

- Action list with all of the possible variations of the actions:
FireShot Capture 129 - Colony CDapp - localhost

Resolves #743


thomassdk (approved):
Great work, everything looks kosher to me :raised_hands:

PR #853, opened 2023-07-21
fix: modal width, text color, select styles

CzarekDryl:
Fixed modal width, description text color, select styles


src/components/v5/common/Fields/Select/Select.module.css
```diff
@@ -3,17 +3,17 @@
...
+  @apply w-full pl-3.5 pr-4 py-3 rounded bg-base-white flex justify-between items-center text-md font-normal transition-all duration-normal border;
 }
 
 .options {
   @apply border-gray-200 hidden;
 }
 
 .show {
-  @apply flex flex-col py-2 px-3.5 border border-gray-300 rounded relative left-0 z-[1] bg-base-white w-full mt-1 max-h-[12.5rem] overflow-x-scroll;
+  @apply flex flex-col py-2 px-2 border border-gray-300 rounded relative left-0 z-[1] bg-base-white w-full mt-1 max-h-[12.5rem] overflow-x-auto;
```
KatarzynaStefanowiczPagepro, line 14: I think we have to remove this style overflow-x-auto at all. I did this but in the other PR.

KatarzynaStefanowiczPagepro: Did you check this component (UeerHubMobile) after changes with paddings?
image

src/components/v5/common/Modals/ManageMemberModal/ManageMemberModal.tsx
```diff
@@ -60,6 +60,7 @@ const ManageMemberModal: FC<ManageMemberModalProps> = ({
         list={manageMemberActions}
         placeholderText={{ id: 'members.modal.selectActions' }}
         isListRelative
+        openButtonClass="outline-blue-100 outline !outline-[0.1875rem] border-blue-200"
```
KatarzynaStefanowiczPagepro, line 63: Do you need outline and !outline-[0.1875rem]? Maybe outline-[0.1875rem] will be enough here?

src/components/v5/shared/MembersSelect/MembersSelect.tsx
```diff
@@ -35,6 +35,7 @@ const MembersSelect: FC<MemberSelectProps> = ({ user }) => {
       list={members}
       placeholderText={user ? undefined : { id: 'members.modal.selectMember' }}
       showAvatar
+      openButtonClass="outline-blue-100 outline !outline-[0.1875rem] border-blue-200"
```
KatarzynaStefanowiczPagepro, line 38: The same.

Also maybe we should create a special class in global for these styles because they are repeatable in a few components.

src/components/v5/common/Fields/Select/Select.tsx
```diff
@@ -18,6 +18,7 @@ const Select = <T extends any[]>({
   isLoading,
   isListRelative,
   showAvatar,
+  openButtonClass,
```
krawczykkatarzyna, line 21: please use different name

src/components/v5/common/Fields/Select/Select.module.css
CzarekDryl: We have to leave overflow-x-auto because we are reducing max-height. UserHubMobile works fine, because I've added missing padding to item.

image

krawczykkatarzyna (approved):

KatarzynaStefanowiczPagepro (approved):

PR #856, opened 2023-07-21
Port contract interaction transaction

### LINKED ISSUE #736: Port "Contract interaction" transaction type section
Clarifications:
- This section displays a contract address picker. It's not populated, the user should always manually enter the target contract.
- This section displays an ABI/JSON text field which is automatically populated with the ABI of the selected contract address. Otherwise, the user can also manually edit/copy-paste the ABI.
- This section displays a function selector. This function selector is automatically populated with contract functions if the ABI in the text field is valid and has functions that can be used.
- Once a function is selected, if that function has params, those should dynamically be added to the form (validation included) in order to allow the user to modify them.
- To get ABIs of contract addresses, we use etherscan's API.
- If the contract is not verified or the ABI couldn't be found, an error should appear.

Image
Image
Image

thomassdk:
## Description

- This PR ports the contract interaction functionality of safe control ~It doesn't port over the validation~ including validation

[Screencast from 2023-08-01 16-06-07.webm](https://github.com/JoinColony/colonyCDapp/assets/32598350/3df8bd64-16d3-4f5e-ad8d-07a2f3fecd47)

**New stuff** ✨

* Two new env vars have been added to the `.env.example` 
```
BSCSCAN_API_KEY=<bscscan-api-key>
ETHERSCAN_API_KEY=<etherscan-api-key>
```

these need to be added to test this PR

Resolves #736

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -15,3 +15,7 @@ METATRANSACTIONS=true
 # The endpoint of the reputation oracle.
 # DOES NOT NEED TO BE SET WHEN DEVELOPING LOCALLY
 REPUTATION_ORACLE_ENDPOINT=
+
+# Required for safe control
+BSCSCAN_API_KEY=<bscscan-api-key>
+ETHERSCAN_API_KEY=<etherscan-api-key>
--- a/src/components/common/CreateColonyWizard/StepColonyName.tsx
+++ b/src/components/common/CreateColonyWizard/StepColonyName.tsx
@@ -57,11 +57,7 @@ const StepColonyName = ({
             appearance={{ weight: 'medium' }}
           />
           <div className={styles.nameForm}>
-            <NameInputs
-              disabled={isSubmitting}
-              isMobile={isMobile}
-              cleaveDefaultValue={defaultValues.colonyName}
-            />
+            <NameInputs disabled={isSubmitting} isMobile={isMobile} />
             <SubmitFormButton
               disabled={!isValid || isSubmitting}
               loading={isSubmitting}
--- a/src/components/common/CreateColonyWizard/StepColonyNameInputs.tsx
+++ b/src/components/common/CreateColonyWizard/StepColonyNameInputs.tsx
@@ -39,13 +39,11 @@ const formatting = {
 interface StepColonyNameInputsProps {
   disabled: boolean;
   isMobile: boolean;
-  cleaveDefaultValue: string;
 }
 
 const StepColonyNameInputs = ({
   disabled,
   isMobile,
-  cleaveDefaultValue,
 }: StepColonyNameInputsProps) => (
   <>
     <Input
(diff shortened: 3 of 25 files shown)


thomassdk:
There is an issue here where if the user has selected a contract and then goes to re select but doesn't choose anything, that contract disappears but all the data remains. 

[Screencast from 2023-08-01 16-10-30.webm](https://github.com/JoinColony/colonyCDapp/assets/32598350/617c0df2-9107-438b-a666-d22bf6871a7c)

What should be the correct behaviour in this case?

ArmandoGraterol:
> What should be the correct behaviour in this case?

 I think we can leave it as it is. This is also how the original dialog in the Dapp behaves and we haven't received any complaints about it :) I don't think it hinders the UX.

thomassdk:
> > What should be the correct behaviour in this case?
> 
> I think we can leave it as it is. This is also how the original dialog in the Dapp behaves and we haven't received any complaints about it :) I don't think it hinders the UX.

i which case this PR is ready for your eagle eyes now that the niggling validation bug has been fixed

ArmandoGraterol (changes requested):
Nice work porting over the most complex part of the dialog :)

For now, I've found some weird behavior.

- If I remove the section above the contract interaction section then the function parameters input disappear:
[26cea28f-39b8-4c2e-9cc0-c10543f8c2b6.webm](https://github.com/JoinColony/colonyCDapp/assets/18473896/70c54ad2-ca22-43cb-b78a-cd7e1cbc5a23)

- If I remove a contract interaction section then the one above loses the values on their param inputs:
[1660c3a0-c662-4e84-8bf2-657aa65a1c27.webm](https://github.com/JoinColony/colonyCDapp/assets/18473896/fe5a2e1c-6293-40a8-b5a1-49518ebc1f47)

That's all I've found for now but knowing this form, I wouldn't be surprised if we found more edge cases :)

thomassdk:
@ArmandoGraterol ive fixed that bug that you found and found another one as well that took me a while to get to the bottom of!

[Screencast from 2023-08-09 17-48-19.webm](https://github.com/JoinColony/colonyCDapp/assets/32598350/ff45128c-5597-48c8-b588-17cf18366acb)


ive also found another bug but this seems to do more with the transaction header adding and removing tabs i think it would be best to open another issue to track it and handle it separately what do you reckon?

[Screencast from 2023-08-09 17-49-09.webm](https://github.com/JoinColony/colonyCDapp/assets/32598350/dbc0cae8-f36c-45cc-8a22-3f7770fed298)

ArmandoGraterol (approved):
Great work 🎉

ArmandoGraterol:
> ive also found another bug but this seems to do more with the transaction header adding and removing tabs i think it would be best to open another issue to track it and handle it separately what do you reckon?

I can't see what's the bug in the video you posted, is it that the tabs are not being removed when you click? Anyways, yes we could create an issue for this 👍

PR #858, opened 2023-07-22
Bump colony version to v12

ArmandoGraterol:
This PR bumps the commit of `colonyNetwork` to support v12. This version is required to allow meta transactions for Control Safe.

To test this PR, we just need to make sure nothing broke and that the warning in the Control Safe dialog disappeared :)

### DIFF
--- a/docker/colony-cdapp-dev-env-network
+++ b/docker/colony-cdapp-dev-env-network
@@ -1,6 +1,6 @@
 FROM colony-cdapp-dev-env/base:latest
 
-ENV NETWORK_HASH=0b78037610a96c0f27b1d78122b467a6c413afae
+ENV NETWORK_HASH=618722ed9ae30900ab717e36621c57289c363042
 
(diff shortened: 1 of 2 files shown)


thomassdk (changes requested):
This looks great but we just have to undo the inverted `disabledInputs` in the `ControlSafeForm` component for this to work.

https://github.com/JoinColony/colonyCDapp/blob/52643a28ae680c53ddfcd3739d983d339aef8221/src/components/common/Dialogs/ControlSafeDialog/ControlSafeForm.tsx#L124-L129

ArmandoGraterol:
@thomassdk 
> This looks great but we just have to undo the inverted `disabledInputs` in the `ControlSafeForm` component for this to work.

Can you clarify? I don't get how reverting it would fix the warning not appearing when it shouldn't.

thomassdk (approved):
Sublime work, everything is working well for me

PR #861, opened 2023-07-24
Fix: Wallet switching to Metamask when page refreshed

jakubcolony:
## Description

This PR fixes a bug where the wallet would always switch to Metamask when page is refreshed, even when the dev wallet is connected.

## Testing

Connect any of the dev wallets and refresh the page. You should see the connected wallet staying the same, instead of switching to Metamask.


willm30 (approved):
LGTM

PR #885, opened 2023-07-31
Port Transfer NFT section into control safe dialog

### LINKED ISSUE #851: Port "Transfer NFT" transaction type
Clarifications:
- If we have a safe selected and the transaction type is "Transfer NFT", we should fetch all the NFTs within the safe using the endpoint `/v2/safes/{address}/collectibles`
- The NFT selector should only contain NFTs inside the safe.
- If there are no NFTs, then we should display a message stating so.
- If the fetching failed, we should display a message letting the user know.

Image

Image

Image

Image

Image

ArmandoGraterol:
- The Ethereum safe should have 1 NFT to test, which is the one I've used in the screenshots. 
- The Binance safe shouldn't, so an appropriate message should appear when we select it.

FireShot Capture 136 - Colony CDapp - localhost

FireShot Capture 137 - Colony CDapp - localhost

FireShot Capture 138 - Colony CDapp - localhost

Resolves #851


thomassdk (approved):
Once again a stellar job Armando, just a minor comment about an untyped function

PR #901, opened 2023-08-02
Bump colony js version in lambdas

willm30:
Bump colony js version in lambdas to use latest.


src/cache/colony/balances.ts
```diff
@@ -113,7 +113,7 @@ const balancesFieldCache = {
               domain: null,
               id: `${colonyChainId}_${colonyAddress}_${COLONY_TOTAL_BALANCE_DOMAIN_ID}_${token.id}_balance`,
               token,
-              balance: totalTokenBalances[token.id].toString(),
+              balance: totalTokenBalances[token.id]?.toString(),
```
jakubcolony, line 116: 👍

PR #902, opened 2023-08-02
Add ipfs to annotations

willm30:
**Note**: Large diff due to package.lock

## Description

This PR adds all of the library code from the dapp necessary to store annotations in ipfs. 

It tests with the mintTokens motion. 

Test with: https://github.com/JoinColony/colonyCDapp/pull/909


src/redux/actionTypes.ts
```diff
@@ -191,4 +191,10 @@ export enum ActionTypes {
   META_UNWRAP_TOKEN = 'META_UNWRAP_TOKEN',
   META_UNWRAP_TOKEN_ERROR = 'META_UNWRAP_TOKEN_ERROR',
   META_UNWRAP_TOKEN_SUCCESS = 'META_UNWRAP_TOKEN_SUCCESS',
+
+  // IPFS
+
```
thomassdk, line 196: why does this comment header differ from the other ones in this file?

thomassdk (approved):
:confetti_ball:

PR #905, opened 2023-08-02
Wire all sagas with annotations

willm30:
## Description

This pr wires up annotations with all actions and motions. Test with: https://github.com/JoinColony/colonyCDapp/pull/909


src/redux/sagas/transactions/createTransaction.ts
```diff
@@ -99,13 +99,17 @@ export interface ChannelDefinition {
...
+export type TransactionChannel = {
+  channel: Channel<any>;
+  index: number;
+  id: string;
+};
+
 export function* createTransactionChannels(
   batchId: string,
   ids: string[],
   customIndex = 0,
```
thomassdk, line 111: why not extract all of the types outside the function? like so:

```ts
export type TransactionChannel = {
  channel: Channel<any>;
  index: number;
  id: string;
};

type TransactionChannelMap = { [id: string]: TransactionChannel };

export function* createTransactionChannels(
  batchId: string,
  ids: string[],
  customIndex = 0,
): IterableIterator<TransactionChannelMap> {
```

thomassdk (approved):
LGTM

PR #913, opened 2023-08-03
Retrieve expenditure balance

### LINKED ISSUE #911: Expenditure balance retrieval
This issue is to explore and implement a way for the UI to retrieve the current balance of the expenditure's funding pot. This is needed to decide if user should be allowed to claim an expenditure payout.

Two options considered so far:
1. Picking up ColonyFundsMovedBetweenFundingPots and PayoutClaimed events, checking if the funding pot belongs to an expenditure (by querying the database), and if so, updating that expenditure entry with the current balance
2. Checking the expenditure balance on the go, using a lambda function

jakubcolony:
## Description

This PR adds `fetchExpenditureBalances` lambda that allows to retrieve the total balance of the expenditure's funding pot grouped by token address.

Resolves #911


amplify/backend/.eslintrc.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  ignorePatterns: ['*'],
+};
```
jakubcolony, line 3: Disabling eslint in lambdas

amplify/backend/function/fetchExpenditureBalances/src/index.js
```diff
@@ -0,0 +1,78 @@
...
+      tokenRequiredAmounts.set(
+        tokenAddress,
+        BigNumber.from(amount).add(currentAmount).toString(),
+      );
+    });
+  });
+
+  // Get the balances for each token address
+  const balances = await Promise.all(
+    [...tokenRequiredAmounts.entries()].map(
```
willm30, line 61: nit: `tokenRequiredAmounts.entries().map` would be equivalent here, but without the extra iteration

willm30 (approved):
LGTM 👍👍

rdig (approved):
This is all good, can't even find anything proper to nitpick about :sweat_smile: 

I thought the `balances` prop on the `Expenditures` model would cause a naming conflict with our local Apollo cache, but turns out that's not the case.

So this is all good to go!

amplify/backend/function/fetchExpenditureBalances/src/index.js
jakubcolony: @willm30 Map's `entries()` method returns an `IterableIterator` rather than an array (unlike Object's `entries()`). If I want to use `map` I need to convert it into an array first

amplify/backend/function/fetchExpenditureBalances/src/index.js
```diff
@@ -0,0 +1,78 @@
...
+    } = require('../../../../mock-data/colonyNetworkArtifacts/etherrouter-address.json');
+    networkAddress = etherRouterAddress;
+  }
+};
+
+/**
+ * @type {import('@types/aws-lambda').APIGatewayProxyHandler}
+ */
+exports.handler = async (event) => {
+  setEnvVariables();
```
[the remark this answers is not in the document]
jakubcolony: thanks 👍

PR #925, opened 2023-08-07
Claiming expenditure payout

### LINKED ISSUE #912: Claiming an expenditure payout
If an expenditure is finalized, there should be a button allowing to "Release" the funds. The new UI calls for the "Release" button to trigger the claim of _all_ payouts, not just a single one so the saga should account for that.

jakubcolony:
## Description

This PR adds claiming expenditure payout functionality.

After finalization, you should see a Claim funds button appear that lets you claim all the payouts for which the claim delay has elapsed (see below for more info on how claim delay works in dev).

## Testing

1. Make sure you restart your entire dev environment as there was a network version upgrade.
2. Create and advance expenditures through all the stages to finalization.
3. Attempt to claim the funds. Note: It is expected that attempts to claim a locked token (e.g. from another colony) will fail.
4. If you set your wallet address as a recipient, you should see your balance increase by nearly the entire amount of all your payouts (less the network fee).

Scenarios I recommend testing:
- Multiple payouts
- Edited payouts (especially removing one of the payouts)
- Payouts with different claim delays

### Note on claim delay in dev

Time works differently in dev blockchain. This makes testing a bit tricky, since the UI uses the real time, whereas the dev blockchain time is "frozen" until the next block is mined. 

The consequence of this is that when you finalize an expenditure with non-zero claim delay, the UI should show the correct time at which you can claim the next payout, e.g.:
image

But you won't be able to claim it until another block is mined and the blockchain time updates. To force mining a block, run the following command in truffle:

```js
web3.currentProvider.send({method: "evm_mine", params: []}, () => {})
```

You can also advance time by an arbitrary number of seconds:

```js
web3.currentProvider.send({method: "evm_increaseTime", params: [<seconds>]}, () => {})
// You need to mine a block after increasing time
web3.currentProvider.send({method: "evm_mine", params: []}, () => {})
```

Resolves #912


jakubcolony:
Marking as blocked by the bug in network - waiting for https://github.com/JoinColony/colonyNetwork/pull/1154 to be merged and deployed

willm30 (commented):
What's the expected behaviour when the recipient is a colony? I am able to click claim on the expenditure (even though I'm not the recipient -- is that right?), but I don't see an incoming funds claim in the colony.

jakubcolony:
@willm30 The button is supposed to trigger the claim of all claimable payouts, even if you're not the recipient, so that sounds correct

The issue with incoming funds not showing up is most likely due to the fact that we are not storing fund claims from expenditures in the DB yet - I'm about to speak to Raul about this but in the meantime you can verify that the funds have actually been transferred in truffle (after you claim the expenditure):
```js
// the recipient colony
c = await IColony.at()
ta = await c.getToken()
c.claimColonyFunds(ta)
```
That should update the colony balance. 

Thanks for spotting this, it would've gone under the radar otherwise. Everything else should work now so ready for you to have another look 👍

src/components/common/Expenditures/ExpenditureDetailsPage/ExpenditureClaimButton/ExpenditureClaimButton.tsx
```diff
@@ -0,0 +1,68 @@
...
+    return <div>There are no more payouts to claim</div>;
+  }
+
+  let nextClaimableAt: number | null = null;
+  const claimableSlots = expenditure.slots.filter((slot) => {
+    const claimableFrom = new Date(
+      ((expenditure.finalizedAt ?? 0) + (slot.claimDelay ?? 0)) * 1000,
+    ).getTime();
+
+    if (Date.now() >= claimableFrom) {
```
willm30, line 35: I see what you mean about testing this being tricky. I suppose the solution is to only advance block time by a few seconds, so system time can catch up. I guess it's not a problem at all in production since block time can't be meddled with.

scripts/upgrade-colony-js.sh
willm30, line 1: 🙌

willm30 (approved):
LGTM. I confirmed the tokens were appearing in the colony I was sending them to, but had to make a normal payment and claim it that way for the funds to be claimed and appear in the UI. Not sure why just doing it in ganache alone wasn't sufficient. 

Everything else working as expected 👍 (although testing the claim delay had me scratching my head for a minute 😅 My fault for advancing block time by a few hours!)

src/components/common/Expenditures/ExpenditureDetailsPage/ExpenditureClaimButton/ExpenditureClaimButton.tsx
willm30: We can get the actual block time like we do in the `fetchMotionTimeoutPeriods` lambda but it's an async call and I suppose an anti-pattern for us. And there shouldn't be a difference between block time and date.now() in production, right?

jakubcolony: I wanna say it shouldn't make a difference (and the worst case is that the transaction will fail during a small time window) but we'll only know once it's deployed to a real chain

PR #930, opened 2023-08-08
Editing an expenditure while in Draft status

### LINKED ISSUE #896: Editing an expenditure while in Draft status
To the expenditure details page, add a form allowing to edit the recipient(s) while the expenditure is in a Draft status.

Blocked by #893 and #892

jakubcolony:
## Description

This PR adds the ability to edit expenditure payouts (recipients, tokens and amounts) while the expenditure is in Draft status.

## Testing
I recommend testing the following scenarios:
### Creating an expenditure
There's been a bit of refactoring to the `createExpenditure` saga to make it call `setExpenditureValues` contract function, so it's worth testing that creating an expenditure with different recipients, tokens, and amounts still works.

### Adding new payouts while in Draft status
From expenditure details page, try adding new payouts. They should appear after page refresh.

### Removing payouts while in Draft status
Try removing some of the existing payouts. Since there's no option to remove slots/payouts on a contract level, the saga will set the amount of the payout you wish to remove to 0. All payouts with 0 amount are then ignored in the UI.

### Changing recipient, token address, and amount of the existing payouts
Changing recipients and amounts is very straightforward, whereas when changing token address we need to use the trick with setting amount to 0 described above.

### Handling states that are possible to reach via contract interaction, but not the UI
The UI dictates there should only be one payout per slot (so that we can be display them one per row), but the contracts allow multiple payouts per slot, as long as they have different token addresses.
We should be able to handle this situation while editing by transforming the payouts in such a way there's only one per slot.
To test, create an expenditure with one payout and run the following in truffle:
```
c = await IColony.at('<your colony address>')
c.setExpenditurePayouts(<expenditure native id>, [1], '<token address>', ['1000000000000000000'])
```
Note the `<token address>` should be different from the one used when creating the expenditure. This should add another payout to the only existing slot of the expenditure:
image
Simply submitting the edit form should "move" the second payout to its own slot (note the slot id in the screenshot):
image
You can also try modifying some values.

Resolves #896


amplify/backend/function/fetchExpenditureBalances/src/index.js
```diff
@@ -58,8 +58,9 @@ exports.handler = async (event) => {
 
   // Get the balances for each token address
   const balances = await Promise.all(
-    [...tokenRequiredAmounts.entries()].map(
-      async ([tokenAddress, requiredAmount]) => {
+    [...tokenRequiredAmounts.entries()]
+      .filter(([_, requiredAmount]) => BigNumber.from(requiredAmount).gt(0))
+      .map(async ([tokenAddress, requiredAmount]) => {
```
jakubcolony, line 63: Ignoring token addresses with required amount of 0

willm30 (commented):
I'm getting the following contract error when changing the amount:
underflow

To reproduce: 

-> Create an expenditure with 2 recipients
-> Edit the expenditure and delete one of the recipients
-> Save changes and refresh. For the recipient that remains, try to edit the amount (as shown in the gif)

This also happens when trying to edit the token address in the same scenario. Everything else works as described 👌

src/redux/sagas/utils/expenditures.ts
```diff
@@ -0,0 +1,71 @@
...
+    // 2-dimensional array mapping token addresses to amounts
+    [...payoutsByTokenAddresses.values()].map((payoutsByTokenAddress) =>
+      payoutsByTokenAddress.map((payout) =>
+        BigNumber.from(payout.amount).mul(
+          // @TODO: This should get the token decimals of the selected token
+          BigNumber.from(10).pow(DEFAULT_TOKEN_DECIMALS),
+        ),
+      ),
+    ),
+  ];
```
rdig, line 70: Holy hell. Good thing you left comments

Also I think this is the place where the error that Will highlighted originates, since it's a `math-ds` library error, which usually is due to some params being passed in wrong.

Anyway, ask Alex as he most likely has the answer for that.

rdig (approved):
Nothing sticks out, nice work.

I'm approving it, but just note that I'm experiencing the same `math-ds` lib error that Will recorded

jakubcolony:
@willm30 Thanks for spotting this issue, I'm leaning towards thinking this is an issue with the way I remove the payouts (by setting the amount to 0). I'll talk to Alex about that at the earliest opportunity and open a separate PR with a fix, in the meantime I'll merge this PR as everything else works.

Issue #918, opened 2023-08-07
Store all `UInt256` in DB as string


willm30:
In Solidity, one of the data types is `UInt256` (unsigned 256-bit integer, i.e. a very large number). In Javascript, the `number` data type can only represent integers up to 2^53 – 1 ([source](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER)). 

Therefore, if we store `UInt256` values in the database as `Int`, implying a conversion to a javascript `Number` type beforehand, we run the risk of overflow. Generally, this risk is quite small, but nevertheless, it would be an easy win to avoid it entirely by storing all `UInt256` values as strings in the db.

This issue is to enforce this consistently across the CDapp and block ingestor by ensuring that every value contract-level `UInt256` is stored as a `String`, not an `Int`, in the db. 

A hint for what to look for in the block ingestor would be all event arguments that are assumed to be a `BigNumber`. This is highly likely to be a `UInt256` on the contracts.

rdig:
Closing as not relevant anymore

PR #943, opened 2023-08-10
Setting expenditure claim delay

### LINKED ISSUE #927: Setting expenditure claim delay
This issue is to add the ability to set an expenditure slot's claim delay. It should be possible to modify the delay during both expenditure creation and while the expenditure is in Draft status.

Blocked by #896

jakubcolony:
## Description

This PR modifies the expenditure form to allow setting claim delay both while creating and updating an expenditure.

## Testing
Creating an expenditure: Try setting claim delay for one or more recipients when creating an expenditure. You should see the correct delays displayed next to slots on the expenditure details page.

Editing an expenditure: On the expenditure details page, while an expenditure is in Draft status, modify some of the claim delays. After refreshing the page, you should see the updated values.

Note there's no validation of the form inputs at the moment, so entering anything that's not a number will break things.

Resolves #927

### DIFF
--- a/src/components/common/Expenditures/ExpenditureForm/helpers.ts
+++ b/src/components/common/Expenditures/ExpenditureForm/helpers.ts
@@ -12,6 +12,7 @@ export const getInitialPayoutFieldValue = (
   recipientAddress: '',
   tokenAddress,
   amount: '0',
+  claimDelay: 0,
 });
 
 export const getExpenditurePayoutsFieldValue = (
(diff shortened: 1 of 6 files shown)


willm30 (approved):
LGTM

src/components/common/Expenditures/ExpenditureForm/helpers.ts
```diff
@@ -12,6 +12,7 @@ export const getInitialPayoutFieldValue = (
   recipientAddress: '',
   tokenAddress,
   amount: '0',
+  claimDelay: 0,
```
rdig, line 15: Nitpick. Should these two values e of the same type? Ie: either `int`'s or `string`'s

rdig (approved):
All good

src/components/common/Expenditures/ExpenditureForm/helpers.ts
jakubcolony: There's an issue to store all `uint256` values into strings: https://github.com/JoinColony/colonyCDapp/issues/918

PR #951, opened 2023-08-15
Port safe transaction event list items

### LINKED ISSUE #742: Update action list with safe transaction event
- Status should show "Action needed" when the safe transaction hasn't been executed.
- Otherwise, show "Completed".
- Every action list item should be prefixed with a `Safe transaction: ` and then the custom title of the transaction.
- In the Dapp, we look for `ArbitraryTransaction` events to represent safe transactions.

Image

thomassdk:
## Description

This PR ports over the action list item for Safe Transactions

image

image

## TODO

~- [ ] Hook up dynamic checking for whether a safe transaction is complete (should this go in the block ingestor?)~

The ability to check whether a safe transaction is complete will be completed in a separate issue #966

Resolves #742


src/components/common/ColonyActions/ActionsListItem/ActionsListItem.tsx
```diff
@@ -69,6 +76,27 @@ const ActionsListItem = ({
...
+        theme:
+          safeTransactionStatus === TRANSACTION_STATUS.PENDING
+            ? 'golden'
+            : 'primary',
+        colorSchema: 'fullColor',
+      }}
+    />
+  );
+
+  const tag = safeTransaction ? SafeTag : <MotionTag />;
```
ArmandoGraterol, line 98: FYI: Both tags should be shown in the case of safe transactions that are motions. You can check this issue for examples: https://github.com/JoinColony/colonyCDapp/issues/963

In the scope of this PR, making sure that the tag looks good with motion is not a concern so you can leave this in if you want. I'm just saying this so you are aware that this will be updated in the future :)

ArmandoGraterol (approved):
Great job! I did notice that the initiator was the colony instead of the user but that's my fault :) I've fixed it [here](https://github.com/JoinColony/block-ingestor/pull/116/commits/937f36f61cc7a4d4155df6069adfc8b34a7f6deb) already and tested it with this PR. You don't need to include it, I'm mentioning it just so you know.

FireShot Capture 139 - Colony CDapp - localhost

src/components/common/ColonyActions/ActionsListItem/ActionsListItem.tsx
```diff
@@ -69,6 +76,27 @@ const ActionsListItem = ({
 
   const status = ListItemStatus.Defused;
 
+  const fromDomain = safeTransaction
+    ? initiatorColony?.domains?.items[0]
+    : itemDomain;
```
ArmandoGraterol, line 81: According to my review about the colony address wrongfully being the initiator, you might want to update this so it _assumes_ that the Root domain is the `fromDomain` instead of getting it from `initiatorColony`

src/components/common/ColonyActions/ActionsListItem/ActionsListItem.tsx
thomassdk: Thanks for the future notice :eyes:

PR #962, opened 2023-08-16
Integrate motions with ControlSafeDialog

### LINKED ISSUE #744: Integrate motions with Control Safe dialog
Clarifications:
- The components should only appear in the last step of the Control Safe dialog.
- The respective saga should be ported over in this issue.
- This issue doesn't cover updating the motion page.

Image

Image

ArmandoGraterol:
## Testing time!

- It's the same process for testing the normal actions but with the exception that you need the voting reputation extension enabled.
- Make sure to test how the first step and the preview look:
  - With and without reputation in the colony. (If there's no rep, you shouldn't be able to create the motion and a warning should appear explaining so. Unless you have permission to force the action through.)
  - With and without permissions. (If you have permission, you should see the `ForceToggle` on the preview, otherwise, it should be hidden.)
  - With and without forcing the action (You will need the ROOT permission to force it.)
- Once you submit a motion, you should be redirected to the motion's page, which is not wired up, for now.

Block ingestor branch used by this PR: https://github.com/JoinColony/block-ingestor/pull/116

Resolves #744


src/redux/sagas/motions/initiateSafeTransactionMotion.ts
```diff
@@ -0,0 +1,289 @@
...
+    const apolloClient = getContext(ContextModule.ApolloClient);
+    const colonyManager = yield getColonyManager();
+    const colonyClient = yield colonyManager.getClient(
+      ClientType.ColonyClient,
+      colonyAddress,
+    );
+
+    const ZODIAC_BRIDGE_MODULE_ADDRESS = isDev
+      ? process.env.ZODIAC_BRIDGE_MODULE_ADDRESS
+      : safe.moduleContractAddress;
```
thomassdk, line 59: This should now be
```
    const ZODIAC_BRIDGE_MODULE_ADDRESS = isDev
      ? getDevZodiacBridgeAddress()
      : safe.moduleContractAddress;
```

thomassdk (approved):
LGTM, nice changes and refactoring in here

src/redux/sagas/motions/initiateSafeTransactionMotion.ts
ArmandoGraterol: Good idea!

Issue #966, opened 2023-08-17
Safe transaction status check


thomassdk:
This PR should port over the capability to check whether a safe transaction has been completed via the Dapp and the hook this to where it is needed in the current safe PRs (eg. the Colony action list and action page)

It should be based on the `getTransactionStatuses` hook but it would be good to explore moving this check to a lambda function

PR #990, opened 2023-08-22
Integrate Safe Transactions with motion page

### LINKED ISSUE #963: Add support of safe transactions to the motion page and action list
The requirements are quite similar to the action page and normal action list item with a few differences detailed here.

Motion page:
- The red banner should appear AFTER the motion has been finalized.

Image

Action list:
- The motion tags should appear alongside the transaction status tags:
- The transaction status tag should appear AFTER the motion has been finalized

Image

Image

ArmandoGraterol:
**NOTE:** "_Armando what the heck this PR is 4,6k lines of code!!_" Don't worry, around 3,9k lines come from auto-generated files and ABIs.

- This PR adds a `getSafeTransactionStatus` which takes a transaction hash and a chain ID and returns an array with each transaction's status.

- The safe transaction is executed ONLY after the motion is finalized, so the transaction status or the banner shouldn't appear before that.

- As always, when testing motions, make sure you have a reputation in Root. You should be able to create motions with and without permission.

FireShot Capture 140 - Colony CDapp - localhost

FireShot Capture 141 - Colony CDapp - localhost

Resolves #963 and #966


src/components/common/ColonyActions/ActionDetailsPage/staticMaps.ts
```diff
@@ -97,6 +97,7 @@ export const ACTIONS_EVENTS: ActionsEventsMap = {
   [ColonyActionType.EmitDomainReputationReward]: [
     ColonyAndExtensionsEvents.ArbitraryReputationUpdate,
   ],
+  [ColonyActionType.MakeArbitraryTransaction]: SAFE_EVENTS,
```
thomassdk, line 100: I think this is wrong as only one of these events should be shown depending on the type of safe transaction that was carried out.

Probably best to remove this line as #990 handles this integration.

Unless I am missing something and if so sing out :bird:

src/components/common/ColonyActions/ActionsListItem/ActionsListItem.tsx
```diff
@@ -75,23 +83,27 @@ const ActionsListItem = ({
...
-          safeTransactionStatus === TRANSACTION_STATUS.PENDING
-            ? 'golden'
-            : 'primary',
-        colorSchema: 'fullColor',
-      }}
-    />
+  const ActionListItemTags = (
+    <>
+      <MotionTag />
+      {safeTransaction && !isEmpty(safeTransactionStatus) && (
```
thomassdk, line 89: Would be nicer to extract these two checks into a single boolean variable above

src/hooks/useSafeTransactionStatus.ts
```diff
@@ -0,0 +1,53 @@
+import { defineMessages } from 'react-intl';
+
+import { ETHEREUM_NETWORK } from '~constants';
+import { useGetSafeTransactionStatusQuery } from '~gql';
+import { ColonyAction } from '~types';
+
+export enum TRANSACTION_STATUS {
+  COMPLETED = 'Completed',
+  ACTION_NEEDED = 'Action neeeded',
```
thomassdk, line 9: typo :slightly_smiling_face: 
```suggestion
  ACTION_NEEDED = 'Action needed',
```

src/hooks/useSafeTransactionStatus.ts
```diff
@@ -0,0 +1,53 @@
...
+const displayName = 'utils.safe';
+
+export const MSG = defineMessages({
+  [TRANSACTION_STATUS.COMPLETED]: {
+    id: `${displayName}.${[TRANSACTION_STATUS.COMPLETED]}`,
+    defaultMessage: 'Completed',
+  },
+  [TRANSACTION_STATUS.ACTION_NEEDED]: {
+    id: `${displayName}.${[TRANSACTION_STATUS.ACTION_NEEDED]}`,
+    defaultMessage: 'Action needed',
```
thomassdk, line 21: Would be nice to use the value of the enums here to save some repetition. Whadda you reckon?

thomassdk (approved):
LGTM nice work, just some minor comments

src/components/common/ColonyActions/ActionDetailsPage/staticMaps.ts
ArmandoGraterol: As far as I understand those mappings, they only indicate the type of events that _could_ be displayed on the action page. So we are not necessarily telling it that all those events will appear at the same time. For example, `MOTION_EVENTS` has these:

```
const MOTION_EVENTS = [
  ColonyAndExtensionsEvents.MotionCreated,
  ColonyAndExtensionsEvents.MotionStaked,
  ColonyAndExtensionsEvents.ObjectionRaised,
  ColonyAndExtensionsEvents.MotionFinalized,
  ColonyAndExtensionsEvents.MotionRewardClaimed,
];
```

The only event in this array that is 100% guaranteed to appear on the motion page is `MotionCreated`. The rest _could_ appear depending on certain conditions (This also happens with the recovery mode events). I treat `MakeArbitraryTransaction` the same way as you can see :)

thomassdk: perhaps I don't understand this properly then, lets have a quick chat but my understanding is that each event will be mapped over and displayed in `ActionDetailsPageFeed.tsx` with:
```tsx
  const events =
    JSON.parse(actionData.individualEvents as string) ||
    arbitraryTransactionEventParser(actionData) ||
    ACTIONS_EVENTS[actionData.type];

  return (
    <>
      {events?.map((event) => (
        <ActionsPageEvent
          actionData={actionData}
          eventName={event?.type || event}
          key={event?.id || event}
          eventId={event?.id || event}
        />
      ))}
    </>
  );
```

I experimented with setting up `SAFE_EVENTS` like you have here and could not see a way for them to appear only with certain conditions.

ArmandoGraterol: Sorry, I will try to explain myself better.

What I said was in response to this comment specifically:
> I think this is wrong as only one of these events should be shown depending on the type of safe transaction that was carried out.

It's my understanding (I could be wrong) that when we group events together as we do with `SAFE_EVENTS` or `MOTION_EVENTS`, we are _not_ telling `ActionDetailsPageFeed` that _every single_ event inside those objects should appear, instead, we are telling it "Hey, for this type of action page, every event inside this object is an event that COULD BE displayed."

PR #994, opened 2023-08-23
Creating staked expenditure

### LINKED ISSUE #952: Creating staked expenditures
This issue is to add the ability to create a staked expenditure.
When a user is missing Administration permissions in the domain they wish to create an expenditure in, and they have the Staked Expenditure extension installed, they should be able to create a "staked expenditure" by staking a particular amount of tokens.

jakubcolony:
## Description

This PR adds the functionality of creating a staked expenditure.

For context, you can only create an expenditure if you have the Administration permissions in the colony. Staked Expenditure extension allows anyone to create an expenditure, as long as they stake it. The required stake amount is a fraction of the total domain reputation that an expenditure is created in, which is configured when enabling the extension.

### Creating a regular (non-staked) expenditure
Before installing any extensions, try creating a new expenditure. It should work exactly the same as previously.

### Creating a staked expenditure
1. Enable reputation auto mining: http://localhost:3001/reputation/monitor/toggle
2. Mint some tokens, pay yourself and wait for the mining cycle to finish (you can do step 3 while you're waiting).
3. Install and enable Staked Expenditure extension. Enter the desired stake fraction.
4. Activate all your tokens.
5. Create an expenditure. You should see a dialog appear with the required stake amount. Since at this point you're the only user with reputation, the amount should be roughly equal to the fraction percentage times the amount of tokens you've paid yourself:
image

7. The saga with 3 transactions (approve stake, make expenditure, and set expenditure values) should go through and you should be redirected to the expenditure page.
8. Check the token popover to confirm the stake amount is shown there.

### Not enough active tokens 
Deactivate all your tokens and try to create an expenditure. The stake dialog should inform you you don't have enough active tokens:
image

### Editing, locking, funding & finalizing
Should work exactly as previously.

### Claiming a payout
The function determining which payouts are claimable now uses the block time instead of machine time. If you disable auto mining, you should see the correct time until next claim (relative to the machine time):
image

Note this will only be correct straight after reload, as the machine time advances while the dev blockchain time doesn't. In the example above you can advance the time by 60 seconds using the approach from #925.

### Creating an expenditure as a non-admin user
Try paying some tokens to another dev wallet/metamask, then follow the steps above to confirm you can create a staked expenditure even if you don't have the Administration permissions. You should also be able to edit, lock and finalize, but not fund (since you're missing permissions to move funds).

Resolves #952


src/components/common/Expenditures/ExpenditureForm/CreateExpenditureForm.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const { isStakedExpenditureEnabled } = useEnabledExtensions();
+
+  const [isStakeDialogOpen, setIsStakeDialogOpen] = useState(false);
+
+  if (!colony) {
+    return null;
+  }
+
+  /**
+   * @TODO: This should include a permissions check as users with
```
willm30, line 33: `getUserRolesFroDomain` should serve you here (in `src/transformers/index.ts`)

willm30 (commented):
LGTM, aside from a couple of observations:

1. Claiming the funds is not removing them from "staked" in the token activation popover

2. I believe the wrong slots are being shown in expenditures.

E.g. Expenditure 3 UI:

expenditure3

Versus Expenditure 3 DB:

expend3

The recipient address and amount are actually from expenditure 2:

expend2

Possibly caused by slots not having unique ids and apollo mixing them up? That would be my first guess. 

There's definitely something funky going on since when I click back on expenditure 3, it has the correct slot:

slot3

Let me know if you want to address these elsewhere and I can approve this PR, or if you'll address them here.

jakubcolony:
@willm30 

> 1. Claiming the funds is not removing them from "staked" in the token activation popover
There's a separate issue for that: https://github.com/JoinColony/colonyCDapp/issues/1006

I couldn't reproduce the wrong slots issue myself but indeed all the slots were being cached under a single cache entry. I disabled caching of that type entirely which should hopefully fix the problem.

willm30 (approved):

PR #1004, opened 2023-08-24
Port Metatransactions

rdig:
This PR adjusts and fixes _(more than actually porting)_ metatransactions to work within the CDapp environment.

Overview of what this PR does:
- update the transaction encoding method _(due to the Ethers upgrade from the Dapp)_
- fixes ganache web3-onboard module message signing _(so that metatransactions work in dev mode, Metamask had not issue)_
- refactor sagas that used "withProofs" contract call helpers, to use the actual methods and provide own proofs _(the "withProofs" methods are not "real" contract methods, but helpers introduced by `colonyJS`)_
- minor fixes

### Issues found
- Setting permissions is flaky and sometimes the action doesn't show up in the list #1008 
- #1009

### Changes:
- Fixed `ganacheWalletModule` personal signature method
- Fixed `getMetatransactionPromise` to update the function encode method
- Created proofs saga util (for `getMoveFundsPermissionProofs` and `getMultiPermissionProofs`)
- Updated the add domain saga to not use the "withProofs" method
- Updated the edit domain saga to not use the "withProofs" method
- Updated the create payment saga to not use the "withProofs" method
- Updated the move funds saga to not use the "withProofs" method
- Updated the manage permissions saga to not use the "withProofs" method
- Updated the smite saga to not use the "withProofs" method
- Fixed the create colony saga to actually work for metatransactions
- Refactored manage permissions saga to use `intArrayToBytes32` for generating hex permissions string
- Updated extension enable saga to not use "withProofs" method
- Fixed `getMetatransactionPromise` to properly set the token client's `metatransactionVariation` _(this was done due to a bug being identified and fixed on the Dapp side https://github.com/JoinColony/colonyDapp/commit/56e605d7171d6b6e33ab2430a1a8ef140ecb7362)_
- Refactor the stake motion saga to not use "withProofs" method
- Refactor the claim motion saga to not use "withProofs" method
- Refactor the claim all motion saga to not use "withProofs" method
- Refactor the escalate motion saga to not use "withProofs" method _(even though this is not really wired in properly, because a lack of UI/feature specs)_

### Docs

**[Add Metatransactions Documentation](https://github.com/JoinColony/ColonyFrontEndLivingStandard/pull/20)**

### Testing

Execute one _(or preferably all)_ the following transactions, with _(and without)_ metatransactions enabled:
- Add Domain Action
- Edit Domain Action
- Create Payment Action
- Move Funds Between Domains Action
- Manage Permissions Action _(Note the bug I've mentioned above, which will make this harder to test, see #1008 )_
- Smite Action
- Create a colony
- Extension Enable
- Once you have a Motion created, test Staking said motion
- Once you have a Motion finalized, test Claiming motion winnings/losses
- Once you have a Motion finalized, test Claiming All motions winnings/losses _(from the user token activation menu)_

_Note: that while I refactored the motion escalate saga as well, that was never wired properly in the Dapp and since it was ported that way, it cannot be tested._

### Demo

_The call will be scheduled once I'm back from my time off_


thomassdk (approved):
LGTM (though most of it goes over my head tbh :slightly_smiling_face:  ) 

One thing I did notice was that it is still not possible to turn on meta transactions via the advanced user settings

image

Is this something this PR should enable as well?

willm30:
@thomassdk 

> One thing I did notice was that it is still not possible to turn on meta transactions via the advanced user settings

Can you check you have the following in your `.env`?

```
METATRANSACTIONS=true
```

I think that's responsible for what you're seeing.

willm30:
Just noting here if you see the following while testing motions:

motionStaking

Try turning off the reputation monitor. It's probably that blocktime forwarding has caused the motion to end.

src/components/common/UserProfileEdit/UserAdvancedSettings.tsx
```diff
@@ -123,7 +123,16 @@ const UserAdvancedSettings = ({ user: { walletAddress, profile } }: Props) => {
...
+                toggleDisabled={(() => {
+                  if (row.name === 'metatransactionsEnabled') {
+                    return !metatransactionsAvailable;
+                  }
+                  // Not implemented currently, so disable it outright
+                  if (row.name === 'decentralizedModeEnabled') {
+                    return true;
+                  }
+                  return true;
+                })()}
```
willm30, line 135: I think this logic should be moved up to the `getAdvancedSettingsRows` fn. In there you can add a `disabled` field that's `true` for `decentralizedModeEnabled` and `!metatransactionsAvailable` for `metatransactionsEnabled`

src/redux/sagas/actions/managePermissions.ts
```diff
@@ -94,6 +111,28 @@ function* managePermissionsAction({
       );
     }
 
+    yield put(transactionPending(setUserRoles.id));
```
willm30, line 114: Any reason to prefer doing it like this instead of moving the logic up and adding the params directly?

src/redux/sagas/colony/colonyCreate.ts
```diff
@@ -529,6 +545,31 @@ function* colonyCreate({
...
+        colonyAddress,
+      );
+
+      let oneTxPaymentRoleProofs;
+      // This has two potential permissions, so we try both of them
+      try {
+        oneTxPaymentRoleProofs = yield getPermissionProofs(
+          colonyClient,
+          Id.RootDomain,
+          ColonyRole.Architecture,
```
willm30, line 562: Can you explain this? Under what circumstances would one fail and not the other?

willm30 (approved):
Working for me across a number of actions / motions. Good job!

ArmandoGraterol (approved):
LGTM 💯

src/redux/sagas/actions/managePermissions.ts
rdig: Consistency mostly, as for some sagas this is actually required since you don't have the params available when creating the transaction channel _(ie: you need data from the first transaction that gets executed)_

The second reason, is UX. For sagas/transactions that require us computing their proofs, on slower machines it might take some time, so I've chose this pattern since `transactionPending` will start showing the spinner next to the transaction (in the gas station), while we compute data, and send it, in the background.

Otherwise, the user will just sit there with a transaction entry with no state.

src/redux/sagas/colony/colonyCreate.ts
rdig: Yeah, if I think real hard about it, this should just be root in this context.

Don't get me wrong, the logic is still correct, as you can assign permissions to this extension both with Architecture or Root. 

However in the context of this saga, this will be a newly deployed colony, and this a newly deployed extension, so the current user's address will always have Root, since they were the ones that just deployed the colony.

src/redux/sagas/motions/claimAllMotionRewards.ts
```diff
@@ -67,6 +74,17 @@ function* claimAllMotionRewards({
...
+    const colonyClient = yield colonyManager.getClient(
+      ClientType.ColonyClient,
+      colonyAddress,
+    );
+
+    const votingReputationClient: AnyVotingReputationClient = yield call(
+      [colonyManager, colonyManager.getClient],
+      ClientType.VotingReputationClient,
+      colonyAddress,
+    );
```
jakubcolony, line 86: This only seems to be used to get the voting reputation address, should we get it from the DB and pass it to the saga instead?

src/redux/sagas/motions/claimMotionRewards.ts
```diff
@@ -78,13 +85,37 @@ function* claimMotionRewards({
...
+    const colonyClient = yield colonyManager.getClient(
+      ClientType.ColonyClient,
+      colonyAddress,
+    );
+
+    const votingReputationClient: AnyVotingReputationClient = yield call(
+      [colonyManager, colonyManager.getClient],
+      ClientType.VotingReputationClient,
+      colonyAddress,
+    );
```
jakubcolony, line 97: Same as above

jakubcolony (approved):
Looking good 👍

rdig:
@willm30 

> Just noting here if you see the following while testing motions:

I've ran into that as well when enabling/disabling extensions and when switching between metatransactions on/off. I don't exactly know how to replicate it every time, but I think it's to do with the block ingestor not the cdapp's logic

PR #1012, opened 2023-08-28
Connect safe token to main token table

### LINKED ISSUE #964: Improve safe balance token storing
We should look into improving how we store the safe balance's tokens data once the user submits a safe transaction by exploring if we can store those tokens the "Colony way". (Using `getTokenFromEverywhere`)

thomassdk:
## Description

This PR improves the token storing of the safes by associating the token stored by the safe by one contained by the Token model.

To do so, it was necessary to make the `SafeTransactionData` a model in its own right:
```graphql
type SafeTransactionData @model {
  id: ID!
  transactionType: SafeTransactionType!
  tokenAddress: ID
  token: Token @hasOne(fields: ["tokenAddress"])
  amount: String
  rawAmount: String
  recipient: SimpleTarget
  data: String
  contract: SimpleTarget
  abi: String
  contractFunction: String
  nft: NFT
  nftData: NFTData
  functionParams: [FunctionParam]
}
```

A lot of files were changed, but they are mainly minor changes to fix types and conform the functions to the new fragments. 

The main work is done in the safe saga's which work slightly differently now

Resolves #964


src/redux/sagas/actions/initiateSafeTransaction.ts
```diff
@@ -192,11 +197,30 @@ function* initiateSafeTransactionAction({
...
+      yield apolloClient.mutate<
+        CreateSafeTransactionDataMutation,
+        CreateSafeTransactionDataMutationVariables
+      >({
+        mutation: CreateSafeTransactionDataDocument,
+        variables: {
+          input: {
+            ...omit(transaction, 'token'),
+            tokenAddress: transaction.token?.tokenAddress,
+            safeTransactionTransactionsId:
```
ArmandoGraterol, line 217: I feel like this name could be better, how about just `safeTransactionId`? or perhaps `transactionHash` considering that if this is inside a `SafeTransactionData` model then we could conclude that it's a safe transaction so there wouldn't be a need to be so specific.

ArmandoGraterol (changes requested):
Unfortunately, I'm getting an error whenever I try to make a second Transfer Funds transaction:
Screenshot 2023-08-28 at 15 37 37 (2)

This may be related to what I mentioned in one of my comments below about the  `GetTokenFromEverywhereQuery`

src/redux/sagas/actions/initiateSafeTransaction.ts
thomassdk: sounds good I was struggling with naming this so I just fell back on amplify's auto naming. Thanks for the suggestions pana!

src/graphql/mutations/token.graphql
```diff
@@ -15,3 +15,9 @@ mutation DeleteColonyTokens($input: DeleteColonyTokensInput!) {
     id
   }
 }
+
+mutation CreateToken($input: CreateTokenInput!) {
+  createToken(input: $input) {
+    id
+  }
+}
```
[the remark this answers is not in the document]
thomassdk: ah okay i missed the fact that this was calling an amplify function.

I have since updated the amplify function to introduce the concept of chains so that it knows what rpc url to call when checking for the token existence.

This still needs to be uploaded to the aws param store but you can test locally by applying this dirty patch that simulates what would happen if we were working live on chain

[safe-token.patch.gz](https://github.com/JoinColony/colonyCDapp/files/12463892/safe-token.patch.gz)

which adds these lines to the `fetchTokenFromChain` lambda:
```ts

  /* For dirty testing */
  switch (network) {
    case BINANCE_NETWORK.shortName:
      rpcURL = 'https://bsc.meowrpc.com';
      break;
    case ETHEREUM_NETWORK.shortName:
    default:
      rpcURL = 'https://eth.drpc.org';
  }
```

Importantly though, you have to apply this patch **after** you have created your colony as otherwise the setup process breaks as it cant find the token on the `ganache` chain

ArmandoGraterol (commented):
I'm getting this error when I try to create a colony.

Screenshot 2023-08-29 at 14 59 03

Also getting it when I run the dummy data setup script.

thomassdk:
My bad, I didn't notice this when I added it. 

Coinmachine's build system for lambda functions was different in that it allowed importing constants from the main project, and I wrongly made the assumption that this would work here.

I've added in stubs for these constants as I've noticed some comments indicating that this will ideally be implemented in the future, and it should all work now :crossed_fingers:

amplify/backend/function/fetchTokenFromChain/src/index.js
```diff
@@ -24,29 +29,44 @@ const baseToken = {
...
   if (ENV === 'qa') {
+    let chainRpcParam = 'chainRpcEndpoint';
+
+    switch (network) {
+      case BINANCE_NETWORK.shortName:
+        chainRpcParam = 'bnbRpcEndpoint';
+        break;
+      case ETHEREUM_NETWORK.shortName:
+      default:
+      // Use default chainRpcParam ie Ethereum to set `rpcURL`
```
ArmandoGraterol, line 44: Shouldn't this switch also happen in `prod` and not just in QA?

ArmandoGraterol (approved):
LGTM 👍 Nice work!

I've left a question but other than that I'm good with this.

amplify/backend/function/fetchTokenFromChain/src/index.js
thomassdk: so at the moment we only have the QA environment up and running and a proper `prod` isnt yet running.

I've added it here in this case as this is the only condition that is checked so far ie such as what happens below. This is so that there is no confusion when we get around to adding in the case `if (ENV = 'prod')`

thomassdk:
> If this patch is needed for the dev env to work properly, why is not included inside a conditional that executes the code if the env is the dev env? Would this be addressed by what you mentioned about uploading the changes to was param store?

So this is slightly more tricky than just checking that the lambda function is running in the `dev` env.

As this patch is required for a safe testing only (via querying live chain RPC's as opposed to our local Ganache chain) due to the safe token contract addresses not existing on our local ganache chain. When we apply it generally it will break the CDapp as now all `fetchTokenFromChain` queries will check the live chain RPC's and not our local environment. This is the reason for only applying the patch **after** the first colony creation. A more accurate condition for when to apply the patch would be only **before** a safe transfer funds transaction is being tested, and after that is done, the patch should be removed. 

This is a tricky question to answer as it is a trade off. Do we add a special condition to the `fetchTokenFromChain` lambda to signal to that this is a request from the safe feature, or do we go down the manual patch apply / remove annoyance for the developers? On one hand it seems wrong to add a general condition just for testing the safe but on the other hand developers are ~lazy~ human and I can see this manual patch/de-patch (not sure de-patch is right but sounds good to me :stuck_out_tongue_closed_eyes: ) being a major source of confusion and setbacks when we test it later.

So to end with a proposal: How about we add this flag to the `fetchTokenFromChain` lambda that signals that this is solely used to test the safe feature with a note to remove it when we have a ganache development environment which forks the history of the chains so that these token contracts exist at the addresses that correspond to the ones we receive from querying the safe balance API?

amplify/backend/function/fetchTokenFromChain/src/index.js
ArmandoGraterol: True! Then perhaps we could add a note or a todo comment to remind us to update this once we have a prod env?

src/components/common/Dialogs/ControlSafeDialog/TransactionTypesSection/TransferFundsSection.tsx
```diff
@@ -109,22 +110,24 @@ const TransferFundsSection = ({
...
         const data = (await response.json()) as SafeBalanceApiData[];
-        const formattedSafeBalances: SafeBalance[] = data.map(
-          (balanceData) => ({
+
+        const tokenFromDbPromises = data.map(async (balanceData) => {
+          const network = getNetworkFromChainName(chainName);
+          const token = await getTokenFromEveryWhereQuery(
+            balanceData?.tokenAddress || ADDRESS_ZERO,
+            network,
+          );
```
ArmandoGraterol, line 119: Why this approach instead of the old one? The token data is already available in `balanceData`.

ArmandoGraterol (commented):
Left a question and noticed that the token selector is not working properly now:
FireShot Capture 149 - Colony CDapp - localhost

But other than this, the rest seems to be working as expected, with no amplify errors like before :)

amplify/backend/function/fetchTokenFromChain/src/index.js
thomassdk: I dont personally think this is necessary as I imagine we will just copy the entire `qa` if statement to the `prod` env because as you rightly pointed out this will need to be set there and I dont see anything changing

src/components/common/Dialogs/ControlSafeDialog/TransactionTypesSection/TransferFundsSection.tsx
thomassdk: My logic in making this change was that at the moment no tokens have their thumbnail string filled out, so in order to correctly set the thumbnail to show to the user in this step we need to fetch it from the database (as the `logoUri` returned via the safe API will in all likelihood not be the same as the thumbnail that we choose to store).

However, on coming back to this today I have reverted these changes as it does not seem worth the extra fetch to the database and this will be a problem to tackle when we figure out what thumbnails we are using for tokens and how we are storing them (I'm hoping that there will be a quicker way to get the thumbnail than just querying the database, but I may be wrong)

This is also why the DAI token will no longer show, as on second thought I believe it might be confusing if we set token thumbnails to tokens **only** when they have been fetched from a safe

thomassdk:
OK hopefully the final round on this PR!

I've changed things so that we still have no concept of native tokens in the database to stay with the mapping of token address's as ID's for tokens and reduce complexity.

I've added in some logic to the `getTokenFromEverywhere` that correctly sets the rpc to use depending on the network passed to the lambda (defaulting to ganache) thus removing the need for that icky safe only variable we talked about before.

In the UI we now get the chain Id from the safe and infer the native token if the token field is null.
  
image

Two things immediately spring to the eye.

One is that all the possible safe events are shown (this speaks to our previous conversation on the static mapping of ActionTypes to Events). I believe the reason it was working for you was that we were parsing the `arbitraryTransaction` **before** the static map so the mapping of those Safe Events never occured for you in the previous branch.  ie `arbitraryTransactionEventParser(actionData);`

I still do not see how this mapping will work as my interpretation of those events is that it should show **all** events that need to show on the page. I've left it as is so we can discuss but this is the main outstanding issue.

The other thing to note is more minor and is the regression with regards to the dai token no longer being displayed and a fallback being utilized. I answer why this happens in a comment above.

The token picker is also now fixed:

image

thomassdk:
Another thing I have noticed is that if we make multiple transactions I get an amplify error like so:
```
amplify   | Error while executing Local DynamoDB
amplify   | {
amplify   |     "version": "2018-05-29",
amplify   |     "operation": "PutItem",
amplify   |     "attributeValues": {
amplify   |         "id": {
amplify   |             "S": "0x2cdbe41c323715877cc1f07275dc42e25f42f7cf171a022260701fb6559918f6"
amplify   |         },
amplify   |         "createdAt": {
amplify   |             "S": "2023-09-01T13:35:45.000Z"
amplify   |         },
amplify   |         "updatedAt": {
amplify   |             "S": "2023-09-01T13:35:49.279Z"
amplify   |         },
amplify   |         "colonyId": {
amplify   |             "S": "0x3a157280ca91bB49dAe3D1619C55Da7F9D4438c3"
amplify   |         },
amplify   |         "type": {
amplify   |             "S": "MAKE_ARBITRARY_TRANSACTION"
amplify   |         },
amplify   |         "blockNumber": {
amplify   |             "N": "1030"
amplify   |         },
amplify   |         "showInActionsList": {
amplify   |             "BOOL": true
amplify   |         },
amplify   |         "initiatorAddress": {
amplify   |             "S": "0xb77D57F4959eAfA0339424b83FcFaf9c15407461"
amplify   |         },
amplify   |         "recipientAddress": {
amplify   |             "S": "0xbc71B7AD34817986D99d1a5c78A02C44fdE3236F"
amplify   |         },
amplify   |         "fromDomainId": {
amplify   |             "S": "0x3a157280ca91bB49dAe3D1619C55Da7F9D4438c3_1"
amplify   |         },
amplify   |         "__typename": {
amplify   |             "S": "ColonyAction"
amplify   |         }
amplify   |     },
amplify   |     "condition": {
amplify   |         "expression": "attribute_not_exists(#id)",
amplify   |         "expressionNames": {
amplify   |             "#id": "id"
amplify   |         }
amplify   |     },
amplify   |     "key": {
amplify   |         "id": {
amplify   |             "S": "0x2cdbe41c323715877cc1f07275dc42e25f42f7cf171a022260701fb6559918f6"
amplify   |         }
amplify   |     }
amplify   | }
amplify   | ConditionalCheckFailedException: The conditional request failed
amplify   |     at Request.extractError (/snapshot/repo/build/node_modules/aws-sdk/lib/protocol/json.js:52:27)
amplify   |     at Request.callListeners (/snapshot/repo/build/node_modules/aws-sdk/lib/sequential_executor.js:106:20)
amplify   |     at Request.emit (/snapshot/repo/build/node_modules/aws-sdk/lib/sequential_executor.js:78:10)
amplify   |     at Request.emit (/snapshot/repo/build/node_modules/aws-sdk/lib/request.js:686:14)
amplify   |     at Request.transition (/snapshot/repo/build/node_modules/aws-sdk/lib/request.js:22:10)
amplify   |     at AcceptorStateMachine.runTo (/snapshot/repo/build/node_modules/aws-sdk/lib/state_machine.js:14:12)
amplify   |     at /snapshot/repo/build/node_modules/aws-sdk/lib/state_machine.js:26:10
amplify   |     at Request.<anonymous> (/snapshot/repo/build/node_modules/aws-sdk/lib/request.js:38:9)
amplify   |     at Request.<anonymous> (/snapshot/repo/build/node_modules/aws-sdk/lib/request.js:688:12)
amplify   |     at Request.callListeners (/snapshot/repo/build/node_modules/aws-sdk/lib/sequential_executor.js:116:18)
amplify   |     at Request.emit (/snapshot/repo/build/node_modules/aws-sdk/lib/sequential_executor.js:78:10)
amplify   |     at Request.emit (/snapshot/repo/build/node_modules/aws-sdk/lib/request.js:686:14)
amplify   |     at Request.transition (/snapshot/repo/build/node_modules/aws-sdk/lib/request.js:22:10)
amplify   |     at AcceptorStateMachine.runTo (/snapshot/repo/build/node_modules/aws-sdk/lib/state_machine.js:14:12)
amplify   |     at /snapshot/repo/build/node_modules/aws-sdk/lib/state_machine.js:26:10
amplify   |     at Request.<anonymous> (/snapshot/repo/build/node_modules/aws-sdk/lib/request.js:38:9)
amplify   |     at Request.<anonymous> (/snapshot/repo/build/node_modules/aws-sdk/lib/request.js:688:12)
amplify   |     at Request.callListeners (/snapshot/repo/build/node_modules/aws-sdk/lib/sequential_executor.js:116:18)
amplify   |     at callNextListener (/snapshot/repo/build/node_modules/aws-sdk/lib/sequential_executor.js:96:12)
amplify   |     at IncomingMessage.onEnd (/snapshot/repo/build/node_modules/aws-sdk/lib/event_listeners.js:380:13)
amplify   |     at IncomingMessage.emit (events.js:412:35)
amplify   |     at IncomingMessage.emit (domain.js:475:12)
amplify   |     at endReadableNT (internal/streams/readable.js:1333:12)
amplify   |     at processTicksAndRejections (internal/process/task_queues.js:82:21) {
amplify   |   code: 'ConditionalCheckFailedException',
amplify   |   time: 2023-09-01T13:35:49.284Z,
amplify   |   requestId: 'fb3d9dd3-671d-4693-a2ea-20a1eec3eef6',
amplify   |   statusCode: 400,
amplify   |   retryable: false,
amplify   |   retryDelay: 19.35558239733075
amplify   | }
```

It seems to refer to duplicate txHash in this mutation
```
    const safeTransaction = yield apolloClient.mutate<
      CreateSafeTransactionMutation,
      CreateSafeTransactionMutationVariables
    >({
      mutation: CreateSafeTransactionDocument,
      variables: {
        input: {
          id: txHash,
          title,
          safe,
        },
      },
    });
```

Are you getting this in your branches or is it something that has been introduced in this branch.

Just a note everything seems to be working fine even though this error is cropping up

ArmandoGraterol (approved):
Excellent work improving the code further 🎉 

Yeah, I see what you mean now regarding the events, I thought the logic that displays the event was still working like in the Dapp but something definitely changed. I can fix it in my PR, same with taking a look at why the multiple transactions are throwing an amplify error.

For me, this is good to go tho!

PR #1017, opened 2023-08-30
Ensure errors propagate to ui from extension sagas

willm30:
## Description

This PR tweaks the extension sagas logic so that errors correctly propagate to the UI. 

**Changes** 🏗

* Listen for the success / error tx action, before resolving the saga as successful

Contributes to #1018

### DIFF
--- a/src/components/shared/Button/ActionButton.tsx
+++ b/src/components/shared/Button/ActionButton.tsx
@@ -55,15 +55,13 @@ const ActionButton = ({
         typeof values == 'function' ? await values() : values;
       result = await asyncFunction(asyncFuncValues);
       if (isMountedRef.current) setLoading(false);
+      if (typeof onSuccess == 'function') onSuccess(result);
     } catch (err) {
       setLoading(false);
 
       /**
        * @todo : display error somewhere
        */
-      return;
-    } finally {
-      if (typeof onSuccess == 'function') onSuccess(result);
     }
   };
 
--- a/src/redux/sagas/extensions/extensionDeprecate.ts
+++ b/src/redux/sagas/extensions/extensionDeprecate.ts
@@ -27,13 +27,15 @@ function* extensionDeprecate({
 
     yield takeFrom(txChannel, ActionTypes.TRANSACTION_CREATED);
 
-    yield put<AllActions>({
-      type: ActionTypes.EXTENSION_DEPRECATE_SUCCESS,
-      payload: {},
-      meta,
-    });
+    const result = yield waitForTxResult(txChannel);
 
-    yield waitForTxResult(txChannel);
+    if (result.type === ActionTypes.TRANSACTION_SUCCEEDED) {
+      yield put<AllActions>({
+        type: ActionTypes.EXTENSION_DEPRECATE_SUCCESS,
+        payload: {},
+        meta,
+      });
+    }
   } catch (error) {
     return yield putError(ActionTypes.EXTENSION_DEPRECATE_ERROR, error, meta);
   }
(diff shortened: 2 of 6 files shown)


src/components/shared/Button/ActionButton.tsx
```diff
@@ -55,15 +55,13 @@ const ActionButton = ({
...
+      if (typeof onSuccess == 'function') onSuccess(result);
     } catch (err) {
       setLoading(false);
 
       /**
        * @todo : display error somewhere
        */
-      return;
-    } finally {
-      if (typeof onSuccess == 'function') onSuccess(result);
```
willm30, line 66: This is a stupid bug I introduced when porting decisions, incorrectly assuming that the catch return would skip the finally block.

src/redux/sagas/extensions/extensionDeprecate.ts
```diff
@@ -27,13 +27,15 @@ function* extensionDeprecate({
...
-    yield put<AllActions>({
-      type: ActionTypes.EXTENSION_DEPRECATE_SUCCESS,
-      payload: {},
-      meta,
-    });
-
-    yield waitForTxResult(txChannel);
+    const result = yield waitForTxResult(txChannel);
+
+    if (result.type === ActionTypes.TRANSACTION_SUCCEEDED) {
```
willm30, line 32: Here we wait for `waitForTxResult`, which is listening for error, success, and cancelled. 

If success, then we emit saga's success message. `waitForTxResult` throws if there's a tx error, which triggers the saga's catch block.

jakubcolony: This made me think, there are also some sagas (including expenditure sagas) that are using such approach instead of `waitForTxResult`:
```
yield takeFrom(createDomain.channel, ActionTypes.TRANSACTION_SUCCEEDED);

// put the success action
```
It looks like they need to be switched over to use `waitForTxResult` in order to properly handle errors too.

jakubcolony: https://github.com/JoinColony/colonyCDapp/issues/1018

src/redux/sagas/extensions/extensionEnable.ts
```diff
@@ -100,20 +104,21 @@ function* extensionEnable({
...
+      const result = yield waitForTxResult(initialise.channel);
 
       if (setUserRolesWithProofs) {
-        yield takeFrom(
-          setUserRolesWithProofs.channel,
-          ActionTypes.TRANSACTION_SUCCEEDED,
-        );
+        yield waitForTxResult(setUserRolesWithProofs.channel);
+        // assume if this doesn't error, the transaction has succeeded.
+        // @TODO: Handle state in which it gets cancelled
```
jakubcolony, line 112: Looks like right now if it gets cancelled the loading will go on forever. Is there a reason why we couldn't throw an error when the tx is cancelled? We could skip the extra if checks too

src/redux/sagas/extensions/extensionDeprecate.ts
willm30: Yep, I'll make a note that this contributes to that issue

src/redux/sagas/extensions/extensionEnable.ts
willm30: @jakubcolony Yep, I think it makes sense to do this on the new ui branch as it's intrinsically connected to the cancel functionality in the ui

jakubcolony (approved):

Issue #1013, opened 2023-08-28
Fix multi transaction tab behaviour


thomassdk:
There are some weird behaviours that happen with the multi transaction tabs presently.

Some of these issues were pointed out here: https://github.com/JoinColony/colonyCDapp/pull/960#pullrequestreview-1596627455

PR #1023, opened 2023-08-30
Include various fixes to Safe Control

ArmandoGraterol:
- Fixed removal of tabs in the Control Safe dialog. They weren't being deleted properly and that caused a bunch of other weird behaviours.

- Fixed safe balance validation of the Transfer Funds transaction type. The validation wasn't taking into account the safe's balance at all.

- Fixed numeral inputs in the Contract Interaction transaction type not being updated properly after a re-render caused by a bug of Cleave.js. To reproduce in the feature branch or any other branch besides this one, add a contract interaction section in the second or more position in the dialog that has a`UINT` or `INT` param input filled and remove the transaction tab that comes before the contract interaction one, and try to change the number input value. 

- Fixed the CDapp crashing when a custom wallet address is selected in the NFT picker.

- Fixed permission icons missing in the action page for safe transaction events.

- Added `excludeFilterValue` to `SingleUserPicker` to filter out the value from the selection in the NFT picker.

- Updated raw amount validation to fix it not updating correctly after a re-render (Like what happened with the numeral inputs of the contract interaction section.)

- Fixed errors being persistent when you change transaction types, therefore, not allowing you to continue with the flow if you had errors before changing transaction type.

- Improved contract interaction `useEffects` performance, before the `useEffects` where hitting the API to get the contract's ABI too often, exceeding the limits.

Resolves #[1013](https://github.com/JoinColony/colonyCDapp/issues/1013)

### DIFF
--- a/src/components/common/Dialogs/ControlSafeDialog/TransactionTypesSection/RawTransactionSection.tsx
+++ b/src/components/common/Dialogs/ControlSafeDialog/TransactionTypesSection/RawTransactionSection.tsx
@@ -49,11 +49,6 @@ const RawTransactionSection = ({
         labelValues={{
           span: labelValues,
         }}
-        formattingOptions={{
-          numeral: true,
-          numeralPositiveOnly: true,
-          numeralDecimalScale: 0,
-        }}
       />
     </DialogSection>
     <DialogSection appearance={{ theme: 'sidePadding' }}>
--- a/src/components/common/Dialogs/ControlSafeDialog/validation.ts
+++ b/src/components/common/Dialogs/ControlSafeDialog/validation.ts
@@ -193,7 +193,8 @@ export const getValidationSchema = (
               then: number()
                 .typeError(() => MSG.notPositiveIntegerError)
                 .required(() => MSG.requiredFieldError)
-                .positive(() => MSG.notPositiveIntegerError),
+                .min(0, () => MSG.notPositiveIntegerError)
+                .integer(() => MSG.notPositiveIntegerError),
               otherwise: number().nullable(),
             })
             .nullable(),
--- a/src/components/common/Dialogs/ControlSafeDialog/validation.ts
+++ b/src/components/common/Dialogs/ControlSafeDialog/validation.ts
@@ -35,9 +35,9 @@ const MSG = defineMessages({
     id: `${displayName}.notHexError`,
     defaultMessage: 'Value must be a valid hex string',
   },
-  notIntegerError: {
-    id: `${displayName}.notIntegerError`,
-    defaultMessage: 'Amount must be an integer',
+  notPositiveIntegerError: {
+    id: `${displayName}.notPositiveIntegerError`,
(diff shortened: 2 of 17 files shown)


src/components/common/Dialogs/ControlSafeDialog/TransactionTypesSection/RawTransactionSection.tsx
```diff
@@ -49,11 +49,6 @@ const RawTransactionSection = ({
         labelValues={{
           span: labelValues,
         }}
-        formattingOptions={{
-          numeral: true,
-          numeralPositiveOnly: true,
-          numeralDecimalScale: 0,
-        }}
```
thomassdk, line 56: why is this removed?

src/components/common/Dialogs/ControlSafeDialog/validation.ts
```diff
@@ -149,9 +149,8 @@ export const getValidationSchema = (
                         }),
                       });
                     }
-                    const selectedToken = this.parent.tokenData?.tokenAddress;
-                    const selectedTokenDecimals =
-                      this.parent.tokenData?.decimals;
+                    const selectedToken = this.parent.token?.tokenAddress;
+                    const selectedTokenDecimals = this.parent.token?.decimals;
```
thomassdk, line 153: I also noticed and fixed this in the `improve token storage` PR :smile:

thomassdk (approved):
Works really well and the changes are a pleasure to interact with. Nice work

src/components/common/Dialogs/ControlSafeDialog/TransactionTypesSection/RawTransactionSection.tsx
ArmandoGraterol: Providing a `formattingOptions` prop to the Input component makes the component use `cleave.js` and unfortunately, `cleave.js` which is the library that we use to format the values of numeral inputs, has a bug in which it doesn't update itself after a re-render.

So in context what would happen is this:

- You want to make a multi-transaction, you add a Transfer NFT and a Transfer Funds transactions.
- You decide that you don't want to transfer the NFT so you remove the tab from the dialog.
- Now the Transfer Funds goes from having a `transactionIndex` of 1 to 0.
- This change of state will not make `cleave.js` automatically update the input in the re-render, so if it had a `transactionIndex` of 1 before, and you interact with it after removing the tab, it will behave as if the `transactionIndex` was the old one, and that causes more problems like the validation breaking because it thinks that there's still 2 transactions in the form when there's only 1.


Unluckily, `Cleave.js` is no longer maintained so to solve this we would need to either find some hacky way of making it work or outright looking for another library and refactoring the Input component

src/components/common/Dialogs/ControlSafeDialog/validation.ts
ArmandoGraterol: Great minds think alike 🧠

PR #1024, opened 2023-08-30
Reclaiming staked expenditure stake

### LINKED ISSUE #1006: Reclaiming staked expenditure stake
Staked expenditure owners should be able to reclaim their stake:
 - When cancelling an expenditure
 - After expenditure finalization

jakubcolony:
## Description

This PR adds the ability to reclaim the tokens staked against a staked expenditure after it has been finalized or cancelled.

### Reclaiming after cancellation
1. Create a staked expenditure as described in #994.
2. Cancel the expenditure while in draft. After refreshing, the reclaim button should appear:
image
3. Check your staked tokens amount before and after clicking the button. If the saga succeeds, you should get the stake for this expenditure back.
4. Reload the page, the button should disappear.

### Reclaiming after finalization
1. Create a staked expenditure as described in #994.
2. Progress the expenditure through statuses to Finalized. Similarly, after refreshing the reclaim button should appear.
Follow steps 3-4 from the previous point.

Resolves #1006


willm30 (commented):
Reclaiming working well!

Do we know whether we'll need to display the user's stake in a "Stakes" tab, much like currently in the token activation popover? 

Currently, that stakes tab assumes it's only going to be showing motion stakes, since previously you could only stake motions.

However, you might want to coopt that field on the Colony model for these stakes too? 

Captura de pantalla 2023-09-01 090224

willm30 (approved):

PR #1035, opened 2023-09-01
Fix add existing safe validation

### LINKED ISSUE #812: Improve "Add Safe" step 1 validation
> When you change from one chain to another with a valid safe address in the input, the status message Safe found on X chain will appear for some milliseconds before being replaced by the correct "This safe doesn't exist in this chain". Unfortunately, I haven't been able to fix this after trying several approaches, if you guys have any ideas then they are 100% welcome.

Image

thomassdk:
## Description

This PR fixes the validation for adding a safe. It removes the flickering that was present before as well as the brief flash of incorrect messages. All while removing code :scissors: 

[Screencast from 2023-09-01 19-07-07.webm](https://github.com/JoinColony/colonyCDapp/assets/32598350/5a8e7862-18d2-45fe-946e-df65d58e1fa7)

Resolves #812

### DIFF
--- a/src/components/common/Dialogs/AddExistingSafeDialog/AddExistingSafeDialog.tsx
+++ b/src/components/common/Dialogs/AddExistingSafeDialog/AddExistingSafeDialog.tsx
@@ -29,7 +29,6 @@ const AddExistingSafeDialog = ({
   close,
 }: Props) => {
   const navigate = useNavigate();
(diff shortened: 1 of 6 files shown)


ArmandoGraterol (approved):
Wow, nice job finding a fix for this 🎉 I've tried several times and couldn't fix it with the approaches I was trying.

Next time we see each other, I must buy you a beer :)

PR #1036, opened 2023-09-03
Cache expenditure payout claimed status

### LINKED ISSUE #982: Cache expenditure payout claimed status
To correctly display whether an expenditure payout is claimable or not, we need to cache that information in the database.

There's a new `ExpenditurePayoutClaimed` event being added in https://github.com/JoinColony/colonyNetwork/pull/1157 that will allow it.

This issue includes:
- Adding an appropriate schema field to the `ExpenditurePayout` type
- Handling the `ExpenditurePayoutClaimed` event in block-ingestor
- Dynamically showing/hiding Claim funds button depending on whether there are payouts to claim
- Modifying the `claimExpenditure` saga to ensure only unclaimed payouts are claimed

jakubcolony:
## Description

This PR brings in the necessary changes to store expenditure payout claimed status in the DB.

## Testing

- Create an expenditure with 2 slots, one with no claim delay, the other with an arbitrary claim delay.
- Progress the expenditure to finalization. Both slots should have claimed status of "No".
- Click on the claim button. If the second slot claim delay is far enough in the future, only the first slot would be claimed. 
- See that the claimed status changed to "Yes". You should also be unable to claim the same slot again.

Resolves #982

### DIFF
--- a/amplify/backend/api/colonycdapp/schema.graphql
+++ b/amplify/backend/api/colonycdapp/schema.graphql
@@ -2566,6 +2566,7 @@ type ExpenditureSlot {
 type ExpenditurePayout {
   tokenAddress: ID!
   amount: String!
+  isClaimed: Boolean!
 }
 
 type ExpenditureMetadata @model {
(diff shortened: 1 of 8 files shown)


willm30 (approved):
LGTM

PR #1044, opened 2023-09-06
Changed decision method to generic component, added container ref to …

TomaszChmielPagepro:
https://tw.auroracreation.com/app/tasks/15580162

image


src/components/v5/common/Fields/CardSelect/FormCardSelect.tsx
```diff
@@ -0,0 +1,32 @@
...
+    field: { onChange },
+    fieldState: { invalid, error },
+  } = useController({
+    name,
+  });
+  const value = useWatch({ name });
+
+  return (
+    <CardSelect<TValue>
+      {...{ ...props, value, onChange }}
```
joanna-pagepro, line 23: props -> rest

src/components/v5/common/Fields/CardSelect/types.ts
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { BaseFieldProps } from '../types';
+
+interface CardSelectOption<TValue> {
```
joanna-pagepro, line 4: Shouldn't this be exported as well?

webpack.dev.js
```diff
@@ -50,11 +50,6 @@ module.exports = () => ({
     new webpack.WatchIgnorePlugin({
       paths: [/css\.d\.ts$/],
     }),
-    new BundleAnalyzerPlugin({
```
joanna-pagepro, line 53: Why is this removed?

joanna-pagepro (changes requested):

webpack.dev.js
TomaszChmielPagepro: This is not needed in dev at all and it only slows the build process

src/components/v5/common/Fields/CardSelect/FormCardSelect.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import { useController, useWatch } from 'react-hook-form';
+import { FormCardSelectProps } from './types';
+import CardSelect from './CardSelect';
+import { FIELD_STATE } from '../consts';
+
+const displayName = 'v5.common.Fields.FormCardSelect';
```
krawczykkatarzyna, line 7: 'v5.common.Fields.CardSelect.FormCardSelect'

src/components/v5/common/ActionsContent/consts.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import { FormattedMessage } from 'react-intl';
+
+export const DECISION_METHOD_OPTIONS = [
```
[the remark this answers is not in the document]
TomaszChmielPagepro: Types added

src/components/v5/common/Fields/CardSelect/FormCardSelect.tsx
TomaszChmielPagepro: Renamed

src/components/v5/common/Fields/CardSelect/types.ts
TomaszChmielPagepro: Exported

webpack.dev.js
joanna-pagepro: But this wasn't added by us, so I guess that it would be good to talk to someone who can decide if this should stay or not.

src/components/v5/common/Fields/CardSelect/CardSelect.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const [uncontrolledValue, setUncontrolledValue] = useState<TValue>();
+  const uncontrolledOnChange = useCallback((newValue: TValue) => {
+    setUncontrolledValue(newValue);
+  }, []);
+
+  const value = valueProp ?? uncontrolledValue;
+  const onChange = onChangeProp ?? uncontrolledOnChange;
+
+  const selectedOption = useMemo(() => {
+    if (value === undefined) {
```
CzarekDryl, line 40: Can't this be `if (!value)`?

webpack.dev.js
krawczykkatarzyna: this was added a long time ago by some client side developer

src/components/v5/common/Fields/CardSelect/FormCardSelect.tsx
TomaszChmielPagepro: Name changed

src/components/v5/common/Fields/CardSelect/CardSelect.tsx
TomaszChmielPagepro: It will fail if `TValue` will be number and one of options value would be for example `0`

webpack.dev.js
TomaszChmielPagepro: I still think this is a good change as analyzing dev bundle makes no sense because the results are way different than those from the production bundle which is something that really matters

CzarekDryl (approved):

joanna-pagepro (approved):

PR #1057, opened 2023-09-07
Improve upgrade banner dismissal

### LINKED ISSUE #372: Improve `ColonyUpgrade` banner dismissal
The dismiss button in the upgrade banner only closes the banner temporarily and the banner returns after a refresh, which can be annoying, especially if the user can't do something about it (They don't have the Root permission to upgrade the colony).

We could save the state of the banner so it remains closed until the next colony version is released, the upgrade dialog can still be accessed through the actions button.

Improvement initially suggested by @Julianahlk

ArmandoGraterol:
Added improvements to one of the upgraded colony version banners.

A long time ago, Juliana correctly pointed out that the upgrade colony version banner comes back after a refresh even if you dismiss it, which is bad UX. This PR should remedy that.

To test the banner, just invert the `canUpgrade` in line 103 of `ColonyUpgrade`, click the `Dismiss` button, and reload the page to check if the banner reappears.

Resolves #372

### DIFF
--- a/src/components/common/ColonyHome/ColonyUpgrade/ColonyUpgrade.tsx
+++ b/src/components/common/ColonyHome/ColonyUpgrade/ColonyUpgrade.tsx
@@ -39,7 +39,13 @@ const MSG = defineMessages({
 
 const ColonyUpgrade = () => {
   const { colony } = useColonyContext();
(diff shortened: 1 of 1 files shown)


thomassdk (approved):
LGTM :rocket:

PR #1058, opened 2023-09-07
Remove unnecessary memoization from routes

### LINKED ISSUE #175: Tidy Routes
1. We're copying and pasting the "Routes" component every time we add a new route. I suggest we instead extract route data to a separate file and map through it, thus removing the need to repeat ourselves in Routes.tsx.

2. This comment is no longer CDapp specific and therefore may present an opportunity to un-memoize the routes. At the least, we should update the comment. Best case is we can remove the memoized switch.

Routes.tsx 
memoizedswitch

## TODO
- [ ] Refactor Routes at:
  - [ ] Routes.tsx
  - [ ] ColonyHome.tsx
- [ ] Test colony creation and app use in general without memoized routes.

ArmandoGraterol:
Self-explanatory PR, this memoization was needed in the Dapp and that's no longer the case in the CDapp. I've tested by creating several colonies, doing some actions, and enabling extensions and I didn't notice any extra re-render or weird behaviors.

Resolves #175

### DIFF
--- a/src/components/common/ColonyHome/ColonyHome.tsx
+++ b/src/components/common/ColonyHome/ColonyHome.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useState } from 'react';
+import React, { useState } from 'react';
 import {
   Outlet,
   Route,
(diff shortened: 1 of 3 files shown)


src/components/common/ColonyHome/ColonyHome.tsx
```diff
@@ -37,57 +36,54 @@ const ColonyHome = () => {
...
   );
 
-  const memoizedSwitch = useMemo(() => {
-    if (colony) {
-      return (
-        <RoutesSwitch>
-          <Route
-            path={COLONY_EVENTS_ROUTE}
-            element={
+  if (colony) {
```
jakubcolony, line 39: Changing this to a guard clause such as:
```
if (!colony) {
  return null;
}
```
would reduce nesting

src/routes/Routes.tsx
```diff
@@ -69,210 +67,185 @@ const MSG = defineMessages({
...
-    () => (
-      <RoutesSwitch>
-        <Route path="/" element={<Navigate to={LANDING_PAGE_ROUTE} />} />
-        <Route path={NOT_FOUND_ROUTE} element={<FourOFour />} />
-        <Route
-          path={LANDING_PAGE_ROUTE}
-          element={
+  // if (isAppLoading) {
+  //   return <LoadingTemplate loadingText={MSG.loadingAppMessage} />;
+  // }
```
jakubcolony, line 80: Is this still relevant? It might be worth creating an issue for it or removing the comment from code

jakubcolony (approved):
All good 👍

src/components/common/ColonyHome/ColonyHome.tsx
ArmandoGraterol: Nice catch!

src/routes/Routes.tsx
ArmandoGraterol: Good catch! I think this code is really old and no longer relevant

PR #1064, opened 2023-09-12
Feat/15567808 recipient user and wallet

krawczykkatarzyna:
Screen Shot 2023-09-15 at 10 11 53
Screen Shot 2023-09-15 at 10 12 11
Screen Shot 2023-09-15 at 10 35 52


src/components/v5/shared/IconWithTooltip/IconWithTooltip.tsx
```diff
@@ -0,0 +1,29 @@
+import React, { FC, PropsWithChildren } from 'react';
+import clsx from 'clsx';
+
+import Tooltip from '~shared/Extensions/Tooltip';
+import Icon from '~shared/Icon';
+import { IconWithTooltipProps } from './type';
```
CzarekDryl, line 6: type => types

src/components/v5/shared/IconWithTooltip/IconWithTooltip.tsx
```diff
@@ -0,0 +1,29 @@
...
+const IconWithTooltip: FC<PropsWithChildren<IconWithTooltipProps>> = ({
+  iconName,
+  tooltipContent,
+  className,
+  children,
+  onClick,
+}) => (
+  <Tooltip tooltipContent={tooltipContent} className="flex items-center">
+    {children}
+    <button type="button" onClick={onClick}>
```
CzarekDryl, line 19: I think there should be aria-label if there is no text inside button

src/components/v5/common/ActionsContent/partials/UserSelect/hooks.ts
```diff
@@ -26,3 +31,30 @@ export const useUserSelect = (): UserSelectHookProps => {
     isAccordion: false,
   };
 };
+
+export const useVerifyRecipient = () => {
```
CzarekDryl, line 35: Maybe it will be better to add isVerified prop to user object in `useUserSelect` hook? I think this will reduce the amount of logic in other components

src/components/common/Extensions/NotificationBanner/NotificationBanner.tsx
```diff
@@ -31,10 +33,13 @@ const NotificationBanner: FC<PropsWithChildren<NotificationBannerProps>> = ({
         {
           'bg-success-100 border-success-200': status === 'success',
           'bg-warning-100 border-warning-200': status === 'warning',
+          'bg-warning-100 border-warning-400': status === 'notVerified',
```
TomaszChmielPagepro, line 36: I don't think `notVerfied` is the correct status name for a banner

src/components/v5/shared/UserAvatar/UserAvatar.tsx
```diff
@@ -16,6 +16,7 @@ const UserAvatar: FC<UserAvatarProps> = ({
   size = 'xxs',
   userStatus,
   isContributorsList,
+  avatarClassName,
```
TomaszChmielPagepro, line 19: Why not just `className`

src/components/v5/common/ActionSidebar/partials/UserSelect/UserSelect.tsx
```diff
@@ -38,9 +41,10 @@ const UserSelect: FC<UserSelectProps> = ({ name }) => {
     HTMLButtonElement,
     HTMLDivElement
   >([isUserSelectVisible]);
+  const fieldValueFormat: string = isHexString(field.value) ? field.value : '';
```
TomaszChmielPagepro, line 44: Why do we need this? `field.value` should be user address or undefined

src/components/v5/common/ActionSidebar/partials/UserSelect/hooks.ts
```diff
@@ -18,11 +43,21 @@ export const useUserSelect = (): UserSelectHookProps => {
...
+
   return {
     loading,
     options: users || [],
     key: 'users',
     title: { id: 'actions.recipent' },
     isAccordion: false,
+    isUserVerified,
+    isRecipientNotVerified,
+    userFormat,
```
TomaszChmielPagepro, line 61: These things should be tied together with this
image
in a single hook as they are related to a selected user not to a list of options

src/components/v5/shared/SearchSelect/SearchSelect.tsx
```diff
@@ -153,7 +155,19 @@ const SearchSelect = React.forwardRef<HTMLDivElement, SearchSelectProps>(
...
+                      "
+                      onClick={() => onSelect?.(searchValue)}
+                    >
+                      <Avatar />
+                      <span className="max-w-[15.625rem] truncate">
+                        {searchValue}
+                      </span>
+                    </button>
+                  ) : (
                     <EmptyContent
```
TomaszChmielPagepro, line 171: Didn't you just break `showEmptyContent` prop?

krawczykkatarzyna: in that case empty content shoudn't be display here. In other places empty content works fine.

src/components/v5/common/ActionSidebar/partials/UserSelect/UserSelect.tsx
krawczykkatarzyna: we need this because in this task user have possibility to put some value in search and then this value should be display in every needed places.

src/components/common/Extensions/NotificationBanner/NotificationBanner.tsx
krawczykkatarzyna: what name do you suggest?

src/components/v5/shared/SearchSelect/SearchSelect.tsx
TomaszChmielPagepro: But in @CzarekDryl PR if the user pastes in the token address then the dropdown content is hidden

src/components/v5/common/ActionSidebar/partials/UserSelect/hooks.ts
TomaszChmielPagepro: I mean you can put this logic in the same hook no?

src/components/v5/common/ActionSidebar/partials/UserSelect/UserSelect.tsx
TomaszChmielPagepro: Can you check how @CzarekDryl did this in his `TokenSelect` component? Would that solve your issue?

src/components/common/Extensions/NotificationBanner/NotificationBanner.tsx
```diff
@@ -56,6 +58,7 @@ const NotificationBanner: FC<PropsWithChildren<NotificationBannerProps>> = ({
           <div
             className={clsx('text-sm max-w-[50rem] mt-1.5', {
               'text-negative-400': status === 'error',
+              'text-warning-400 font-semibold': status === 'warning',
```
TomaszChmielPagepro, line 61: I think you can add `font-semibold` to children that you are passing to this component

src/components/v5/common/ActionSidebar/partials/UserSelect/UserSelect.tsx
```diff
@@ -32,15 +35,21 @@ const UserSelect: FC<UserSelectProps> = ({ name }) => {
     },
   ] = useToggle();
   const { user: userByAddress } = useUserByAddress(field.value);
-  const userDisplayName = userByAddress?.profile?.displayName;
+  const { user: userByName } = useUserByName(field.value);
+  const userDisplayName =
+    userByAddress?.profile?.displayName ||
+    userByName?.profile?.displayName ||
+    field.value;
```
TomaszChmielPagepro, line 42: `field.value` will be a hex address, do we want to show it here? If we do I think we should also check for loading of `useUserByAddress` and `useUserByName` to avoid flashing the hex address while user details are loading

src/components/v5/common/ActionSidebar/partials/UserSelect/UserSelect.tsx
```diff
@@ -32,15 +35,21 @@ const UserSelect: FC<UserSelectProps> = ({ name }) => {
...
+  const userWalletAddress =
+    userByAddress?.walletAddress || userByName?.walletAddress || field.value;
```
TomaszChmielPagepro, line 44: I think it's safe to just use `field.value` here as it will always keep wallet address

src/components/v5/common/ActionSidebar/partials/UserSelect/UserSelect.tsx
```diff
@@ -52,16 +61,26 @@ const UserSelect: FC<UserSelectProps> = ({ name }) => {
...
+              className={
+                usersOptions.isRecipientNotVerified ? 'text-warning-400' : ''
+              }
+            />
+            {usersOptions.isUserVerified && (
+              <span className="flex ml-2 text-blue-400">
+                <Icon name="verified" />
+              </span>
+            )}
+          </>
```
TomaszChmielPagepro, line 81: What if `field.value` is set but the user is not found?

src/components/v5/common/ActionSidebar/partials/UserSelect/UserSelect.tsx
```diff
@@ -32,15 +35,21 @@ const UserSelect: FC<UserSelectProps> = ({ name }) => {
     },
   ] = useToggle();
   const { user: userByAddress } = useUserByAddress(field.value);
-  const userDisplayName = userByAddress?.profile?.displayName;
+  const { user: userByName } = useUserByName(field.value);
```
TomaszChmielPagepro, line 38: Are you sure we need `useUserByName` here? I believe we keep only the wallet address in `field.value`

src/components/v5/common/ActionSidebar/partials/UserSelect/hooks.ts
```diff
@@ -1,28 +1,51 @@
...
+    skip: !colonyAddress,
+  });
+  const isWhitelistActivated = metadata?.isWhitelistActivated;
 
-  const users = members?.map((member) => {
-    const { label, walletAddress, avatar } = member || {};
+  const verifiedUsers: SearchSelectOption[] = useMemo(
+    () =>
+      (data?.getContributorsByColony?.items ?? []).map((item, index) => ({
+        value: item?.user?.profile?.displayName,
```
TomaszChmielPagepro, line 21: The value should never be `diplayName` as it's not a good indicator of uniqueness. Use `walletAddress` or id instead

src/components/v5/shared/SearchSelect/SearchSelect.tsx
```diff
@@ -153,6 +155,19 @@ const SearchSelect = React.forwardRef<HTMLDivElement, SearchSelectProps>(
...
+                      className="text-sm md:hover:text-blue-400 flex items-center gap-2 min-h-[3.125rem]
+                      "
+                      onClick={() => onSelect?.(searchValue)}
+                    >
+                      <Avatar />
+                      <span className="max-w-[15.625rem] truncate">
+                        {searchValue}
+                      </span>
+                    </button>
+                  )}
```
TomaszChmielPagepro, line 170: I would try to find a more meaningful name for this prop. Something like `showSearchValueAsOption` or something like that

src/components/v5/shared/SearchSelect/hooks.ts
```diff
@@ -9,15 +9,15 @@ export const useSearchSelect = (
...
 
           return (
-            optionValue.toLowerCase().includes(searchValue.toLowerCase()) ||
+            optionValue?.toLowerCase()?.includes(searchValue?.toLowerCase()) ||
             optionWalletAddress
-              .toLowerCase()
-              .startsWith(searchValue?.toLowerCase() ?? '')
+              ?.toLowerCase()
+              ?.startsWith(searchValue?.toLowerCase() ?? '')
           );
```
TomaszChmielPagepro, line 21: Are all these null checks needed?

src/components/v5/shared/SearchSelect/partials/SearchItem/SearchItem.tsx
```diff
@@ -39,13 +37,14 @@ const SearchItem: FC<SearchItemProps> = ({
...
           missingPermissions,
+          walletAddress,
           token,
         }) => {
           const firstDisabledOption = options.filter(
             (option) => option.isDisabled,
           )[0];
           const labelText =
-            typeof label === 'string' ? label : formatMessage(label);
+            typeof label === 'string' ? label : formatText(label || '');
```
TomaszChmielPagepro, line 47: I think you do not need to check if label is string you can just pass label to `formatText`

src/components/v5/shared/SearchSelect/types.ts
```diff
@@ -22,10 +23,10 @@ export interface SearchSelectOptionProps {
 }
 
 export interface SearchSelectOption {
-  label: MessageDescriptor | string;
-  value: string;
+  label?: MessageDescriptor | string | null;
```
TomaszChmielPagepro, line 26: Does this make any sense to have an option without label?

src/components/v5/shared/SearchSelect/types.ts
```diff
@@ -22,10 +23,10 @@ export interface SearchSelectOptionProps {
...
+  value?: string | null;
```
TomaszChmielPagepro, line 27: Value should be required

src/components/v5/shared/UserAvatarPopover/UserAvatarPopover.tsx
```diff
@@ -46,31 +49,25 @@ const UserAvatarPopover: FC<UserAvatarPopoverProps> = ({
 
   const button = (
     <button
-      onClick={isMobile ? onOpenModal : noop}
-      onMouseEnter={isMobile ? noop : () => onOpenModal()}
-      onMouseLeave={isMobile ? noop : () => onCloseModal()}
+      onClick={isMobile ? () => setIsOpen(true) : noop}
+      onMouseEnter={isMobile ? noop : () => setIsOpen(true)}
+      onMouseLeave={isMobile ? noop : () => setIsOpen(false)}
```
TomaszChmielPagepro, line 54: Can you use `useToggle` here?

src/components/v5/common/ActionSidebar/partials/UserSelect/hooks.ts
krawczykkatarzyna: I know but when I'll change it on `walletAddress` then for sec next to user avatar instead of user name will be displayed wallet address and then will be replaced by user name if exist

src/components/v5/shared/SearchSelect/types.ts
krawczykkatarzyna: yes you're right but what this hook `useGetVerifiedMembersQuery` returns can be null | string | undefined

src/components/v5/shared/SearchSelect/types.ts
krawczykkatarzyna: yes you're right but in this hook `useGetVerifiedMembersQuery` label can be string | null | undefined

src/components/v5/common/ActionSidebar/partials/UserSelect/UserSelect.tsx
TomaszChmielPagepro: In which case?

src/components/v5/common/ActionSidebar/partials/UserSelect/UserSelect.tsx
TomaszChmielPagepro: Not sure what you mean by that, can you show me an example of how that looks when `field.value` is set and `userByName` and `userByAddress` is undefined?

src/components/v5/shared/SearchSelect/types.ts
TomaszChmielPagepro: So maybe you should filter options that are not possible to give them value? You can use `reduce` for that

src/components/v5/shared/SearchSelect/types.ts
TomaszChmielPagepro: https://github.com/JoinColony/colonyCDapp/pull/1064#discussion_r1354228669

CzarekDryl (approved):

src/components/v5/common/ActionSidebar/partials/UserSelect/hooks.ts
```diff
@@ -1,28 +1,52 @@
...
-      showAvatar: true,
-    };
-  });
+  const preparedUserOptions: SearchSelectOption[] = isWhitelistActivated
+    ? verifiedUsers
+    : members || [];
+
+  const isUserVerified = preparedUserOptions.some(
+    ({ label, walletAddress }) =>
+      label === inputValue || walletAddress === inputValue,
```
TomaszChmielPagepro, line 38: I don't think it is correct to compare the label to the field value

src/components/v5/common/ActionSidebar/partials/UserSelect/hooks.ts
```diff
@@ -1,28 +1,52 @@
...
-export const useUserSelect = (): UserSelectHookProps => {
+export const useUserSelect = (inputValue: string): UserSelectHookProps => {
   const { colony } = useColonyContext();
-  const { members, loading } = useGetColonyMembers(colony?.colonyAddress);
+  const { colonyAddress = '', metadata } = colony ?? {};
+  const { members, loading } = useGetColonyMembers(colonyAddress);
+  const { data, loading: verifiedMembersLoading } = useGetVerifiedMembersQuery({
+    variables: { colonyAddress },
+    skip: !colonyAddress,
+  });
```
TomaszChmielPagepro, line 15: Shouldn't this query be skipped when `isWhitelistActivated === false`  ?

src/components/v5/shared/SearchSelect/SearchSelect.tsx
```diff
@@ -153,6 +155,19 @@ const SearchSelect = React.forwardRef<HTMLDivElement, SearchSelectProps>(
...
```
TomaszChmielPagepro, line 170: How does this work now? This will be shown when no option is found for the query but when the search value is not an address it will do nothing?

src/components/v5/common/ActionSidebar/partials/UserSelect/hooks.ts
krawczykkatarzyna: fixed

src/components/v5/common/ActionSidebar/partials/UserSelect/hooks.ts
krawczykkatarzyna: fixed

src/components/v5/shared/SearchSelect/SearchSelect.tsx
TomaszChmielPagepro: What's the point of allowing invalid values in this input? I think we need to discuss this with the team

PR #1071, opened 2023-09-13
Remove updatedExtensionByColonyAndHash lambda function

### LINKED ISSUE #339: Move `updateExtensionByColonyAndHash` lamda to block ingestor
Currently, the `updateExtensionByColonyAndHash` is a lamda function that is only used by the block ingestor. Which means we don't need it as a lamda function at all. It can be removed and the functionality copied over to the block ingestor itself.

Before you do this, just triple check it is indeed not being called anywhere in the CDapp.

ArmandoGraterol:
Moved `updatedExtensionByColonyAndHash` lambda function logic to the block-ingestor as it's only used there.

Make sure creating a colony, installing, deprecating, enabling, re-enabling, upgrading, and uninstalling still work as expected. 

Resolves #339


willm30 (approved):
LGTM

jakubcolony (approved):
LGTM 👍

PR #1088, opened 2023-09-19
Implement "Release Expenditure Stage" motion

### LINKED ISSUE #1062: Releasing expenditure stage using motions
If a user is missing the permissions needed to release expenditure stage, and voting reputation extension is enabled in the colony, they should be able to create a motion to achieve that.

ArmandoGraterol:
## Description

- [] TODO write description

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

* Changes go here

**Deletions** ⚰️

* Deletions go here

## TODO

- [ ] Add TODOs

Resolves #1062


jakubcolony:
@ArmandoGraterol Could you change the base to `feat/advanced-payments-2` please?

jakubcolony (commented):
I'm unable to test this because I can't assign the required permissions to the Staged Expenditure extension. The banner says it succeeded but no tx has actually happened:

https://github.com/JoinColony/colonyCDapp/assets/112586815/2708e53c-d2f8-40b8-ab83-30172b89ebe4

@willm30 is this expected, ie. it's yet to be implemented?

willm30:
@jakubcolony You should still be able to assign permissions manually in the old UAC.

willm30:
> Before finalizing the release expenditure motion, you must advance a couple of weeks, otherwise, you will receive a voting-execution-failed-not-one-week.

This is a sign that there is something wrong. Specifically, it tells you the action is failing to execute. I would double check it has been correctly encoded and that it is targeting the right contract address. It may also be a gas problem, which you can test by manually setting the gas limit in the `finalizeMotion` saga to the network maximum. If that fails, confirm the extension is deployed correctly (i.e. that the method actually exists on the contract at the address you're targeting). Perhaps also make sure the colony network hash associated with this branch is up to date. 

If all else fails, I'd speak with Alex.

jakubcolony (changes requested):
Thanks @willm30, got it to work that way 👍 

Overall looks good @ArmandoGraterol, but we need to allow creating another motion for a stage if the previous one failed. We should also store the history of the motions for UI display purposes:
image

It should be just like the fund/cancel staked expenditure motions

src/redux/sagas/motions/expenditures/releaseExpenditureStageMotion.ts
```diff
@@ -0,0 +1,144 @@
...
+      type: ActionTypes.MOTION_RELEASE_EXPENDITURE_STAGE_ERROR,
+      payload: {
+        name: ActionTypes.MOTION_RELEASE_EXPENDITURE_STAGE_ERROR,
+        message: JSON.stringify(error),
+      },
+      meta,
+      error: true,
+    });
+  } finally {
+    createMotion.channel.close();
```
willm30, line 135: Also down here we need to put the tx hash on the url params. This is a fairly recent product requirement but it enables users to see the txhash and come back to it in the new ui. I have done this for all other sagas, for reference.

willm30:
@ArmandoGraterol What was the problem with the tx failing to execute in the end?

src/redux/sagas/utils/expenditureMulticallData.ts
```diff
@@ -136,3 +137,49 @@ export const getMulticallDataForPayout = (
 
   return encodedMulticallData;
 };
+
+const convertNumberToBytes32 = (number: number) =>
+  hexZeroPad(hexlify(number), 32);
```
jakubcolony, line 142: There's function doing that already - `toB32`

src/redux/sagas/utils/expenditureMulticallData.ts
```diff
@@ -136,3 +137,49 @@ export const getMulticallDataForPayout = (
...
+
+const mask = [false, true];
```
jakubcolony, line 144: I believe this mask is specific for a particular multicall, so it should be put inside the relevant method or renamed

src/redux/sagas/utils/expenditureMulticallData.ts
```diff
@@ -136,3 +137,49 @@ export const getMulticallDataForPayout = (
...
+
+export const getMulticallDataForStageRelease = (
+  expenditure: Expenditure,
+  slotId: number,
+  colonyClient: AnyColonyClient,
+  permissionDomainId: BigNumber,
+  childSkillIndex: BigNumber,
+  tokenAddresses: string[],
+) => {
+  const keys = [convertNumberToBytes32(slotId), convertNumberToBytes32(1)];
```
jakubcolony, line 154: This can use `EXPENDITURESLOT_CLAIMDELAY` defined above

jakubcolony (approved):
Nice work, I left a few comments about the multicall helper implementation, feel free to merge after addressing them

PR #1097, opened 2023-09-21
Fix permissions action not being created

### LINKED ISSUE #1008: Permissions fail to create actions list entry if setting all of them at once
I noticed this while working on #1004 whoever if seems very flaky and hard to reproduce.

- if you set all of the permissions to a new address they get set, and action shows up in the list
- if you then remove all of the permissions from that address they get removed, however the action does not show up in the list
- If you then try to set / remove permissions individually (ie: one permission at a time), they start showing up in the list

I think this is more to do with the Block Ingestor, rather than the UI.

So considerations / notes:
- Same behavior either with metatransactions on or off
- only noticed this developing locally
- no errors shown in the console, both client or server
- permissions always get set/removed and reflected accordingly to the transaction you've sent, just the entry in the actions list fails to get created

willm30:
## Description

This is a bug fix as identified here: https://github.com/JoinColony/colonyCDapp/issues/1008

I believe it's caused by setting the recovery role. The recovery role event is handled by the same handler as all other roles, but it has a different event signature: specifically, it doesn't include the `msg.sender`. We were using this as the `initiatorAddress` on the action, but since that field is non-nullable, the mutation fails. 

There are many potential solutions to this, but I feel this is the easiest and simplest.

**Changes** 🏗

* Make `initiatorAddress` nullable.

## Testing

Confirm you're not experiencing the same issues as colonyCDapp#1008 and that the action is created for all permutations of setting permissions.

Resolves #1008


rdig (approved):

willm30:
See https://github.com/JoinColony/block-ingestor/pull/134

PR #1118, opened 2023-09-29
feat: added styles for table on mobile

CzarekDryl:
https://tw.auroracreation.com/app/tasks/15583746

image


src/components/v5/common/Table/types.ts
```diff
@@ -1,7 +1,8 @@
-import { TableOptions } from '@tanstack/react-table';
+import { TableOptions, Row } from '@tanstack/react-table';
 
 export interface TableProps<T>
   extends Omit<TableOptions<T>, 'getCoreRowModel'> {
   getCoreRowModel?: TableOptions<T>['getCoreRowModel'];
   className?: string;
+  getRowClassName?: (row: Row<T>) => string | undefined;
```
krawczykkatarzyna, line 7: getRowClassName?: (row: Row<T>) => string;

src/components/v5/common/TableWithMeatballMenu/TableWithMeatballMenu.tsx
```diff
@@ -17,7 +17,15 @@ const TableWithMeatballMenu = <T,>({
     [columns, helper, getMenuProps],
   );
 
-  return <Table<T> columns={columnsWithMenu} {...rest} />;
+  return (
+    <Table<T>
+      getRowClassName={() =>
```
krawczykkatarzyna, line 22: Can't it just be a string?

src/components/v5/common/TableWithMeatballMenu/utils.tsx
```diff
@@ -12,6 +12,11 @@ export const makeMenuColumn = <T,>(
     cell: ({ row }) => {
       const props = getMenuProps(row);
 
-      return props ? <MeatBallMenu {...props} /> : undefined;
+      return props ? (
+        <MeatBallMenu
+          className="absolute top-4 right-4 md:top-auto md:right-auto"
```
joanna-pagepro, line 17: Is absolute needed on desktop?

joanna-pagepro (changes requested):

src/components/v5/common/TableWithMeatballMenu/utils.tsx
CzarekDryl: There is `md:relative` added in MeatBallMenu component

joanna-pagepro: So maybe you should add a wrapper here that will have absolute-related classes. This way you will separate the component's logic from the place where it's used

joanna-pagepro (approved):

src/components/v5/common/Table/types.ts
joanna-pagepro: The type is correct here. `?` doesn't mean that the function can also return undefined

PR #1122, opened 2023-10-01
Proposing changes to advanced payments 

### LINKED ISSUE #1080: Proposing changes to advanced payments
As a user, I want to be able to propose changes to expenditures once they're out of draft state. The scope of possible changes should be the same as editing an expenditure while in draft state.

This issue covers the case when user has permissions to propose changes. For proposing changes using motions, see #1081

jakubcolony:
## Description

This PR introduces the functionality of proposing changes to expenditures in locked state, which in case of using permissions is the same as simply editing locked expenditures.

### Testing
When you lock an expenditure, an Edit button should still be shown next to the payouts:
image

You should be able to edit the recipient, token address, amount and claim delay. You should also be able to add and remove payouts. 

All these interactions (apart from editing token address and amount) rely on the `setExpenditureValues` contract function which requires the caller to explicitly specify the storage slot of the values to be edited. You will see the code setting those params in `expenditureMulticallData` util file.

Resolves #1080


willm30 (commented):
Looking good, just noticing that the claims ui isn't updating following the deletion of a payment: 
Captura de pantalla 2023-10-17 091211

To reproduce:
Create an expenditure with two slots
Lock it
Delete one
Fund and finalize
Claim still showing as 2 slots

jakubcolony:
@willm30 good spot, fixed

willm30 (approved):

PR #1132, opened 2023-10-04
Implement `createStreamingPayment` motion

### LINKED ISSUE #1077: Creating streaming payments using motions
If a user is missing permissions to create a streaming payment, they should be able to create a motion to achieve that.

ArmandoGraterol:
- Enable the relevant extensions before testing it.
- After creating the motion, you should refresh and see at the bottom of the form the link for the newly created motion.
- Once the motion has passed, you should see all the data per the screenshot in the DB.

FireShot Capture 161 - Amplify GraphiQL Explorer - localhost

FireShot Capture 160 - Actions - Colony - wonker - localhost

Resolves #1077


src/redux/sagas/motions/expenditures/createStreamingPaymentMotion.ts
```diff
@@ -0,0 +1,185 @@
...
+    interval,
+    endCondition,
+    limitAmount,
+  },
+  meta,
+  meta: { setTxHash },
+}: Action<ActionTypes.MOTION_STREAMING_PAYMENT_CREATE>) {
+  const apolloClient = getContext(ContextModule.ApolloClient);
+
+  const batchId = 'motion-cstreaming-payment-create';
```
jakubcolony, line 51: typo

src/redux/sagas/motions/expenditures/createStreamingPaymentMotion.ts
```diff
@@ -0,0 +1,185 @@
...
+        type: ActionTypes.MOTION_STREAMING_PAYMENT_CREATE_SUCCESS,
+        meta,
+      });
+
+      window.history.replaceState(
+        {},
+        '',
+        `${window.location.origin}${window.location.pathname}?tx=${txHash}`,
+      );
+    }
```
jakubcolony, line 163: This doesn't seem to work for me, not getting redirected anywhere

jakubcolony (changes requested):
Spotted a few issues but overall nice job, I managed to create a stream using motions 👍 

I was only able to create one motion, subsequent attempts fail due to an error in `syncTransactionWithDb` util. It seems like the ID used to store the tx in the DB is always the same: `motion-cstreaming-payment-create-createMotion`, so it only works the first time.

I think it's probably due to using `batchId` in the saga instead of `meta.id`:
```ts
 const { createMotion /* annotationMessage */ } = yield call(
    createTransactionChannels,
    batchId,
    ['createMotion', 'annotateMotion'],
  );
```
And we have the same issue with cancel staked expenditure motion, would you be able to fix that as part of your changes?

Also can we keep 2 separate buttons for creating a streaming payment using permissions and motions?

src/redux/sagas/motions/expenditures/createStreamingPaymentMotion.ts
ArmandoGraterol: I think the idea of this pattern is for us to easily access the tx hash once we submit a motion, and use it to manually go to the motion page instead of being redirected which always results in a 404, is this right @willm30?

ArmandoGraterol:
Thank you for the review @jakubcolony 

> I think it's probably due to using `batchId` in the saga instead of `meta.id`:

 It should be fixed 👍 

> Also can we keep 2 separate buttons for creating a streaming payment using permissions and motions?

I noticed we are using an action form in the case of streaming payment, so to make the change easier I've added a force toggle alongside the submit button.

willm30 (approved):
LGTM 👍

ArmandoGraterol:
> @ArmandoGraterol is this now ready for re-review?

@jakubcolony It is Mr. Jakub! Review away

jakubcolony (approved):
Nice work, the force button works great 👌

PR #1147, opened 2023-10-10
Fix: `caches` undefined

willm30:
## Description

This pr checks `caches` for `undefined` in the event the app is being run in an insecure context.

This resolves an error when building for production.

### DIFF
--- a/lib/pinata/PinataCache.ts
+++ b/lib/pinata/PinataCache.ts
@@ -9,7 +9,7 @@ class PinataCache {
 
   async init(): Promise<Cache> {
     if (!this.cache) {
-      this.cache = await caches.open(PINATA_CACHE_STORE);
+      this.cache = await caches?.open(PINATA_CACHE_STORE);
     }
     return this.cache;
   }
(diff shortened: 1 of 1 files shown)


willm30:
@chmanie Can you double check the code change I've made to fix the problem you mentioned yesterday, and then we can confirm whether it actually resolves the problem after merging?

lib/pinata/PinataCache.ts
```diff
@@ -72,7 +72,11 @@ class PinataCache {
          * upload the IPFS blob ourself, in which case, we'll have to request it from
          * the Gateway
          */
-        await this.cache?.add(new Request(HASH_URL));
+        if (!this.cache) {
```
willm30, line 75: this prevents an infinite loop as `getCacheObject` is called from `setCacheObject` (which is called from `getCacheObject`)

chmanie:
Hey Will, thanks for the fix! I'll apply it to my branch and check again!

chmanie (approved):
This works now, great!

PR #1150, opened 2023-10-10
Add notifications web worker

### LINKED ISSUE #1138: Create web worker client to receive web push notifications
One avenue that users will receive notifications will be via Web Push Notifications.

https://web.dev/push-notifications-overview/

This issue is to create the client side service worker that will receive the web push notification when the app is in the background

thomassdk:
## Description

This PR adds the firebase notification messaging web worker to the project

https://firebase.google.com/docs/cloud-messaging/js/receive

It is built using [workbox](https://developer.chrome.com/docs/workbox/) by Chrome team as I could not get any other method of building it to work. 

Currently it is setup to get the data from the firebase message and create a notification using it.

To test go to the `useNotifications` PR and send a notification while the app is in the background. A notification with the colony logo should appear.

**New stuff** ✨

- A workers folder in `src` where we can but any future web workers
- The firebase web worker 

Resolves #1138

### DIFF
--- a/src/workers/firebase-messaging-sw.ts
+++ b/src/workers/firebase-messaging-sw.ts
@@ -1,7 +1,15 @@
-/* eslint-disable no-restricted-globals */
+/// <reference lib="WebWorker" />
 
+// export empty type because of tsc --isolatedModules flag
 import { initializeApp } from 'firebase/app';
 import { getMessaging, onBackgroundMessage } from 'firebase/messaging/sw';
+import { precacheAndRoute } from 'workbox-precaching';
+
+export type {};
+declare const self: ServiceWorkerGlobalScope;
+
+// eslint-disable-next-line no-underscore-dangle
+precacheAndRoute(self.__WB_MANIFEST);
 
 const firebaseConfig = process.env.FIREBASE_CONFIG
   ? JSON.parse(process.env.FIREBASE_CONFIG)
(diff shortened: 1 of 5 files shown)


src/workers/firebase-messaging-sw.ts
```diff
@@ -0,0 +1,65 @@
+/// <reference lib="WebWorker" />
+
+// export empty type because of tsc --isolatedModules flag
```
jakubcolony, line 3: That comment should probably be a few lines lower

webpack.base.js
```diff
@@ -180,6 +181,11 @@ const config = {
       process: 'process/browser',
       Buffer: ['buffer', 'Buffer'],
     }),
+    new InjectManifest({
+      swSrc: './src/workers/firebase-messaging-sw.ts',
+      swDest: 'firebase-messaging-sw.js',
+      exclude: [/./],
+    }),
```
jakubcolony, line 188: So we will need to add one of these for each service worker we create?

jakubcolony (approved):

webpack.base.js
thomassdk: indeed however there is a different version of creating a web worker with the google web work called `generateSW` but that method doesnt work with the notifications api

https://developer.chrome.com/docs/workbox/the-ways-of-workbox/

PR #1151, opened 2023-10-10
Add notifications to schema

### LINKED ISSUE #1124: Create notifications in amplify schema
All notifications need to be sorted via date and each notification consists of:

- From (which Colony the notification comes from)
- Time
- Read/Unread flag
- Text

The text can contain mentions to users and actions and these should be linked

https://docs.amplify.aws/guides/api-graphql/query-with-sorting/q/platform/js/

thomassdk:
## Description

This PR adds the notification types and models to the amplify schema including the notification settings.

These models will undoubtably change as the feature develops but this PR gives a solid baseline to work from.

Resolves #1124

### DIFF
--- a/amplify/backend/api/colonycdapp/schema.graphql
+++ b/amplify/backend/api/colonycdapp/schema.graphql
@@ -1020,7 +1020,7 @@ type NotificationSettings {
   Device registration tokens for push notifications
   If this is null or the array is empty the user is not subscribed to push notifications
   """
-  notificationTokens: [String]
+  notificationTokens: [String!]
   """
   Indicates whether the user wants to be emailed
   """
@@ -1082,7 +1082,7 @@ type Notification @model {
   The links connected to what is signified in the text
   The order matches what is in the string
   """
-  links: [String]
+  links: [String!]
 }
 
 """
--- a/amplify/backend/api/colonycdapp/schema.graphql
+++ b/amplify/backend/api/colonycdapp/schema.graphql
@@ -1011,6 +1011,30 @@ type Profile @model {
   """
   notificationSettings: NotificationSettings
 }
+
+"""
+An object holding the personal notification settings of a user
+"""
+type NotificationSettings {
+  """
+  Device registration tokens for push notifications
+  If this is null or the array is empty the user is not subscribed to push notifications
+  """
+  notificationTokens: [String]
+  """
+  Indicates whether the user wants to be emailed
+  """
+  enableEmail: Boolean
+  """
+  Indicates whether the user wants to be notified on mentions
+  """
+  enableMention: Boolean
+  """
+  Indicates whether the user wants to be notified on assignments
+  """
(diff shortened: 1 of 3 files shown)


amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -998,11 +998,91 @@ type Profile @model {
...
+
+"""
+An object holding the personal notification settings of a user
+"""
+type NotificationSettings {
+  """
+  Device registration tokens for push notifications
+  If this is null or the array is empty the user is not subscribed to push notifications
+  """
+  notificationTokens: [String]
```
jakubcolony, line 1023: Can elements of this array be `null`?

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -998,11 +998,91 @@ type Profile @model {
...
+Notifications that a user may receive from a Colony
+"""
+type Notification @model {
+  """
+  User ID the Notification was sent to
+  """
+  userId: ID!
+    @index(
+      name: "byUser"
+      queryField: "notificationsByDate"
```
jakubcolony, line 1048: `getNotificationsByDate` for consistency

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -998,11 +998,91 @@ type Profile @model {
...
+  """
+  The text of the notification with indications which string portions are links
+  eg. 'This signifies a {link}'
+  """
+  text: String!
+  """
+  The links connected to what is signified in the text
+  The order matches what is in the string
+  """
+  links: [String]
```
jakubcolony, line 1085: Same as above, can the elements be `null`?

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -998,11 +998,91 @@ type Profile @model {
...
   user: User! @belongsTo(fields: ["id"])
+  """
+  List of a user's notifications
+  """
+  notifications: [Notification] @hasMany
+  """
+  Notification settings
+  """
+  notificationSettings: NotificationSettings
+}
```
jakubcolony, line 1013: What's the reasoning behind putting notifications on the profile model while the notification model itself has a relationship with the user model?

jakubcolony (commented):
Left a few comments

amplify/backend/api/colonycdapp/schema.graphql
thomassdk: None this is an oversight on my part which was left over from a previous experimentation where I had attached the Notifications to the user model.

Over time as I worked on this feature I decided that the `Profile` model was the correct place for Notifications as it uses elements of that model (such as the user's email) and generally seems to be where we store information about the user. Happy to be argued against however :)

I've updated the model to change the notification model to have a relationship with the *profile* now.

PR #1153, opened 2023-10-10
Subscribe to colony push notifications

### LINKED ISSUE #1140: Add ability to subscribe to colony wide notifications
The user should be able to subscribe to all notifications from a colony.

thomassdk:
## Description

This PR adds a lambda that handles subscribing (and unsubscribing) users to colony wide push notifications.

A contributor now has notification settings as part of their model.

**NOTE** There is a boolean in this case to track a users subscription to a colony as we have no indication from the firebase messaging SDK who is subscribed to a topic. (A topic is the name they use for broadcasting messages)

Resolves #1140


jakubcolony (approved):
LGTM, but you are missing an updated `generated.ts` following adding a new mutation. Make sure to run `npm run codegen` and commit it

PR #1170, opened 2023-10-12
Change active installs to use db 

### LINKED ISSUE #1134: Add numbers of installs metadata for extensions
https://www.notion.so/colony/Will-s-Specs-f92c55ff64b3424ca6b08b9c4457b876

Todo:

-- add metrics model
-- initialise it in block ingestor on startup (dev only)
-- add it to the qa env
-- update counter on extension installation

willm30:
## Description

This PR enables the storing of global extension installation information in the db, so users can see how many times an extension has been installed on the chain.

**New stuff** ✨

* Add `ExtensionInstallationsCount` to schema

Block ingestor: https://github.com/JoinColony/block-ingestor/pull/141

Resolves #1134

### DIFF
--- a/src/components/frame/Extensions/pages/partials/ActiveInstalls.tsx
+++ b/src/components/frame/Extensions/pages/partials/ActiveInstalls.tsx
@@ -1,6 +1,6 @@
 import React, { FC } from 'react';
 
-import { MAX_INSTALLED_NUMBER } from '~constants';
+import { EXTENSIONS_COUNT_THRESHOLD } from '~constants';
 import { ActiveInstallsProps } from './types';
 import ExtensionStatusBadge from '~v5/common/Pills/ExtensionStatusBadge';
 import { formatText } from '~utils/intl';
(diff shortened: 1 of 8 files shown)


src/components/frame/Extensions/pages/partials/ActiveInstalls.tsx
```diff
@@ -1,35 +1,25 @@
...
-              text={formatText({ id: 'status.new' })}
-            />
-          ) : (
-            <p className="text-gray-400 text-sm text-right">
-              {activeInstalls.toLocaleString('en-US')}{' '}
-              {formatMessage({ id: 'active.installs' })}
-            </p>
-          )}
-        </>
+      {activeInstalls < MAX_INSTALLED_NUMBER ? (
```
willm30, line 13: If number of `activeInstalls` is a below this constant (currently 39 (?) ), it will display a "new" status pill, else it'll show the actual count.

jakubcolony: Could we rename this const to something more meaningful?

jakubcolony (approved):
LGTM, I guess the alternative could be to experiment with the `@searchable` directive and the count it returns but that's still in a PR. Good to go!

PR #1171, opened 2023-10-12
Feat/15622636 permission pills

CzarekDryl:
https://tw.auroracreation.com/app/tasks/15622636

image
image

### DIFF
--- a/src/components/v5/common/Pills/PermissionsBadge/PermissionsBadge.tsx
+++ b/src/components/v5/common/Pills/PermissionsBadge/PermissionsBadge.tsx
@@ -3,6 +3,8 @@ import React, { FC } from 'react';
 import { PermissionsBadgeProps } from './types';
 import PillsBase from '../PillsBase';
 
+const displayName = 'v5.common.Pills.PermissionsBadge';
+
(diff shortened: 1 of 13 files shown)


src/components/v5/common/Pills/PermissionsBadge/PermissionsBadge.tsx
```diff
@@ -0,0 +1,15 @@
+import React, { FC } from 'react';
+
+import { PermissionsBadgeProps } from './types';
+import PillsBase from '../PillsBase';
+
```
krawczykkatarzyna, line 5: missing displayname

krawczykkatarzyna (approved):

src/components/v5/common/Pills/PermissionsBadge/types.ts
```diff
@@ -0,0 +1,5 @@
+import { PillsProps } from '../types';
+
+export interface PermissionsBadgeProps extends PillsProps {
+  isMultiSig?: boolean;
```
TomaszChmielPagepro, line 4: This prop name is a bit too tied to its use case. As this is a global component can we try to find another solution? Maybe we could just pass the icon?

TomaszChmielPagepro (approved):

PR #1175, opened 2023-10-16
Port sendNotification lambda to typescript

ink-pot-monkey:
## Description

This PR ports the `sendNotification` lambda to typescript and in the process sets up a bundling and compiling pipeline that can be used for all lambdas.

`Esbuild` is used to compile and bundle the lambda, resulting in a much smaller size of lambda which should result in faster startup times and thus less cost per lambda invocation.

To setup a lambda to be built with typescript the following steps have to be followed:

1. Add the lambda to the package.json scripts section in the following format: `"amplify:<function-name>": "./scripts/bundle-lambda.sh \"$npm_lifecycle_event\""`. This is so that when `amplify push` is run amplify knows how to bundle the lambda to be compressed and pushed to AWS.
2. Move all files in the `src` folder of the lambda to a `lib` folder and convert them to typescript files. Remove the `src` folder as esbuild will created when bundling and compiling the typescript files into a single js file. 

**Note** to add a package to a typescript lambda install it to the main project's dev dependencies and esbuild will treeshake and bundle any dependencies in the lambda for us.

**New stuff** ✨

1. The script `./scripts/bundle-lambda.sh` which parses the lambda name either from the package json format `amplify:<function-name>` (when called during `amplify push` or from the path format (when called from the lambda watcher).
2. A [chodikar](https://github.com/paulmillr/chokidar) watcher that rebundles the lambdas whenever changes are made (this is added to the `npm run dev` command to ease development)

**Changes** 🏗

* Common typescript lambda util functions have been moved to `amplify/backend/function/` to be shared by all lambdas


amplify/backend/function/sendNotification/lib/index.ts
```diff
@@ -0,0 +1,238 @@
...
+    [apiKey, graphqlURL, firebaseAdminConfig, mailJetApiKey, mailJetApiSecret] =
+      await getParams([
+        'appsyncApiKey',
+        'graphqlUrl',
+        'firebaseAdminConfig',
+        'mailJetApiKey',
+        'mailJetApiSecret',
+      ]);
+  }
+};
```
rdig, line 45: Part of this logic (api keys, configs, etc) should be fetched inside the `email` file, so it would make calling the specific messaging functions more readable.

Ie. something like:
```
await sendEmail(params);
```
Would become
```
await sendEmail(title, body);
```

amplify/backend/function/sendNotification/lib/index.ts
```diff
@@ -0,0 +1,238 @@
...
+
+const removeInterpolationMarkers = (message: string) => {
```
rdig, line 47: There's a lot of "heavy" logic inside this file which is hard to grasp at cursory glance. 

Please add some comments.

amplify/backend/function/sendNotification/lib/index.ts
```diff
@@ -0,0 +1,238 @@
...
+  >(GetColonyName_SnDocument, { id: colonyId || '' }, graphqlURL, apiKey);
+
+  if (colonyQuery.errors || !colonyQuery.data) {
+    const [error] = colonyQuery.errors;
+    throw new Error(error || 'Could not fetch colony data from DynamoDB');
+  }
+
+  const { name, metadata } = colonyQuery?.data?.getColony || {};
+
+  const title =
```
rdig, line 88: I know this was not spec'd out properly (or at all), but I'm making a case, that since notifications are app breaking, they should not have fallbacks of last resort.

So if a proper title cannot be sourced (either custom or default), the system should just stop and not send anything else out.

This is true for titles, body, names, etc, both in users and colonies

package.json
```diff
@@ -183,6 +193,8 @@
     "@tiptap/pm": "^2.0.4",
     "@tiptap/react": "^2.0.4",
     "@tiptap/starter-kit": "^2.0.4",
+    "@types/graphql": "^14.5.0",
+    "@types/node-mailjet": "^3.3.9",
```
rdig, line 197: Why is this here? This is a dev dependency

package.json
```diff
@@ -215,6 +227,7 @@
     "move-decimal-point": "^0.0.4",
     "nanoid": "^3.1.31",
     "node-fetch": "^2.6.7",
+    "npm-run-all": "^4.1.5",
```
rdig, line 230: Same with this one, a dev dependency

src/graphql/queries/lambdas/sendNotifications.graphql
```diff
@@ -0,0 +1,53 @@
+query GetUserDetails_Sn($id: ID!) {
```
rdig, line 1: There has to be a better place for this file.

Even with it's own folder, we're mixing up queries to go in the client, with queries that go on the "server"

src/graphql/queries/lambdas/sendNotifications.graphql
```diff
@@ -0,0 +1,53 @@
...
+          id
+          displayName
+          email
+        }
+      }
+    }
+  }
+}
+
+mutation CreateNotification_Sn($input: CreateNotificationInput!) {
```
rdig, line 49: Also, this is mutation, placed inside the "queries" folder

rdig (changes requested):
Overall seems good, and on preliminary tests it seems to be working.

I'm just rejecting in based on opportunities for improvement. I've left comments where I noticed parts that can be either refactored or done better

amplify/backend/function/getParams.ts
```diff
@@ -0,0 +1,64 @@
+const { default: fetch } = require('node-fetch');
```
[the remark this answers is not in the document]
thomassdk: This is intentional, as we will need both approaches as we migrate to the typescript approach outlined in this PR (if it is agreed that this is the correct approach)

Eventually `colonycdappSSMAccess` wont be needed but only once this transition is complete

amplify/backend/function/getParams.ts
```diff
@@ -0,0 +1,64 @@
...
+    throw Error(`No AWS Session Token found`);
+  }
+
+  const decrpytParam =
+    paramName === 'appsyncApiKey' ? '&withDecryption=true' : '';
+
+  // Retrieve param from Parameter Store
+  try {
+    const res = await fetch(
+      `http://localhost:2773/systemsmanager/parameters/get?name=${ParamNames[paramName]}${decrpytParam}`,
```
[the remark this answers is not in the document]
thomassdk: done this is now:

`http://${AWS_PARAMS_ENDPOINT}/systemsmanager/parameters/get?name=${ParamNames[paramName]}${decrpytParam}`,`

amplify/backend/function/sendNotification/lib/email.ts
```diff
@@ -1,8 +1,15 @@
-const Mailjet = require('node-mailjet');
-const { graphqlRequest } = require('./utils');
-const { getColonyContributors } = require('./graphql');
+import Mailjet from 'node-mailjet';
```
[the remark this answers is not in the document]
thomassdk: nice fun fact :nerd_face: 

I have changed the file to `sendEmailNotifications.ts`

amplify/backend/function/sendNotification/lib/index.ts
thomassdk: Ive added comments but more importantly refactored the code so that the push notification capability lives in its own file which makes the readability much clearer

amplify/backend/function/sendNotification/lib/index.ts
thomassdk: At the moment this is true for the body, however for the title I am not convinced. 

Does a colony have to have a display name, or is this optional? Does it have to have a name or again is this optional? If we allow them to be optional (as the schema suggests) I don't think we should break notifications if they are left blank.

The same goes for the username, do we allow the username to be blank? If so, I think it is reasonable to fall back onto the user address in order to send the notification.

To me the only real breakage is if a body cannot be constructed and if all the fallbacks for title etc. fail

src/graphql/queries/lambdas/sendNotifications.graphql
thomassdk: This was a concious decision as I wanted to keep all the graphql queries for a lambda in one file, now that this file is moved to `amplify/backend/function/sendNotification/lib/sendNotifications.graphql` do you think it makes sense to keep queries and mutations together like this?

PR #1177, opened 2023-10-16
Port subscribe to colony push notifications lambda to typescript

ink-pot-monkey:
## Description

This PR ports the `subscribeToColonyPushNotifications` lambda to typescript using the system set out in #1175.

**New stuff** ✨

* The output from the `watchLambda` script has been modified to be prefixed with `lambda    |` to match the output from the docker containers.
* `setEnvVariables` has been added to the `getParams` util file to reduce code duplication


amplify/backend/function/getParams.ts
```diff
@@ -62,3 +62,15 @@ export const getParams = async (params: Array<keyof typeof ParamNames>) => {
 
   return Promise.all(params.map((param) => getParam(param)));
 };
+
+export const setEnvVariables = async (
```
rdig, line 66: Am I misunderstanding this? This method returns either the qa params array, or an empty array?

So why it called `set` instead of `get` ?

amplify/backend/function/subscribeToColonyPushNotifications/lib/index.ts
```diff
@@ -1,60 +1,47 @@
...
+  GetProfile_StcpnQueryVariables,
+  GetProfile_StcpnDocument,
+  UpdateColonyContributorMutation,
+  UpdateColonyContributorMutationVariables,
+  UpdateColonyContributorDocument,
+} from './types';
 
 let apiKey = 'da2-fakeApiId123456';
 let graphqlURL = 'http://localhost:20002/graphql';
 let firebaseAdminConfig;
```
rdig, line 19: Weren't these extracted into a util somewhere ? So that they don't need to be re-declared everywhere ?

package.json
```diff
@@ -30,7 +30,8 @@
     "postinstall": "./scripts/lambda-functions-dependencies.sh && npm run build-lambdas",
     "watch-lambdas": "./scripts/watch-lambdas.js",
     "build-lambdas": "npm-run-all \"amplify:*\"",
-    "amplify:sendNotification": "./scripts/bundle-lambda.sh \"$npm_lifecycle_event\""
+    "amplify:sendNotification": "./scripts/bundle-lambda.sh \"$npm_lifecycle_event\"",
+    "amplify:subscribeToColonyPushNotifications": "./scripts/bundle-lambda.sh \"$npm_lifecycle_event\""
```
rdig, line 34: Are these scripts required for all lambdas that we convert to typescript? Or just the ones that deal with notifications ?

rdig (changes requested):
Just some changes here, as most work will be done in the branch this is based off #1175

amplify/backend/function/getParams.ts
thomassdk: this was a "copy and not think"™ about error, ive since refactored it so that it is much cleaner

amplify/backend/function/subscribeToColonyPushNotifications/lib/index.ts
thomassdk: They are indeed (well the fallback for graphql queries is these local host varaibles) this code has been updated to reflect that

package.json
thomassdk: Unfortunately, we will need to do this for every lambda that we convert to typescript, as it is how amplify knows how to build these lambdas before sending them to AWS lambda on the `amplify push` command. 

It's a bit of a drawback to have these clutter the package.json, but I think having typescript in the lambdas is worth the trade-off

PR #1182, opened 2023-10-16
Streaming payment details UI

### LINKED ISSUE #1106: Streaming payment details UI
This issue is to add a list of colony streaming payments, in which clicking on a payment would open a streaming payment details page, similarly to how the current expenditure list does it.

The following details should be shown:
- Recipient address
- Start date/time
- Token address and amount for all payouts
- Interval 
- End condition
- Depending on end condition:
 - End date/time
 - Limit amount

jakubcolony:
## Description

This PR adds a streaming payment details page which shows all the data we store in the DB:
image

You should get redirected to the streaming payment details page after creating a streaming payment.

Resolves #1106


willm30 (approved):
LGTM 👍

PR #1197, opened 2023-10-23
Re-skin create colony flow for beta invite

### LINKED ISSUE #1161: Colony creation flow (UI)
## Goal

To implement the colony creation flow in the new UI. 

As far as I'm concerned, that's basically just changing the UI (from the CDapp's) but nothing else. 

This issue includes creating the new UI and wiring it up like in the CDapps old UI, with the corresponding checks and validations.

Furthermore the gas station UI needs to be replicated to some extent (not sure yet how this will be handled in the new UI).

Colonies should be created in 5 steps (4 if the user brings their own token).

This flow will be treated in isolation from the user registration flow for now.

## Notes / hints

We will need the custom Colony URL component

Image

## Figma link and preview

[Figma link](https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-V5?type=design&node-id=11978-121935&mode=design&t=SVQOALwEYzDMuqwb-4)

Image

thomassdk:
## Description

This PR reskins the create colony flow for the beta launch.

To test you need to generate an invite code this can be done in the graphql playground http://localhost:20002/ with the following query:
```graphql
mutation CreateInviteCode {
  createPrivateBetaInviteCode(input: { shareableInvites: 2}) {
    id
  }
}
```

This will return an invite code that you can use in the path http://localhost:9091/create-colony/invite-code

**Changes** 🏗

* The Private beta invite validity is now determined by its shareable invite parameter the default of which is 2 (1 for the initial invite and then a subsequent on which is attached to the initial user
* The `createUniqueColony` lambda has been updated to check for invite validity
* You can now go back and forth in steps of the wizard

**Deletions** ⚰️

* All the css files for the wizard have been deleted

Resolves #1161


amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -1775,6 +1783,11 @@ type User @model {
   """
   transactionHistory: [Transaction!]
     @hasMany(indexName: "byInitiatorAddress", fields: ["id"])
+
+  """
+  A user who has been invited by colony will be able to pass on the private beta invite
+  """
+  PrivateBetaInviteCode: PrivateBetaInviteCode @hasOne
 }
```
ArmandoGraterol, line 1791: This prop should start with a lowercase like all the others, right?

ArmandoGraterol (changes requested):
A good chunk of work! I think it would have been best to separate and make an individual PR for each step so that it's easier to review and provide feedback, nonetheless:


- Something I've noticed is that the descriptions of the steps have a lighter color than in the specs
- Also the inputs have a lot of space between them, is that because we are accounting for the space of the input pills beforehand? If that's the case, I don't think the first input needs the padding as it's not displaying any pill below it

FireShot Capture 184 - Create Colony - Colony - localhost (1)

Specs: 
FireShot Capture 185 - Colony V5 – Figma - www figma com (1)

- The input pill in the URL is appearing and overlapping the max length error:
FireShot Capture 186 - Create Colony - Colony - localhost

- Same as the first step, the description is in a lighter color than the specs
- I've noticed that the icon is bigger too but I think we don't have a 24x24 size for icons anyways.
- When selected, there's no blue border in the inputs.
FireShot Capture 187 - Create Colony - Colony - localhost

Specs:
FireShot Capture 188 - Colony V5 – Figma - www figma com

- I tried to upload a random image as the token icon and the CDapp crashed:
FireShot Capture 189 - Create Colony - Colony - localhost
- Also, even after this error, when I closed it, I noticed that there was still a popup telling me that the change was made successfully.
FireShot Capture 190 - Create Colony - Colony - localhost

- The token symbol input has a generic error message that is not very user-friendly:
FireShot Capture 192 - Create Colony - Colony - localhost

- Btw, I think the specs changed between the time this was being implemented. Because in the specs the description is different and there's a popup that is shown below the form when the token address is not found:
FireShot Capture 193 - Colony V5 – Figma - www figma com

- I think we shouldn't attempt to fetch the data of a token address if the address is not valid:
FireShot Capture 194 - Create Colony - Colony - localhost
- On that note, I wasn't able to see this error that appears in the specs, maybe because the fetching above is overriding it?
FireShot Capture 197 - Colony V5 – Figma - www figma com

- The description in the last step also has a lighter color than in the specs:
FireShot Capture 195 - Create Colony - Colony - localhost

- Is this plus sign icon a temporary placeholder? From what the specs say, I think the placeholder should be a blockie if the user doesn't upload the icon. For a second I thought I was able to click it and add the token icon:
FireShot Capture 196 - Create Colony - Colony - localhost

src/components/v5/common/CreateColonyWizard/TokenSelector.tsx
```diff
@@ -0,0 +1,83 @@
...
+      input: {
+        tokenAddress,
+      },
+    },
+    skip: !isAddress(tokenAddress),
+  });
+
+  useEffect(() => {
+    setIsLoading(isFetchingAddress);
+  }, [isFetchingAddress]);
```
ArmandoGraterol, line 53: Missing dependencies here

src/components/v5/common/CreateColonyWizard/TokenSelector.tsx
```diff
@@ -0,0 +1,83 @@
...
+    if (!tokenAddressDirty) {
+      return;
+    }
+
+    // When token is updated (either found or null), clear errors and set the values in hook-form
+    clearErrors(addressFieldName);
+    setValue(tokenFieldName, token);
+    trigger(addressFieldName);
+  }, [
+    tokenAddressField,
```
ArmandoGraterol, line 67: Same here

ArmandoGraterol (approved):
Nice work addressing everything! It's a lot of code

PR #1206, opened 2023-10-24
Enable local testing of amplify's @searchable directive

thomassdk:
## Description

This PR modifies the local dev environment so that it is possible to do local development with amplify's [@searchable](https://docs.amplify.aws/cli/graphql/search-and-result-aggregations/) directive.

> Add the @searchable directive to an @model type to enable OpenSearch-based data search and result aggregations. This gives you the ability to:
> search for data using advanced filters, such as substring matching, wildcards, regex, and/or/not conditions
> get aggregation values, such as sum, average, min, max, terms
> retrieve total search result count
> sort the search results across one or multiple fields 

This PR is in answer to the question for how to get colony stats for the colony dashboard page.

### Testing

To test this PR some file permissions need to change to enable the user inside the docker container to correctly read and write mounted files.

To do so first reset all the permissions by running the following command in the root of the project:

```shell
sudo chown -R $(whoami):$(ls -l package.json | awk '{print $4}') ./ && sudo chmod  -R g+rws  ./amplify
```

This is because some files in the project were committed under the root ownership and with the new docker configuration this means amplify will fail when it starts. 

**Changes** 🏗

* The Python is been set to 3.8.18 using pyenv in the base image for opensearch
* Codegen now exists outside the docker container
* `@seachable` is now enabled which can be added to any graphql model in the schema
* The user that runs amplify in the amplify container is `node`

## TODO

- [x] Set up amplify opensearch for [production environments](https://docs.amplify.aws/cli/graphql/search-and-result-aggregations/#set-up-opensearch-for-production-environments) 
- [x] Compile python for arm architecture in the amplify dev image for macs


ArmandoGraterol (approved):
Still working as expected for me! Nice job 👌

codegen.js
```diff
@@ -30,7 +31,10 @@ const codegen = async () => {
   try {
     await fetchSchema();
 
-    const graphqlFiles = './src/graphql/**/*.graphql';
+    const graphqlFiles = [
+      './src/graphql/**/*.graphql',
+      './amplify/backend/api/colonycdapp/schema.graphql',
```
jakubcolony, line 36: Why do we need this file here? This array lists files that declare queries, not the schema itself

jakubcolony (commented):
~~Hey @thomassdk, I can't get this to work, maybe we can jump on a call tomorrow to set it up?~~

Docker caching got me, it did work after clearing all the images - nice work! I'll wait for your answer to the questions and approve

codegen.js
```diff
@@ -56,8 +60,36 @@ const codegen = async () => {
...
+      callback();
+    } else {
+      setTimeout(() => {
+        waitForServiceToBeResponsive(callback);
+      }, 1000); // Check every second
+    }
+  });
+}
+
+waitForServiceToBeResponsive(codegen);
```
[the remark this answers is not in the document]
thomassdk: Waiting for amplify as above worked while we this was part of the docker container, however now that the codegen exists outside the docker container this version of waiting means that we get a **lot** of SPAM messages from the codegen, the purpose of this extra layer of waiting is to replicate the wait until the docker container service is up and running

codegen.js
thomassdk: you are completely correct this was a mistake on my end as I was trying to get changes to the schema to be automatically picked up. I have now removed this extra file and discovered that it is possible to regenerated the `generated.ts` file if we run codegen in a separate terminal (and then close it down) *after* the amplify container has correctly generated a new schema from the changes. This could be done nicer but I think it is good enough for now

jakubcolony: I don't think there's much value in running codegen together with amplify as you're most likely to need it when you change the schema, and as you've noticed that requires running it separately anyway. I'll raise this at the dev call next week

jakubcolony (approved):

PR #1231, opened 2023-10-31
Changed event to initialValues state

TomaszChmielPagepro:
Changed how initial action in ActionSidebar is set

### DIFF
--- a/src/components/frame/Extensions/layouts/Header.tsx
+++ b/src/components/frame/Extensions/layouts/Header.tsx
@@ -20,6 +20,7 @@ const Header: FC<HeaderProps> = ({ navBar = null, txButtons, userHub }) => {
       isActionSidebarOpen,
       { toggleOn: toggleActionSidebarOn },
     ],
+    actionSidebarInitialValues,
   } = useActionSidebarContext();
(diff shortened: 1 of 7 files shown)


src/components/frame/Extensions/layouts/Header.tsx
```diff
@@ -35,7 +36,10 @@ const Header: FC<HeaderProps> = ({ navBar = null, txButtons, userHub }) => {
   const userHubComponent = userHub || <HeaderAvatar />;
 
   const userMenuComponent = isActionSidebarOpen ? (
-    <ActionSidebar transactionId={transactionId || undefined}>
+    <ActionSidebar
+      transactionId={transactionId || undefined}
```
krawczykkatarzyna, line 40: Should we pass undefined here?

TomaszChmielPagepro: image
`transationId` is nullable so yes

KornelKwiatkowski (approved):

CzarekDryl (approved):

krawczykkatarzyna (approved):

PR #1232, opened 2023-10-31
Merge New UI w/ Master

willm30:
I will ensure this branch has the latest commits from `extensions-new-ui` before merging, after which point `master` will be the new branch from which everything should be based.

P.S. 

I created a copy of `extensions-new-ui` to test out merging before doing it on the real thing, but turns out it was simpler than I expected and this can be merged straight into master, after which point we can deprecate `extensions-new-ui` (instead of merging `extensions-new-ui` directly).


rdig (approved):
Good to go!

PR #1247, opened 2023-11-03
Add custom action titles

ArmandoGraterol:
Now when you create an action, you should be able to access a prop called `metadata` inside the `ColonyAction` data structure, and inside it, you can use `customTitle` to get the title the user is going to input in the future UI.

You can test this just by inserting the title you want the action to have in their respective payload, of course, there's still no place in the UI where this title is shown but you can check that the title was stored successfully by using the playground. For example, I've used the payment payload as a test for this:
FireShot Capture 199 - Amplify GraphiQL Explorer - localhost

The Safe Control action is also an example of how this could look when wired up. (Right now it seems like adding a safe is not working properly but that's something not related to this branch)


jakubcolony (approved):
Nice work, LGTM! 👍

arrenv (dismissed):
Wow Armando, this required touching way more files then expected.

arrenv (approved):

PR #1257, opened 2023-11-06
Hot fix for the inability to create colony

thomassdk:
## Description

This PR is a hot fix for the current inability to create a colony. 

It was breaking due to a rebase failing to completely remove the `AvatarUploaders` dependency on the user context.

Until the whole beta create colony flow is finished it is important to remember that you will need to run the `temp-create-colony` script in the scripts folder for the new beta flow to work as we are currently not creating a user before the colony and a user is now a hard requirement in the creation of a colony due to associating said user with beta invite codes


jakubcolony (approved):
Good to go 👍

PR #1258, opened 2023-11-06
Add analytics tracking context

arrenv:
## Description

This is mainly focused on adding some analytic tracking to the app. However, I have taken some liberties to fix some other issues that came up during testing this. Including the phospher icon library updates and updates to the the UserHub.

### Changes
- Adds a context for analytics tracking in the app to help with performance measurement and to help improve the user experience.
- Adds two event tracking calls for the action panel opening and UserHub opening, to test you will need to have an .env variable set and access to the GTM preview mode.
- Adds page view tracking with the `RouteTracker`.
- Adds the ability for Beamer to be triggered to open the embedded updates feed and bug reporting.
- Styling and copy fixes to the Userhub.
- Solves the issue with the overflow of submenu switching.
- Adds translation to menus that were missing them.
- Updates external URLs.
- Fixed height of Stakes tab in UserHub.

Please let me know if you have any questions.


src/context/AnalyticsContext.tsx
```diff
@@ -0,0 +1,80 @@
...
+// AnalyticsContextProvider component
+export const AnalyticsContextProvider: FC<PropsWithChildren> = ({
+  children,
+}) => {
+  const gtmId = process.env.GOOGLE_TAG_MANAGER_ID; // GTM ID from environment variable
+  useBeamer();
+
+  useEffect(() => {
+    // Inject GTM script only if gtmId is available
+    if (gtmId && (!window.dataLayer || window.dataLayer.length === 0)) {
```
rdig, line 40: `gtmId` should short-circuit early, rather than having repetitive checks all over the place

arrenv: Good pickup, I used to have that in, but forgot to add it back in after losing my updates.

src/components/common/Extensions/UserNavigation/partials/UserSubmenu/consts.ts
```diff
@@ -1,64 +1,129 @@
...
+      external: true,
     },
     {
       id: 'featureRequest',
-      label: 'Feature request',
-      url: '/feature-request',
+      label: menuMessages.whatsNew,
+      url: WHATS_NEW,
       icon: 'star',
+      onClick: useWhatsNew,
```
bassgeta, line 70: maybe we can not name it as a hook a la `useWhatsNew`, name it `openWhatsNew` and just move it to a helper? Since I see that just calls the Beamer instance, maybe a `utils/beamer` file wouldn't be a bad place for this

src/hooks/useBeamer.ts
```diff
@@ -0,0 +1,33 @@
...
+      if (typeof Beamer !== 'undefined') {
+        Beamer.init();
+      }
+    };
+
+    initializeBeamer();
+  }, []);
+};
+
+export const useWhatsNew = (
```
bassgeta, line 17: as mentioned in a different comment, maybe we can move this and `useFeaturesBugs` into a utils file since they are not hooks and we can call them anywhere

bassgeta (changes requested):

src/components/common/Extensions/UserNavigation/partials/UserSubmenu/consts.ts
arrenv: Adding here discussion from Discord. Basically, I want to contain this implementation as much as possible, so, I will rename but keep in the same location.

src/components/common/Extensions/UserHubButton/UserHubButton.tsx
```diff
@@ -68,6 +70,15 @@ const UserHubButton: FC<UserHubButtonProps> = ({
...
 
+  const handleButtonClick = () => {
+    trackEvent({
+      event: 'custom_event',
+      category: 'User',
+      action: 'Click',
+      label: 'Open UserHub',
+    });
+  };
+
```
rdig, line 81: This is not ok. Too many hardcoded values. Everything in here should be `enum` or otherwise a type. 

The only one that I can maybe accept as being a plain string is the `label`'s value.

Generally I'm not ok with patterns like this all strung about the codebase. Unless a super specific case calls for it, they should all be added to top-level / core components (buttons, forms, etc), not on the specific common components, as this way it's not maintainable.

rdig (changes requested):
Something in the styles changes to the `UserHub` broke the stakes tab, as it now renders as tall as the content with no overflow.

Screenshot from 2023-11-27 15-52-13
Screenshot from 2023-11-27 16-02-57

src/context/AnalyticsContext.tsx
```diff
@@ -0,0 +1,89 @@
...
+export const AnalyticsContext = createContext<AnalyticsContextValue>(
+  defaultAnalyticsContext,
+);
+
+// AnalyticsContextProvider component
+export const AnalyticsContextProvider: FC<PropsWithChildren> = ({
+  children,
+}) => {
+  const [isAnalyticsActive, setIsAnalyticsActive] = useState(false);
+  const gtmId = process.env.GOOGLE_TAG_MANAGER_ID; // GTM ID from environment variable
```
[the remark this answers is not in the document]
arrenv: With the gtmID .env set and while logged into Google Tag manager, you can trigger Preview mode with the URL as `http://localhost:9091/` to test it locally.

src/components/v5/common/Fields/Select/Select.tsx
```diff
@@ -80,12 +80,21 @@ const Select = <T extends any[]>({
     selectedItem &&
     list.filter(({ value }) => value !== selectedItem.value);
   const optionsList = isPlaceholder ? list : filteredList;
-  const placeholder =
-    typeof placeholderText === 'string'
-      ? placeholderText
-      : placeholderText && formatMessage(placeholderText);
 
-  const labelText = isPlaceholder ? placeholder : selectedItem?.label;
+  const formatLabel = (label) => {
```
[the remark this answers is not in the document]
arrenv: Do you have an example of what you searched for the other patterns? I have moved to a utils to be reused.

arrenv:
@rdig I have added max height to the stakes tab in this PR.

src/components/common/Extensions/UserHubButton/UserHubButton.tsx
arrenv: This has now been updated.

src/hooks/useBeamer.ts
```diff
@@ -0,0 +1,33 @@
+import { useEffect } from 'react';
+
+declare const Beamer: any;
+
+export const useBeamer = (): void => {
+  useEffect(() => {
+    const initializeBeamer = () => {
+      if (typeof Beamer !== 'undefined') {
```
bassgeta, line 8: I think this should run if beamer is undefined

bassgeta (changes requested):

src/components/v5/shared/Button/Button.tsx
```diff
@@ -28,6 +28,8 @@ const Button: FC<PropsWithChildren<ButtonProps>> = ({
   iconName,
   iconSize = 'tiny',
   isIconRight,
+  onTrackClick,
```
bassgeta, line 31: I think `onTrackClick` is not necessary, we can just track the event in the button's own `onClick` handler

src/hooks/useTracking.ts
```diff
@@ -0,0 +1,22 @@
+import {
+  useAnalyticsContext,
+  AnalyticsEventAction,
+  AnalyticsEventCategory,
+  AnalyticsEventLabel,
+  AnalyticsEventType,
+} from '~context/AnalyticsContext';
+
+const useTracking = (
```
bassgeta, line 9: hope it's not too nitpicky, but I don't think we need this wrapper. We pass it the same props as we do just to the `trackEvent` function. We can just always use `trackEvent` which works nicely

src/utils/intl.ts
```diff
@@ -104,3 +104,20 @@ export function formatText(
 
   return message;
 }
+
+// Handle labels that are either strings or message descriptors
+export const useFormattedText = () => {
```
bassgeta, line 109: I am not 100% sure of this one, but I think we could just use `formatText` here? cc @rdig, you know more of the inner workings of these helpers :)

bassgeta (changes requested):
I checked functionality wise, so Beamer works, it opens, just these 2/3 minor changes and I think we can :ship: it

src/components/v5/common/Fields/Select/Select.tsx
```diff
@@ -83,9 +82,9 @@ const Select = <T extends any[]>({
     list.filter(({ value }) => value !== selectedItem.value);
   const optionsList = isPlaceholder ? list : filteredList;
 
-  const placeholder = getFormattedText(placeholderText);
-  const selectedLabel = selectedItem
-    ? getFormattedText(selectedItem.label)
+  const placeholder = placeholderText ? formatText(placeholderText) : undefined;
```
rdig, line 85: Why doesn't `formatText` handle this case internally as well? 

You feed it some data, and if it's not what it expects, it just returns undefined. That way, you won't have these checks all over the place.

package-lock.json
chmanie, line 1: Whoever that was, please use `npm ci` exclusively as opposed to `npm install` (in the future)

src/components/common/Extensions/UserHub/UserHub.tsx
```diff
@@ -19,16 +19,12 @@ export const displayName = 'common.Extensions.UserHub.partials.UserHub';
...
 const UserHub: FC = () => {
   const isMobile = useMobile();
-  const [selectedTab, setSelectedTab] = useState(0);
+  const [selectedTab, setSelectedTab] = useState(UserHubTabs.Balance);
 
   const handleTabChange = (newTab: UserHubTabs) => {
     setSelectedTab(newTab);
   };
 
-  useEffect(() => {
```
chmanie, line 28: Very nice

src/components/common/Extensions/UserHub/partials/ReputationTab/partials/Balance.tsx
```diff
@@ -60,7 +60,7 @@ const Balance: FC<BalanceProps> = ({ nativeToken, wallet, onTabChange }) => {
 
   return (
     <div>
-      <TitleLabel text={formatMessage({ id: 'balance.in.colony' })} />
+      <TitleLabel text={formatMessage({ id: 'userHub.nativeToken' })} />
```
chmanie, line 63: Thank you for this. In the future, I'd be happy if we moved this kind of stuff back into the individual components using `defineMessages`

src/components/common/Extensions/UserHub/partials/StakesTab/partials/StakesList.css
```diff
@@ -0,0 +1,38 @@
+@value query700 from "~styles/queries.css";
+
+.stakesListContainer {
+  padding-right: 5px;
```
chmanie, line 4: All of these can surely be done with tailwind, don't you think?

src/components/common/Extensions/UserHub/partials/StakesTab/partials/StakesList.css
```diff
@@ -0,0 +1,38 @@
...
+  height: 60vh;
+  max-height: 356px;
+  width: 100%;
+  overflow-x: hidden;
+  overflow-y: scroll;
+}
+
+/* The scrollbar itself */
+.stakesListContainer::-webkit-scrollbar {
```
chmanie, line 13: There's also https://www.npmjs.com/package/tailwind-scrollbar. Maybe make a note for later.

src/components/common/Extensions/UserHub/partials/StakesTab/partials/StakesList.tsx
```diff
@@ -29,7 +30,7 @@ const StakesList = ({ stakes, loading, colony }: StakesListProps) => {
   }
 
   return (
-    <>
+    <div className={styles.stakesListContainer}>
```
chmanie, line 33: I would suggest to apply all classes that are actually supported by tailwind.

src/components/common/Extensions/UserHubButton/UserHubButton.tsx
```diff
@@ -56,6 +72,11 @@ const UserHubButton: FC<UserHubButtonProps> = ({
     );
 
   useDisableBodyScroll(visible && isMobile);
+  const handleButtonClick = () => {
```
chmanie, line 75: Is this the pattern we are using moving forward? I was hoping we could maybe be a bit smarter with the tracking. E.g. you could create a function, that wraps a function.

Here's an example:


```tsx
// This is also a bit annoying, maybe we can come up with a better, shorter abstraction?
// At least the advantage here is that this doesn't necessarily have to live inside the component (or this file, even)
const trackUserButtonClick = [
    AnalyticsEventType.CUSTOM_EVENT,
    AnalyticsEventCategory.USER,
    AnalyticsEventAction.CLICK,
    AnalyticsEventLabel.OPEN_USER_HUB,
];

// ...

const { track } = useAnalyticsContext();
// This will look much cleaner with only one function being passed in
return (
<div
  onClick={track(trackUserButtonClick, () => {
    setOpenItemIndex(undefined);
    toggleOff();
  })}
/>
)
```

src/components/common/Extensions/UserNavigation/partials/UserSubmenu/types.ts
```diff
@@ -8,7 +10,10 @@ export interface UserSubmenuItems {
 
 export interface UserSubmenuItem {
   id: string;
-  label: string;
+  label: MessageDescriptor;
```
chmanie, line 13: We discussed earlier to move the formatting outside of the components. Maybe you can do that here as well (sorry, this essentially reverts this).

src/components/v5/common/Modals/ManageMemberModal/consts.ts
```diff
@@ -1,32 +1,65 @@
...
+  editPermissions: {
+    id: 'ManageMembers.Actions.EditPermissions',
+    defaultMessage: 'Edit this member’s permissions',
+  },
+});
+
 export const manageMemberActions: SelectOption[] = [
   {
-    label: 'Ban this member from chat access',
+    label: formatText(actionMessages.banMember) || '',
```
chmanie, line 38: If we're merging #1455 before this one, we have to remember to remove the fallbacks.

chmanie (changes requested):
Quite some good work in here, I learned a lot! I left some comments :)

src/components/common/Extensions/UserHubButton/UserHubButton.tsx
bassgeta: I honestly don't see a benefit in making a funcition call a callback after an event was tracked, it just makes it seem more complicated than it actually is.
I however agree on reworking the `trackEvent` API, I'll try to make it receive an object, and then we can reuse objects across the app if ever necessary

src/components/common/Extensions/UserHub/partials/StakesTab/partials/StakesList.tsx
bassgeta: would something like:
` <div className={clsx(styles.stakesListContainer, 'pr-2 .....')}>` suffice? I would love to rework the scrollbar while I'm at it, but this PR is a bit old, so I'm getting merge conflicts every time anybody touches any file tracking was added to :sweat_smile: 
Can we open an issue to tackle that, I can do it whenever, since the package you've linked looks like it should do the job

src/components/common/Extensions/UserHub/partials/ReputationTab/partials/Balance.tsx
```diff
@@ -19,6 +19,13 @@ import styles from '../ReputationTab.module.css';
 const displayName =
   'common.Extensions.UserHub.partials.ReputationTab.partials.Balance';
 
+const MSG = defineMessages({
```
chmanie, line 22: :)

src/components/common/Extensions/UserHub/partials/StakesTab/partials/StakesList.tsx
```diff
@@ -29,7 +31,12 @@ const StakesList = ({ stakes, loading, colony }: StakesListProps) => {
   }
 
   return (
-    <>
+    <div
```
chmanie, line 34: Great!

src/components/common/Extensions/UserHubButton/consts.ts
```diff
@@ -0,0 +1,14 @@
+import {
+  AnalyticsEvent,
+  AnalyticsEventAction,
+  AnalyticsEventCategory,
+  AnalyticsEventLabel,
+  AnalyticsEventType,
+} from '~context/AnalyticsContext';
+
+export const OPEN_USER_HUB_EVENT: AnalyticsEvent = {
```
chmanie, line 9: :+1:

chmanie (approved):
Great! Everything was addressed, I'm happy. There's a linter issue that wants to be fixed :)

melyndav (dismissed):
Approving this one. Great work @bassgeta and @arrenv on the updates.

src/components/common/Extensions/UserNavigation/partials/UserSubmenu/UserSubmenu.tsx
```diff
@@ -10,21 +11,59 @@ const displayName = 'common.Extensions.UserNavigation.partials.UserSubmenu';
...
+              rel="noopener noreferrer"
+              onClick={item.onClick}
+            >
+              <span className="flex items-center shrink-0 mr-2 sm:mr-0 flex-grow">
+                <Icon name={item.icon} appearance={{ size: iconSize }} />
+                <p className="ml-2 ">{item.label}</p>
+              </span>
+            </a>
+          ) : (
+            <NavLink
```
chmanie, line 53: This doesn't need to be a NavLink if we're not really using it, do we? Or is this avout the `active` class?

src/components/v5/frame/NavigationSidebar/partials/NavigationFeedbackWidget/NavigationFeedbackWidget.tsx
```diff
@@ -31,6 +32,9 @@ const NavigationFeedbackWidget = () => (
     <button
       type="button"
       className="text-xs font-medium underline hover:text-gray-900"
+      onClick={(e) => {
```
chmanie, line 35: You should be able to just pass it in `onClick={openFeaturesBugs}`

src/components/common/Extensions/UserNavigation/partials/UserSubmenu/UserSubmenu.tsx
arrenv: This is likely going to be needed in future.

chmanie: What do you mean by "this"? Can you expand please?

bassgeta: probably gonna link somewhere inside the app at some point? dunno, we can keep it like this and realign when we get there. I'd like to have just a "button" component here too, but this PR is getting old and it's constantly getting new conflicts :')

chmanie: Yeah, well I really don't like this kind of stuff, it gets out of hand quickly. But if you create an issue for it we can leave it like this.
A button is a button and a link is a link :)

melyndav (approved):
Nice one, working well!

chmanie (approved):
Ok it's time to merge this bad boi

PR #1267, opened 2023-11-07
Factor out the ColonyContext again

chmanie:
## Description

In short: `useColonyContext` can only be used in a route that is a child of `ColonyRoute`.

I made a few changes to existing components, but tried to keep the general style. The main change is that `SharedRoute` is now `MainRoute` again and `ColonyRoute` has the `ColonyContext`.

This introduces a little bit of code duplication but that can be improved further down the line. It is late.

There will be a follow up to this as we need a more flexible sidebar component (e.g. for the create user and create colony wizards). I suggest just passing in a component instead of all of these props.

@joanna-pagepro, please give this a review and let me know if you're ok with it.

### DIFF
--- a/src/components/frame/Extensions/layouts/ColonyLayout.tsx
+++ b/src/components/frame/Extensions/layouts/ColonyLayout.tsx
@@ -1,13 +1,29 @@
 import React, { FC, PropsWithChildren } from 'react';
+import { Navigate } from 'react-router-dom';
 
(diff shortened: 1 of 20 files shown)


joanna-pagepro (approved):
Looks good!
`PageLayout` component can be changed in the next iterations. I didn't know about all of the versions of the sidebar and that's why it's done this way, but feel free to modify it :)

PR #1282, opened 2023-11-08
Various colony creation fixes

jakubcolony:
## Description

This PR fixes the following issues with colony creation:
- Setting the newly created token decimals to 0, which was causing the payment action to fail.
- Caching the results of query checking if colony name is already in use, which resulted in 404 after colony create redirection.
- Being unable to activate tokens due to token authority not being set correctly. This had to be fixed on the contracts side so the network image hash was updated.

It also introduces a simpler query to check if colony name is in use as well as refactors `ColonyContext` to remove some of the redundant code.


src/components/v5/common/CreateColonyWizard/validation.ts
```diff
@@ -19,7 +19,7 @@ const COLONY_NAME_REGEX = /^[A-Za-z0-9]\w+$/;
 const TOKEN_SYMBOL_REGEX = /^[A-Za-z0-9]+$/;
 
 const isNameTaken = createYupTestFromQuery({
-  query: GetFullColonyByNameDocument,
+  query: CheckColonyNameExistsDocument,
```
jakubcolony, line 22: Little point in getting the massive colony object just to see if it exists

arrenv (approved):
Nice work!! Fixes it for me!

src/context/ColonyContext/ColonyContext.tsx
```diff
@@ -51,31 +46,13 @@ const MSG = defineMessages({
 });
 
 const MIN_SUPPORTED_COLONY_VERSION = 5;
-export const PREV_COLONY_LOCAL_STORAGE_KEY = 'prevColonyName';
```
chmanie, line 54: Very good, removing this, I approve!

src/context/ColonyContext/usePreviousColony.ts
chmanie, line 1: :+1:

chmanie (approved):
Really nice work and good clean-up @jakubcolony! I approve!

PR #1290, opened 2023-11-09
Fix: Missing OneTx permissions in colonyCreate saga

jakubcolony:
## Description

When creating a colony OneTxPayment wasn't being given the Arbitration permissions

### DIFF
--- a/src/redux/sagas/colony/colonyCreate.ts
+++ b/src/redux/sagas/colony/colonyCreate.ts
@@ -40,7 +40,11 @@ import {
   GetTokenFromEverywhereQueryVariables,
 } from '~gql';
 import { ColonyManager, ContextModule, getContext } from '~context';
-import { ADDRESS_ZERO, DEFAULT_TOKEN_DECIMALS } from '~constants';
+import {
+  ADDRESS_ZERO,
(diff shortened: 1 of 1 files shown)


ArmandoGraterol (approved):
LGTM 👍

src/redux/sagas/colony/colonyCreate.ts
```diff
@@ -482,13 +486,16 @@ function* colonyCreate({
...
+        (config) => config.extensionId === Extension.OneTxPayment,
+      );
       yield put(
         transactionAddParams(setOneTxRoles.id, [
           permissionDomainId,
           childSkillIndex,
           oneTxPaymentExtension.address,
           Id.RootDomain,
-          colonyRoles2Hex([ColonyRole.Administration, ColonyRole.Funding]),
+          colonyRoles2Hex(extensionConfig?.neededColonyPermissions ?? []),
```
rdig, line 498: I like this, however this makes `neededColonyPermissions` to only have a certain format always, otherwise this saga breaks. So we'll have to ensure that it's format never changes.

rdig (approved):
:+1:

PR #1291, opened 2023-11-09
Run amplify folder repermissioning at container level

chmanie:
## Description

So here's an unexpected PR. I suddenly experienced some weird permission problems, I almost went crazy until I started some [helpful conversations](https://chat.openai.com/share/db30a4d9-e1ca-4f91-8c1c-c403528dec44) with ChatGPT. It was a little bit of a trial and error thing until _we_ realized that the permissions for mounted volumes are always mounted into the container using the `root` user. This can't be changed at the image build level but has to be done during the start up time of the container.

So what happens here is I basically moved the `chown` command and the privilege drop to the run script. 

~~I'm not exactly sure but this _should_ remove the need for the infamous~~
```bash
sudo chown -R $(whoami):$(ls -l package.json | awk '{print $4}') ./ && sudo chmod  -R g+rws  ./amplify
```
~~command.~~

Update: it does not.

### DIFF
--- a/docker/colony-cdapp-dev-env-amplify
+++ b/docker/colony-cdapp-dev-env-amplify
@@ -32,14 +32,12 @@ ADD docker/files/amplify/amplify-meta.json.base /colonyCDappBackend/scripts/ampl
 ADD docker/files/amplify/config.base /home/node/.aws/config
 ADD docker/files/amplify/credentials.base /home/node/.aws/credentials
 
-RUN chown -R node:users /colonyCDapp
-RUN chown -R node:users /colonyCDappBackend
-
 # Opensearch cannot be run by root so we use the built in node user
 # We also add the node user to the users group so that the node user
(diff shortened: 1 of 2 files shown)


rdig (approved):
Seems to be working for me, however I'd like someone who's on a Mac to also test this. Maybe even @arrenv since he's on Windows via WSL

arrenv (approved):
Running fine for me.

Although, @chmanie, with your update, are you saying this doesn't actually solve the permission issue for you?

chmanie:
Have you tried switching branches? That's somehow where it breaks for me. So this is definitely not a great solution :(

arrenv:
> Have you tried switching branches? That's somehow where it breaks for me. So this is definitely not a great solution :(

I have been switching branches without any issues.

chmanie:
Closed in favor of #1316

PR #1292, opened 2023-11-09
Fix user profile creation redirect

ArmandoGraterol:
This PR is quite straightforward! I just moved the contents of the user profile edit form to its own component in order to use `reset`. What we should test for in this PR is that when you create a user, you should be redirected to your user profile page successfully and the form inside it should show the corresponding data.

Also, you shouldn't be able to get to or remain on this page if you are not registered or your wallet is not connected.

### DIFF
--- a/src/components/frame/v5/pages/UserProfilePage/partials/UserProfilePageForm.tsx
+++ b/src/components/frame/v5/pages/UserProfilePage/partials/UserProfilePageForm.tsx
@@ -1,0 +1,258 @@
+import React, { useEffect } from 'react';
+import clsx from 'clsx';
+import { useFormContext } from 'react-hook-form';
+import axios from 'axios';
+import { toast } from 'react-toastify';
+
+import Toast from '~shared/Extensions/Toast';
+import Input from '~v5/common/Fields/Input';
+import Textarea from '~v5/common/Fields/Textarea';
+import Button from '~v5/shared/Button';
+import AvatarUploader from '~v5/common/AvatarUploader';
+import Avatar from '~v5/shared/Avatar';
+import { formatText } from '~utils/intl';
+import { UseAvatarUploaderProps } from '~v5/common/AvatarUploader/hooks';
+import { useAppContext, useMobile } from '~hooks';
+import { UserFragment, useUpdateUserProfileMutation } from '~gql';
+
+import styles from '../UserProfilePage.module.css';
+import { MAX_BIO_CHARS, MAX_DISPLAYNAME_CHARS } from '../consts';
+
+import LeftColumn from './LeftColumn';
+
+const displayName = 'v5.pages.UserProfilePage';
+
+interface Props {
+  user: UserFragment | null | undefined;
+  avatarUrl: string | null | undefined;
+  canChangeUsername: boolean;
+  daysTillUsernameChange: number;
+}
+
+const UserProfilePageForm = ({
+  user,
+  avatarUrl,
+  canChangeUsername,
+  daysTillUsernameChange,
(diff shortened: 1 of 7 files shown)


src/components/frame/v5/pages/UserProfilePage/partials/UserProfilePageForm.tsx
```diff
@@ -0,0 +1,258 @@
...
+    });
+  }, [user, reset]);
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const axiosUpload: UseAvatarUploaderProps['updateFn'] = async (
+    _,
+    thumbnail,
+    setProgress,
+  ) => {
+    axios.post(
```
rdig, line 67: I know you didn't put it in here, but why is axios being used here? Why not strait up apollo ?

src/hooks/useCanEditProfile.ts
```diff
@@ -1,24 +1,36 @@
...
-  const { user, userLoading, walletConnecting } = useAppContext();
+  const { user, userLoading, walletConnecting, updateUser, wallet } =
+    useAppContext();
 
   const navigate = useNavigate();
   // If there's a last wallet label, it means we're logged in.
   // If null, we're logged out and should be redirected to landing page.
   const lastWalletLabel = getLastWallet();
+  const { state: locationState } = useLocation();
+  const isRedirect = locationState?.isRedirect;
```
rdig, line 17: The redirect is not working properly, as it always redirects (when going directly to the url), even if we have a wallet logged in.

I suspect it's because of the wallet not being available at the first render of the page

src/components/frame/v5/pages/UserProfilePage/partials/UserProfilePageForm.tsx
ArmandoGraterol: No idea tbh, I wondered the same when I saw this

chmanie (dismissed):
I have nothing to add here. Works flawlessly.

src/components/frame/v5/pages/UserProfilePage/partials/UserProfilePageForm.tsx
thomassdk: this was something that page pro added that was broken when I refactored the `AvatarUploader` to be independent from the the user profile avatar uploading. I assumed it was something that had been agreed upon with us that this would be how uploading of files would eventually work. Feel free to remove it if this was not the case

src/components/v5/common/CreateUserForm/CreateUserForm.tsx
```diff
@@ -12,6 +12,7 @@ import { HeaderRow } from '../CreateColonyWizard/shared';
 import { validationSchema } from './validation';
 import { CreateUserFormValues } from './types';
 import CreateUserFormInputs from './CreateUserFormInputs';
+import { useAppContext } from '~hooks';
```
thomassdk, line 15: `Up` this needs to go

thomassdk (approved):
Great work, just a minor comment about imports and if nobody knows about why axios is here I think we should just remove it

chmanie (approved):

src/components/frame/v5/pages/UserProfilePage/partials/UserProfilePageForm.tsx
rdig: I don't remember this being agreed with us. And frankly I don't really like it. We need to find out more about this.

rdig (approved):
All good now. :+1:

PR #1300, opened 2023-11-10
Feat/15645086 join colony button

CzarekDryl:
https://tw.auroracreation.com/app/tasks/15645086

Added join colony button with animation.


src/components/common/Extensions/UserNavigation/UserNavigation.tsx
```diff
@@ -17,10 +18,10 @@ const UserNavigation: FC<UserNavigationProps> = ({
   txButtons = null,
 }) => {
   const { wallet, user, connectWallet } = useAppContext();
-  const { formatMessage } = useIntl();
   const isTablet = useTablet();
   const { setOpenItemIndex, mobileMenuToggle } = useNavigationSidebarContext();
   const [, { toggleOff }] = mobileMenuToggle;
+  const isOnColonyRoute = window.location.pathname.includes('/colony/');
```
TomaszChmielPagepro, line 24: Use `useMatch` instead

rdig (changes requested):
- Icon is misaligned
Screenshot from 2023-11-13 13-36-50

- Join functionality is not triggering correctly, as it needs to fire off the mutation at the start, and run in parallel to the animation, not wait until the animation finishes and then fire it off

- Leaving a colony (unjoin) will bring back the button with the wrong state: text "Joined" (vs. "Join") and the icon being visible

krawczykkatarzyna (approved):

src/components/common/Extensions/UserNavigation/UserNavigation.tsx
CzarekDryl: `useMatch` used

CzarekDryl:
@rdig All issue mentioned by you are fixed, please check it once more :)

KornelKwiatkowski (approved):

rdig (changes requested):
It's now working (more or less) as expected but has a bit of weird behavior.

- Join a colony
- Leave a colony
- Join a colony again -- triggers an Apollo error

ArmandoGraterol (dismissed):
LGTM 👍

CzarekDryl:
@rdig  it was the same in the old UI, it's only in the case of a new user. I think it's better for someone from your team to check it.

image

rdig:
>  I think it's better for someone from your team to check it.
?

 I did try it with new users, 3 of them actually.

TomaszChmielPagepro:
@rdig Didn't just @CzarekDryl say that it only occurs for new users? 😅
This seems to be a problem with the old UI as well and the error suggests that it might be something wrong on the backend side

melyndav (changes requested):
Hey @CzarekDryl thanks for your work on this!

A minor update:

1. The 16px padding is missing for both the wallet connected and not connected states:
image

melyndav (approved):
Great work @CzarekDryl, works as expected.

arrenv (approved):
Looks good, thank you @CzarekDryl

PR #1302, opened 2023-11-13
Storybook fix, code clean up

joanna-pagepro:
- Updated Storybook webpack config
- Removed redundant story that required Colony Context, etc.
- Removed redundant dependencies from dependency array and unused import of react intl hook from Select component
- Removed alerts from stories


krawczykkatarzyna (dismissed):

CzarekDryl (dismissed):

.storybook/preview.tsx
```diff
@@ -1,13 +1,12 @@
 import React from 'react';
-import type { Preview } from '@storybook/react';
 import { MockedProvider } from '@apollo/client/testing';
 
 import '~utils/yup/customMethods';
 import '../src/styles/main.global.css';
 import { reactIntl } from './reactIntl.js';
 import { applyTheme } from "../src/components/frame/Extensions/themes/utils";
 
-const preview: Preview = {
```
TomaszChmielPagepro, line 10: Why tho? This is actually helpful as it verifies if all options have the correct types/shape

joanna-pagepro: You're right, added back

CzarekDryl (dismissed):

krawczykkatarzyna (approved):

TomaszChmielPagepro (approved):

KornelKwiatkowski (approved):

ArmandoGraterol (commented):
Is it expected that the Meat Ball menu story crashes?

FireShot Capture 213 - Shared _ Meat Ball Menu - Base ⋅ Storybook - localhost

joanna-pagepro:
@ArmandoGraterol Meat Ball menu story should be ok now. Can you review it again?

ArmandoGraterol (commented):
Thank you for the changes! Just noticed that `Empty Menu` is crashing with a `There are no items added to the menu.`
FireShot Capture 217 - Shared _ Meat Ball Menu - Empty Menu ⋅ Storybook - localhost

If the component is supposed to always have an item, at least according to the error message, we should remove this "sub-story". Otherwise, we need to fix this.

The other stories are working as expected.

ArmandoGraterol (approved):
Nice work cleaning everything up! Looks good to me 👍

TomaszChmielPagepro (approved):

PR #1309, opened 2023-11-13
Show Payment actions incoming funds

### LINKED ISSUE #991: Store new `ColonyFundsClaim` when an expenditure payout is claimed by a colony
If an expenditure payout recipient is a colony, the claimed funds will not automatically appear in its balance. Due to how contracts work, they have to be explicitly claimed with `claimColonyFunds` function. We provide a way of doing this in the UI already, for things such as transferring or minting tokens:
image

It relies on the "pending" claims being stored in the DB which is not yet the case for expenditure payouts. 

This issue is to add logic to the `ExpenditurePayoutClaimed` event handler that will store the pending payout claim in the database, so that it's claimable from the UI.

Marking as blocked by #982

jakubcolony:
## Description

Please see https://github.com/JoinColony/block-ingestor/pull/144 for details. In CDapp this PR simply updates the ingestor hash.

Resolves #991

### DIFF
--- a/docker/colony-cdapp-dev-env-block-ingestor
+++ b/docker/colony-cdapp-dev-env-block-ingestor
@@ -1,6 +1,6 @@
 FROM colony-cdapp-dev-env/base:latest
 
-ENV BLOCK_INGESTOR_HASH=36c2b5f63d32e8e690cd74049e0cb7067eb36bc4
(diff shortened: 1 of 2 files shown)


ArmandoGraterol (approved):
LGTM 👍

thomassdk (approved):
Also approved from me

PR #1316, opened 2023-11-13
Improve docker compatibility

chmanie:
## Description

This PR improves the compatibility of our docker setup with newer docker versions and tools like podman which in turn enables us to run our setup as non-root users in Linux.

Most importantly the `links` entries were removed from the docker-compose files which is an option that's [long deprecated](https://docs.docker.com/network/links/) and not necessary anymore. 

-----

To use with `podman` if you like (rootless, on Linux - **DO NOT DO THIS IF YOU'RE ON A MAC OR WINDOWS**):

- Remove docker from your system :scream: 
- We need `podman` in version 4.7.x to support the `docker compose` command
- Use either `docker-compose` (standalone, without docker) or `podman-compose` (untested)
- Make sure you have the podman [`dnsname`](https://github.com/containers/dnsname/blob/main/README_PODMAN.md) plugin installed and configured
- Make sure you have `docker` aliased to `podman` ([`podman-docker`](https://packages.debian.org/sid/podman-docker) might help)
- Before running `npm run dev`, make sure to `export DOCKER_HOST=unix:///run/user/1000/podman/podman.sock` (this will be fixed in a future version of podman - https://github.com/containers/podman/pull/19917#issuecomment-1802805687)

### DIFF
--- a/amplify/backend/function/fetchColonyBalances/src/index.js
+++ b/amplify/backend/function/fetchColonyBalances/src/index.js
@@ -12,7 +12,7 @@ Logger.setLogLevel(Logger.levels.ERROR);
 
 let apiKey = 'da2-fakeApiId123456';
 let graphqlURL = 'http://localhost:20002/graphql';
-let rpcURL = 'http://network-contracts.docker:8545'; // this needs to be extended to all supported networks
+let rpcURL = 'http://network-contracts:8545'; // this needs to be extended to all supported networks
 let network = Network.Custom;
 let networkAddress;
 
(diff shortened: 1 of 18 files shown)


thomassdk (dismissed):
Works for me, nice work :rocket:

ArmandoGraterol (approved):
As the President of the "Colony Devs with Mac" association, I approve this PR

arrenv (approved):
Local dev environment is working for me.

thomassdk (approved):

PR #1318, opened 2023-11-13
Create Action new main menu item

### LINKED ISSUE #1271: Quick action navigation item
## Description 

A new navigation menu item should be added to the new navigational bar for the CDapp to allow users to create a quick action.

**Specs:**

- The menu item should be placed above the dashboard menu item and below the Colony switcher. 
- The item will have a light-base-bg colour and a text-gray-900 icon colour.
- The item will have the same hover state as the other menu items.
- When triggered, a default action side panel will be triggered. There is no menu draw for this navigation item. 
- The nav item should not show in the out of colony context since a user is not within a Colony to create an action.
- The nav item should appear during beta in the beta navigation version.

**Figma links**
Component: https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-V5?type=design&node-id=11290%3A413385&mode=dev&t=XOXSKM9TdW35BwEq-1

arrenv:
## Description

This PR adds a new menu item that provides a more convenient way to trigger the Action side panel.

- This required updating the type and also some props in the navigation components.

- There is also a small fix to the Colony's network icon size.

- I have now also updated all icons in the main nav to be more consistent with the design, and the sizes. They are slightly smaller, but the alternative is to introduce a new icon size in the Icon component. This has been confirmed by design.

- Vertical padding has also been increased slightly with confirmation from design.

### Design
[Figma link](https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-V5?type=design&node-id=11290%3A413385&mode=dev&t=XOXSKM9TdW35BwEq-1)

image

image

Resolves #1271


src/components/v5/frame/NavigationSidebar/partials/NavigationSidebarMainMenu/types.ts
```diff
@@ -4,11 +4,14 @@ import { NavigationSidebarThirdLevelProps } from '../NavigationSidebarThirdLevel
 
 export interface NavigationSidebarItem extends NavigationSidebarButtonProps {
   key: string;
-  secondLevelMenuProps: Omit<
+  secondLevelMenuProps?: Omit<
     NavigationSidebarSecondLevelProps,
     'isExpanded' | 'onArrowClick'
   >;
   relatedActionsProps?: NavigationSidebarThirdLevelProps;
+  onClick?: () => void;
```
joanna-pagepro, line 12: onClick and className should already exist in `NavigationSidebarButtonProps` interface, so you don't have to add it here

src/components/v5/frame/NavigationSidebar/partials/NavigationSidebarButton/types.ts
```diff
@@ -1,10 +1,13 @@
...
+import { NavigationSidebarSecondLevelProps } from '../NavigationSidebarSecondLevel/types';
 
 export interface NavigationSidebarButtonProps
   extends ButtonHTMLAttributes<HTMLButtonElement> {
   iconName: string;
   label: string;
   isActive?: boolean;
   isExpanded?: boolean;
+  hideMobile?: boolean;
+  secondLevelMenuProps?: NavigationSidebarSecondLevelProps;
```
joanna-pagepro, line 11: This prop is redundant here, because you're using it as a boolean and you are passing the whole second level menu object. You should simplify that and just change it to boolean that can be named `withCaretIcon`. By default you can set it to true, so that every button will have it by default. Then inside the component that is using this one, just set it to false when  `secondLevelMenuProps` is undefined.

src/components/v5/frame/NavigationSidebar/partials/NavigationSidebarButton/NavigationSidebarButton.tsx
```diff
@@ -13,17 +13,23 @@ const NavigationSidebarButton: FC<NavigationSidebarButtonProps> = ({
...
+  secondLevelMenuProps,
+  hideMobile,
   isActive,
   isExpanded,
   ...rest
 }) => {
   const isTablet = useTablet();
 
+  // Don't show item if hideMobile is true and we're on mobile
+  if (isTablet && hideMobile) {
```
joanna-pagepro, line 25: You should hide the whole item (li), and not just the button. I think that this prop should be moved to a different component

joanna-pagepro (changes requested):

rdig (dismissed):
Fix the render logic on mobile (to be outside of the component itself), after which feel free to merge.

melyndav (approved):
Nice work @arrenv. Approved.

rdig (dismissed):

ArmandoGraterol (dismissed):
LGTM 👍

melyndav (approved):
Looks great @arrenv. Approved.

melyndav (approved):

jakubcolony (approved):
LGTM 👍

PR #1336, opened 2023-11-15
`useActivityFeed` hook

jakubcolony:
## Description

Hook handling fetching, filtering and paginating actions shown in the activity list.
image

For demo purposes, the old UI was switched over to use the new hook and provides filters for decision method and status. Other filters can be added by following the existing ones.

## Testing

For testing, create different actions, motions and play around with filters, sorting, and pagination.


src/hooks/useActivityFeed/useActivityFeed.ts
```diff
@@ -0,0 +1,125 @@
...
+  );
+
+  if (!currentPageActions.length && pageNumber > 1) {
+    setPageNumber((number) => number - 1);
+  }
+
+  return {
+    loading: loading || fetchMoreActions,
+    actions: currentPageActions,
+    sortDirection,
```
TomaszChmielPagepro, line 116: I know this is a draft but will we be able to change which property are we sorting by?
Seems like it should be possible to sort by date and status

Story: 
image

Designs:
image

jakubcolony: Hi Tomasz, apologies I missed your comment. Yes, a consuming component will be able to provide which field actions should be sorted by.

src/hooks/useActivityFeed/useActivityFeed.ts
```diff
@@ -0,0 +1,147 @@
...
+  getActionsByPageNumber,
+  getSearchActionsFilterVariable,
+} from './helpers';
+import {
+  ActivityFeedFilters,
+  SortDirectionChangeHandler,
+  UseActivityFeedReturn,
+} from './types';
+
+const ITEMS_PER_PAGE = 2;
```
jakubcolony, line 23: temporary value, makes it easier to test without having tens of actions

src/components/common/ColonyActions/ColonyActions.tsx
```diff
@@ -1,14 +1,19 @@
...
+import Button from '~shared/Button';
 
 import { ActionsListHeading } from '.';
 
 import styles from './ColonyActions.css';
+import {
+  ActivityDecisionMethod,
+  ActivityFeedFilters,
+} from '~hooks/useActivityFeed/types';
+import { MotionState } from '~utils/colonyMotions';
```
thomassdk, line 16: the classic import order mixup

src/utils/colonyMotions.ts
```diff
@@ -13,17 +13,18 @@ export enum MotionVote {
...
 export enum MotionState {
-  Staked = 'Staked',
+  Supported = 'Supported',
   Staking = 'Staking',
   Voting = 'Voting',
   Reveal = 'Reveal',
-  Objection = 'Objection',
-  Motion = 'Motion',
+  Opposed = 'Opposed',
+  Motion = 'Motion', // @TODO: This was used by the old MotionTag component and should be removed
```
thomassdk, line 21: Can we not remove this as part of this PR, we are in a strange state in the CDapp where we have left in obsolete parts that are then used by others for the v5, my preference would just be to remove this entirely to avoid this problem in the future

thomassdk (approved):
Works well and the code looks good to me, good work. 

One thing that is unclear to me is if this hook should provide the data for the activity feed widgets. https://github.com/JoinColony/colonyCDapp/pull/1358

It seems to me like it should but would be interested to hear your thoughts

jakubcolony:
@thomassdk This hook will only provide the data for activity feed, the widget you posted I'm not sure of, but it seems like it should be able to just call `searchActions` independently to get the aggregates it needs.

amplify/backend/api/colonycdapp/resolvers/Query.searchColonyActions.req.vtl
```diff
@@ -0,0 +1,105 @@
+#set( $args = $util.defaultIfNull($ctx.stash.transformedArgs, $ctx.args) )
```
rdig, line 1: What's the change in this file? As opposed to the original version of this resolver

rdig (approved):
Code-wise I'm good with this. Functionally I've tested this as part of #1384 

Nice work! :100:

amplify/backend/api/colonycdapp/resolvers/Query.searchColonyActions.req.vtl
jakubcolony: Line 96, the `$args.nextToken` bit wasn't wrapped in double quotes which caused the query to fail

arrenv (changes requested):
Nice work @jakubcolony. Everything looks and works as expected. Just a questions around the status, these should be all the states:

- Staking
- Supported
- Opposed (missing)
- Voting (missing)
- Reveal (missing)
- Finalizable (missing)
- Passed
- Failed
- Escalated (missing)

image

arrenv (approved):
Discussed with @jakubcolony, UI in this example is only for demo purposes.

This is good to go!

PR #1345, opened 2023-11-16
Fix Message Signing, User Transaction Persistance

rdig:
This PR is two-parter:
- First, fix UserHub transaction persistence
- Second, prevent messages from showing up in the UserHub transactions list

**UserHub Transaction Persistence**

This was implemented using a wrong spec, as it tried to only load transactions based on the currently selected colony, which should not be case.

I refactored this to fetch all transactions from the database that were initiated by this user.

**Prevent messages from showing up in UserHub**

Messages should not show up (at least for now) in the transactions list. So I refactored the hook to use the `transactionsGroup` selector vs. the original `transactionsAndMessagesGroup`. 

This was both a UI issue, as the messages that showed up in the list were unstyled, and both a logic one, as the mutation would try to save the message to the database, where the format is unsupported

### DIFF
--- a/src/components/common/Extensions/UserNavigation/hooks.ts
+++ b/src/components/common/Extensions/UserNavigation/hooks.ts
@@ -148,7 +148,6 @@ export const useGroupedTransactionsAndMessages = (): {
 
   const { data, fetchMore, refetch } = useGetUserTransactionsQuery({
     variables: {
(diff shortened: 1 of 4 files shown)


bassgeta (approved):

arrenv (approved):
Working great, nice one Raul!

PR #1346, opened 2023-11-17
fix: remove token dropdown from value field in mint token form

joanna-pagepro:
https://tw.auroracreation.com/app/tasks/15622227

image

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/forms/MintTokenForm/MintTokenForm.tsx
+++ b/src/components/v5/common/ActionSidebar/partials/forms/MintTokenForm/MintTokenForm.tsx
@@ -1,21 +1,22 @@
 import React, { FC } from 'react';
 
+import { useColonyContext } from '~hooks';
+import { formatText } from '~utils/intl';
 import ActionFormRow from '~v5/common/ActionFormRow';
 import AmountField from '~v5/common/ActionSidebar/partials/AmountField';
 import TeamsSelect from '~v5/common/ActionSidebar/partials/TeamsSelect';
 import { FormCardSelect } from '~v5/common/Fields/CardSelect';
 
-import { formatText } from '~utils/intl';
-
-import { useMintToken } from './hooks';
 import { ActionFormBaseProps } from '../../../types';
 import { useDecisionMethods } from '../../../hooks';
 import DescriptionRow from '../../DescriptionRow';
+import { useMintToken } from './hooks';
 
 const displayName = 'v5.common.ActionSidebar.partials.MintTokenForm';
 
 const MintTokenForm: FC<ActionFormBaseProps> = ({ getFormOptions }) => {
   const { decisionMethods } = useDecisionMethods();
+  const { colony } = useColonyContext();
 
   useMintToken(getFormOptions);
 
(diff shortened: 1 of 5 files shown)


CzarekDryl (dismissed):

KornelKwiatkowski (dismissed):

TomaszChmielPagepro (approved):

TomaszChmielPagepro (approved):

src/components/v5/common/ActionSidebar/partials/forms/MintTokenForm/hooks.ts
```diff
@@ -28,11 +28,9 @@ export const useMintToken = (
     defaultValues: useMemo<DeepPartial<MintTokenFormValues>>(
       () => ({
         createdIn: Id.RootDomain.toString(),
-        amount: {
-          tokenAddress: colony?.nativeToken.tokenAddress,
-        },
+        amount: {},
       }),
-      [colony?.nativeToken.tokenAddress],
+      [],
```
ArmandoGraterol, line 33: If we are assuming that the amount will correspond to the native token of the colony, then this field doesn't need to be an object like the validation suggests:
```
amount: object()
  .shape({
    amount: number()
      .required(() => 'required field')
      .transform((value) => toFinite(value))
      .moreThan(0, () => 'Amount must be greater than zero.'),
  })
  .required(),
```

 This could be just, which is more or less what we do already in the old mint tokens form:
 ```
 amount: number()
  .required(() => 'required field')
  .transform((value) => toFinite(value))
  .moreThan(0, () => 'Amount must be greater than zero.'),
 ```

Unless I'm missing something

ArmandoGraterol (dismissed):
Working as expected! Just wanted to point out that the validation/how we store the amount could be simplified further.

src/components/v5/common/ActionSidebar/partials/forms/MintTokenForm/hooks.ts
joanna-pagepro: @ArmandoGraterol I agree with you, it would be better to have a different structure here, but I can't make such change, because we are using `AmountField` component here which is used also in other forms and it contains 2 fields inside: one for `amount` and one for `tokenAddress`, so it has to be an object. 
To make the change requested by you, I would have to create a new component (which will cause code duplication) or modify this one (it will require a lot of changes in the code of other forms in this case).

src/components/v5/common/ActionSidebar/partials/forms/MintTokenForm/MintTokenForm.tsx
```diff
@@ -11,11 +11,13 @@ import { useMintToken } from './hooks';
 import { ActionFormBaseProps } from '../../../types';
 import { useDecisionMethods } from '../../../hooks';
 import DescriptionRow from '../../DescriptionRow';
+import { useColonyContext } from '~hooks';
```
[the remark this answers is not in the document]
joanna-pagepro: Done

rdig (approved):

krawczykkatarzyna (approved):

arrenv (approved):
Working as expected now, thank you!

PR #1351, opened 2023-11-17
fix: mint token in root team only

joanna-pagepro:
https://tw.auroracreation.com/app/tasks/15670128

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/TeamsSelect/TeamsSelect.tsx
+++ b/src/components/v5/common/ActionSidebar/partials/TeamsSelect/TeamsSelect.tsx
@@ -13,7 +13,7 @@ import { useAdditionalFormOptionsContext } from '~context/AdditionalFormOptionsC
 
 const displayName = 'v5.common.ActionsContent.partials.TeamsSelect';
 
-const TeamsSelect: FC<TeamSelectProps> = ({ name }) => {
+const TeamsSelect: FC<TeamSelectProps> = ({ name, readonly: readonlyProp }) => {
   const {
     field,
     fieldState: { error },
@@ -44,7 +44,7 @@ const TeamsSelect: FC<TeamSelectProps> = ({ name }) => {
 
   return (
     <div className="sm:relative w-full">
-      {readonly ? (
+      {readonly || readonlyProp ? (
         <TeamBadge
           teamName={
             typeof selectedOption?.label === 'object'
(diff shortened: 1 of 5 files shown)


TomaszChmielPagepro (dismissed):

KornelKwiatkowski (approved):

krawczykkatarzyna (approved):

src/components/v5/common/ActionSidebar/partials/TeamsSelect/TeamsSelect.tsx
```diff
@@ -44,7 +44,7 @@ const TeamsSelect: FC<TeamSelectProps> = ({ name }) => {
 
   return (
     <div className="sm:relative w-full">
-      {readonly ? (
+      {readonly || readonlyProp ? (
```
ArmandoGraterol, line 47: What's the difference between both `readonly`? Is the first one a "form-wide" read-only instead of the `readonlyProp` that only marks this particular input as read-only? If so, I would suggest that we change the name so that it clears that this is the case. Otherwise a person without context, like me :P Wouldn't understand why there are 2 props with kind of the same name, seemingly achieving the same functionality.

ArmandoGraterol (commented):
Looks good to me! I just have 1 question

src/components/v5/common/ActionSidebar/partials/TeamsSelect/TeamsSelect.tsx
joanna-pagepro: @ArmandoGraterol one `readonly` comes from the context and another that I've added is a prop passed to this component, so one is form level and another one is field level, and both of them have to stay

ArmandoGraterol: Thank you for clarifying. In that case, I would argue that we need to rename both variables, at least inside this component, so that it's more clear that one is form level and the other one is input level.

The form level could be something as simple as `readonlyForm`. 

The name of the input prop can still be `readonly`, but once inside the component we could rename it and use it as `readonlyInput`.

This is an example, they could have other names but the important bit is that we can tell the difference between both at first glance.

joanna-pagepro: Isn't it clear with `Prop` suffix?

rdig (approved):

krawczykkatarzyna (approved):

PR #1354, opened 2023-11-17
Refactor AppContext, improve wallet initialization

chmanie:
## Description

This PR refactors the `AppContext`, so that 

- all functions and booleans of the context are now always defined (when in context)
- the login and autoLogin are now more predictable
- `connectWallet` and `disconnectWallet` take care of everything there needs to be
- the login code is not scattered across multiple files anymore
- `walletConnecting` starts with `true` (as the attempt of an autoLogin will happen by default)


jakubcolony (approved):
Great work, much cleaner 👌

bassgeta (approved):
Looks good, much better dev UX

bassgeta (approved):

PR #1355, opened 2023-11-17
Fix sagas with incorrect getPermissionProofs calls

rdig:
Exactly as it says on the tin, this PR fixes two instances of the `getPermissionProofs` call, which didn't pass in the first argument (network client)

Because of this, both the `TransferFundsBetweenPots` and `FundExpenditureMotion` transactions would fail.

Checked all the other instances and everything seems to be correct, so I have no idea how these slipped between the cracks.


chmanie:
Yeah that was due to the ColonyJS update. This probably was copied later.

chmanie (approved):

jakubcolony (approved):

ArmandoGraterol (approved):
LGTM 🆗 💯

PR #1356, opened 2023-11-17
Fix UserAvatar showing alternating blockies

chmanie:
## Description

On my journey through the dapp I noticed that the UserAvatar would show a new blockie on every reload. That's not cool, so I fixed it.

Also I think I improved the UserAvatar in a way that it takes some of the mental load off the developers. The experience should be pretty consistent now, basically pass in a user object or a wallet address as string as the `user` prop. Then `showUsername` will do the right thing.
Mind that `user` is a _required_ prop now. I would argue if we don't have a wallet connected, we shouldn't render this component in any case.

@joanna-pagepro can you maybe make sure that I didn't break any of the stories?

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/Motions/steps/StakingStep/hooks.tsx
+++ b/src/components/v5/common/ActionSidebar/partials/Motions/steps/StakingStep/hooks.tsx
@@ -91,12 +91,10 @@ export const useStakingInformation = (
         return result;
       }
 
-      const userName = users?.length
-        ? users?.find((user) => user?.walletAddress === item.address)?.profile
-            ?.displayName
-        : undefined;
+      const user = users?.find((u) => u?.walletAddress === item.address)
+        ?.profile?.displayName;
 
-      if (!userName) {
+      if (!user) {
         return result;
       }
 
@@ -117,7 +115,7 @@ export const useStakingInformation = (
             },
           ),
           userProps: {
-            userName,
+            user,
           },
         },
       ];
--- a/src/components/v5/shared/UserAvatar/types.ts
+++ b/src/components/v5/shared/UserAvatar/types.ts
@@ -1,16 +1,17 @@
+import { Optional } from 'utility-types';
 import { User } from '~types';
 import { UserStatusMode } from '~v5/common/Pills/types';
 import { AvatarSize } from '../Avatar/types';
 
 export interface UserAvatarProps {
-  userName?: string | null;
-  preferThumbnail?: boolean;
-  user?: User | null;
-  isLink?: boolean;
-  size?: AvatarSize;
   avatarSize?: AvatarSize;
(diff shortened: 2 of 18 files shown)


src/components/v5/common/ActionSidebar/partials/Motions/steps/StakingStep/hooks.tsx
```diff
@@ -117,7 +115,7 @@ export const useStakingInformation = (
             },
           ),
           userProps: {
-            userName,
+            user,
```
chmanie, line 118: I'm unsure whether this was the right thing to do. Please let me know.

src/components/v5/shared/UserAvatar/types.ts
```diff
@@ -1,16 +1,18 @@
...
 import { UserStatusMode } from '~v5/common/Pills/types';
 import { AvatarSize } from '../Avatar/types';
 
 export interface UserAvatarProps {
-  userName?: string | null;
-  preferThumbnail?: boolean;
-  user?: User | null;
+  avatarSize?: AvatarSize;
+  className?: string;
+  hideUsernameOnMobile?: boolean;
```
bassgeta, line 9: I think this one is not needed anymore since you unified everything under `showUsername`

bassgeta (changes requested):

bassgeta (dismissed):

src/components/v5/common/ActionSidebar/partials/Motions/steps/StakingStep/hooks.tsx
rdig: I think it's fine

rdig (changes requested):
Still needs updating the other places we render the avatar that exhibit this behavior: update avatar (user profile), and the user dropdown

Screenshot from 2023-11-20 17-22-39
Screenshot from 2023-11-20 17-22-47

However it seems to be working fine on the Colony Dashboard Widget, in the members list and verified members list

Screenshot from 2023-11-20 17-23-46
Screenshot from 2023-11-20 17-24-06
Screenshot from 2023-11-20 17-26-25

chmanie:
> Still needs updating the other places we render the avatar that exhibit this behavior: update avatar (user profile), and the user dropdown

I can tell you why I missed these places. There, an entirely different user avatar component is rendered (and I have no idea why). Do you want me to consolidate?

chmanie:
Shit, I opened a can of worms there, didn't I.

chmanie:
@rdig I made the changes you requested, even though only half-heartedly. I left this there for later though: https://github.com/JoinColony/colonyCDapp/issues/1372

bassgeta (approved):

ArmandoGraterol (approved):
Nice one comrade 🥇

PR #1362, opened 2023-11-20
Fix arn referenced in createPrivateBeta lambda

inkpot-monkey:
## Description

For some reason the arn of the lambda layers that give access to aws params has changed... this PR updates the arn to the latest version


bassgeta (approved):
Is there a good way of testing this locally? Or is it too resource intensive, because in any case, I trust you got the layer id right, :ship:

thomassdk:
Unfortunately there is no way to test this locally as it is to do with aws resource's ill message you the function url via the backchannels so which should give confirmation that you are getting a code back

willm30 (approved):

PR #1376, opened 2023-11-21
Fix full wallet initialization

chmanie:
## Description

#1357 broke wallet initialization, so this here fixes it for good hopefully.

So now when only a basic wallet is initialized, we run a background task to also set up the full wallet again. No need to run `connectWallet` twice.


jakubcolony (approved):
I was able to create a colony using both dev wallet and metamask 👍

willm30 (approved):
LGTM

rdig (commented):
Late to the party, but tested and works fine. Nice work!

PR #1385, opened 2023-11-22
Improve Static Data Creation Script

rdig:
This PR "improves", by way of replicating every transaction and mutation the CDapp fires, in the form of a script, and with 12 colonies worth of seeded data.

The script itself is still a spaghetti mess, but there's no reason to sync more time into it to make it better. It does it's job, and it's good enough for now.

So just please do a functional test, not an actual code review of it.

How to run:
- start your dev envrironment (it needs to be clean, a fresh start)
- run `node scripts/create-data.js`
- wait for it to finish (it takes a while)
- run webpack
- enjoy QA-ing everything locally

Not that while the script is running you can still browse the CDapp (if you started webpack in parallel), however do note that the reputation miner will be active, so any transaction you send out will likely skip time on your chain. Also, turning it off won't really work, as the script will turn it back on when it needs it to be on. (the script will perform cleanup at the end and stop the miner itself)

Run time is about 6 minutes flat on my machine, so your time might vary. 

_Warning: do not run this on a non-fresh dev environment, as that's just asking for trouble!_


arrenv (approved):
Very nice! This is impressive.

It took 8 minutes on my machine. Maybe a setting at the start could be the number of colonies you need to create.

Not a big issue, but I see 2 broken images.
image

jakubcolony (dismissed):
Life saver!

rdig:
@arrenv 

Already fixed the images (you missed the one missing from the first dev account), plus I made the time-out shorter, so on my machine it dropped down to 2 minutes

I debated if I should make a choice for the number of colonies, but the reality is that for a proper QA you need at least 12 colonies, and have interactions between them

Ways to create less colonies:
- edit the [`tempColonyData`](https://github.com/JoinColony/colonyCDapp/blob/e50e5057d5a50554fd119b62d83f3abde6c95b6c/scripts/tempColonyData.js) file and remove some of the colonies from that list
- stop the script execution after a number of colonies (if you choose this route, make sure to manually toggle off the reputation miner)

rdig:
@arrenv 

I've thought this through some more and put in a limit of `2` colonies when running the script by default, a value that can now be changed via the `--coloniesCount <number>` cli argument, up to a max number of colonies as defined in the `tempColonyData` file

arrenv:
> @arrenv
> 
> I've thought this through some more and put in a limit of `2` colonies when running the script by default, a value that can now be changed via the `--coloniesCount <number>` cli argument, up to a max number of colonies as defined in the `tempColonyData` file

Very nice! Cool addition.

PR #1390, opened 2023-11-23
Remove obsolete members code and replace it with new contributors model

thomassdk:
## Description

This PR fixes a serious problem we have at the moment where the old watchers/contributors model was being used all over the new CDapp that we are building. 

It rectifies this by removing all references to the old code to make sure this can not happen again and to make sure that all references have been removed.

*Note: The Member Context could do with a refactor to extract the paginated code for usage only in the pages where this is necessary but this PR is large enough so that should be tackled in a separate PR*

*Note 2: You will need to refresh your application to get the correct reputation state as this is only trigger when a user accesses a page and is not propagated to the live application. If this is an issue it will need to be addressed in a separate PR*

*Note 3 We are now using the **isVerified** boolean on the contributor model to ascertain user verification which means the old way of tracking addresses on the whitelist in the metadata is now obsolete. This should be removed in a separate PR the issue for tracking is #1472*

**Testing**

To test just interact with the app as usual (testing the join button is a little trickier due to the private beta checks, contact @thomassdk for in depth guidance for how to do this)

**Changes** 🏗

All code that was using the old Watchers/Contributors models have been changed to now use the new `Contributors` model that combines these two concepts. This has the implication that there are some `any` types introduced for the old UI to keep the type checker happy, but on the whole the old UI has been left in for reference.

The most major change that this PR introduces (aside from the changing of models) is that **all** contributors to a colony are fetched when the Colony is loaded. This is necessary in order to enable the contributor select options that are embedded across the application.

**Deletions** ⚰️

There are a lot of these...

- All schema code related to old Watchers/Contributors
- All code that relied on these old models including the lambdas `getMembersForColony` and `getTotalMemberCount` (`getTotalMemberCount` has not be refactored as due to the need to fetch all members for a colony for usage in multiple places, instead this has been replaced with a simple `length` of the array of contributors found.
- Colony Users Page (followers route) (obsolete according to product)


arrenv (approved):
Great work, working well!

melyndav (dismissed):
Nice work on this one today @rdig. Approved.

PR #1393, opened 2023-11-23
Feat/15629289 members page

joanna-pagepro:
https://tw.auroracreation.com/app/tasks/15629289

- tabs
- new member card for contributors and all members
- placeholder card with invite members button to copy colony url
- member actions (copy wallet address, make payment, etc.)
- colony members actions (invite to colony, manage members, etc.)
- filters

Animation while loading more items wasn't added, because I wasn't able to do it. All of the items are loaded again on `load more` action.

Out of scope
- fitering in breadcrumbs
- fix team reputation breakdown

Additional
- new select component used in MembersSelect, UserHubMobile and Navigation (that was used in the old UI)


melyndav (changes requested):
Hey @joanna-pagepro super work on this! 

Some UI issues I've found:

1. The numbers are not centred aligned in each tab count:

image

2. Usernames in both the contributing and members tabs should have a hover(text-blue-400) and when selected, trigger the user pop-over. So it's both the avaar and the username. 

image

3. All verticle dots should be text-gray-400 (both tabs)
image

4. Padding between the tabs and the content below should be 24px (both tabs).
image

5. Only 4px padding between the heading and sub heading text (both tabs) - 
image

6. The invite members empty box isn't the correct height as per Figma:
image

7. View on Gnosis Scan button has a different hover state (showing blue) 
image

8. Can we make the copy wallet address be the correct green variation:
image

9. The horizontal gap should be the same as the vertical gab with both set to 1.5 on both tabs

```
 .gap-y-4 {
    row-gap: 1rem;
}
```

10. Permission pills should trigger a corresponding tooltip:
image

11. The dashed border on the invite members empty box doesn't match Figma. 
image

12. The button insides the invite members empty box also does't match Figma's styling including font and icon size and border colour. 

13. This tooltip should also be the green variation with the check icon and the text inside the tooltip should say 'Link copied'

image

14. The permission pills don't match Figma (font, icon and padding wrong size - 
image

15. All avatars should be 60 x 60px.
image

16. On load more it is reloading the whole list as opposed to just the new ones. This was previously working correctly in the existing design.

This PR:
reloading-not-adding

Existing design:
reloading-adding

src/components/v5/shared/Avatar/Avatar.tsx
```diff
@@ -19,17 +19,12 @@ const Avatar: FC<PropsWithChildren<AvatarProps>> = ({
   title,
   mode = 'general',
 }) => {
-  // eslint-disable-next-line no-param-reassign
-  title = title ?? '';
-  const source = notSet ? null : avatar || getIcon(seed || title);
+  const source = notSet ? null : avatar || getIcon(seed || (title ?? ''));
```
jakubcolony, line 22: Are the extra parentheses necessary?

src/components/v5/shared/MeatBallMenu/MeatBallMenu.tsx
```diff
@@ -30,8 +34,11 @@ const MeatBallMenu: FC<MeatBallMenuProps> = ({
     throw new Error('There are no items added to the menu.');
   }
 
+  const menuItemClassName =
+    'flex items-center text-md transition-colors duration-normal md:hover:bg-gray-50 hover:font-medium rounded py-2 px-4 gap-2 w-[calc(100%+2rem)] -mx-4';
```
jakubcolony, line 38: This is a very long line, can it be split into multiple lines?

src/components/v5/shared/MembersSelect/MembersSelect.tsx
```diff
@@ -1,41 +1,96 @@
...
+        },
+      ];
+    },
+    [],
   );
 
-  const handleChange = (selectedOption: number | undefined) => {
-    setSelectedMember(selectedOption);
-  };
+  const onChange = useCallback<
```
jakubcolony, line 52: It should remain `handleChange`, makes it easier to distinguish from the prop too

src/hooks/members/useAllMembers.ts
```diff
@@ -1,15 +1,17 @@
-import { ColonyRole } from '@colony/colony-js';
 import { useMemo, useState } from 'react';
+import { ColonyRole } from '@colony/colony-js';
+import range from 'lodash/range';
```
jakubcolony, line 3: All lodash imports should be re-exported and imported from `~utils/lodash`

jakubcolony (changes requested):
Nice work overall, a few things I spotted:

The count badge looks bad if there are too many members, there should be a limit:
image

When clicking Load more on the members page, it should be possible to only show loading for the button, not the entire page, which I feel is a nicer UX:

https://github.com/JoinColony/colonyCDapp/assets/112586815/6913fe73-dd37-4194-a9af-9c5162c04211

joanna-pagepro:
@melyndav I wasn't able to add the requested animation effect. I've added a simple one for now. Maybe we can skip it for now and add it later?

package.json
```diff
@@ -244,6 +245,7 @@
     "react-responsive-masonry": "^2.1.7",
     "react-router-dom": "^6.4.2",
     "react-scrollbars-custom": "^4.1.1",
+    "react-select": "^5.8.0",
```
rdig, line 248: Why is this needed, don't we already have a select component ?

Unless there's a good reason for having it, I'm not ok with it, as it adds 700kb to the bundle

src/i18n/en.json
```diff
@@ -1466,5 +1466,19 @@
     "dashboard.agreements.widget.title": "Agreements",
     "dashboard.agreements.widget.noData": "There are no agreements to display.",
     "dashboard.agreements.widget.createObjective": "Create a new agreement",
-    "dashboard.agreements.widget.comingSoon": "Coming soon "
+    "dashboard.agreements.widget.comingSoon": "Coming soon",
+    "badge.new": "New",
+    "badge.top": "Top",
+    "badge.dedicated": "Dedicated",
```
rdig, line 1472: We already have this exact set of message descriptor ids defined, in this very file: 
https://github.com/JoinColony/colonyCDapp/blob/fdfe45b23021a23ed956bae83ba39d6ce27c7675/src/i18n/en.json#L941-L943

What's the difference between them ?

src/routes/UserRoute.tsx
```diff
@@ -10,24 +11,21 @@ import {
...
 
-const navigationItems = [
+const navigationItems: SelectOption[] = [
   {
-    id: 0,
-    linkTo: USER_EDIT_PROFILE_ROUTE,
+    to: USER_EDIT_PROFILE_ROUTE,
     label: 'Profile',
-    value: 'profile',
+    value: 0,
```
rdig, line 18: What do these values mean?

Wound't a `enum` make them descriptive ?

rdig (dismissed):
- Are the "All members" section user popovers no longer showing reputation / domains ?

Screenshot from 2023-11-26 03-44-25

In that case the last divider, after the user's description should be removed

package.json
joanna-pagepro: This lib has implemented a lot of features related to accessibility - you can easily navigate between the options using a keyboard, open the select using it, etc. It shows the menu in a place where there's enough space for it, for example, if you place a dropdown at the end of the page, then the menu would appear on the top of the dropdown and not on the bottom causing an overflow. It's also generic, so it's pretty easy to use it everywhere because you can create a custom option and pass it to the component without any weird if statements in the code. It's also a lot easier to modify than the component that we had in the app. You can also use it with `react-hook-form` lib.

Also, it's only about 30kb - https://bundlephobia.com/package/react-select@5.8.0

src/i18n/en.json
joanna-pagepro: I wanted to make label translations consistent. This `dedicated` translation can be used anywhere and can be changed later to something that won't fit the badge. A lot of translations are related to the component/page that they are used on.

src/routes/UserRoute.tsx
joanna-pagepro: This nav will be removed soon. because of a new design. I think that we can skip it right now. Unless there's a reason to spend time on it right now.

melyndav (changes requested):
Hey @joanna-pagepro! Great work on the updates. A couple of things:

1. I see the 'View on..' menu link is not working for both tabs.
image

2. Can we increase the stroke on the empty widget so it's more obvious like Figma? It's a bit hard to see it's a dashed line currently.
image

3. Can you change the text within the green tool tip across two place to be 'Link copied'.

Place one: the empty members widget
image

Place two: the kebab menu next to the heading:
image

melyndav (changes requested):
Hey @joanna-pagepro I appreciate the further updates. A couple more:

1. On hover of any 'invite' in the dropdown menus across members, the grey tooltip shown here should not show. The green copied tooltip only shows once the menu item in the dropdown has been selected. 

image

2. Also, there should be no underlined on the success tooltip. It should match the other in this area:
image

Correct without underline:
image

3. Can the tooltip come from the right side of the dropdown so it does not cover the other dropdown items below? 

image

arrenv (changes requested):
@joanna-pagepro In addition to other issues, one thing I noticed is that the permission filters are not working as expected.

When filtering by `Payer` I get `Owner` showing. It should only show those matching the exact abstracted permissions, i.e. if I select `Payer` it will only show those with the exact `Payer` permission pill.

joanna-pagepro:
@arrenv @melyndav can you check if it's ok now?

melyndav (changes requested):
Hey @joanna-pagepro!

Thanks again for the updates. Just a minor issue, when I try to copy the users wallet address from the dropdown menu, the menu closes without first showing the success tooltip. And when I open the menu again, it shows the tooltip. See clip below:

arrenv (changes requested):
Great work @joanna-pagepro, filtering is working a lot better now, a few minor things:

1. The `Custom` permission pill is showing even if the contributor does not have any permissions.

image

2. The tabs should still work with routing so that they can be directly linked to as it worked previously.
   - Contributors should be `/members/contributors`
   - Members should be `/members`
   
3. When filtering by Payer, the `Invite members` box disappears, it is there for `Owner` though and if I filter by `Custom` > ` Funding`.

invite-members

joanna-pagepro:
@arrenv @melyndav it should be ok now. Can you check?

arrenv (changes requested):
> @arrenv @melyndav it should be ok now. Can you check?

Thanks @joanna-pagepro, so very close.

Looks like removing the incorrect permission pills broke the alignment.

image

arrenv (dismissed):
All my issues have been resolved, looks great, thank you @joanna-pagepro

PR #1397, opened 2023-11-24
fix: updated cleave package, formatted input prefix, chart colors

CzarekDryl:
https://tw.auroracreation.com/app/tasks/15649439
https://tw.auroracreation.com/app/tasks/15655944

Things that I've fixed in this PR:
- Cleave.js package update to make tailPrefix work
- Fixed colors of staking chart progress bar
- Hiding indicator of staking chart when for vote is higher than minimum value

image
image

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/Motions/consts.ts
+++ b/src/components/v5/common/ActionSidebar/partials/Motions/consts.ts
@@ -26,6 +26,6 @@ export const opposeOption = {
 };
 
 export const STAKING_RADIO_BUTTONS: ButtonRadioButtonItem<MotionVote>[] = [
-  supportOption,
   opposeOption,
+  supportOption,
 ];
--- a/src/components/v5/common/Fields/InputBase/FormattedInput.tsx
+++ b/src/components/v5/common/Fields/InputBase/FormattedInput.tsx
@@ -32,6 +32,8 @@ const FormattedInput: FC<FormattedInputProps> = ({
     buttonRef,
     wrapperRef,
     customPrefixRef,
+    handlePrefixTrim,
+    inputRef,
   } = useFormattedInput(value, options);
 
   const stateClassNames = useStateClassNames(
@@ -87,6 +89,9 @@ const FormattedInput: FC<FormattedInputProps> = ({
         )}
         <Cleave
           {...rest}
+          htmlRef={(ref) => {
+            inputRef.current = ref;
+          }}
           disabled={disabled}
           key={dynamicCleaveOptionKey}
           placeholder={placeholder || ''}
@@ -96,7 +101,10 @@ const FormattedInput: FC<FormattedInputProps> = ({
            */
           options={options}
           onInit={(cleaveInstance) => setCleave(cleaveInstance)}
-          onChange={onChange}
+          onChange={(e) => {
+            handlePrefixTrim(options.prefix);
+            onChange?.(e);
+          }}
           className={clsx(
             className,
             state ? stateClassNames[state] : undefined,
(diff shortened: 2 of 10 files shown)


src/components/v5/common/Fields/InputBase/hooks.ts
```diff
@@ -83,6 +79,18 @@ export const useFormattedInput = (
     );
   }, []);
 
+  const handlePrefixTrim = useCallback((prefix?: string) => {
+    if (inputRef.current) {
+      const input = inputRef.current;
```
TomaszChmielPagepro, line 84: I think you can get input element from onChange event

src/components/v5/common/Fields/InputBase/FormattedInput.tsx
```diff
@@ -96,7 +101,10 @@ const FormattedInput: FC<FormattedInputProps> = ({
            */
           options={options}
           onInit={(cleaveInstance) => setCleave(cleaveInstance)}
-          onChange={onChange}
+          onChange={(e) => {
+            handlePrefixTrim(options.prefix);
```
TomaszChmielPagepro, line 105: Name of this function is quite misleading as this does not trim the prefix but it's moving the cursor to correct position

KornelKwiatkowski (dismissed):

krawczykkatarzyna (dismissed):

src/i18n/en.json
```diff
@@ -1221,6 +1221,7 @@
     "motion.staking.activateTokens": "Activate tokens",
     "motion.staking.staked": "Staked {value}",
     "motion.staking.input.label": "Stake amount",
+    "motion.staking.input.balance": "Balance {balance}",
```
rdig, line 1224: Why was this added? 

You have it in the exact next line

src/components/v5/common/Fields/InputBase/FormattedInput.tsx
rdig: As I understand it we're no longer doing prefix/suffix token symbol for the staking input. 

@arrenv please confirm if that's still the case

src/components/v5/common/ActionSidebar/partials/Motions/consts.ts
```diff
@@ -26,6 +26,6 @@ export const opposeOption = {
 };
 
 export const STAKING_RADIO_BUTTONS: ButtonRadioButtonItem<MotionVote>[] = [
-  supportOption,
   opposeOption,
+  supportOption,
```
rdig, line 30: This branch will cause a merge conflict with #1394 

Just a heads up

rdig (changes requested):
- "Not public until min 10% is staked" line should only disappear if the motion has that amount of stake, not just when entering a value higher than it

Screenshot from 2023-11-26 02-50-45

Screenshot from 2023-11-26 02-50-49

src/components/v5/common/ActionSidebar/partials/Motions/consts.ts
arrenv: > As I understand it we're no longer doing prefix/suffix token symbol for the staking input.
> 
> @arrenv please confirm if that's still the case

Yes, that is still the case. It was never in the design as far as I understand.

CzarekDryl: @arrenv @rdig There is token symbol on designs - https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-V5?type=design&node-id=2729-220432&mode=design&t=wSu6J9exWt1VH7yU-0, or these are old ones?

ArmandoGraterol (changes requested):
Prefix still shows for me in the input when I input a number. It has some confusing behavior when deleting what I input but I guess that will be fixed once we take care of the prefix

CzarekDryl:
I'm closing this PR and moving changes here - https://github.com/JoinColony/colonyCDapp/pull/1427

PR #1399, opened 2023-11-24
Add token activation to stake motion saga

### LINKED ISSUE #1350: Activate tokens automatically when staking
# Description

In order to help streamline the motion process for first time users, we would like the Staking process to also activate the number of tokens the user would like to stake towards the motion.

This is a new flow we are trying, so may require grouping the activating tokens, approving tokens and staking token transactions.

Note: if this is not possible, or causes weird behavior, at the very least we need to handle validation on the staking flow.

image

ArmandoGraterol:
With this PR you shouldn't need to activate the tokens first before staking. Now the staking saga should activate them for you if you have enough inactive tokens, otherwise, if you have enough active tokens it should work as before.

If you don't have enough active + inactive tokens, then the good ol' `You don't have enough tokens to stake` error message should appear and it shouldn't let you stake.

Resolves #1350


rdig (changes requested):
Having some tokens/rep, but not enough to cover the max amount, and entering an amount larger than what you have (ie: press the max button), will crash the app, as it tries to approve more tokens than you actually have.

Max, in this case, should limit you to the amount you have available for staking at the current moment (non active tokens + active non staked tokens)

Also, the input needs to be validated for the same limit, as the user might just enter the value manually, bypassing the max button

Screenshot from 2023-11-26 03-18-22
Screenshot from 2023-11-26 03-18-30
Screenshot from 2023-11-26 03-18-44

Also, once you do this, the app constantly crashes and cannot be loaded anymore for the user that attempted this.

Screenshot from 2023-11-26 03-25-01

rdig:
Note that the latter error might not be specific to your branch, as I can still replicate it on master / any other branch I'm testing, however, It's the "activating more tokens than available" issue that triggers it, so it's somewhat related.

This needs to be investigated and fixed, as it's app breaking. We cannot launch until we find and fix that.

ArmandoGraterol:
@rdig 

- Okay, so the CDapp was crashing because `UserTransactionContext` was stuck in an infinite loop after a transaction failed it seems. It should be fixed now.

- The form should now take into account the user balance, both for the max button and for when you manually enter an amount, and disable the submit button if it exceeds the balance.

- I've fixed the balance message in the staking form using the wrong ID and therefore not displaying the message.

Let me know if you find something else related that needs fixing :)

jakubcolony (approved):
Looking good, nice work!

bassgeta (approved):
looks good, tried it out and it also works good, :shipit:

chmanie (approved):
Approving this!

PR #1414, opened 2023-11-28
Connnect colony stats on activity page and on colony home

thomassdk:
## Description

This PR connects the stats on the colony page (the number of active actions) and the stats on the activity page (Active actions, Recent actions and most active team)

[Screencast from 2023-11-28 10-11-28.webm](https://github.com/JoinColony/colonyCDapp/assets/32598350/20fc6e3c-06c6-4c75-bcd5-f3e3b35d5c63)

image

To test create actions and these pages should be populated with the results

**New stuff** ✨

There have been some fun opensearch aggregate queries added to the schema


arrenv (changes requested):
@thomassdk 

1. How is this Recent actions value calculated? I am getting 42 in the widget, 40 in the activity list, the database is showing 49 actions.

Also, the date of actions is in the future in the activity feed and in the database, so it seems odd that it is counting it in the widget. With only 8 that fall in the past 30 days.

2. I have an motion the the staking phase, however, the active actions value is not showing any value on the dashboard and activity pages.

thomassdk:
Thankyou for your review @arrenv 

1. recent actions values actions are calculated counting everything from 30 days ago as discussed. The issue with future dates seems like a non problem as putting in code to discount motions that happen in the future won't be an issue when the app is live no?  We can have a call to discuss this further.

2. We are counting all motions that are not finalized, perhaps this is wrong I will double check now

src/components/frame/v5/pages/ActivityPage/hooks.tsx
```diff
@@ -1,17 +1,69 @@
...
+  const midnightThirtyDaysAgo = startOfDay(thirtyDaysAgo);
+  return midnightThirtyDaysAgo.toISOString();
+};
 
 export const useActivityFeedWidgets = (): WidthBoxItem[] => {
+  const { colony } = useColonyContext();
+  const { domains, colonyAddress = '' } = colony ?? {};
+  const { data: motionData } = useGetActiveMotionsQuery({
+    variables: {
+      colonyId: colonyAddress ?? '',
```
bassgeta, line 26: I guess we don't need ?? if we have a fallback here

src/components/frame/v5/pages/ActivityPage/hooks.tsx
```diff
@@ -1,17 +1,69 @@
...
+    },
+  });
+
+  const activeActionsNumber = motionData?.searchColonyMotions?.total ?? 0;
+
+  const { data: actionData } = useGetTotalColonyActionsQuery({
+    variables: {
+      colonyId: colonyAddress ?? '',
```
bassgeta, line 34: same with the ?? operator here

src/components/frame/v5/pages/ActivityPage/hooks.tsx
```diff
@@ -1,17 +1,69 @@
...
+      since: getThirtyDaysAgoIso(),
+    },
+  });
+
+  const recentActionsCount = actionData?.searchColonyActions?.total ?? 0;
+
+  const { data: domainData } = useGetTotalColonyDomainActionsQuery({
+    variables: {
+      colonyId: colonyAddress ?? '',
```
bassgeta, line 43: same with the ?? operator here

bassgeta (changes requested):

src/components/frame/v5/pages/ActivityPage/hooks.tsx
thomassdk: very good point lazy copy and paste strikes again :)

arrenv (dismissed):
Thanks for the updates, stats are showing correctly from what I have tested.

I created two new motions, staked one past 10% and did not stake another, the count increased by one as expected. Nice work!

bassgeta (approved):

src/components/frame/v5/pages/ActivityPage/hooks.tsx
```diff
@@ -1,25 +1,82 @@
...
+    variables: {
+      colonyId: colonyAddress,
+    },
+  });
+
+  const domainsActionCount =
+    // Buckets exists on this type but for some reason typescript is not
+    // recognising this
+    // @ts-ignore
+    domainData?.searchColonyActions?.aggregateItems[0]?.result?.buckets ?? [];
```
jakubcolony, line 56: I suspect it's the codegen that can't pick up union types without a separate plugin: https://the-guild.dev/graphql/codegen/plugins/other/fragment-matcher
It was pretty straightforward last time I set it up so I think it's worth getting sorted in this PR as we're gonna have the same problem with all GQL unions. Feel free to give me call if you need a hand with this

src/graphql/queries/actions.graphql
```diff
@@ -43,12 +43,31 @@ query GetSafeTransactionStatus($input: GetSafeTransactionStatusInput!) {
...
+    filter: { colonyId: { eq: $colonyId } }
+    aggregates: { name: "FromDomainIdCount", type: terms, field: fromDomainId }
+  ) {
+    aggregateItems {
+      result {
+        ... on SearchableAggregateBucketResult {
+          __typename
+          buckets {
+            key
+            doc_count
```
jakubcolony, line 63: Can we rename this to be camelCase? Should be able to do 
```gql
docCount: doc_count
```

jakubcolony (changes requested):

src/components/frame/v5/pages/ActivityPage/hooks.tsx
thomassdk: turns out I wasn't understanding this issue clearly enough as the type was generated correctly have fixed this issue now :)

melyndav (approved):
Nice work @thomassdk! Get this baby merged! ;)

arrenv (approved):
Good to go!

jakubcolony (approved):
LGTM, nice work! 💪

PR #1428, opened 2023-11-29
Fix broken form error display

### LINKED ISSUE #1415: Broken inputs after refactor
## Description

After refactor the existing input fields, there are some issues now with inputs across other areas of the CDapp, such as the extensions area with error states as seen below.

This issue should address these UI problems and ensure all inputs are displaying correctly across all areas, models, motion inputs etc. 

image

thomassdk:
## Description

This PR fixes the broken form error display in #1415 by allowing in some cases the original design of layout shifting to occur on error:

image

Resolves #1415

### DIFF
--- a/src/components/v5/common/Fields/Input/Input.tsx
+++ b/src/components/v5/common/Fields/Input/Input.tsx
@@ -131,7 +131,7 @@ const Input: FC<InputProps> = ({
             {isDecoratedError ? (
               <InputPills message={customErrorMessage} status="error" />
             ) : (
-              <FormError isFullSize alignment="left">
+              <FormError isFullSize alignment="left" allowLayoutShift={false}>
                 {customErrorMessage ||
                   (errorMaxChar
                     ? formatMessage(
--- a/src/components/v5/shared/FormError/FormError.tsx
+++ b/src/components/v5/shared/FormError/FormError.tsx
@@ -8,14 +8,19 @@ const FormError: FC<PropsWithChildren<FormErrorProps>> = ({
   alignment = 'right',
   isFullSize,
   children,
+  allowLayoutShift = true,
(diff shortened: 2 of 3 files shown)


src/components/v5/common/Fields/Input/Input.tsx
```diff
@@ -131,7 +131,7 @@ const Input: FC<InputProps> = ({
             {isDecoratedError ? (
               <InputPills message={customErrorMessage} status="error" />
             ) : (
-              <FormError isFullSize alignment="left">
+              <FormError isFullSize alignment="left" allowLayoutShift={false}>
```
bassgeta, line 134: should we maybe just add the classname override here? maybe there's no need for additional props if we can use CSS overrides for this

bassgeta (changes requested):

src/components/v5/common/Fields/Input/Input.tsx
thomassdk: I prefer the additonal prop here as it clearly states the meaning of the extra CSS, it is also less code as we dont have to go into every `FormError` component to set these classNames 
```
allowLayoutShift ? 'mt-1' : 'absolute',
```

bassgeta: yeah fair enough, I thought about this too, it's a different mode of error :D i can't think of a better name than `allowLayoutShift` , so lgtm

bassgeta (approved):

arrenv (approved):
Fixes it on my end, tested also Colony creation inputs and account inputs and no issues.

ArmandoGraterol (approved):
Nice work Thomas! All inputs displaying as expected as far as I can tell

PR #1433, opened 2023-11-29
Hotfix for user query's in validate user lamdba

thomassdk:
Hotfix for QA so that the validate user lamdba does not crash


chmanie (approved):

arrenv (approved):
Good to go!

PR #1434, opened 2023-11-29
fix: changed tx buttons order in nav, mobile styling

CzarekDryl:
https://tw.auroracreation.com/app/tasks/15633537

Changes in this PR:
- Moved tx buttons to the left of user nav
- Changed mobile styling to show only icons


src/components/frame/Extensions/layouts/ColonyLayout.tsx
```diff
@@ -80,14 +81,24 @@ const ColonyLayout: FC<PropsWithChildren> = ({ children }) => {
...
+    ),
+    groupState === TransactionGroupStates.AllCompleted && (
+      <CompletedButton
+        text={isMobile ? undefined : { id: 'button.completed' }}
+        className={clsx({
+          '!min-w-0': isMobile,
+        })}
+      />
+    ),
+  ];
```
arrenv, line 101: Do these need to be two separate components? Can it not just be one component with the pending and completed states?

arrenv (changes requested):
Thank you @CzarekDryl great to have the button back in there. A few things I noticed:

1. The transaction tab should still open up when you click on the pending button.
image

2. The pending button is blurred in the background when activating tokens.
pending-button


3. When creating an action using the Reputation decision method, the Pending button does not seem to come up until after awhile.
pending-button-2

arrenv (changes requested):
@CzarekDryl All good regarding point 3.

I just noticed that the Completed button does not work on click, it should work the same as the pending state of the button.

image

CzarekDryl:
@arrenv This should work now as expected, please check it :)

melyndav (approved):
Nice work @CzarekDryl, approved.

arrenv (approved):
Everything looks good, thank you @CzarekDryl !

PR #1439, opened 2023-11-29
Don't create contributor if already in existence

thomassdk:
Colony contributor's can come into existence in multiple ways not just via this lambda

This PR does not make the lambda error on existence

### DIFF
--- a/amplify/backend/function/validateUserInvite/src/index.js
+++ b/amplify/backend/function/validateUserInvite/src/index.js
@@ -121,7 +121,7 @@ exports.handler = async (event) => {
 
   if (contributorExistenceCheckQuery?.data) {
     // contributor already exists no need to create one
-    return;
+    return true;
   }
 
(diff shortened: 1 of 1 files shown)


amplify/backend/function/validateUserInvite/src/index.js
```diff
@@ -120,7 +120,8 @@ exports.handler = async (event) => {
   );
 
   if (contributorExistenceCheckQuery?.data) {
-    throw new Error('Colony contributor already exists!');
+    // contributor already exists no need to create one
+    return;
```
chmanie, line 124: Should we also return `true` here?

thomassdk: good point well made

thomassdk: fixed

chmanie (approved):

bassgeta (approved):

Issue #1411, opened 2023-11-27
Default Colony avatars upon Colony creation


melyndav:
## Description

**Issue one:** 
When a new Colony is created, the default avatar should be a blockie that is displayed at the top of the navigation bar and across other core areas of the Colony. Currently, it's a fake avatar that was designed for a fake colony for design purposes in Figma.

![image](https://github.com/JoinColony/colonyCDapp/assets/82127904/75b4ff33-ce0f-4723-8ab3-12fd2167bf26)

**Issue two:**
When I am connected and on the landing page, I also shouldn't see this default Colony avatar and I should see our default Colony logo that's used across the create profile and create colony wizards.  

Incorrect:
<img width="1113" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/615e9071-01d5-4a84-b633-fd9c1e0474fe">

Correct:
![image](https://github.com/JoinColony/colonyCDapp/assets/82127904/2ce46c60-b8fa-4123-be87-f915656f2268)

## Important

Anywhere this fake Colony avatar has been added by the PP devs, should be removed. It was purely for a placeholder in Figma for a fake Colony. 
![image](https://github.com/JoinColony/colonyCDapp/assets/82127904/9fcca75b-b761-49fe-8674-17d5afe432b7)

PR #1440, opened 2023-11-29
Add missing colony logo and breadcrumb to landing page

### LINKED ISSUE #1412: Colony landing page missing UI elements
## Description

The Colony landing page is missing the breadcrumb placeholder:

image

And the Colony logo in the navigation draw:

image

Both should be added in to the Colony landing page.

ArmandoGraterol:
I've updated the page headers to allow the breadcrumbs to show without the need for a title, which is the case for the landing page. The changes introduced should be specifically for this, but they are made in general components so make sure to test the breadcrumbs and titles of other pages in order to check that they haven't changed with this.

FireShot Capture 239 - Welcome to Colony - localhost

Resolves #1412

### DIFF
--- a/src/components/frame/Extensions/layouts/MainLayout.tsx
+++ b/src/components/frame/Extensions/layouts/MainLayout.tsx
@@ -23,12 +23,13 @@ const MainLayout: FC<PropsWithChildren<MainLayoutProps>> = ({
   return (
     <PageLayout
       headerProps={{
-        pageHeadingProps: pageHeadingTitle
-          ? {
-              title: pageHeadingTitle,
-              breadcrumbs,
-            }
-          : undefined,
+        pageHeadingProps:
+          pageHeadingTitle || breadcrumbs.length
+            ? {
+                title: pageHeadingTitle,
+                breadcrumbs,
+              }
+            : undefined,
         userNavigation: <UserNavigationWrapper />,
       }}
       sidebar={Sidebar}
--- a/src/components/frame/LandingPage/LandingPage.tsx
+++ b/src/components/frame/LandingPage/LandingPage.tsx
@@ -100,7 +100,6 @@ const LandingPage = () => {
       {
         key: 'landing-page',
         label: 'Colony App',
-        href: '/',
       },
     ]);
   }, [setBreadcrumbs]);
--- a/src/components/frame/LandingPage/LandingPage.tsx
+++ b/src/components/frame/LandingPage/LandingPage.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import { defineMessages, FormattedMessage } from 'react-intl';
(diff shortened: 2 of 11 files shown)


thomassdk (changes requested):
Code looks good to me but im getting the logo flashing in the top left corner

[Screencast from 2023-11-30 08-35-28.webm](https://github.com/JoinColony/colonyCDapp/assets/32598350/0eca21d5-35e8-45f0-9ba3-1f303a6ca47f)

thomassdk:
[Screencast from 2023-11-30 08-36-44.webm](https://github.com/JoinColony/colonyCDapp/assets/32598350/7623df7c-d970-44ea-b266-c2bda8212633)

melyndav (changes requested):
Heya @ArmandoGraterol! Thanks for picking up this issue. A couple of issues picked up.

1. Padding top and bottom in the nav bar has changed when my wallet is disconnected but isn't an issue when it's connected. The should be the same.

2. There should be no hover or interaction on the title in this state (it should work in other states) for both disconnected and connected states:
image

3. When my wallet is connected, I should see the colony avatr and not this placeholder colony avatar that's been added here by PagePro:

image

ArmandoGraterol:
@thomassdk

> Code looks good to me but im getting the logo flashing in the top left corner

Unfortunately, changing doesn't seem trivial. This happens because the user is not connected fast enough, that's why it flashes. It happens in other parts as well, for example in one of the items on the landing page.

The first solution that came to my mind was adding a spinner to the landing page, but if we check for `walletConnecting || userLoading` then the spinner will show when we open the modal to connect a wallet, which is terrible in this case, and that is ignoring the fact that we are adding yet another spinner to the CDapp.

I think I'll have to talk about this with Chris to see how can we proceed.

@melyndav
> 1. Padding top and bottom in the nav bar has changed when my wallet is disconnected but isn't an issue when it's connected. The should be the same.
> 2. There should be no hover or interaction on the title in this state (it should work in other states) for both disconnected and connected states:

Fixed!

> 3. When my wallet is connected, I should see the colony avatr and not this placeholder colony avatar that's been added here by PagePro:

There's a separate issue for this: https://github.com/JoinColony/colonyCDapp/issues/1411 which I'm working on right meow.

melyndav (approved):
Looking slick @ArmandoGraterol! Thanks for your work on this one!

arrenv (approved):
Looks good my man @ArmandoGraterol !! Even adds it on the users account page.

ArmandoGraterol:
@thomassdk I've added a loading spinner to the landing page for when the user or wallet is loading, this should prevent the flashing when reloading with autologin. Before this spinner would even show when opening the dialog to connect a wallet so I've moved the handler to a bit further down the road when you actually click and connect a wallet. Let me know if this is nicer than the previous solution.

arrenv (approved):

PR #1441, opened 2023-11-29
Added activity feed on colony dashboard

TomaszChmielPagepro:
This PR adds an activity feed table on the colony home page and filtering by team.
Optimized refetch method of `useNetworkMotionStates`, the consumer can now decide which motions ids should be refetched.

image
image
image


arrenv (changes requested):
@TomaszChmielPagepro Thank you adding this is, coming together now!

From my side, we just need to align it better with the design.

1. The bottom of the activity feed should align with the bottom of the widgets:
image


2. There should be 7 items per page on the dashboard version of the activity feed. The height of each row is 70px as well. This is different to the dedicated activity page where each row is 58px, and there are 10 items per page.


3. When clicking on an item in the activity feed, it re-renders the page changing the team and refetching the list.
rerender-when-clicking-item


4. The width of the widgets does not match the design. The activity feed on the dashboard should have a width of 61% at the max container width.
The design:
image
This PR:
image
It causes the chart in the widget to disappear:
image

TomaszChmielPagepro:
@arrenv Issues fixed

melyndav (changes requested):
Hey @TomaszChmielPagepro. Great start on this one.

1. I see the issue Arren raised about the width of the table and it impacting the three left side widgets is still visible.
image

2. Is it possible to align the bottom of the table with the left side widget
image

3. Can the skeleton loader also be applied to the avatars to keep it consistent with the activity feed?
image

TomaszChmielPagepro:
@melyndav @arrenv Issues fixed

melyndav (dismissed):
Great work @TomaszChmielPagepro! Thank you and approved. :)

tailwind.config.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
     fontSize: {
       xs: ['0.625rem', 1.6],
       sm: ['0.75rem', 1.5],
-      md: ['0.875rem', 1.4],
+      md: ['0.875rem', '1.25rem'],
```
arrenv, line 13: This looks to better match the design, we looked in as many places as we could find it was being used and seems to not cause any alignment issues.

arrenv (approved):
Nice work @TomaszChmielPagepro, looks and works great!

PR #1442, opened 2023-11-29
Wire up PermissionsNeededBanner component

### LINKED ISSUE #1196: Missing extension permissions
## Steps to reproduce

1. Install an extension and then remove it's permissions.
2. Return to the extension page and note the missing permissions banner.

## Expected behaviour

* You should be able to trigger an action to award the extension it's required permissions.
* TODO text should not be visible in the UI.

You can see the design here - https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-V5?type=design&node-id=1864%3A189234&mode=design&t=ENM1S5QhU4zQBbnH-1

## Actual behaviour

* When an extension does not have the correct permissions, the following callout appears: "This extension is missing some or all of the permissions it needs to work."
* On clicking "Enable permissions" another callout appears with the text: "The required permissions have been updated. TODO: Wire this in with Actions / Motions"

ArmandoGraterol:
Wired up `PermissionsNeededBanner` with extension enabling saga. There's no need for an `Enable extension` button in the banner as this saga already takes care of that. Setting permissions is actually the last step of the saga when you are enabling an extension so if that is missing, there's nothing else to do other than that.

To test, you can install any of the expenditure-related extensions as some of them do not assign the permissions on initialization, otherwise, you can manually remove the permissions from any enabled extension.

Resolves #1196


arrenv (approved):
Nice one @ArmandoGraterol ! I tested on the Staged Expenditure and also by removing permissions manually from the Reputation Weighted extension, working as expected.

PR #1445, opened 2023-11-29
Keep track of network fee in actions

jakubcolony:
## Description

This PR modifies the schema to store the network fee amount for a payment action. It also modifies existing components to make sure they're no longer subtracting fee from the amount.

To test, create payment actions and verify the amount shown in the sidebar and the activity feed doesn't include network fee. Network fee should be stored in a separate field instead.


thomassdk (commented):
Maybe I am misunderstanding this PR but it seems to me that the amount that I sent is modified in the activity feed et al. is this due to the network fee?

Screenshot from 2023-11-30 08-56-40
Screenshot from 2023-11-30 08-57-02
Screenshot from 2023-11-30 08-57-20

jakubcolony:
Sorry @thomassdk, it was just me forgetting to update ingestor hash! 🙄

thomassdk (approved):
Works well now :raised_hands:

PR #1448, opened 2023-11-30
fix: 15690133 - external url

krawczykkatarzyna:
Fix issue with building external links

### DIFF
--- a/src/utils/isUrlExternal.ts
+++ b/src/utils/isUrlExternal.ts
@@ -1,7 +1,15 @@
 import { Path } from 'react-router-dom';
 
 export const isUrlExternal = (url: string | Partial<Path>) => {
-  const regEx = /^https/;
+  if (typeof url !== 'string') {
+    return false;
+  }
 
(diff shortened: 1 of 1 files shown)


TomaszChmielPagepro (approved):

arrenv (changes requested):
I still seem to have an issue, the URL is being generated as of the screenshot and going to a 404.

image

Validation in the social links should handle this, but it is still possible to create a different type of link when using the create data script.

Where are you seeing this fix the issue and I can check there as well?

chmanie:
@arrenv are you saying we should detect it as URLs if someone just gives "example.com" or "www.example.com" (as opposed to "https://www.example.com")

arrenv:
> @arrenv are you saying we should detect it as URLs if someone just gives "example.com" or "www.example.com" (as opposed to "https://www.example.com")

If possible, yes, otherwise, at least we should hide links that are won't work due being added to the route path.

PR #1381, opened 2023-11-22
feat: banner flex direction change depending on container width

bassgeta:
## Description

If the notification banner is less than 600px wide, it should display the call to action underneath the text not to the right.
I didn't touch the colors as it was already done in [this PR](https://github.com/JoinColony/colonyCDapp/pull/1207)

**New stuff** ✨

* `@tailwindcss/container-queries` added to dependencies

Resolves #1254 , but only after that PR is also merged

PR #1454, opened 2023-12-01
fix: wire up permissions needed banner again

bassgeta:
## Description

Rebase didn't alert me of any changes, nor did the build fail, so my merge of #1381 overrided work on #1442.
This PR realigns everything as it should.
Mea culpa :facepalm: 

**Changes** 🏗

* realign the permissions needed banner to work again


arrenv (changes requested):
@bassgeta I believe this is an issue that was not caused by you, and was missed previously, but if you don't have root permissions, the button still appears and you can click it, but nothing happens other then an error in the console.

Do you think it is worth tackling this in this PR?

image

ArmandoGraterol (approved):
Best code I've ever seen! Let me know if you want me to take care of the permissions issues Arren pointed as it is technically my bad :) The solution shouldn't be too complicated , it's just a matter of hiding the button when the user doesn't have the permissions

arrenv (approved):
Thanks for clarifying, good to go on my end!

ArmandoGraterol (dismissed):
@bassgeta @arrenv The button should now be hidden when the user doesn't have the root permission :) Let me know if it's the same for you guys.

arrenv (dismissed):
Very nice @ArmandoGraterol & @bassgeta, works like a charm.

In future, it would be good to have it open the Manage permission action panel with all details pre-filled, but for now I have added that to the list for future improvements.

src/components/frame/Extensions/pages/ExtensionDetailsPage/PermissionsNeededBanner.tsx
```diff
@@ -1,31 +1,74 @@
...
+
+const PermissionsNeededBanner = ({ extensionData }: Props) => {
+  const { colony } = useColonyContext();
+  const { user } = useAppContext();
 
-const PermissionsNeededBanner = () => {
-  // @TODO: Change extension missing permissions functionality
   const [isPermissionEnabled, setIsPermissionEnabled] = useState(false);
+  const errorAction = getFormAction(ActionTypes.EXTENSION_ENABLE, 'ERROR');
+  const successAction = getFormAction(ActionTypes.EXTENSION_ENABLE, 'SUCCESS');
```
jakubcolony, line 41: IIRC, these utils were put in so that components like `ActionForm` and `ActionButton` can get default error and success actions to allow the consumer to omit providing them. Since you have to pass them to `useAsyncFunction` anyway, I don't think there's any benefit in using those over passing `EXTENSION_ENABLE_ERROR` and `EXTENSION_ENABLE_SUCCESS` explicitly.

jakubcolony (approved):
Good to go!

melyndav (dismissed):
Everything looks good on my end @bassgeta great work!

arrenv (approved):

PR #1420, opened 2023-11-28
user profile redesign

krawczykkatarzyna:
Note: 
- breadcrumbs will be added in this pull request - https://github.com/JoinColony/colonyCDapp/pull/1440
- navigation on menu sidebar will be added later  

Screen Shot 2023-12-01 at 18 16 30

PR #1456, opened 2023-12-01
feat: add new page loader

### LINKED ISSUE #1217: Update app main loader animation
### Story

As a *user*, I want to *see a visual indicator of activity/loading*, so that *I know I should wait for it to finish.*

### Description
With the updated app design, we want to show an updated loading state to align with it.

image

[Design - Loader - Figma](https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-V5?type=design&node-id=12582%3A207698&mode=dev)

As with the current setup, we also want to have the loading text change to give feedback to the user as to what is being loaded.

[Design - Loading step change - Figma](https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-V5?type=design&node-id=12654%3A102852&mode=dev)

We will also need to keep the same service loading statuses at the timed intervals with some updates

#### If after 15 seconds it hasn’t loaded: 

- Show a message to the user with the following copy: 
“"It’s taking a while to connect."
"Please hold tight while we keep trying.”

- Only add the copy, there is no need to change the existing loading animation.

[Design - Text after 15 sec - Figma](https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-V5?type=design&node-id=13873%3A535567&mode=dev)

#### If after 30 seconds it hasn’t loaded: 

- Show a message to the user with the following copy: 
“Oh no! We tried but couldn’t get any response from the service."
"Please try again later.”

[Design - Text after 30 sec - Figma](https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-V5?type=design&node-id=13873%3A535588&mode=dev)

### Acceptance criteria
- Loader has been updated to match the latest designs.
- Copy has been updated for each loading status state.

bassgeta:
## Description

Add a new page loader as per [Figma](https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-V5?type=design&node-id=11978-400122&mode=dev).
Use it in the `LoadingTemplate` component, so it's essentially our app loader.
Also the `!h-12 !w-12` thing was added because our `Icon` component uses its own CSS file, therefore it overrides tailwind by specificity, as tailwind is always injected first.

**New stuff** ✨

* `PageLoader` component done with SVG and CSS :sunglasses: 

**Changes** 🏗

* `LoadingTemplate` now uses a new loader and simplified the rendering a bit

Resolves  #1217


melyndav (changes requested):
Hey @bassgeta! You're smashing these out, thanks for your work on these PRs!

Some feedback for this:

1. The old coloured gradient is still showing
2. The bg of the circle is showing as white for me (should be text-gray-100)
3. Check the tracking (letter spacing) on the all variations of the text (loading app, loading colony etc) colony text - we have one in the design system (it's the same as the top breadcrumb one) so it should match this (example below)
image

image

chmanie:
Shouldn't we update the loading spinner everywhere so that we have a consistent design? We are also using `<Spinner>` and `<SpinnerLoader>` in a bunch of other places. Maybe even consolidate the two while we're at it (there also might be other, unused Spinners that can be removed).

bassgeta:
> Shouldn't we update the loading spinner everywhere so that we have a consistent design? We are also using `<Spinner>` and `<SpinnerLoader>` in a bunch of other places. Maybe even consolidate the two while we're at it (there also might be other, unused Spinners that can be removed).

So this new loader is a splash loader for our app loading screens, not an inline spinner. 
So I would keep `SpinnerLoader` as it is, but I can safely remove `Spinner`

melyndav:
> Shouldn't we update the loading spinner everywhere so that we have a consistent design? We are also using `<Spinner>` and `<SpinnerLoader>` in a bunch of other places. Maybe even consolidate the two while we're at it (there also might be other, unused Spinners that can be removed).

Hey @chmanie as @bassgeta it's just the main page loader (with old UI styling - gradient bg and green teal spinner) that is the scope here. There are smaller inline spinners placed in the motions side-panel and other content places but these don't need touching in this PR as they align to the design. Inline example below. We can update these inline ones later on. The focus is just to remove any old UI styling. 

image

chmanie:
I was actually talking about the SpinnerLoader which we use in a lot of places and it looks like this
image

This for example is the loading screen for the Colony extensions page.

melyndav:
> I was actually talking about the SpinnerLoader which we use in a lot of places. Maybe it's just in the old design, but we should probably double check.
> 
> Here's for example the Spinner for the Colony Preview Page (splash page):
> 
> image
> 
> We can however create another PR for this, I guess

@bassgeta please include this one in the update as well. Anything that has the old UI styling should be included here. Good fine @chmanie.

melyndav (approved):
Everything looks good on my end, I tried to check all the places so hopefully we've got all the cases. Thanks for getting this one up today @bassgeta!

arrenv (approved):
Nice @bassgeta. Approving for now, but we need to re-review after https://github.com/JoinColony/colonyCDapp/pull/1420 has been merged.

image

melyndav (approved):
Approved! Let's get this baby merged @bassgeta!. Nice work.

src/components/frame/LoadingTemplate/LoadingTemplate.css
```diff
@@ -3,7 +3,7 @@
   position: absolute;
```
chmanie, line 3: I also would have liked to see this being moved over to tailwind, but we can tackle that another time.

src/components/frame/LoadingTemplate/LoadingTemplate.tsx
```diff
@@ -55,49 +55,36 @@ const LoadingTemplate = ({ children, loadingText }: Props) => {
     return () => clearTimeout(failedTimer);
   }, []);
 
-  const { formatMessage } = useIntl();
+  const getLoadingDescription = () => {
+    if (loadingState === 'delayed') {
```
chmanie, line 59: I think this would be a good case for an enum.

src/components/v5/common/PageLoader/PageLoader.css
```diff
@@ -0,0 +1,23 @@
...
+  0% {
+    transform-origin: 50% 50%;
+  }
+
+  100% {
+    transform: rotate(360deg);
+  }
+}
+
+.pageLoaderCircle {
```
chmanie, line 11: At least these should be tailwind, shouldn't they?

src/components/frame/v5/pages/OnboardingPage/OnboardingPage.tsx
```diff
@@ -11,7 +11,7 @@ import { LANDING_PAGE_ROUTE } from '~routes';
 import { formatText } from '~utils/intl';
 import CardConnectWallet from '~v5/shared/CardConnectWallet';
 import NotificationBanner from '~v5/shared/NotificationBanner';
-import Spinner from '~v5/shared/Spinner';
```
chmanie, line 14: Are you sure we're not using this anywhere else? I will believe you if you have checked and the answer is yes ;)

chmanie (commented):
Great work, a few comments, I do have though :)

src/components/frame/v5/pages/OnboardingPage/OnboardingPage.tsx
bassgeta: image
I double checked, the `Spinner` was just wrapping `SpinnerLoader` and applying a large size for the colony preview page and user onboarding loader :+1:

chmanie (approved):

jakubcolony (dismissed):
Good to go!

ArmandoGraterol (dismissed):
LGTM 👍 Good work with the loader! It's nice

jakubcolony (approved):

melyndav (approved):
Great work!! Approved :)

PR #1458, opened 2023-12-02
Build, Setup and Wire up, Auth Logic

rdig:
This PR puts in place a authentication proxy service that lives between the client (CDapp) and the GraphQL API (Amplify/AppSync)

This is done via the help of the [CDapp Authentication Proxy](https://github.com/JoinColony/colonyCDappAuthProxy)

The basic way this works is that when a request comes in, if it's a query (or one of the whitelisted mutations) it just goes through and forwards the response from the GraphQL API, while, if it's a mutation, it checks if the user is authenticated (see below) and if it is, it verifies if the specific user can indeed execute those mutation, at which point it either forwards the response again, or denies it.

```
   --------4--------
   v               |
Client --1---> Auth Proxy ---2--> GraphQL 
                   ^                |
                   --------3---------
```

### Authentication

Authentication is done using [SIWE](https://login.xyz/) and it's pretty similar to the original JWT system we had in the Dapp, just a bit more modernized.

The way it works is you first get a nonce _(which is more a token rather than a nonce)_, you compose it into a message object, you sign it with your wallet, then send both the message and the signature to the server.

On the server side, the message is compared against the nonce and signature and your wallet address is decoded. That is then being stored in a session cookie and the cookie is being sent back to the client to use for all the future requests.

Screenshot from 2023-12-05 01-35-54

As long as you'll have proper authentication, you will see a cookie, named `auth` inside your domain's cookie store:

Screenshot from 2023-12-05 01-35-16

### Proxy Service

Locally, the authentication proxy runs on `http://localhost:3005` on which you'll have a couple of endpoints:
- `/health` simple endpoint, returns the health status of the service _(useful for external monitoring)_
- `/check` will give you your authentication status, based on the cookie you send along with the request
- `/nonce` will generate a nonce for you to sing
- `/auth` send your message, and signature to create your session and cookie for authentication
- `/deauth` call it if you want to destroy your local session and expire your cookie _(useful for logging out)_
- `/graphql` the actual proxied endpoind which will respond, barring operations you are not allowed to execute, the same way the Amplify/AppSync endpoint responds

Screenshot from 2023-12-05 02-04-28

### Operation Restrictions

All queries are allowed to pass by default, as well as some [whitelisted mutations](https://github.com/JoinColony/colonyCDappAuthProxy/blob/master/src/routes/graphql/mutations.ts#L127-L135).

Everything else, will require, besides having a valid authentication session, a combination of correct wallet address or permissions to execute certain mutation operations.

You can see the list, as well as read more into how they were set up in here: https://github.com/JoinColony/colonyCDappAuthProxy/blob/master/src/routes/graphql/mutations.ts#L21-L139

### Client

On the client the changes are not extensive. Some logic was added to trigger signing when a wallet is connected and deauthentication when the wallet is disconnected.

Screenshot from 2023-12-05 02-07-13

For the dev environment a new docker image has been added to the orchestration, the one that starts the `auth-proxy`.

Alongside that, the CDapp has now be linked directly to the proxy endpoint, meaning you'll always go though autentication via the client.

While the above is true, that doesn't mean you can't directly manipulate the database / graphql api via the old `http://localhost:20002/graphql` playground.

Because of this change, one `env` file variable has changed: `AUTH_PROXY_ENDPOINT` to `http://localhost:3005` _(note that there's no trailing `/graphql` path)_ while `AWS_APPSYNC_KEY` was removed completely as it's no longer needed. **Please make sure to update your `.env` file accordingly.**

### Lambda Functions / Block Ingestor

The authentication proxy and it's mechanism is only relevant for the client, the CDapp, every other service we run, both server and serverless will still use a `API_KEY` to connect directly to Amplify/AppSync and run mutations and queries as they've always had.

### Future

This was always meant to be a stepping stone towards better authentication. Once Amplify supports local `AWS_LAMBDA` authentication configuration and Javascript based resolvers we can transition the authentication to it an have native control over our schema and mutation. 

Once we can do that, we can remove the authentication proxy entirely as it has the potential to be a bottleneck once user traffic picks up.

### Honorable Mentions

A massive shout-out goes to @area for providing a ton of help and input during the development of the [CDapp Authentication Proxy](https://github.com/JoinColony/colonyCDappAuthProxy) service


src/redux/sagas/setupUserContext.ts
```diff
@@ -21,6 +21,7 @@ import { getGasPrices, putError } from './utils';
 import { getBasicWallet, getWallet } from './wallet';
 // import vestingSagas from './vesting';
 import getOnboard from './wallet/onboard';
+import { authenticateWallet } from '~auth';
```
thomassdk, line 24: import ordering

thomassdk (dismissed):
Any idea why I am getting this error?

image

However this doesnt seem to affect anything and it works great :star:

bassgeta (approved):
Code wise looks great, login and logout is more predictable, Metamask pops up for login, LGTM :tada:

PR #1469, opened 2023-12-04
Fix: create new team validation

krawczykkatarzyna:
https://tw.auroracreation.com/app/tasks/15687546

- validation on the Team color field
- fix warnign with missing keys

- fix fields size
Screen Shot 2023-12-06 at 11 46 17

- fix state after creating a team 
Screen Shot 2023-12-06 at 17 36 59


arrenv (changes requested):
@krawczykkatarzyna Thank you. I noticed these while reviewing.

1. The validation on the name is misaligned, the number limit should not break the layout:
image

2. The width of the fields is not using the full width:
image

3. We actually need to have validation on the Team color field as well. This is the error without it:
image

4. I saw this error in the console when creating a new team. I'm not sure why, but did not get the same in `master`.
image


5. The state after creating a team does not show the team details, title or description and the created in is empty and the buttons are still showing in the footer:
image
image

ArmandoGraterol (changes requested):
Nice work! For the motion it's showing the summary correctly but for the _action_ summary it's still missing data like Arren pointed out, other than that. I didn't find anything weird

krawczykkatarzyna:
@arrenv @ArmandoGraterol  fixed the case when the Permission decision method is used. 
Regarding to buttons section - it's global issue in every action types. As I spoke with @AdrianDudko he mentioned that will be separate task to fix that. And I checked the API doesn't return an action title ( for permission decision method ), it's also issue in every action types. 

Screen Shot 2023-12-07 at 17 14 50

arrenv (changes requested):
Hi again @krawczykkatarzyna,

I noticed the team colour icon is missing the change to red in the error state. 
2023-12-07 18_18_38-Planet Express _ Colony - Brave

krawczykkatarzyna:
@arrenv color icon fixed. I didn't use icon from phosphor-react because the component where the icon is passed not supported that.

Screen Shot 2023-12-07 at 18 40 20

ArmandoGraterol (dismissed):
Looks good to me 👍 Nice work

arrenv (dismissed):
Working great, nice work @krawczykkatarzyna !!

src/components/frame/Extensions/layouts/hooks.tsx
```diff
@@ -357,48 +357,52 @@ export const useGetTxButtons = () => {
...
-            })}
-          >
-            <Icon name="white-tick" appearance={{ size: 'tiny' }} />
-          </span>
-        }
-        data-openhubifclicked // see UserReputation for usage
-      />
-    ),
-  ];
+  const txButtons = (
```
bassgeta, line 360: I'd much rather see this be a component instead of a hook returning JSX so we avoid antipatterns

src/components/frame/Extensions/layouts/hooks.tsx
```diff
@@ -357,48 +357,52 @@ export const useGetTxButtons = () => {
...
+    <>
+      {groupState === TransactionGroupStates.SomePending && (
+        <TxButton
+          key="1"
```
bassgeta, line 364: do we need keys here since it's no rendered in an array?

src/components/v5/common/Fields/InputBase/InputBase.tsx
```diff
@@ -16,13 +18,15 @@ const InputBase = React.forwardRef<HTMLInputElement, InputBaseProps>(
...
       prefix,
+      value,
       suffix,
       mode = 'primary',
       disabled,
       stateClassNames: stateClassNamesProp,
       autoWidth = false,
       label,
       id: idProp,
+      maxLength,
```
bassgeta, line 29: do we want to add `maxLength` to the input too?

src/components/v5/common/Fields/TextareaBase/TextareaBase.tsx
```diff
@@ -17,8 +19,8 @@ const TextareaBase = React.forwardRef<HTMLTextAreaElement, TextareaBaseProps>(
       value,
       stateClassNames: stateClassNamesProp,
       wrapperClassName,
-      maxCharNumber,
       message,
+      maxLength,
```
bassgeta, line 23: do we want to add maxLength to the textarea too?

bassgeta (changes requested):
A few things, but otherwise good job! If you want to get these merged, we can open an issue for my comments and tackle it later on

src/components/frame/Extensions/layouts/hooks.tsx
krawczykkatarzyna: @bassgeta yes, otherwise in browser console warning appears with missing key, this component is used in several places and in one of theme them it's rendered as an array.

src/components/v5/common/Fields/InputBase/InputBase.tsx
krawczykkatarzyna: @bassgeta  yes, it's necessary to display message error

src/components/v5/common/Fields/TextareaBase/TextareaBase.tsx
krawczykkatarzyna: @bassgeta  yes, the same as above

src/components/v5/common/Fields/InputBase/InputBase.tsx
bassgeta: sorry, misunderstaning, I mean that if we could add the maxLength to the `<input />` element as an HTML property too

src/components/v5/common/Fields/TextareaBase/TextareaBase.tsx
bassgeta: sorry, misunderstaning, I mean that if we could add the maxLength to the `<input />` element as an HTML property too

src/components/frame/Extensions/layouts/hooks.tsx
krawczykkatarzyna: @bassgeta 
I suggest to refactor it in separate pull request, not here. This pull request is responsible to display error messages and that's all.

bassgeta: makes sense yeah

src/components/v5/common/Fields/InputBase/InputBase.tsx
krawczykkatarzyna: @bassgeta  when we add `maxLength` also to <input /> element then information about min/max characters will appears only when value in the input will be empty. So the question is should this work in that way?

bassgeta: I see, well I don't know what the UX should be like, but it's pretty normal to show a character counter underneath fields with a maximum length, and disabling further input when you hit it. If that is not how our UX works, then we can not add the `maxLength` attribute onto the HTML tags

krawczykkatarzyna: yes I aggree with you.
but regarding to this task and the design it looks like user will able to add more characters so as I understand we can't add `maxLength` attribute onto the HTML tags 
https://tw.auroracreation.com/app/tasks/15687546
Screen Shot 2023-12-13 at 13 52 04

bassgeta: alright if we don't show the max length before it's an error, then this behaviour is correct, I'll resolve both comments

krawczykkatarzyna: @bassgeta are we agree to merge this pull request?

bassgeta (dismissed):

melyndav (changes requested):
Hey @krawczykkatarzyna I found some issues with this as you can see in the screenshot below. 

image

1. I can break the side panel by entering too much copy into the fields. This shouldn't happen. 

2. When I click out of the team name there is no error validation. But the team purpose has this for max char limits. These two fields should have the same behaviour so it's consistent for the user. 

3. The transaction is now not completing and is stuck in a continual loading state. I'm seeing this error in my console. 

image

src/components/v5/common/ActionSidebar/partials/ActionSidebarContent/ActionSidebarContent.tsx
```diff
@@ -178,19 +183,20 @@ const ActionSidebarContent: FC<ActionSidebarContentProps> = ({
...
-              md:border-l-gray-200
-              bg-gray-25
-            `}
+            w-full
+            md:w-[35%]
+            md:h-full
+            md:overflow-y-auto
+            md:flex-shrink-0
+            px-6
+           y p-8
```
bassgeta, line 192: y the y :eyes: ?

bassgeta (changes requested):

src/components/v5/common/ActionSidebar/partials/ActionSidebarContent/ActionSidebarContent.tsx
krawczykkatarzyna: @bassgeta fixed

bassgeta (dismissed):

krawczykkatarzyna:
@melyndav @bassgeta  reported issues fixed

src/components/v5/common/ActionSidebar/partials/ActionSidebarContent/ActionSidebarContent.tsx
```diff
@@ -83,7 +83,7 @@ const ActionSidebarFormContent: FC<ActionSidebarFormContentProps> = ({
           `}
           message={false}
         />
-        <div className="text-gray-900 text-md flex gap-1">
+        <div className="text-gray-900 text-md flex gap-1 break-all">
```
bassgeta, line 86: do we want break-all or our custom `break-word` here?

bassgeta (approved):

src/components/v5/common/ActionSidebar/partials/ActionSidebarContent/ActionSidebarContent.tsx
krawczykkatarzyna: yes, otherwise when the content of header is too long size of sidebar is broken

melyndav (approved):
All the fixes look great @krawczykkatarzyna this is approved from my side!

PR #1473, opened 2023-12-05
fix: transfer funds form missing fields added

joanna-pagepro:
https://tw.auroracreation.com/app/tasks/15687541


arrenv (changes requested):
@joanna-pagepro  Mostly looks good, a few things I picked up.

1. The value is not correct in the transaction state:
image

2. The Created In is not needed for the Permissions decision method, as you can see in the before and after transaction, it does not do anything:
Before transaction:
image

After transaction:
image

3. `Recipient` is spelled wrong here:
image

joanna-pagepro:
@arrenv 
1. Will be fixed here: https://github.com/JoinColony/colonyCDapp/pull/1483
2. The `created in` field should be hidden now when no `decision method` is selected and when the `permission` decision method is selected
3. fixed

ArmandoGraterol (changes requested):
Nice work addressing the changes! Here's some things I've noticed regarding the sidepanel after the action is created:

This is my panel before creating the motion:
FireShot Capture 262 - Details - Colony - Planet Express - localhost

This one is after:
FireShot Capture 263 - Transaction - 0x0d8dec96b7a263918b4f6c5ed7864078562cd4667db8702c69479_ - localhost

- ~~The amount now has a lot of zeroes for some reason~~ Saw that this is gonna be tackled in another PR
- The `Created In` field is wrong as it's showing that it was created in `Root` and not in `Andromeda`

joanna-pagepro:
@ArmandoGraterol I think that `createdIn` is not send anywhere in this case and it's just set to Root domain all the time, and that's why it's not updated here. @ArmandoGraterol @arrenv can you check this? I think that it should be added to `src/redux/sagas/motions/moveFundsMotion.ts`, but I can't find it.

joanna-pagepro:
@arrenv @ArmandoGraterol issue with created in field fixed, please check this PR again

jakubcolony (approved):
LGTM, nice work

jakubcolony (approved):

arrenv (approved):
All looks good to me, thanks @joanna-pagepro

PR #1483, opened 2023-12-05
Generate description metadata from action data when present, hide createdIn when empty, align edit team payload

TomaszChmielPagepro:
https://tw.auroracreation.com/app/tasks/15687538

image
image

### DIFF
--- a/src/components/shared/Fields/Form/Form.tsx
+++ b/src/components/shared/Fields/Form/Form.tsx
@@ -1,5 +1,10 @@
 import { yupResolver } from '@hookform/resolvers/yup';
-import React, { useCallback, useEffect, useImperativeHandle } from 'react';
+import React, {
+  useCallback,
+  useEffect,
+  useImperativeHandle,
+  useState,
+} from 'react';
 import {
   useForm,
   UseFormProps,
@@ -11,7 +16,8 @@ import {
 import { Schema } from 'yup';
 
 import AdditionalFormOptionsContextProvider from '~context/AdditionalFormOptionsContext/AdditionalFormOptionsContext';
-import { AdditionalFormOptionsContextValue } from '~context/AdditionalFormOptionsContext/types';
+import { AdditionalFormState } from '~context/AdditionalFormOptionsContext/types';
+import { merge, omit } from '~utils/lodash';
 
 const displayName = 'Form';
 
@@ -35,7 +41,7 @@ export interface FormProps<FormData extends FieldValues> {
   validationSchema?: Schema<FormData>;
   defaultValues?: UseFormProps<FormData>['defaultValues'];
   mode?: UseFormProps<FormData>['mode'];
-  options?: UseFormProps<FormData> & AdditionalFormOptionsContextValue;
+  options?: UseFormProps<FormData> & AdditionalFormState;
   /** Pass true to reset the default values to latest values on form submission. This will reset the isDirty prop. */
   resetOnSubmit?: boolean;
   className?: string;
@@ -55,7 +61,8 @@ const Form = <FormData extends FieldValues>(
   }: FormProps<FormData>,
   ref: React.ForwardedRef<UseFormReturn<FormData, any, undefined>>,
 ) => {
-  const { readonly, ...formOptions } = options || {};
+  const [formState, setFormState] = useState<AdditionalFormState>({});
+  const { readonly, ...formOptions } = merge({}, options || {}, formState);
   const formHelpers = useForm({
     resolver: validationSchema ? yupResolver(validationSchema) : undefined,
     defaultValues,
@@ -73,6 +80,14 @@ const Form = <FormData extends FieldValues>(
   } = formHelpers;
   const values = watch();
 
+  useEffect(() => {
+    if (isSubmitting) {
+      setFormState((prevOptions) => ({ ...prevOptions, readonly: true }));
+    } else {
+      setFormState((prevOptions) => omit(prevOptions, 'readonly'));
+    }
+  }, [isSubmitting]);
+
   /*
    * Effect resets default values to latest values, which resets the isDirty prop.
    * Useful in user settings.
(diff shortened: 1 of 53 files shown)


joanna-pagepro (dismissed):

jakubcolony:
@TomaszChmielPagepro I think this needs a rebase

TomaszChmielPagepro:
@jakubcolony Reabsed

src/components/v5/common/ActionSidebar/partials/ActionSidebarContent/ActionSidebarContent.tsx
```diff
@@ -51,25 +52,25 @@ const MSG = defineMessages({
...
     formState: {
+      isSubmitting,
       errors: { title: titleError },
     },
   } = useFormContext();
-
   const { noPermissionsError } = usePermissionsValidation();
   const { noReputationError } = useReputationValidation();
 
+  useDisableFormOnSubmit();
```
jakubcolony, line 72: Instead of a separate hook, could this be an option on `ActionForm` component for example?

src/components/v5/common/ActionSidebar/partials/forms/EditTeamForm/utils.tsx
```diff
@@ -33,17 +37,34 @@ export const editTeamDescriptionMetadataGetter: DescriptionMetadataGetter<
...
+  {
+    description,
+    domainColor,
+    domainPurpose,
+    teamName,
+    decisionMethod: decisionMethodValue,
+  }: EditTeamFormValues,
+  domain: Domain,
+) => {
+  const payload: ActionPayloadWithoutCustomTitle<ActionTypes.ACTION_DOMAIN_EDIT> =
```
jakubcolony, line 49: This shouldn't omit custom title

src/hooks/useActivityFeed/helpers.ts
```diff
@@ -19,7 +19,7 @@ const getActivityFeedMotionState = (
 
   const networkMotionState = motionStatesMap.get(action.motionData.motionId);
 
-  return networkMotionState
+  return networkMotionState !== undefined && networkMotionState !== null
```
jakubcolony, line 22: What's the reason for making this check stricter?

jakubcolony (changes requested):
@TomaszChmielPagepro thanks for your work on this PR, I think we don't want to omit the custom title anymore so could you please change that?

src/components/v5/common/ActionSidebar/partials/forms/EditTeamForm/utils.tsx
TomaszChmielPagepro: The title is added by parent transform here https://github.com/JoinColony/colonyCDapp/blob/f643066e6b3d1ff80d3b29804a4e06685ebef891/src/components/v5/common/ActionSidebar/hooks/useActionFormProps.ts#L40

I have omitted it as it's required in sagas but not mandatory in transforms for each of the forms (in fact it would be overwritten by parent transform). I can make it optional but I think it's better to have it omitted here to avoid confusion. You can still get the original payload type by using `ActionPayload` type.

src/hooks/useActivityFeed/helpers.ts
TomaszChmielPagepro: Probably this wouldn't cause any issues but there is a motion state with a value `0` which would cause this condition to return undefined. If this was intended let me know
image

src/components/v5/common/ActionSidebar/partials/ActionSidebarContent/ActionSidebarContent.tsx
TomaszChmielPagepro: Moved to `Form` component

ArmandoGraterol (commented):
I'm still seeing the `Created In` field when I use `Permissions` as the method

arrenv:
@TomaszChmielPagepro Is this ready for review? It may need a rebase.

src/context/AdditionalFormOptionsContext/types.ts
```diff
@@ -1,3 +1,9 @@
-export interface AdditionalFormOptionsContextValue {
+import { Dispatch, SetStateAction } from 'react';
+
+export interface AdditionalFormState {
   readonly?: boolean;
 }
+
+export interface AdditionalFormOptionsContextValue extends AdditionalFormState {
+  setFormState: Dispatch<SetStateAction<AdditionalFormState>>;
```
bassgeta, line 8: let's name this `updateFormState` and let it receive the form state, and not do direct react setters

bassgeta (changes requested):

src/components/shared/Fields/Form/Form.tsx
```diff
@@ -100,7 +117,7 @@ const Form = <FormData extends FieldValues>(
   );
 
   return (
-    <AdditionalFormOptionsContextProvider value={{ readonly }}>
+    <AdditionalFormOptionsContextProvider value={{ readonly, setFormState }}>
```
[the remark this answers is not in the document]
joanna-pagepro: What is the reason for that?

src/components/v5/common/ActionSidebar/partials/Motions/types.ts
```diff
@@ -11,3 +9,7 @@ export enum CustomStep {
 }
 
 export type Steps = MotionState | CustomStep;
+
+export type MotionsProps = Omit<UseGetColonyActionReturnType, 'action'> & {
```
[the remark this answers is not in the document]
joanna-pagepro: Why is this one wrong?

src/hooks/useActivityFeed/helpers.ts
jakubcolony: Yes it's fine for it to return undefined

jakubcolony (changes requested):
Generally all looking good, it's just the additional context that should be simplified further/removed.

src/components/v5/common/ActionSidebar/partials/Motions/types.ts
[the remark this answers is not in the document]
bassgeta: I wouldn't extend a hook's interface directly, it's spaghettifying the codebase. Imo it's better to just use what's necessary for the component to work, otherwise we're coupling `useGetColonyAction`'s internal workings to the `Motions` component

PR #1487, opened 2023-12-05
Motion styling fixes #2

### LINKED ISSUE #1480: Motions UI issues - #2
## Description

When QA'ing motion flows, a number of UI issues have been identified that require fixing. These include: 

- [x] Check icon doesn't match the weight in Figma 
image

- [ ] Icons inside both buttons should be 18x18px
image

- [x] Icon isn't correct size in the support pill
image

- [x] Remove the hover on the staking component when it's the only active step.
image

- [x] Space between percentage inside voting widget should be removed. Font should also be 10px.
image

- [x] When a vote is selected, this heading shouldn't show above either the support or oppose buttons
image

- [x] Reveal timer isn't aligned correctly and icons not the correct size as spec'd above
image

- [x] Revealed icon in reveal step should be text-gray900.
image

- [x] Voting widget when completed shows empty padding which should be removed.
image

arrenv:
## Description

Adds various style fixes to motions as listed out here https://github.com/JoinColony/colonyCDapp/issues/1480 including an additional fix for the 0 appearing when the voting threshold progress bar loads.

Resolves #1480

### DIFF
--- a/src/components/v5/shared/ProgressBar/ProgressBar.tsx
+++ b/src/components/v5/shared/ProgressBar/ProgressBar.tsx
@@ -12,6 +12,10 @@ const ProgressBar: FC<ProgressBarProps> = ({
   barClassName,
   className,
 }) => {
+  if (threshold == null || threshold === 0) {
+    return null;
+  }
+
   if (progress > max || progress < 0) {
     throw new Error(`Progress bar value must be between between 0 and ${max}`);
   }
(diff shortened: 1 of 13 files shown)


chmanie (dismissed):
Looks reasonable to me!

src/components/v5/shared/ProgressBar/ProgressBar.tsx
```diff
@@ -12,6 +12,10 @@ const ProgressBar: FC<ProgressBarProps> = ({
   barClassName,
   className,
 }) => {
+  if (threshold == null || threshold === 0) {
```
jakubcolony, line 15: Why non-strict equality? If you want to cover null, undefined and 0 you could consider `!threshold`, unless of course you want to allow empty strings

jakubcolony (changes requested):
Looks good, just a small request regarding the enum

melyndav (dismissed):
Nice work on this @arrenv! Approved.

jakubcolony (approved):
LGTM 👍

rdig (approved):
All good here

rdig:
Since @arrenv is off, I'm going to take over this, fix @jakubcolony 's code suggestion then merge this

PR #1489, opened 2023-12-06
Added webpack dev proxy to avoid CORS errors

TomaszChmielPagepro:
image


joanna-pagepro (approved):

TomaszChmielPagepro:
@rdig I completely disagree with that as you would have to disable CORS for QA env for `localhost` which shouldn't be the case. This allows us to use QA API while developing the application without sacrificing security on QA API side (or any backend in general)

PR #1502, opened 2023-12-06
Updates team colors to match design

arrenv:
## Description

In order to align team colors to match the design and also to reduce accessibility issues with the previously used colors, this PR updates the color at a surface level.

To avoid confusion with the name of `DomainColor` mappings, it would be ideal to update the names to more generic ids which would also need to be done in the block-ingestor.


thomassdk (approved):

melyndav (approved):
Nice work on this one @arrenv

bassgeta (approved):

PR #1503, opened 2023-12-06
Fix annotating transactions on actions

### LINKED ISSUE #1451: Annotating Transactions is Failing on actions
## Steps to reproduce
- Create a Minting tokens or Edit colony details action
- Be sure to include a Description.
- Check the Description after completing the transaction.
- Check the transactions in the UserHub transactions list.

This was noticed on both Minting tokens & Edit colony details actions, however all actions should be tested.

## Expected behaviour
- You should be able to annotate transactions, i.e. fill out the description and it works correctly.

## Actual behaviour
- You are not able to  annotate transactions and thus no descriptions are working.

image

jakubcolony:
## Description

This PR grew a bit unexpectedly, here's a summary of changes:
- You should now be able to annotate all action types
- Custom title should now be stored for all action types
- `colony` field on `ColonyContributor` has been made non-nullable 
- `GetContributorsByAddress` query performance was significantly improved 
- Removed `userColonies` from AppContext, instead created `useJoinedColonies` hook

## Testing

Until #1513 gets merged, the change to make `colony` field non nullable makes the contributors query fail because there's no metacolony stored in the DB. **You need to run this mutation as otherwise no colonies will show:**
```
mutation Delete {
  deleteColonyContributor(input: {
    id: "0x1133560dB4AebBebC712d4273C8e3137f58c3A65_0xb77D57F4959eAfA0339424b83FcFaf9c15407461"
  })   {
    id
  }
}
```

Resolves #1451


src/redux/sagas/actions/unlockToken.ts
```diff
@@ -100,6 +98,8 @@ function* tokenUnlockAction({
       meta,
     });
 
+    setTxHash?.(txHash);
```
TomaszChmielPagepro, line 101: I wonder if this shouldn't be moved down in every saga to make sure we will get complete data from `getColonyAction` right after creating the action.

Also, I have an issue in the local environment where `createColonyActionMetadata` mutation would get stuck in a pending state and never reach `setTxHash?.(txHash);` call. Do you know if this could be related somehow to payloads sent from forms?

image
image

src/components/v5/common/ActionSidebar/partials/forms/EditColonyDetailsForm/utils.tsx
```diff
@@ -14,3 +17,38 @@ export const editColonyDetailsDescriptionMetadataGetter: DescriptionMetadataGett
...
+    colonyDescription,
+    colonyAvatarImage:
+      typeof colonyAvatarImage === 'string' || colonyAvatarImage === null
+        ? colonyAvatarImage
+        : colony.metadata?.avatar,
+    colonyThumbnail:
+      typeof colonyThumbnail === 'string' || colonyThumbnail === null
+        ? colonyThumbnail
+        : colony.metadata?.thumbnail,
+    colonyExternalLinks: externalLinks.map(({ url, linkType }) => ({
```
bassgeta, line 47: just a heads up that this PR and [this PR](https://github.com/JoinColony/colonyCDapp/pull/1530) will probably have conflicts

src/components/v5/common/ActionSidebar/partials/forms/ManageTokensForm/utils.tsx
```diff
@@ -24,3 +28,35 @@ export const manageTokensDescriptionMetadataGetter: DescriptionMetadataGetter<
...
+  colony: Colony,
+  values: ManageTokensFormValues,
+) => {
+  const {
+    selectedTokenAddresses,
+    description: annotationMessage,
+    title,
+  } = values;
+
+  let addresses = selectedTokenAddresses.map(({ token }) => token) ?? [];
```
bassgeta, line 42: maybe we can do just 
```JSX
const tokenAddresses =  [
  ...new Set(
    selectedTokenAddresses.map(({ token }) => token)
      .map((address) => createAddress(address))
      .filter((address) => {
        return (
          address !== ADDRESS_ZERO &&
          address !== colony.nativeToken.tokenAddress
        );
      }),
    ),
  ];
```
we can always fallback to empty array if selectedTokenAddresses can be undefined a la 
`(selectedTokenAddressed || []).map` but as I see it's always defined

src/components/v5/common/ActionSidebar/partials/forms/SimplePaymentForm/utils.tsx
```diff
@@ -77,3 +82,82 @@ export const simplePaymentDescriptionMetadataGetter: DescriptionMetadataGetter<
...
+        amount: amount.amount.toString(),
+        tokenAddress: amount.tokenAddress,
+        networkInverseFee,
+      }),
+      ...payments.map(({ recipient, amount: paymentAmount }) =>
+        getPaymentPayload({
+          colony,
+          recipientAddress: recipient,
+          amount: paymentAmount.amount.toString(),
+          tokenAddress: amount.tokenAddress,
```
bassgeta, line 152: you know best, I'm dropping a comment just in case of a typo, do we need `paymentAmount` instead of just `amount` here?  :)

bassgeta (dismissed):
Just a potential typo, which you can resolve if it's not a typo, and a small comment, looks good otherwise, nice one!

src/components/v5/common/ActionSidebar/partials/forms/ManageTokensForm/utils.tsx
bassgeta:
```JSX
const tokenAddresses =  [
  ...new Set(
    selectedTokenAddresses
      .map(({ token }) => createAddress(token))
      .filter((address) => {
        return (
          address !== ADDRESS_ZERO &&
          address !== colony.nativeToken.tokenAddress
        );
      }),
    ),
  ];
```
would be even a bit more concise :)

src/components/v5/common/ActionSidebar/partials/forms/SimplePaymentForm/utils.tsx
jakubcolony: Yes, `amount` is a property of the `values` argument, whereas `paymentAmount` is a property of one of the `payments` array elements

bassgeta (dismissed):

bassgeta (approved):

src/components/v5/common/ActionSidebar/partials/forms/UnlockTokenForm/hooks.ts
```diff
@@ -37,18 +37,12 @@ export const useUnlockToken = (
     transform: useCallback(
       pipe(
```
rdig, line 38: Here pipe makes some sense, as you're both adding a key/id and mapping the payload

rdig (approved):
Good to go!

PR #1508, opened 2023-12-07
Fix avatar uploader breaking in production

thomassdk:
## Description

This PR fixes the avatar breaking in production that was caused by passing in a boolean to the format message function call


chmanie (commented):
Can you fix the lint error? Then this can be approved!

bassgeta (approved):

PR #1512, opened 2023-12-07
Add "No colonies" state to navigation bar

### LINKED ISSUE #1506: Landing page - UX changes to the wallet connected state
## Description

When the user has first created a profile but has not joined or created any new colonies, the Colony switcher navigation bar shows empty and doesn't provide an enriched experience of what the user is to expect in this feature. 

Currently:
image

For this state, please update the Colony switcher to the following design in Figma: https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-V5?type=design&node-id=17970-46669&mode=design&t=mePF35rAUSW03Qvi-4

1. Update the UI to match
2. Ensure the beta widget shows at the bottom
3. Remove the 'Create new Colony' button since users can not do their freely in beta. 

_**Important note:**_ This state should not show for any user who has not connected their wallet. In this non-connected wallet state, the Colony switcher is disabled and not selected, so please do not change this. Thank you. :) 

If the user has already created or joined a Colony then the state should remain unchanged as shown below:

image

ArmandoGraterol:
This is how the navbars of users with profile but no joined colonies should look like now:

FireShot Capture 264 - Welcome to Colony - localhost

It shouldn't appear in any other case.

Also, we've removed the "Create a colony" button as it won't be functional in the beta.

Resolves #1506


src/components/frame/Extensions/layouts/partials/ColonySwitcherContent/ColonySwitcherContent.tsx
```diff
@@ -14,6 +15,34 @@ import { ColonySwitcherContentProps } from './types';
...
+    defaultMessage: 'Joined colonies',
+  },
+  emptyFilteredStateTitle: {
+    id: `${displayName}.emptyFilteredStateTitle`,
+    defaultMessage: 'No results to display',
+  },
+  emptyFilteredStateSubtitle: {
+    id: `${displayName}.emptyFilteredStateSubtitle`,
+    defaultMessage:
+      "There are no Colony's that match your search. Try searching again",
```
bassgeta, line 34: no colonies instead of no Colony's probably?

bassgeta (approved):
Looks good, if you want to, you can fix the legacy typo too :D

src/components/frame/Extensions/layouts/ColonySidebar.tsx
```diff
@@ -68,11 +68,11 @@ const ColonySidebar = ({ txButtons, userHub }: Props) => {
         content: {
           title: formatText({ id: 'navigation.colonySwitcher.title' }),
           content: <ColonySwitcherContent colony={colony} />,
-          bottomActionProps: {
-            text: formatText({ id: 'button.createNewColony' }),
-            iconName: 'plus',
-            to: CREATE_COLONY_ROUTE,
-          },
```
chmanie, line 75: Can you add a comment containing @BETA to all your commented out code so that we know it will be used again at some point and isn't dead?

chmanie (commented):
Thank you for all the translation improvements!

src/components/frame/Extensions/layouts/partials/ColonySwitcherContent/ColonySwitcherContent.tsx
jakubcolony: I fixed that in #1503

jakubcolony (approved):
Looks & works well, I'll need to make sure not to break it when merging #1503

melyndav (approved):
Smashing work @ArmandoGraterol thank you so much. Approved with no changes.

PR #1513, opened 2023-12-07
Refactor Colony Creation Logic

### LINKED ISSUE #583: Move redirection after creating a colony to the `colonyCreate` saga
This will allow to remove polling logic from ColonyContext by ensuring the user is only redirected once the colony is in the DB.

### LINKED ISSUE #1574: Refactor Colony Creation Logic
Issue placeholder _(meaning this already has a in-work PR, please reference that for more info)_

rdig:
This PR refactors the colony creation logic, by offloading actual colony creation to the [block-ingestor](https://github.com/JoinColony/block-ingestor) using the events emmited by the contracts.

The basic logic is that the client will send out the create colony transaction, as soon as it's generated (before even being mined) we get the transaction's hash and use that to send a mutation to the database containing the upcoming colony's metadata.

On the block ingestor side, it picks up the colony created event, extracts required data from it, then will trigger a mutation (via a lambda). This lambda is quite complex as it encapsulates everything needed to create a colony, it's metadata, it's token, and it's domains.

Below you'll find a simplified diagram of how this works:

Screenshot from 2023-12-08 02-26-49

#### Custom Provider

This PR also adds in a custom provider, `RetryProvider` which is basically either a jsonrpc provider or a web3 provider (depending on the environment) which has the `perform` method replaced with one that attempts to retry in an event of a failure

#### Lambda Functions

This PR adds a new lambda function: `createColonyEtherealMetadata` as well as heavily refactor an existing one: `createUniqueColony`

The first lambda is in charge of temporarily saving all the upcoming's colony metadata to the database, until the block ingestor has the change to confirm and create the new colony (for which it will use this metadata), while the second lambda has been made to encapsulate all the mutations required to initially create a colony (metadata, tokens, domains), which makes it easier to call it from the block ingestor

#### Graphql / Database schema

A few minor changes here, nothing breaking. Besides wiring in the new lambda, it also adds in a new type for the `ColonyMetadata` model, as the `createColonyEtherealMetadata` lambda will use that underlying model to store it's data

#### Saga

The `colonyCreate` saga has been refactored pretty heavily as well, as it's no longer in charge of directly creating the colony, a lot of mutations got removed and cleaned up, in essence making it simpler.

It's now in charge of saving metadata and firing off the required colony creation transactions.

One thing to note here is that I've moved the redirection logic for after a colony is created in this saga. It will wait for a colony to first exist in the database (as it not longer directly controls that), the it will redirect to it's route

#### UI

Because of the above, the final step of the onboarding flow (the one which lists all the transactions) has now a loading spinner that shows up after all the transactions have completed, up until you are redirected to you new colony's page

Screenshot from 2023-12-08 02-46-31

#### IPFS

While at it, I've also refactored the IPFS and Pinata libraries that are used to handle the _(minimal for now)_ IPFS uploads in production. Nothing major, just move files around, and removed logic that isn't needed now, or isn't going to be needed.

#### Related Content

The following repos and code was also changed to accomodate this refactor:

JoinColony/colonyCDappAuthProxy#4
JoinColony/block-ingestor#151

While not directly targeting them, this PR inadvertendly fixes the following _(older)_ issues:
- Resolves #1574
- Resolves #583


arrenv (dismissed):
Great work @rdig, tested the following scenarios:
- Creating a new token (Using Dev wallet) - worked great, however, UI was indicating I could not mint tokens, I could mint though by going around the UI restriction.
- Using an existing token (Using Dev wallet) - worked great.
- Created a new token (Using Metamask) - worked great, however, UI was indicating I could not mint or unlock tokens, I could though by going around the UI restriction.

rdig:
This is back on hold until we figure out the Token Authority set race condition

rdig:
@JoinColony/frontend this and it's related PRs _(block ingestor, auth proxy)_ are now ready for review

src/types/wallet.ts
```diff
@@ -1,11 +1,15 @@
...
-import { providers } from 'ethers';
+
+import RetryRpcProvider from '~redux/sagas/wallet/RetryProvider';
+import { CustomEIP1193Provider } from '~redux/sagas/wallet/ganacheModule';
 
 export type ColonyWallet = BasicWallet | FullWallet;
 
 export interface FullWallet extends WalletState, Account {
-  ethersProvider: providers.Provider | null;
+  // @ts-ignore
```
jakubcolony, line 10: Why do we need that? It's not showing any errors when I remove it

src/redux/sagas/colony/colonyCreate.ts
```diff
@@ -507,6 +366,43 @@ function* colonyCreate({
       yield takeFrom(setOneTxRoles.channel, ActionTypes.TRANSACTION_SUCCEEDED);
     }
 
+    /*
+     * Wait for the colony to exist, then navigate to it.
+     */
+    const colonyExistsSubscrition = yield apolloClient
```
jakubcolony, line 372: typo

src/redux/sagas/colony/colonyCreate.ts
```diff
@@ -507,6 +366,43 @@ function* colonyCreate({
...
+          name: givenColonyName,
+        },
+        pollInterval: 1000,
+      })
+      .subscribe({
+        next: ({ data: { getColonyByName } }) => {
+          const [colony] = getColonyByName?.items ?? [];
+          const { displayName: existingColonyDisplayName } =
+            colony?.metadata ?? {};
+          if (existingColonyDisplayName && navigate) {
```
jakubcolony, line 388: Is checking for colony not being undefined not enough?

jakubcolony (approved):
Great work, created a few colonies without an issue 👍

src/types/wallet.ts
rdig: Can't remember why this is here. I'll investigate.

src/redux/sagas/colony/colonyCreate.ts
rdig: Good eyes :eagle:

src/redux/sagas/colony/colonyCreate.ts
rdig: Yes it is, however I wanted to also check that `navigation` is being passed through to the saga, in the event that UI will get changed

src/types/wallet.ts
rdig: Oh, I do remember now. This was added before I added the types to the `RetryProvider` class and it kept complaining, and you are correct that it's pointless now.

Have removed it.

src/redux/sagas/colony/colonyCreate.ts
jakubcolony: @rdig I was referring just to the fact you're checking for the display name property rather than just the metadata or even the colony itself, is there a particular reason?

jakubcolony (approved):

PR #1514, opened 2023-12-08
Hide invite members button initially

arrenv:
## Description

In order to roll out access to the UI more consistently, we would like to hide the self serve "Invite members" button from the UI.

This PR is intended to do just that.


chmanie (approved):
LGTM

melyndav (approved):
@arrenv nice work! Approved. :)

PR #1526, opened 2023-12-11
feat: filter members by team

joanna-pagepro:
https://tw.auroracreation.com/app/tasks/15684419

### DIFF
--- a/src/hooks/members/useMemberFilters.ts
+++ b/src/hooks/members/useMemberFilters.ts
@@ -13,24 +13,25 @@ import { useSearchContext } from '~context/SearchContext';
 import { ColonyContributor } from '~types';
 import { notNull } from '~utils/arrays';
 import { UserRole } from '~constants/permissions';
+import { useGetSelectedTeamFilter } from '~hooks/useTeamsBreadcrumbs';
 
 const useMemberFilters = ({
   nativeDomainIds,
   filterPermissions,
   filterStatus,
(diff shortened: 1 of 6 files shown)


TomaszChmielPagepro (approved):

src/hooks/members/useMemberFilters.ts
```diff
@@ -58,7 +59,9 @@ const useMemberFilters = ({
 
   const filteredByPermissions = useMemo(() => {
     const databaseDomainIds = new Set(
-      nativeDomainIds.map((id) => `${colonyAddress}_${id}`),
+      selectedTeam
+        ? [selectedTeam.id]
```
chmanie, line 63: I feel like this should probably be of the same format as the other option: `${colonyAddress}_${id}`. What do you think?

joanna-pagepro: @chmanie It contains the same format as you described. It's taken from the colony context. Let me know if it can stay like this, or if you want me to change it anyway.

chmanie: Aaah right I forgot that these are stored in the db that way. Don't mind me then, sorry to interrupt :)

bassgeta (approved):
nice one, readability is improved quite a lot :+1:

jakubcolony (approved):
LGTM, nice work!

PR #1530, opened 2023-12-11
fix: use name and link instead of linkType and url for external links

### LINKED ISSUE #1522: Adding URLs to the Edit Colony Details action type
## Description

Adding URLs via the Social links section in the 'Edit Colony Details' action type is broken. 

When adding the social links to the action, the all content added in the action do not get saved even when the transaction goes through.

If I do not add social links, the action content above saves once the transaction is made.  

image

bassgeta:
## Description

This PR fixes editing of social links.
The issue was that we used `linkType` and `url` as the values, which wasn't the case anymore.
Also fixed a minor CSS alignment on the links modal.

I updated our version of apollo, so we could strip `__typename` on the mutation and so that the `haveExternalLinksChanged` flag isn't always true when editing details. 

**Changes** 🏗
* renamed the attributes accordingly
* updated apollo client version

Resolves  #1522

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/forms/EditColonyDetailsForm/hooks.ts
+++ b/src/components/v5/common/ActionSidebar/partials/forms/EditColonyDetailsForm/hooks.ts
@@ -33,11 +33,7 @@ export const useEditColonyDetails = (
           thumbnail: metadata?.thumbnail,
         },
         createdIn: Id.RootDomain.toString(),
-        externalLinks:
-          metadata?.externalLinks?.map(({ name, link }) => ({
-            link,
-            name,
-          })) || [],
+        externalLinks: metadata?.externalLinks ?? [],
       }),
       [
         metadata?.avatar,
--- a/src/components/v5/common/ActionSidebar/partials/forms/EditColonyDetailsForm/hooks.ts
+++ b/src/components/v5/common/ActionSidebar/partials/forms/EditColonyDetailsForm/hooks.ts
@@ -35,8 +35,8 @@ export const useEditColonyDetails = (
         createdIn: Id.RootDomain.toString(),
         externalLinks:
           metadata?.externalLinks?.map(({ name, link }) => ({
-            url: link,
-            linkType: name,
+            link,
+            name,
           })) || [],
       }),
       [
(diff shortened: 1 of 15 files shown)


src/components/v5/common/ActionSidebar/partials/forms/EditColonyDetailsForm/hooks.ts
```diff
@@ -61,12 +61,10 @@ export const useEditColonyDetails = (
...
-            externalLinks: payload.externalLinks.map(
-              ({ linkType, ...link }) => ({
-                ...link,
-                name: linkType,
-              }),
-            ),
+            externalLinks: payload.externalLinks.map(({ name, link }) => ({
+              link,
+              name,
+            })),
```
jakubcolony, line 67: Do we actually need that map?

jakubcolony (approved):
Looks good, nice work 👍

src/components/v5/common/ActionSidebar/partials/forms/EditColonyDetailsForm/hooks.ts
bassgeta: huh good point, I don't really think we do :D

jakubcolony (dismissed):

src/components/v5/common/ActionSidebar/partials/forms/EditColonyDetailsForm/hooks.ts
chmanie: You only might get extra properties but that shouldn't matter

src/components/v5/common/Fields/RadioButtons/TileRadioButtons/TileRadioButtons.tsx
```diff
@@ -27,7 +27,7 @@ function TileRadioButtons<TValue = string>({
               border
               transition-all
               rounded-lg
-              align-center
+              items-center
```
chmanie, line 30: Nice find

chmanie (commented):
Remove that map and get an approval for free :)

jakubcolony (dismissed):

rdig (approved):
Yup! :+1:

jakubcolony (approved):

ArmandoGraterol (approved):
LGTM 👌

PR #1531, opened 2023-12-11
Store display preferences in local storage

### LINKED ISSUE #1168: User Profile Action Window Width
There is a setting in the user preferences that allows users to set the width of the actions panel as a full width element. This is a very low priority feature, but the idea would be to just store it in local storage. The option is in the user account area, preferences.

Design: https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-V5?type=design&node-id=2042%3A204672&mode=design&t=zNJdFiglhEoGd2Lw-1

image

It does the same thing as when you click this full screen icon, but instead loads it as default.

image

https://www.notion.so/colony/Will-s-Specs-f92c55ff64b3424ca6b08b9c4457b876

willm30:
## Description

This PR stores a user's action panel display preferences in local storage, so they can be retrieved by the component whenever it is rendered. 

## Testing

1. Toggle the display preference in user preferences. It should persist even when reloading.

image

2. Open the actions panel, either by adding `?tx={tx_hash}` to the url or directly via the ui. It should open to the width stored in your preferences.

**New stuff** ✨

* Display preferences toggle 

**Changes** 🏗

* If present, use display preference value from local storage for action panel width

Resolves #1168

### DIFF
--- a/src/components/frame/v5/pages/UserPreferencesPage/UserPreferencesPage.tsx
+++ b/src/components/frame/v5/pages/UserPreferencesPage/UserPreferencesPage.tsx
@@ -7,15 +7,14 @@ import Input from '~v5/common/Fields/Input';
 import Button from '~v5/shared/Button';
 import Icon from '~shared/Icon';
 import { useCopyToClipboard } from '~hooks/useCopyToClipboard';
-// @BETA: Disabled for now
-// import Switch from '~v5/common/Fields/Switch';
+import Switch from '~v5/common/Fields/Switch';
 import { multiLineTextEllipsis } from '~utils/strings';
 import LoadingTemplate from '~frame/LoadingTemplate';
 import { LANDING_PAGE_ROUTE } from '~routes';
 import { formatText } from '~utils/intl';
 
 import LeftColumn from '../UserProfilePage/partials/LeftColumn';
-import { useUserPreferencesPage } from './hooks';
+import { useFullScreenMode, useUserPreferencesPage } from './hooks';
 import { UserPreferencesPageProps } from './types';
 import styles from './UserPreferencesPage.module.css';
 
@@ -35,6 +34,7 @@ const UserPreferencesPage: FC<UserPreferencesPageProps> = ({
   const { user, userLoading, walletConnecting } = useAppContext();
   const isMobile = useMobile();
   const { handleClipboardCopy, isCopied } = useCopyToClipboard(5000);
+  const { isFullScreenMode, toggleFullScreenMode } = useFullScreenMode();
 
   const {
     errors,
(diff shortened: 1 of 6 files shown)


src/components/v5/common/ActionSidebar/ActionSidebar.tsx
```diff
@@ -37,8 +38,15 @@ const ActionSidebar: FC<PropsWithChildren<ActionSidebarProps>> = ({
...
   } = useActionSidebarContext();
-  const [isSidebarFullscreen, { toggle: toggleIsSidebarFullscreen }] =
+  const [isSidebarFullscreen, { toggle: toggleIsSidebarFullscreen, toggleOn }] =
     useToggle();
+
+  useLayoutEffect(() => {
+    if (localStorage.getItem(isFullScreen) === 'true') {
+      toggleOn();
+    }
+  }, [toggleOn]);
```
jakubcolony, line 48: It feels like if `useFullScreen` hook became a global one, this could use its return value instead of reading from local storage

src/components/frame/v5/pages/UserPreferencesPage/UserPreferencesPage.tsx
```diff
@@ -173,20 +173,25 @@ const UserPreferencesPage: FC<UserPreferencesPageProps> = ({
...
+          })}
+        </h5>
+        <div className={styles.switchRow}>
+          <LeftColumn
+            fieldTitle={{ id: 'field.display' }}
+            fieldDescription={{ id: 'description.display' }}
+            width=""
+          />
+          <Switch
+            id="display"
```
jakubcolony, line 189: Is this switch referenced by its id anywhere? If not it could be removed

jakubcolony (approved):
LGTM 👍

src/components/frame/v5/pages/UserPreferencesPage/hooks.tsx
```diff
@@ -93,3 +94,24 @@ export const useUserPreferencesPage = () => {
     errors,
   };
 };
+
+export const useFullScreenMode = () => {
+  const [isFullScreenMode, setIsFullScreenMode] = useState(false);
+
+  useLayoutEffect(() => {
```
bassgeta, line 101: could we just pass this into the initial state value?

src/components/frame/v5/pages/UserProfilePage/partials/LeftColumn.tsx
```diff
@@ -5,11 +5,15 @@ import { LeftColumnProps } from '../types';
 
 const displayName = 'v5.pages.UserProfilePage.partials.LeftColumn';
 
-const LeftColumn: FC<LeftColumnProps> = ({ fieldTitle, fieldDescription }) => {
+const LeftColumn: FC<LeftColumnProps> = ({
+  fieldTitle,
+  fieldDescription,
+  width = 'md:w-[16.375rem]',
```
bassgeta, line 11: could we use `className` instead? feels more natural that way

bassgeta (changes requested):

src/components/frame/v5/pages/UserPreferencesPage/hooks.tsx
willm30: @bassgeta Accessing `localStorage` is a side effect and thus belongs in an effect, see https://ui.dev/c/react/effects (around half way down they show this exact example)

bassgeta: sure thing, it's probably not gonna cause issues since we're using `useLayoutEffect`

src/components/v5/common/ActionSidebar/ActionSidebar.tsx
willm30: It is currently a page specific hook since `toggleFullScreenMode` refers specifically to the display preferneces toggle, so don't think we should use it globally.

I could separate the following into its own hook, but it seems worse than just accessing local storage directly:

```
const [isFullScreenMode, setIsFullScreenMode] = useState(false);

useLayoutEffect(() => {
    setIsFullScreenMode(localStorage.getItem(isFullScreen) === 'true');
  }, []);
  
return { isFullScreenMode, setIsFullScreenMode };

```

And would still need the conditional to call `toggleOn`, so not obviously reducing loc either.

src/components/frame/v5/pages/UserPreferencesPage/UserPreferencesPage.tsx
willm30: No but it's used by the input and label tags it wraps. It's also nice for debugging to be able to reference components by some other means than their tailwind classnames

jakubcolony: If you don't pass it explicitly one gets generated for you, hence my question. Currently it doesn't seem this id is required anywhere else

src/components/v5/common/ActionSidebar/ActionSidebar.tsx
jakubcolony: I'm still thinking that you could just copy and paste it into `hooks` outright as it is, the action sidebar doesn't have to use `toggleFullScreenMode` but just read the current setting value. In my eyes you set it in the user preferences page, but it's a global preference

jakubcolony (approved):

bassgeta (approved):

PR #1535, opened 2023-12-11
add search and token filters options

krawczykkatarzyna:
## Description

- [] TODO write description

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

* Changes go here

**Deletions** ⚰️

* Deletions go here

## TODO

- [ ] Add TODOs


melyndav (approved):
Hey @krawczykkatarzyna this looks & works as expected for the UI/UX. So it's approved from my end.Thank you!

src/components/frame/v5/pages/BalancePage/partials/BalanceTable/hooks.tsx
```diff
@@ -171,3 +176,137 @@ export const useGetTableMenuProps = () =>
...
+      }))
+    : [];
+
+  const filters: FilterProps<BalanceTableFilters> = {
+    onChange: setValue,
+    onSearch: setSearchValue,
+    searchValue,
+    value,
+    customLabel: formatText({ id: 'allFilters' }),
+    items: [
```
bassgeta, line 248: maybe this would be better off composed into the `Filters` component. just do
```JSX
<Filters>
  { rootItems.map(({ iconName, items, label, name, title }) => (
    <RootFilter
      key={name.toString()}
      label={label}
      iconName={iconName}
      title={title}
      name={name}
      items={items}
      path={name.toString()}
      value={value}
      onChange={onChange}
    />
  ));
} </Filters>
```

src/components/frame/v5/pages/BalancePage/partials/BalanceTable/hooks.tsx
```diff
@@ -171,3 +176,137 @@ export const useGetTableMenuProps = () =>
       ],
     };
   }, []);
+
+export const useBalanceTable = () => {
+  const { colony } = useColonyContext();
+  const selectedTeam = useGetSelectedTeamFilter();
+  const { balances, nativeToken } = colony || {};
+  const [searchValue, setSearchValue] = useState('');
+  const [value, setValue] = useState<Partial<BalanceTableFilters>>({
```
bassgeta, line 185: is there an option for better types? what are all the possible values? can we use some attribute names, can the filter state be null? `Partial` is quite broad

src/components/frame/v5/pages/BalancePage/partials/BalanceTable/hooks.tsx
```diff
@@ -171,3 +176,137 @@ export const useGetTableMenuProps = () =>
...
+  const sortedTokens = useMemo(
+    () =>
+      tokensData?.sort((a, b) => {
+        if (!a.balance || !b.balance) return 0;
+        return parseInt(b.balance, 10) - parseInt(a.balance, 10);
+      }),
+    [tokensData],
+  );
+
+  const tokenTypeFilters = sortedTokens
```
bassgeta, line 230: we shouldn't return JSX from hooks, the comment I dropped about composition would probably solve this

src/components/frame/v5/pages/BalancePage/partials/BalanceTable/hooks.tsx
```diff
@@ -171,3 +176,137 @@ export const useGetTableMenuProps = () =>
...
+      ({ token: colonyToken }) =>
+        colonyToken?.tokenAddress === token?.token?.tokenAddress,
+    ),
+    isNative: token.token?.tokenAddress === nativeToken?.tokenAddress,
+  }));
+
+  const visibleTokens = tokens?.filter((visibleToken) => {
+    const { type = {}, attribute } = value;
+
+    if (Object.values(type).some((tokenTypeFilter) => tokenTypeFilter)) {
```
bassgeta, line 287: `if (Object.values(type).length > 0)` ?

bassgeta: and imo we need better refined types here, I am not 100% sure what happens here

src/i18n/en.json
```diff
@@ -1453,6 +1453,10 @@
     "balancePage.modal.title": "Deposit funds to the Colony",
     "balancePage.modal.subtitle": "Send funds to the Colony's treasury.",
     "balancePage.send.funds": "Send funds to the Colony",
+    "balancePage.filter.tokenType": "Token type",
```
bassgeta, line 1456: could we please move these to the `BalancePage` component

bassgeta (changes requested):

src/components/frame/v5/pages/BalancePage/partials/BalanceTable/hooks.tsx
krawczykkatarzyna: @bassgeta  this component uses this item props not only for rendering children and is used in several places. If you 'd like to refactor it we can do it in separate pull request.

bassgeta: Could you point me where this happens? I checked and to me it seems like `items` are just used in the `Filter` component to iterate over them and render multiple `RootFilter` items. Do we use the array for anything else?
In any case, I'd prefer this to be refactored before it's merged into master.

src/components/frame/v5/pages/BalancePage/partials/BalanceTable/hooks.tsx
krawczykkatarzyna: we can't change it in line 287 to this condition` if (Object.values(type).length > 0) ?`
Then this condition is true then we will return filtered data, otherwise all data will be returned. When we change to the condition which you suggested then no data will be displayed as default

bassgeta: ah OK my bad, I misunderstood this codesnip then, we want to check if there are any filters that have a value of true

src/components/frame/v5/pages/BalancePage/partials/BalanceTable/hooks.tsx
krawczykkatarzyna: fixed

krawczykkatarzyna:
@bassgeta reported issue have been fixed

bassgeta (approved):

src/components/frame/v5/pages/BalancePage/partials/BalanceTable/hooks.tsx
```diff
@@ -171,3 +234,137 @@ export const useGetTableMenuProps = () =>
       ],
     };
   }, []);
+
+export const useBalanceTable = (): UseFundsTableReturnType => {
+  const { colony } = useColonyContext();
+  const selectedTeam = useGetSelectedTeamFilter();
+  const { balances, nativeToken } = colony || {};
+  const [searchValue, setSearchValue] = useState('');
+  const [value, setValue] = useState<DeepPartial<BalanceTableFilters>>({
```
jakubcolony, line 243: Is it filters value? Please make the naming more descriptive here

src/components/frame/v5/pages/BalancePage/partials/BalanceTable/hooks.tsx
```diff
@@ -171,3 +234,137 @@ export const useGetTableMenuProps = () =>
...
+
+      return visibleToken;
+    },
+  );
+
+  const searchedTokens = visibleTokens?.filter(
+    ({ token }) =>
+      token?.name.toLowerCase().includes(searchValue.toLowerCase()) ||
+      token?.symbol.toLowerCase().includes(searchValue.toLowerCase()),
+  );
```
jakubcolony, line 367: It'd be nice to extract all the filtering logic into helper functions to keep this clean, e.g.:
```ts
const filteredTokens = tokens
  ?.filter(filterTokensByAttribute)
  .filter(filterTokensBySearchValue)
```

src/components/frame/v5/pages/BalancePage/partials/BalanceTable/hooks.tsx
```diff
@@ -171,3 +234,137 @@ export const useGetTableMenuProps = () =>
...
+
+  const sortedTokens = useMemo(
+    () =>
+      tokensData?.sort((a, b) => {
+        if (!a.balance || !b.balance) return 0;
+        return parseInt(b.balance, 10) - parseInt(a.balance, 10);
+      }),
+    [tokensData],
+  );
```
jakubcolony, line 286: This attempts to convert `balance` into a number, but balance will no longer always represent a number after it went through `getFormattedNumeralValue`. Something like 0.0000006 will be formatted into `6 x 10^-7` so the sorting will not work with it. You need to sort using the raw balance, and pass it through something like `BigNumber`, as you're dealing with numbers in excess of 10^18

src/components/frame/v5/pages/BalancePage/partials/BalanceTable/types.ts
```diff
@@ -1,4 +1,5 @@
 import { TokenFragment } from '~gql';
```
jakubcolony, line 1: It should be `Token` from `~types`

src/utils/object.ts
```diff
@@ -0,0 +1,14 @@
...
+export const objectValues = <T extends { [_: string]: any }, K extends keyof T>(
+  obj: T,
+): T[K][] => Object.values(obj);
+
+export const objectEntries = <
+  T extends { [_: string]: any },
+  K extends keyof T,
+>(
+  obj: T,
+): [K, T[K]][] => Object.entries(obj) as [K, T[K]][];
```
jakubcolony, line 14: How is this different from calling `Object.(keys|values|entries)` directly?

src/components/frame/v5/pages/BalancePage/partials/BalanceTable/types.ts
```diff
@@ -8,3 +9,28 @@ export interface BalanceTableFieldModel {
...
+
+export interface TokensProps extends TokensDataProps {
+  isApproved: boolean;
+  isNative: boolean;
+}
+
+export type BalanceTableFilters = {
+  type: {
+    [key: string]: boolean;
+  };
```
jakubcolony, line 26: Storing this as an object feels like overcomplication, why can't it be an array (or a Set) containing token addresses? You can always map that to the same object shape, but we should aim to work with the smallest and simplest form of data.

src/components/v5/shared/Filters/Filters.tsx
```diff
@@ -0,0 +1,411 @@
+import { DeepPartial } from 'utility-types';
```
jakubcolony, line 1: Is putting this component here the right choice? It's not used by anything else other than Balance and Funds table, and we have alternative implementation of table filters in members page, activity feed, etc. Maybe it's better to keep it specific to the balance/funds tables?

jakubcolony (changes requested):
When there are more than 1 tokens selected in the filters, the pill gets truncated and it never shows the entire token (even when you're back to just 1 selected):

Uploading Screen Recording 2023-12-19 at 11.56.06.mov…

In any case, the name of the field that is filtered on shouldn't get truncated, especially if we have plenty of room.

The pill showing filters in the Incoming funds table is broken:

https://github.com/JoinColony/colonyCDapp/assets/112586815/2747fa11-ce23-4fa2-acde-48703018032d

I'm not sure the complexity of Filters component matches what it's supposed to do. Please attempt to simplify the logic and the data models as currently it is very difficult to follow for such a simple feature.

src/components/frame/v5/pages/BalancePage/partials/BalanceTable/types.ts
krawczykkatarzyna: I used this structure because the types require it:

Screen Shot 2023-12-20 at 12 23 38

ArmandoGraterol (changes requested):
Even thought I selected only 1 token type filter, the pill is already truncated:
Screenshot 2024-01-03 at 15 38 07

I selected `Native` and in the pill it shows `Reputation` as if it was also selected:
Screenshot 2024-01-03 at 15 38 24

AdrianDudko:
Filtering & Searching on the Balance Page will be done from scratch by @CzarekDryl.

PR #1538, opened 2023-12-12
fix: native token in table popup

### LINKED ISSUE #1523: Native locked token broken UI
## Description

On the Balance page, the native token of the Colony if locked, shows a lock icon with a tooltip in the table. This icon and the tooltip are still the old UI styling and require updating as part of this PR.

Old UI as shown:

image

Issue scope:

1.  Update the icon. The icon should use the same Phosphor lock icon as the native token widget on the Dashboard.
image

2. Update the tooltip styling to the new tooltip component styling. The content should remain the same.
image

bassgeta:
## Description

Changed the icon so it fits the designs, also had to use a different popup.
Our table rows had `scale(1)` which brought them to the very forefront of our app, thus displaying over the popup.
The hack was introduced due to Safari not handling `position: relative` on table rows, but I think we shouldn't be relying on those, our popups are mostly on single cells or components in there.

**Changes** 🏗

* new icon and tooltip for the native token popup
* removed the scale transform on the entire table row

Resolves #1523

### DIFF
--- a/src/components/frame/v5/pages/BalancePage/partials/TokenAvatar/TokenAvatar.tsx
+++ b/src/components/frame/v5/pages/BalancePage/partials/TokenAvatar/TokenAvatar.tsx
@@ -1,15 +1,17 @@
 import React, { FC } from 'react';
 import clsx from 'clsx';
 
+import { LockKey } from 'phosphor-react';
 import { useMobile, useRelativePortalElement, useToggle } from '~hooks';
 import CopyableAddress from '~shared/CopyableAddress';
-import IconTooltip from '~shared/IconTooltip';
 import TokenIcon from '~shared/TokenIcon';
 import Modal from '~v5/shared/Modal';
 import MenuContainer from '~v5/shared/MenuContainer';
 import Portal from '~v5/shared/Portal';
 
 import { TokenAvatarProps } from './types';
+import Tooltip from '~shared/Extensions/Tooltip';
+import { formatText } from '~utils/intl';
 
 const displayName = 'v5.pages.BalancePage.partials.TokenAvatar';
 // @TODO: implement token popover according to the design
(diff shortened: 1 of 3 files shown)


chmanie (approved):
Yup, that's much nicer.

melyndav (changes requested):
Hey @bassgeta thanks so much for working on this issue.

My only feedback is with the tool tip. Can this be more inline with the component styling so there is consistency across all tooltips? So making it wider and extend out to the right instead of down covering the other tokens in the table. 

Figma component for styling: https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-V5?type=design&node-id=834-186119&mode=design&t=j9QKipjvSqXI1GXM-4

src/components/frame/v5/pages/BalancePage/partials/TokenAvatar/TokenAvatar.tsx
```diff
@@ -52,13 +54,15 @@ const TokenAvatar: FC<TokenAvatarProps> = ({
...
-            />
-          </span>
+          <Tooltip
+            tooltipContent={
+              <span className="sm:w-64 text-left">
+                {formatText({ id: 'tooltip.lockedToken' })}
+              </span>
+            }
+          >
+            <LockKey size={14} className="mt-[1px]" />
```
CzarekDryl, line 64: Please do `mt-px`

rdig (approved):
I'm good here, however please be more descriptive when composing your commit messages:

Screenshot from 2023-12-14 18-40-34

As the last one you did to address review comments isn't descriptive at all, and a year down the line when you look at this commit through some git blame, you will have no idea what Code Review this commit is mentioning.

melyndav (approved):
Thanks for the updates @bassgeta! Approved from my side. :)

PR #1540, opened 2023-12-12
Fix docker permissions for all OSes

area:
## Description

This PR moves from using [bind mounts](https://docs.docker.com/storage/bind-mounts/) to using [volumes](https://docs.docker.com/storage/volumes/). The reason for this is to make permissions easier to solve across multiple environments that we have (primarily between Linux, where `docker` runs as root by default, and OSX, where that is not the case).

**New stuff** ✨

During the build of the development environment, we create volumes (and, if they already exist, empty them). We include an `.initialised` file to help the permissions stick (otherwise, an empty volume seems to have trouble doing that). In the case of the backend volume, we copy what we need in to it. 

Because these files don't exist in the repository in the host file system, we dodge all the permissions issues. The wrinkle it introduces is that some of the files that are generated inside the containers are required by `webpack` for local dev. The options become either expect people to run `webpack` from inside a docker container (which I suspect isn't without its own issues), or add a small server to make those files accessible to the frontend as required. The latter is the option I've chosen here.

**How to test** 🏗

Either check out a new repo, or, probably with `sudo` required if you get errors, `git clean -xdf`. This cleans the repository, including any files that it would normally not touch due to `.gitignore`, which includes those that give rise to these permissions issues. Try to start up the repository, and see if it does so cleanly (in particular, if something with this is wrong, I would expect errors to be coming out of the `amplify` image).

**Still to do** 🖌️
~~The `orchestration-all` docker compose file has not been touched yet, and would need to be updated once we are confident in the solution we have on all development environments we're catering for.~~

**Acceptance criteria** ✅

Once anything in this PR is changed, must test on all of the following following the above procedure, ideally by two different people.

- [x] OSX (Alex)
- [x] Linux (Alex)
- [x] WSL (Chris and Tomasz)
- [x] Rootless docker with Podman (Chris)

~~Leaving this as a draft PR for now, as while it works for me and others on all of the above, it does not currently work for Tomasz from PagePro on WSL, suggesting there remains a subtlety that has not been accounted for. It's therefore not ready to merge, but I think it's certainly in a place where reviews would be useful.~~

It now seems to work for Tomasz.


area:
The `orchestration-all` file has been updated, and I've [extended](https://docs.docker.com/compose/multiple-compose-files/extends/) the `orchestration` file where I could to keep them as maintainable as possible.

area:
This now also includes the work Chris and I have done that was in #1131

src/redux/sagas/wallet/onboard.ts
```diff
@@ -21,15 +22,18 @@ const { formatMessage } = intl({
...
   // variable injected by webpack
   // @ts-ignore
   // if we're using the webpack.dev config, include dev wallets
   if (!WEBPACK_IS_PRODUCTION) {
-    const { private_keys: ganachePrivateKeys } = await import(
-      // @ts-ignore
-      '../../../../amplify/mock-data/colonyNetworkArtifacts/ganache-accounts.json'
-    );
+    const { private_keys: ganachePrivateKeys } = (
+      await axios.get(`${ganacheAccountsUrl.href}ganache-accounts.json`)
```
chmanie, line 35: I _think_ we want to get rid of axios eventually as it's not really needed. For this simple stuff, maybe you could use the `fetch` API?

chmanie (commented):
I tested this many times and it works great! Also in the on-demand environment.

chmanie (dismissed):
Great!

jakubcolony (approved):
All good on MB with Apple chip

jakubcolony (dismissed):

chmanie (dismissed):

rdig (approved):
Good to go

jakubcolony (approved):

Issue #1552, opened 2023-12-13
Add Currency Conversion


willm30:
This can be done after the beta launch.

This adds the ability for users to be able to display converted currencies in the DApp in the currency that they would like. It is an option that is available for the main user dropdown menu. This required us finding a suitable 3rd party API service suitable for getting the conversion to work as required for the supported currencies.

One great bonus would be able to auto detect a users location and if no preference is set for for their preferred currency, we show the our best guess based on location.

The fallback would be to show values in USD.

Supported currencies:

- **USD** - United States Dollar
- **JPY** - Japanese Yen
- **CAD** - Canadian Dollar
- **KRW** - South Korean Won
- **EUR** - Euro
- **INR** - Indian Rupee
- **GBP** - British Pound Sterling
- **BRL** - Brazilian Real
- **ETH** - Ethereum
- **CLNY** - Colony Network Token

Design: https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-V5?type=design&node-id=9707%3A60973&mode=design&t=zNJdFiglhEoGd2Lw-1

![image](https://github.com/JoinColony/colonyCDapp/assets/64402732/0ae64c31-96af-4f3c-b895-a2c097556fc0)

![image](https://github.com/JoinColony/colonyCDapp/assets/64402732/54b91923-c01a-4dc3-91ec-24e98a15b3b8)

Done when:
- [x] There is a UI component in the user menu that allows a user to express a currency preference  #1572
- [x] There is a util function that can fetch the conversion from a given token to a given currency/token #1565
- [x] User location is auto detected and if no preference is set for their preferred currency, we show our best guess based on location. #1594 

For separate issue:

- [x] #1623

rdig:
@willm30 is this Issues still relevant? b/c as far as I can see, _(and other than the #1717 bug)_ everything else that composes this feature was implemented.

That makes me say that this should now be considered complete.

Thank you for your work! 🥃

rdig:
Since now [#1717](https://github.com/JoinColony/colonyCDapp/issues/1717) has been merged, this issue is now complete and I will close it out.

Thank you again for your work here @willm30  _(and @iamsamgibbs)_

PR #1565, opened 2023-12-15
Add currency conversion util

willm30:
## Description

This PR adds a utility for performing conversions from a given token to a given currency.

It will be used around the app to show, for example, the value of a user's balances in USD (or whatever currency they select).

It's using the [CoinGecko](https://www.coingecko.com/api/documentation) api, which does not require an api key, allows 30 calls/minute from the client, is free and supports multichain. 

It does [require attribution](https://www.coingecko.com/en/branding) though, so thoughts on where/how to add that are welcome. @arrenv 

I've made it clear in the code what does and does not make assumptions about the specific api being used, which should make code easier to change if the api gets changed.

## Testing

Mainly code review. If you want, test with #1594 by calling `useCurrency` inside `CurrencyContext` with the parameters for a token you'd like to test the conversion with (e.g CLNY on mainnet), and log out the result.

**New stuff** ✨

* Add token-to-currency conversion utility.

Contributes to #1552

### DIFF
--- a/src/utils/currency/config.ts
+++ b/src/utils/currency/config.ts
@@ -1,0 +1,30 @@
+import { Network, SupportedCurrencies } from '~gql';
+
+// For walkthrough, see: https://apiguide.coingecko.com/getting-started/10-min-tutorial-guide/1-get-data-by-id-or-address
+export const currencyApiConfig = {
+  endpoint: 'https://api.coingecko.com/api/v3/simple/token_price/',
+  searchParams: {
+    from: 'contract_addresses',
+    to: 'vs_currencies',
+  },
+};
+export const coinGeckoMappings = {
+  // This is a map between our internal reference to a supported currency, and the reference the api uses.
+  // For full list: https://api.coingecko.com/api/v3/simple/supported_vs_currencies
+  currencies: {
+    [SupportedCurrencies.Usd]: 'usd',
+    [SupportedCurrencies.Jpy]: 'jpy',
+    [SupportedCurrencies.Gbp]: 'gbp',
+    [SupportedCurrencies.Eur]: 'eur',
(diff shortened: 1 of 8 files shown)


src/utils/currency/config.ts
```diff
@@ -0,0 +1,30 @@
+import { Network, SupportedCurrencies } from '~gql';
+
+// For walkthrough, see: https://apiguide.coingecko.com/getting-started/10-min-tutorial-guide/1-get-data-by-id-or-address
+export const currencyApiConfig = {
+  endpoint: 'https://api.coingecko.com/api/v3/simple/token_price/',
```
rdig, line 5: Initial reaction would be to have this as an `.env` var, but then again, this shouldn't change that often (or based on the environment) to warrant that

rdig (approved):
This looks and acts good. Very nice work!

The only minor gripe I can have is about `fetchData`. Not about it specifically, just that we seem to have various implementation of `fetch` scattered around the codebase, each one with it's own little details or peculiarities. It would be better served if we would have just one fetch `util` and reuse that across everything.

Either way, it's fine for now, and can stay as it is, I've mostly written the above as a note for future reference

arrenv:
> It's using the [CoinGecko](https://www.coingecko.com/api/documentation) api, which does not require an api key, allows 30 calls/minute from the client, is free and supports multichain.

Mel has added a design here for it - https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-V5?type=design&node-id=834-184888&mode=design&t=EJB3s0tzKNing6CL-11

It should sit within the currency selection drop down.

Link should go to `https://www.coingecko.com/` as they require. Should open in a new tab.

ArmandoGraterol (approved):
Solid work 👌

PR #1568, opened 2023-12-15
Activity feed filters

TomaszChmielPagepro:
This PR adds filters to the activity table on the colony activity feed page.

https://tw.auroracreation.com/app/tasks/15669197

image


joanna-pagepro:
@arrenv 
1. Should be ok now
2. I've added a max date to the datepicker, so you shouldn't be able to select a future date
3. Styles updated
4. I couldn't reproduce this bug

arrenv (changes requested):
@joanna-pagepro Thank you for the updates, all looking better now. However, I have trouble testing the filters.

Could you please provide some details on how you are testing this and getting it to work? I consistently am not able to get any filters to actually work in local dev, as it just continuously loads whenever applying a filter or navigating to the second page of the activity table.
image

I have tried restarting the dev environment several times and I still get the same issue.

I also have trouble testing the Status filters, because using Reputation Decision method gives the following error in the console when creating an action with this method:
image

I also noticed this on the incoming funds page in this PR with the alignment of the filter displaying wrong:
image

src/utils/object.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const objectValues = <T extends { [_: string]: any }, K extends keyof T>(
+  obj: T,
+): T[K][] => Object.values(obj);
+
+export const objectEntries = <
+  T extends { [_: string]: any },
+  K extends keyof T,
+>(
+  obj: T,
```
ArmandoGraterol, line 13: What's the advantage to using `.keys`, `.values` and `.entries` like this?

ArmandoGraterol (changes requested):
Massive amount of work!

I've also noticed the same as Arren. Some filters keep the table loading indefinitely, and others don't.

src/components/common/ColonyActionsTable/ColonyActionsTable.tsx
```diff
@@ -115,7 +125,16 @@ const ColonyActionsTable: FC<ColonyActionsTableProps> = ({
       title={formatText({ id: 'activityFeedTable.table.title' })}
       tableComponent={TableWithMeatballMenu}
       tableProps={tableProps}
-    />
+      headerClassName="w-full"
+    >
+      <Filters
```
bassgeta, line 130: I am not 100% sure we should use a reusable component for any filters. The logic can be proprietary, so can be the interfaces.
Could we for starters do filters separately for this table and see if any patterns emerge? 
There are different ways you can go about it, but maybe just create a dropdown in this folder and connect each item with appropriate handlers to modify the filter state.

src/utils/object.ts
bassgeta: +1, we don't need typings for native methods

bassgeta (changes requested):

joanna-pagepro:
Closing this PR, because a lot of changes are required. It will be easier to review and more readable in a new PR

PR #1572, opened 2023-12-18
Add currency conversion menu

willm30:
## Description

This PR adds the menu that enables the user to select their preferred currency for token value conversions.

Captura de pantalla 2023-12-20 164449

Captura de pantalla 2023-12-20 164504

Note that the order of the currencies isn't as specc'd. Is this important? It would mean we'd need to create a mapping between supported currencies and list order, which would create something extra to remember whenever changing the list of supported currencies. I.e. it's simpler without but not difficult to implement. @arrenv 

## Testing
Test with https://github.com/JoinColony/colonyCDapp/pull/1594

You should be able to select a currency using the menu.

**New stuff** ✨

* Menu for selecting preferred currency
* Currency context provider, enabling consumption of this data across entire app

Contributes to #1552


ArmandoGraterol (dismissed):
LGTM 👍

arrenv (approved):
Nice one @willm30, regarding the order, no big issue with it as it is for the moment as long as it is consistent.

Is there anyway to fix the issue with the edges getting cut off.

image

src/images/icons.json
```diff
@@ -267,7 +267,15 @@
...
-    "rocket"
+    "rocket",
+    "us",
+    "uk",
+    "eu",
+    "japan",
+    "brazil",
+    "canada",
+    "southkorea",
+    "india"
```
rdig, line 278: Maybe prepend theses with `flag-`. Ie: `flag-us`, `flag-japan`

rdig (dismissed):
Minor naming nitpick, all good otherwise

ArmandoGraterol (dismissed):

melyndav (dismissed):
A quick check of this one and all looks good @willm30! Nice work.

rdig (approved):
All good again

rdig (dismissed):

rdig (dismissed):

ArmandoGraterol (dismissed):

ArmandoGraterol (dismissed):

arrenv (changes requested):
@willm30 All looks to be functioning well, nice!

Only thing I noticed was doesn't the new env variable `COINGECKO_API_KEY` need to also be added to the `Dockerfile`? - https://github.com/JoinColony/colonyCDapp/blob/cc2baf5affd5ea57f364a169215d26b75d61a79d/Dockerfile#L76

rdig (approved):

PR #1589, opened 2023-12-19
Fix content alignment on colony pages

### LINKED ISSUE #1507: Content alignment issues
## Description

All internal pages (not including the dashboard, landing page, onboarding) have different heights for when the main content starts. As you can see here in this example, the colony details page sits higher than the incoming funds content.

All inside pages **(not including the dashboard, landing page, profile and onboarding pages)** should be consistent. Basically all pages that are accessible from the left navigation menu when a user is inside the content of a Colony and the user account page as well. 

As you can see in Figma, all content should have a consistent 32px padding from the top header:

Please make these pages consistent, so the content doesn't appear to 'jump' around when the user navigates across the app.

To view all inside pages in Figma, you can enter here and view all pages in the column vertically. If you need help navigating this, please reach out for a call - https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-V5?type=design&node-id=9707-52441&mode=design&t=mePF35rAUSW03Qvi-4

ArmandoGraterol:
This PR addresses the colony pages (Any page other than the dashboard, landing page, profile, and onboarding pages) having different spacing between the content and the header. As far as I know, I've only observed this was the case for the incoming funds page, which had an extra padding between the subtitle and the header. Let me know if you guys find that I've missed a place in any of the other colony pages.

Basically, all pages that are accessible through the navigation of a colony, should have a gap of ONLY `32px` between the header and the start of the content.

Resolves #1507

### DIFF
--- a/src/components/frame/v5/pages/ActivityPage/ActivityPage.tsx
+++ b/src/components/frame/v5/pages/ActivityPage/ActivityPage.tsx
@@ -19,7 +19,7 @@ const ActivityPage: FC = () => {
   const widgets = useActivityFeedWidgets();
 
   return (
-    <div className="flex flex-col gap-4 sm:gap-[1.125rem] w-full">
+    <div className="flex flex-col gap-4 sm:gap-6 w-full">
       <WidgetBoxList items={widgets} />
       <div>
         <ColonyActionsTable />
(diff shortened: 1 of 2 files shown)


rdig (approved):
Seems fine

bassgeta (approved):
ez game, nice one :shipit:

PR #1594, opened 2023-12-20
Add currency default by user ip

willm30:
## Description

This PR sets a user's preferred currency based on their location in the event that it has not already been saved in their user profile.

It uses a couple of apis to do so.

1) It makes a request to [ipify](https://www.ipify.org/), which returns the client's ip address. This cannot be gleamed from the `window` object.
2) It then makes a request to [iplocation](https://api.iplocation.net/) to get the country code of the client's ip.

If its country code is supported, it will set its currency as the user's preferred currency, else it falls back to USD.

## Testing

Assuming the user you're using doesn't yet have a `preferredCurrency` saved in the db, you can verify this works. Open up the user menu, and you should see the currency of your country (if you're in the UK or EU). If so, it has worked.

**New stuff** ✨

* Add ip-based default currency functionality

Contributes to #1552


ArmandoGraterol (approved):
In my case, US dollars was the default, so I guess that's close enough :)

rdig (approved):
Seems to be working as expected, however I am a bit worried about the rate limits these two services impose, either directly or indirectly _(technical limits)_

Screenshot from 2024-01-03 18-07-11
Screenshot from 2024-01-03 18-08-14

`ipify` says "no limit" which is either a very high limit, or they have so little traffic that they don't actually know where they're limit is, regardless it makes me worry

`iplocation` doesn't mention a limit at all, meaning it most likely has a very restrictive one

Regardless, no ip service api has "no limit" at all, which makes this likely that we're going to run into this issue somewhere down the line.

I've left the above notes as a warning to us when we'll eventually run into this, however, for now, this can go in as-is

PR #1598, opened 2023-12-20
Tweak member page filtering

### LINKED ISSUE #1583: Members Page Regression
The members page all members tab no longer shows all the members in the colony, but now erroneously shows just the members that also have reputation.

This is not correct, as that should be shown only by the "Contributors" tab, while "All Members"  should show all the members of a specific colony

Screenshot from 2023-12-19 13-44-55
Screenshot from 2023-12-19 13-45-03

This is not a database issue, as the data is available in the members query _(`MembersContext`)_

Screenshot from 2023-12-19 13-53-50

But some hook or util is filtering out members more that it should

Screenshot from 2023-12-19 13-54-22

ArmandoGraterol:
Fixed members without rep or roles disappearing from the `All members` tab. Now they should appear when you are not filtering by a team. The rest of the filters should work as expected (Tested various combinations, let me know if you find any that is not working as expected and I'll fix it in this PR)

Resolves #1583

### DIFF
--- a/src/hooks/members/useMemberFilters.ts
+++ b/src/hooks/members/useMemberFilters.ts
@@ -35,14 +35,56 @@ const useMemberFilters = ({
 
   const { colonyAddress = '' } = colony ?? {};
 
-  const filteredByStatus = useMemo(() => {
-    if (!filterStatus) {
+  const databaseDomainIds = useMemo(
+    () =>
+      new Set(
+        selectedTeam
+          ? [selectedTeam.id]
+          : nativeDomainIds.map((id) => `${colonyAddress}_${id}`),
+      ),
+    [nativeDomainIds, colonyAddress, selectedTeam],
+  );
+
+  // Always include the root domain, since if the user has a permission in root, they have it in all domains
+  const permissionsDomainIds = useMemo(
+    () => new Set([`${colonyAddress}_${Id.RootDomain}`, ...databaseDomainIds]),
+    [databaseDomainIds, colonyAddress],
+  );
+
+  const filteredByTeam = useMemo(() => {
+    if (!selectedTeam) {
       return members;
     }
-    return members.filter(({ isVerified }) =>
+
+    return (
+      members.filter(({ roles, reputation }) => {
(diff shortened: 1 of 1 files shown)


rdig:
This needs rebasing once again, since after https://github.com/JoinColony/colonyCDapp/pull/1540 merged it can no longer be tested

ArmandoGraterol:
It is done Mr. @rdig

bassgeta (dismissed):

src/hooks/members/useMemberFilters.ts
```diff
@@ -34,14 +34,56 @@ const useMemberFilters = ({
...
+          : nativeDomainIds.map((id) => `${colonyAddress}_${id}`),
+      ),
+    [nativeDomainIds, colonyAddress, selectedTeam],
+  );
+
+  // Always include the root domain, since if the user has a permission in root, they have it in all domains
+  const permissionsDomainIds = useMemo(
+    () => new Set([`${colonyAddress}_${Id.RootDomain}`, ...databaseDomainIds]),
+    [databaseDomainIds, colonyAddress],
+  );
```
jakubcolony, line 51: Could we use `getDomainDatabaseId` util here?

jakubcolony: And is the useMemo necessary?

jakubcolony (approved):
Nice work @ArmandoGraterol!

src/hooks/members/useMemberFilters.ts
ArmandoGraterol: I think so, these const are being used by other parts of the code that are memorized so if I remove the `useMemo` here it tells me that it needs it to not re-render the other functions.

jakubcolony: Fair enough if that's the case, although I bet wherever the "useMemo chain" starts it's not needed there either

jakubcolony (approved):

rdig (approved):
Seems to be working good. Nice work Armando

melyndav (approved):
@ArmandoGraterol thanks for fixing this bug! Works as expected. :)

PR #1603, opened 2023-12-21
Fallback to action metadata description when cutom title is empty

TomaszChmielPagepro:
https://tw.auroracreation.com/app/tasks/15701598

Adds a fallback for activity feed/transaction item titles.

Before
image
 After

image

### DIFF
--- a/src/components/common/ColonyActionsTable/partials/ActionDescription/ActionDescription.tsx
+++ b/src/components/common/ColonyActionsTable/partials/ActionDescription/ActionDescription.tsx
@@ -36,6 +36,10 @@ const ActionDescription: FC<ActionDescriptionProps> = ({
     motionState || null,
   );
 
+  const actionMetadataDescription = colony
+    ? formatText({ id: 'action.title' }, getActionTitleValues(action, colony))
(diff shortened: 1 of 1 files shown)


CzarekDryl (approved):

ArmandoGraterol (approved):
With this fallback the feed is looking better:
FireShot Capture 266 - Planet Express - Colony - localhost

Now I wonder if it would be better to hide the meta subtitle when it's already showing as a title, but that may be a question better directed towards Product (@melyndav / @arrenv)

arrenv:
> Now I wonder if it would be better to hide the meta subtitle when it's already showing as a title, but that may be a question better directed towards Product (@melyndav / @arrenv)

It would be good to not have the duplicated lines, however, the hope is that this is not something that happens for all items in the feed and if we hide the duplicated lines only for a couple, where it fails, then it would be inconsistent with the rest.

arrenv (approved):
Great @TomaszChmielPagepro , nice work!

rdig (approved):
Seems fine. :+1:

PR #1605, opened 2023-12-21
Remove `SetTokenAuthority` from the Colony Creation flow

### LINKED ISSUE #1597: Remove `SetTokenAuthority` transaction step from `createColony` saga
It's not needed as, the `createColonyForFrontend` transaction already does this behind the scenes.

This means that the new Colony Creation flow will one less transaction to send

rdig:
This PR removes the extra Token Client call to `setTokenAuthority` as it's already being sent, behind the scenes by the bundled call of `createColonyForFrontend`, meaning the second call changes nothing while incurring gas costs.

A secondary implication here is that due to how the Metatransactions broadcaster works, this call would actually fail each time, since permissions for it would no longer be available due to the first call going through.

Screenshot from 2023-12-21 12-13-28

Resolves #1597

### DIFF
--- a/src/redux/sagas/colony/colonyCreate.ts
+++ b/src/redux/sagas/colony/colonyCreate.ts
@@ -80,7 +80,6 @@ function* colonyCreate({
    * If the user opted to create a token, define txs to manage the token.
    */
   if (tokenChoice === 'create') {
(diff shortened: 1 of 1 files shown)


jakubcolony (approved):
Looks good 👍

PR #1608, opened 2023-12-21
Refactor: Display created actions in separate components 

### LINKED ISSUE #1567: Metadata is showing the Colony creator and not the author of the transaction
## Description

Currently, in production, when I open a transaction from the activity feed, it's showing the Colony creator in the metadata which doesn't match the metadata in the activity which shows the correct username of the user who created the transaction.

image

The expected behaviour is for the action or motion when triggered in the side panel to show the correct user who created the transaction as per the activity feed.

bassgeta:
## Description

I've refactored displaying of already created actions. 
Currently we display a form with a read only property, and we mix and match handling of values depending if we're creating/viewing them.
It's much more predictable to have separated display components.
This PR starts on that, it would be too massive otherwise.

The jist of it is that we now have a component called `CompletedAction` which fetches an action according to the `transactionId` property passed.
After it's received, it displays a different view.
All the different action views are reusing common rows like action type, created in, etc.
These views can also be used after the transaction is successfully created, but first we need to refactor all of them :)

In order to have backwards compatibility, I've created a helper to determine if we should display the new view or the existing one.

**New stuff** ✨

* `CompletedAction` component which displays an action details

**Changes** 🏗

* Instead of always showing a form, I've dislocated display of actions to the `CompletedAction` component and its `partials`
* If action display is not yet supported, we display the old view (we can change this and always display the old one until we refactor all the actions)
* Refactored `SimplePayment` display
* Refactored `MintTokens` display
* Refactored `TransferFunds` display
* Refactored `UnlockToken` display
* Refactored `UpgradeColonyVersion` display
* Refactored `CreateDecision` display
* Refactored `EditColonyDetails` display (need to add social links too)
* Refactored `CreateNewTeam` display

**How to test**
Just create and view actions for
1. simple payment
2. mint tokens
3. transfer funds
4. unlock native token
5. upgrade colony version
6. create decision
7. create new team

Editing of details isn't enabled for now, we need to tackle the social links table in a different PR, since this is getting out of hand size-wise :D

Resolves #1567


src/components/v5/shared/RichTextDisplay/RichTextDisplay.tsx
```diff
@@ -0,0 +1,77 @@
...
+        StarterKit.configure({
+          heading: false,
+        }),
+        Heading.configure({ levels: [1, 2, 3] }).extend({
+          levels: [1, 2],
+          renderHTML({ node, HTMLAttributes }) {
+            // @ts-ignore
+            const level = this.options.levels.includes(node.attrs.level)
+              ? node.attrs.level
+              : // @ts-ignore
```
ArmandoGraterol, line 35: Could you add some comments explaining why are we ignoring ts errors  here?

ArmandoGraterol (changes requested):
A lot of work went into this!

I've noticed that the action types now appear unformatted on the panel after submit:
Screenshot 2024-01-04 at 16 17 48

This is happening for all the actions in the testing list

src/components/v5/shared/RichTextDisplay/RichTextDisplay.tsx
bassgeta: ah sorry I have no idea why we do this, I just moved this component but Git thinks I created it :D I'll check it out though

ArmandoGraterol (dismissed):
LGTM 👌 Thank you for the changes

arrenv (changes requested):
@bassgeta A lot of hard work here, great work. A few things I noticed:

1. The alignment does not match up between the views. This is an issue on all action types.
image

2. The recipient is not showing in the Completed view.
image

3. I received this error when using rich text in the Description field, which crashed the app.
image

4. It took some time for the description to appear. Perhaps we can use a skeleton loader.
description-delay

5. No description appeared when doing a `Transfer funds` action until I reloaded.

6. Long descriptions in the completed view are not showing a minified view with a max of two lines, or allow expanding, as of the design. - https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-v5?type=design&node-id=2890-251758&mode=design&t=R6drl9uZUGmUobaj-11
image

7. The alignment of the description field should be top aligned, as of the design.
image

8. I see this issue when testing Upgrade Colony version. I could not test the completed state for this though.
image

9. Not sure if you changed anything here, but I get an incorrect permission error when creating a decision. So, I couldn't full test.
image

10. Edit colony details had a few things:
  - It did not show the name or description change. (was this raised elsewhere?)
  - Created in still appears, even though it is not needed for permissions.
  - No social links appeared.
  - Add social links button still appeared.
image

11. The description did not appear for `Create new team` action until I refreshed as well.

src/components/v5/common/ActionSidebar/partials/ActionDescriptionMetadata/ActionDescriptionMetadata.tsx
```diff
@@ -0,0 +1,141 @@
...
+  const { colony } = useColonyContext();
+
+  if (!selectedAction || !colony || !user) {
+    return null;
+  }
+
+  const commonActionData: Omit<ColonyAction, 'type'> = {
+    initiatorAddress: user.walletAddress,
+    initiatorUser: user,
+    blockNumber: 0,
```
rdig, line 72: Minimum block number is `1`, but even that is problematic. A more correct approach would be to have this fallback to the first block the current colony has been seen at _(created at)_

src/components/v5/common/ActionSidebar/partials/ActionDescriptionMetadata/ActionDescriptionMetadata.tsx
```diff
@@ -0,0 +1,141 @@
...
+      nativeToken: {
+        ...colony.nativeToken,
+        nativeTokenDecimals: colony.nativeToken.decimals,
+        nativeTokenSymbol: colony.nativeToken.symbol,
+        tokenAddress: colony.nativeToken.tokenAddress,
+      },
+    },
+    colonyAddress: colony.colonyAddress,
+    showInActionsList: true,
+    transactionHash: ADDRESS_ZERO,
```
rdig, line 85: Transaction hash !== address, they're different things altogether. We don't have a default fallback (as we do with the Null Address for addresses) for transactions

src/components/v5/common/ActionSidebar/partials/ActionDescriptionMetadata/ActionDescriptionMetadata.tsx
```diff
@@ -0,0 +1,141 @@
...
+            ),
+          },
+        );
+      }}
+    />
+  );
+};
+
+ActionDescriptionMetadata.displayName = displayName;
+export default ActionDescriptionMetadata;
```
rdig, line 141: Maybe a new line between them ?

src/components/v5/common/CompletedAction/partials/MintTokens/MintTokens.tsx
```diff
@@ -0,0 +1,88 @@
...
+    id: `${displayName}.subtitle`,
+    defaultMessage: 'Mint {amount} {token} by {user}',
+  },
+});
+
+const MintTokens = ({ action }: MintTokensProps) => {
+  const { customTitle = formatText(MSG.defaultTitle) } = action?.metadata || {};
+  const { initiatorUser } = action;
+  const transformedAmount = moveDecimal(
+    action.amount || '0',
```
rdig, line 40: This is `amount || 1` since it's in WEI

src/components/v5/common/CompletedAction/partials/MintTokens/MintTokens.tsx
```diff
@@ -0,0 +1,88 @@
...
+              {initiatorUser.profile?.displayName}
+            </UserPopover>
+          ) : null,
+        })}
+      </ActionSubtitle>
+      <ActionDataGrid>
+        <ActionTypeRow actionType={action.type} />
+
+        <AmountRow
+          amount={action.amount || '0'}
```
rdig, line 67: I think? this is in WEI too ?

src/components/v5/common/CompletedAction/partials/SimplePayment/SimplePayment.tsx
```diff
@@ -0,0 +1,131 @@
...
+              profile: action.recipientUser?.profile,
+              walletAddress: action.recipientAddress || '',
+            }}
+            size="xs"
+            showUsername
+          />
+        </div>
+
+        <AmountRow
+          amount={action.amount || '0'}
```
rdig, line 110: Same

src/components/v5/common/CompletedAction/partials/TransferFunds/TransferFunds.tsx
```diff
@@ -0,0 +1,116 @@
...
+    defaultMessage:
+      'Move {amount} {token} from {fromDomain} to {toDomain} by {user}',
+  },
+});
+
+const TransferFunds = ({ action }: TransferFundsProps) => {
+  const { customTitle = formatText(MSG.defaultTitle) } = action?.metadata || {};
+  const { initiatorUser } = action;
+  const transformedAmount = moveDecimal(
+    action.amount || '0',
```
rdig, line 46: Same as above, in WEI

src/components/v5/common/CompletedAction/partials/TransferFunds/TransferFunds.tsx
```diff
@@ -0,0 +1,116 @@
...
+              <span>{formatText({ id: 'actionSidebar.recipient' })}</span>
+            </div>
+          </Tooltip>
+        </div>
+        <div>
+          <TeamBadge teamName={action.toDomain?.metadata?.name} />
+        </div>
+
+        <AmountRow
+          amount={action.amount || '0'}
```
rdig, line 95: Same

src/components/v5/common/CompletedAction/partials/UpgradeColonyVersion/UpgradeColonyVersion.tsx
```diff
@@ -0,0 +1,76 @@
...
+    id: `${displayName}.defaultTitle`,
+    defaultMessage: 'Updating the Colony version',
+  },
+  subtitle: {
+    id: `${displayName}.subtitle`,
+    defaultMessage: 'Upgrade Colony version by {user}',
+  },
+});
+
+// @TODO connect actual version number
```
rdig, line 33: @area actually raised an issue for this which we'll tackle in the next cycle #1677

rdig (dismissed):
Left some comments, for some stuff that needs to be addressed, but overall this great work.

Very nice cleanup. Kudos! :100:

arrenv (changes requested):
Thank you @bassgeta, nice work, a couple of things noticed.

1. Inconsistent height of rows, which looks like being caused by the pill, other heights are 30px, the row with the team pill has a height of 36px.
image

2. Inconsistency with with creation form and that is inconsistent with design, although yours does better match the design.
image

3. There is no number formatting anymore.
image

4. Inconsistent tooltips on field labels. Can they all be top, as the natural flow is from top to bottom in movement.
image
image

5. No description appeared until after reloading the page when using rich text.

6. There is no hover state on the Description label when hovering, not it should not highlight blue on the collapsed description content as it currently does. Design is as follows: 
image
Figma: https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-v5?type=design&node-id=2890-251758&mode=design&t=qU8bTBetAhPtjaDR-11

7. You are not able to click on the description text to expand it, like you can when creating the action. There is also an expand button that was added.
image

8. There is no collapse option at the bottom of the description when expanded.
image

9. I'm pretty sure this is needing to be handled elsewhere, but, after editing a team, it shows the Colony Objective action. So, I am not able to test the completed state of Edit Colony details.
image

rdig (dismissed):
This is all good again. My concerns addressed. Thank you for your hard work! :tada:

ArmandoGraterol (approved):

arrenv (approved):
Great work @bassgeta!!! Truly outstanding, everything tested is working great.

PR #1618, opened 2023-12-22
block-ingestor update: increase colony metadata query max retries

jakubcolony:
## Description

This PR updates block-ingestor hash to a commit which increases max retries for the colony metadata query in `ColonyAdded` handler. Without it, the create data script tends to fail intermittently.


ArmandoGraterol (approved):
LGTM 👍

arrenv (approved):
Worked great, I had the script fail 3 times on `master`, worked without an issue on this one.

PR #1619, opened 2023-12-22
add missing banner; purple border on mobile;

krawczykkatarzyna:
https://tw.auroracreation.com/app/tasks/15660303
https://tw.auroracreation.com/app/tasks/15678862

Screen Shot 2023-12-22 at 15 35 13
Screen Shot 2023-12-22 at 15 47 13

Changes:
- added missing info in popover about not verified user
- added missing border in modal for top contributor type

### DIFF
--- a/src/components/v5/shared/UserAvatarPopover/UserAvatarPopover.tsx
+++ b/src/components/v5/shared/UserAvatarPopover/UserAvatarPopover.tsx
@@ -14,7 +14,11 @@ import { UserAvatarPopoverProps } from './types';
 
 const displayName = 'v5.UserAvatarPopover';
 
-const UserAvatarPopover: FC<UserAvatarPopoverProps> = ({ size, ...props }) => {
+const UserAvatarPopover: FC<UserAvatarPopoverProps> = ({
+  size,
+  additionalContent,
+  ...props
+}) => {
   const { walletAddress, isContributorsList } = props;
   const {
     colony: { colonyAddress },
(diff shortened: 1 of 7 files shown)


ArmandoGraterol (changes requested):
Good work so far! I've noticed that the banner is missing in some other instances of the popover:
Screenshot 2023-12-28 at 13 14 06

- On the extension page, when you click on the user that installed the extension
- On the action side panel's overview of what member executed the action with what permission

src/components/v5/shared/UserAvatarPopover/UserAvatarPopover.tsx
```diff
@@ -40,6 +46,17 @@ const UserAvatarPopover: FC<UserAvatarPopoverProps> = ({ size, ...props }) => {
       className={clsx({
         skeleton: !user,
       })}
+      additionalContent={
+        additionalContent || (!isVerified && showVerifiedBox) ? (
```
ArmandoGraterol, line 50: If every place where the user popover is used must look the same (according to what Product wrote in the task at least), then I don't think we need this `showVerifiedBox` as the verified box should always show if the user is not verified.

ArmandoGraterol (commented):
The "Not verified" banner is still missing in this popover:
Screenshot 2024-01-09 at 13 34 42

rdig (changes requested):
Code-wise this seems fairly good, however not all the banners show up _(see Armando's comments)_

Once those are addressed, I'll re-review this.

ArmandoGraterol (approved):
Looks good to me 👍 All popovers seem to be appearing as expected

jakubcolony:
@CzarekDryl can you please rebase on latest master to fix the `build` check, after which it should be good to merge.

PR #1624, opened 2023-12-28
Wire currency conversion util

### LINKED ISSUE #1623: Wire in currency conversion util
This issue tracks the locations in the app that take advantage of the currency conversion utility.

-[ ] Colony funds
image
image
image

willm30:
## Description

This PR wires in the currency conversion util in all places it's needed in the app.

Using coingecko's free api key ensures you get a stable rate limit of 30 calls/minute. Create a [demo account](https://www.coingecko.com/) to get your own or add mine to your `.env` file: `CG-d4y9hVCNMRU53fFfmV7AiyKr`

Production may need higher call / rate limits (see https://www.coingecko.com/en/api/pricing). We can also use a different api if desired, we're not at all wedded to coingecko (it was just easiest for proof of concept).

## Testing

You should see the total funds conversion on ColonyHome, as well as conversions on the balances page. It's a bit tricky to test locally because it's searching for token addresses on gnosis chain, but you can mock a balance that has the same address as a token on gnosis chain (e.g xdai), to see it's working fine.

Places to add mock balances: `useTotalFunds` hook for colony home page (colonyHome/hooks), and the `CurrencyConversion` component for colony balances page.

**Changes** 🏗

* Memoize currency conversion api calls
* Wire in currency conversion util

Resolves #1623

### DIFF
--- a/src/components/common/ColonyHome/hooks.ts
+++ b/src/components/common/ColonyHome/hooks.ts
@@ -1,4 +1,5 @@
 import { Id } from '@colony/colony-js';
+import Decimal from 'decimal.js';
 import {
   // @BETA: Disabled for now
   // BellRinging,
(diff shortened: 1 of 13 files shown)


src/components/common/ColonyHome/hooks.ts
```diff
@@ -317,3 +311,42 @@ export const useDashboardHeader = (): ColonyDashboardHeaderProps => {
...
+  const funds = balances.items
+    ?.filter(notNull)
+    .filter(({ domain }) => !!domain?.isRoot)
+    .reduce(async (total, { balance, token: { tokenAddress } }) => {
+      const currentPrice = await fetchCurrentPrice({
+        contractAddress: tokenAddress,
+        conversionDenomination: currency,
+      });
+
+      return (await total).add(BigNumber.from(balance).mul(currentPrice));
```
ArmandoGraterol, line 328: Could this `currentPrice` be a decimal? Right now it would crash if that's the case

ArmandoGraterol (approved):
Looking good 👍 Tested by using some static conversion rates.

FireShot Capture 275 - Balances - Colony - Planet Express - localhost
FireShot Capture 276 - Planet Express - Colony - localhost

src/components/common/ColonyHome/hooks.ts
willm30: Yep good idea

PR #1625, opened 2023-12-28
Add expanded tooltips for motion steps's pill

ArmandoGraterol:
Motion step pills now with dynamic tooltips! For all the possible cases, check out this notion doc: https://www.notion.so/colony/Actions-775cb0e78eb44783a6f28702b7a9dc33#72f018e16a234ef5944613c02ca1846a

The above doc should contain all the possible variations of the pills, therefore, I would recommend keeping it at hand when testing to make sure everything is working as expected.

Take into consideration that time, as always, will be a bit wonky on the dev env.

Examples of the new tooltips at work:

Screenshot 2024-03-25 at 15 13 39
Screenshot 2024-03-25 at 15 13 47
Screenshot 2024-03-25 at 15 13 54
Screenshot 2024-03-25 at 15 14 02
Screenshot 2024-03-25 at 15 14 07
Screenshot 2024-03-25 at 15 14 28

Block-Ingestor PR counterpart: https://github.com/JoinColony/block-ingestor/pull/148

Resolves [#2046](https://github.com/JoinColony/colonyCDapp/issues/2046)


iamsamgibbs (changes requested):
Great work on this ticket, it's all looking really good, just a couple of incorrect messages I've noticed.

When there are no votes either way, the motion fails but the tooltip says fully supported.

no vote incorrect tooltip

When a motion is fully supported without going to a vote, the finalize tooltip is incorrect.
 
action supported incorrect tooltip

amplify/backend/function/fetchMotionState/src/graphql.js
```diff
@@ -81,6 +81,12 @@ module.exports = {
           hasFailed
           hasFailedNotFinalizable
           inRevealPhase
+          yaySideFullyStakedAt
```
rdig, line 84: Really wish we wouldn't have to use this lambda. 

Pretty much this lambda single handedly  stops us from being able to display motions from uninstalled voting reputation extension contracts

That's because once the extension is uninstalled, you can no longer call state (or any other contract call) on it's contract instance. 

If we find a way to get rid of this, we would be better off.

amplify/backend/function/fetchMotionState/src/utils.js
```diff
@@ -285,6 +285,7 @@ const updateMotionMessagesInDB = async (motionData, motionMessages, flag) => {
   const updatedStateHistory = {
     ...motionStateHistory,
     [flag]: true,
+    endedAt: flag === 'hasFailed' || flag === 'hasPassed' || flag === 'hasFailedNotFinalizable' ? new Date().toISOString() : null,
```
rdig, line 288: Really wish we had typescript and `enum`s in here

src/components/v5/common/ActionSidebar/partials/Motions/Motions.tsx
```diff
@@ -125,6 +110,14 @@ const Motions: FC<MotionsProps> = ({ transactionId }) => {
                 refetchMotionState={refetchMotionState}
               />
             ) : null,
+          tooltipProps: {
+            tooltipContent: getStakingStepTooltipText(
+              networkMotionState,
+              motionData,
+            ),
+            placement: 'left' as Placement,
```
rdig, line 118: Why was this casting necessary ?

src/components/v5/common/ActionSidebar/partials/Motions/utils.ts
```diff
@@ -0,0 +1,348 @@
...
+    });
+  }
+
+  return formatText(MSG.stakingNoSidesOutcome, {
+    timestamp: formattedEndedAt,
+  });
+};
+
+export const getVotingStepTooltipText = (
+  motionState: MotionState | undefined = MotionState.Null,
```
rdig, line 194: I mean this can't really be `undefined` since it has a default value of `MotionState.Null` if it's really undefined, so it should only be `MotionState`

I wonder why `tslint` didn't complain about it

src/components/v5/common/ActionSidebar/partials/Motions/utils.ts
```diff
@@ -0,0 +1,348 @@
...
+  }
+
+  return formatText(MSG.votingEnded, {
+    timestamp: formattedAllVotesSubmittedAt,
+    votersCount: voterRecord?.length || 0,
+  });
+};
+
+export const getRevealStepTooltipText = (
+  motionState: MotionState | undefined = MotionState.Null,
```
rdig, line 228: Same here

src/components/v5/common/ActionSidebar/partials/Motions/utils.ts
```diff
@@ -0,0 +1,348 @@
...
+    return formatText(MSG.revealStarted, {
+      timestamp: formattedAllVotesSubmittedAt,
+    });
+  }
+
+  const revealedVoteCount = voterRecord?.reduce((acc, voter) => {
+    if (typeof voter.vote === 'number') {
+      return acc + 1;
+    }
+    return acc + 0;
```
rdig, line 253:
```suggestion
    return acc;
```

src/components/v5/frame/NavigationSidebar/partials/NavigationSidebarSecondLevel/NavigationSidebarSecondLevel.tsx
```diff
@@ -3,7 +3,7 @@ import clsx from 'clsx';
 import React, { type FC } from 'react';
 
 import { useTablet } from '~hooks/index.ts';
-import { multiLineTextEllipsis } from '~utils/strings/index.ts';
+import { multiLineTextEllipsis } from '~utils/strings.ts';
```
rdig, line 6: Did you actually refactored all these barrel files in this PR or this some rebase nonesense ?

rdig (changes requested):
So I'm mostly good with this. This is quite some nice work.

I left some code changes notes, as well as rejected it until you address Sam's comments.

Screenshot from 2024-03-26 20-24-30
Screenshot from 2024-03-26 20-28-20

src/components/v5/frame/NavigationSidebar/partials/NavigationSidebarSecondLevel/NavigationSidebarSecondLevel.tsx
ArmandoGraterol: It was intentional, I saw that the `strings` folder only had 1 file which was the `index` so I just removed the folder and renamed the `index` to `strings`

src/components/v5/common/ActionSidebar/partials/Motions/Motions.tsx
ArmandoGraterol: The component was expecting a type of `Placement | undefined` and it was recognizing this as not a valid option for `Placement` but just as a `string`, but I think I found a type that we could use instead

amplify/backend/function/fetchMotionState/src/utils.js
ArmandoGraterol: Same here 🙏

amplify/backend/function/fetchMotionState/src/graphql.js
ArmandoGraterol: I would agree that this function needs at minimum a rework, a lot of the logic here we are not even using anymore (motion messages)

iamsamgibbs (commented):
I've tried testing this a few times this morning, but I keep getting this issue when enabling the reputation weighted extension:

Screenshot 2024-04-05 at 09 03 01

Probably just an issue with my environment as it doesn't look like the PR touches any files related to this, so I'll try testing again later

iamsamgibbs (approved):
Thanks for resolving those issues, all looks good now!
 
Now showing the correct tooltip when there are no votes:
Screenshot 2024-04-08 at 08 54 31

Now showing the correct tooltip when fully supported:
Screenshot 2024-04-08 at 08 56 30

Other tooltips still working as expected:
Screenshot 2024-04-08 at 08 47 24

Screenshot 2024-04-08 at 08 52 38

Screenshot 2024-04-08 at 08 53 46

Unrelated to this issue, but noticed that if you do not vote, it is possible to go back to the vote stage and this form is still interactive:
 
Screenshot 2024-04-08 at 08 54 54

rdig (dismissed):
Re-tested, all good now. Nice work Armando!

ArmandoGraterol:
@iamsamgibbs @rdig Rebased. May I get a re-approve?

rdig (approved):

iamsamgibbs (dismissed):

iamsamgibbs (approved):
All still working as intended 👍 

Screenshot 2024-04-15 at 15 43 39

rdig (approved):
All good. Re-approving

PR #1653, opened 2024-01-03
Add import order rule to linter

### LINKED ISSUE #1545: Update our linter with import order
We should align our linter to auto format import order, with something like this.
```JSX
"import/order": [
    "error",
    {
      "alphabetize": {
        "caseInsensitive": true,
        "order": "asc"
      },
      "groups": [
        ["builtin", "external", "internal"],
        "parent",
        "sibling",
        "index"
      ],
      "newlines-between": "always",
      "pathGroupsExcludedImportTypes": ["internal"],
    }
  ]
```
Thus all of our imports will be ordered and we won't need to check that on each code review.
We should also align other issues here, like which `'` character to use.

ArmandoGraterol:
Added `import/order` rule and related config to ESLint.

With this change `npm run lint` should:
- Order the imports like we already did: `library > relative > absolute`
- Order these imports alphabetically
- Add spaces between the import groups

Of course, almost all of the files are touched with this rule now active, but I've looked at some of the changes that had variety in import groups, and it seems to be working as expected. Let me know if you spot a case or import not being covered by the config.

Resolves #1545


bassgeta (approved):
Checking the more complex ones, it seems to behave just as intended, very nice :D

rdig (approved):
I mainly checked the rules added, as I trust that the files were changed automatically by the linter itself.

Other than throwing us off the first few times around _(due to muscle memory)_, I don't think it'll otherwise have any issues.

Nice work.

PR #1656, opened 2024-01-04
Fix crash due to custom yup methods

ArmandoGraterol:
Fixed CDapp crash when custom yup methods aren't available fast enough for components. This was caused due to the recent changes to the import/order.

Also fixed global styles overwriting the custom styles of some components like buttons.


arrenv (commented):
Nice, fixed the issue for me. Thanks @ArmandoGraterol

rdig (approved):
Tested. All good.

arrenv (approved):

PR #1679, opened 2024-01-09
Introduce lambda logging

### LINKED ISSUE #1580: Introduce Lambda Logging
Since production lambda output shows up only in cloudwatch logs it'll be a good addition to introduce various logs at key points  in our various lambdas which can aid in debugging when/if somethings goes wrong in production

jakubcolony:
## Description

This small PR adds logging to what previously was a silent try/catch block. In the past the silenced errors were not the "expected" ones and we had a hard time debugging it without proper logs.

Resolves #1580


rdig (approved):
Looks good. Thanks for adding this in Jakub

bassgeta (approved):

PR #1682, opened 2024-01-10
refactor: UserSubmenu to JSX and "add button support" :)

### LINKED ISSUE #1510: Add support for button type of User navigation sub item
The [UserSubMenu component](https://github.com/JoinColony/colonyCDapp/blob/master/src/components/common/Extensions/UserNavigation/partials/UserSubmenu/UserSubmenu.tsx) currently only has support for external and internal links, button clicks are "hacked" links.
We should add support for different types of components there, maybe use composition instead.

bassgeta:
## Description

Rewrite the `UserSubmenu` component to be less config driven and more JSX.
I've reused styles via the tailwind module approach, since all the different links and buttons technically have the same style.
I've also pulled the submenu id into an enum while I was at it.

**Changes** 🏗

* `UserSubmenu` is now a handler that renders one of three menus and uses an enum

Resolves  #1510


src/components/common/Extensions/UserNavigation/partials/UserSubmenu/blocks/MenuList/MenuList.tsx
```diff
@@ -0,0 +1,13 @@
+import React, { PropsWithChildren } from 'react';
```
jakubcolony, line 1: What's the idea with blocks? How are they different from partials?

jakubcolony (commented):
I've got a question about the new folder structure. If we end up introducing new patterns, it'd be good to have them in living standards too.

src/components/common/Extensions/UserNavigation/partials/UserSubmenu/blocks/MenuList/MenuList.tsx
bassgeta: My thinking was that 'blocks' are components which are reusable among the `partials`, but come to think of it, we can just put them into the partials directory.

jakubcolony:
@bassgeta might need a rebase before I should reapprove

willm30 (changes requested):
"Base Currency" should be the currency symbol, e.g. "GBP"
Captura de pantalla 2024-01-12 140302

This should be a two-column layout.
Captura de pantalla 2024-01-12 140307

I recommend looking at the relevant files as they are currently on master for reference. Also see the [original pr](https://github.com/JoinColony/colonyCDapp/pull/1572).

willm30 (approved):
Currency menu looking fine 👍

jakubcolony (approved):

PR #1685, opened 2024-01-10
Update ColonyContext typing

### LINKED ISSUE #1419: Improve typing of ColonyContext
Currently our `ColonyContext` value is tied to the query state.
```JSX
interface ColonyContextValue {
  colony?: Colony;
  loading: boolean;
  canInteractWithColony: boolean;
  refetchColony: RefetchColonyFn;
  startPolling: (pollInterval: number) => void;
  stopPolling: () => void;
  isSupportedColonyVersion: boolean;
}
```

```JSX
  if (loadingColony) {
    return <LoadingTemplate loadingText={MSG.loadingText} />;
  }

  if (!colony || error) {
    return <NotFoundRoute />;
  }

  return (
    <ColonyContext.Provider value={colonyContext}>
      {children}
    </ColonyContext.Provider>
  );
```

So `children` which are the context consumers are always going to have the `colony` property defined, but due to our type definition, we always need to check for `colony` to exist in the context value, a la:
```JSX
const { domains, colonyAddress, nativeToken, balances } = colony || {};
```

If we just initialize the context value with a valid colony, it is going to make our code simpler and more readable.

ArmandoGraterol:
### What the heck is happening?
- All of the queries that were inside `ColonyContext` have been moved to `ColonyRoute`. This way, we can type the widely used `colony` prop as defined and remove all the fluff around this object being undefined or not from the CDapp.
- Instead of redirecting the user to the loading or not found route from the context, this is done in the route component now. 
- There are many changes made in this PR but most of them are just the removal of fallbacks and optional operators.
- What we need to test with this PR is if the CDapp continues to work correctly, even if the majority of changes are mostly  "dev facing".
- What I did was search for any instances where we used the `useColonyContext` hook and updated that file. Let me know if I missed one somehow.

Resolves #1419

### DIFF
--- a/src/hooks/useExtensionData.ts
+++ b/src/hooks/useExtensionData.ts
@@ -49,10 +49,7 @@ interface UseExtensionDataReturn {
  * and mapping it into Installed or InstallableExtensionData object
  */
 const useExtensionData = (extensionId: string): UseExtensionDataReturn => {
-  const {
-    colony: { colonyAddress },
-    colony,
-  } = useColonyContext();
+  const { colony } = useColonyContext();
 
   const extensionHash = getExtensionHash(extensionId as Extension);
 
@@ -64,7 +61,7 @@ const useExtensionData = (extensionId: string): UseExtensionDataReturn => {
     refetch,
   } = useGetColonyExtensionQuery({
     variables: {
-      colonyAddress,
+      colonyAddress: colony.colonyAddress,
       extensionHash,
     },
     fetchPolicy: 'network-only',
--- a/src/hooks/useExtensionData.ts
+++ b/src/hooks/useExtensionData.ts
@@ -49,7 +49,10 @@ interface UseExtensionDataReturn {
  * and mapping it into Installed or InstallableExtensionData object
  */
 const useExtensionData = (extensionId: string): UseExtensionDataReturn => {
-  const { colony } = useColonyContext();
+  const {
+    colony: { colonyAddress },
+    colony,
+  } = useColonyContext();
 
   const extensionHash = getExtensionHash(extensionId as Extension);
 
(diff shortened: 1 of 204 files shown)


src/components/v5/common/ActionSidebar/partials/forms/ManageColonyObjectivesForm/consts.ts
```diff
@@ -13,9 +13,7 @@ export const validationSchema = object()
       .trim()
       .max(MAX_ANNOTATION_LENGTH)
       .required(() => formatText({ id: 'errors.colonyObjective.description' })),
-    colonyObjectiveProgress: number()
-      .max(100)
-      .required(() => formatText({ id: 'errors.colonyObjective.progress' })),
+    colonyObjectiveProgress: number().max(100).notRequired(),
```
bassgeta, line 16: dropping a comment just in case if this is a straggler :D

src/hooks/useExtensionData.ts
```diff
@@ -49,7 +49,10 @@ interface UseExtensionDataReturn {
  * and mapping it into Installed or InstallableExtensionData object
  */
 const useExtensionData = (extensionId: string): UseExtensionDataReturn => {
-  const { colony } = useColonyContext();
+  const {
+    colony: { colonyAddress },
```
bassgeta, line 53: maybe I'd be fine with just doing colony.colonyAddress below here

bassgeta (dismissed):
Code wise this looks good, I love the new `ColonyRoute` component instead of doing it in a context provider, good stuff! `rest?.In?.Peace?.Nullish?.Coallescence?.Everywhere`

jakubcolony (approved):
Huge effort, great change! 👍

src/components/v5/common/ActionSidebar/partials/forms/ManageColonyObjectivesForm/consts.ts
ArmandoGraterol: Jesus man, how did you catch this? You are scaring me

jakubcolony (approved):

bassgeta (dismissed):

jakubcolony (approved):

bassgeta (approved):

PR #1690, opened 2024-01-11
Fix skeleton loader overlapping dropdown menu

### LINKED ISSUE #1639: Skeleton loader overlapping dropdown menu
## Steps to reproduce
- Open the UserMenu while the skeleton loader is still present in the activity feed.
- Notice that the skeleton loader it overlaps the UserMenu.

## Expected behavior
- Skeleton loaders should not be visible over elements that are above it.

## Actual behavior
- Skeleton loaders are visible over elements that are above it.

skeleton-loader

warrencolliver:
## Description

This PR fixes the issue of the Skeleton loader in the activity feed overlapping the userMenu when it is present. 

**Changes** 🏗

Changed the Z index of the user menu pop up to 12.
1639-skeleton-overlap

Resolves #1639

### DIFF
--- a/src/components/v5/shared/PopoverBase/PopoverBase.tsx
+++ b/src/components/v5/shared/PopoverBase/PopoverBase.tsx
@@ -19,7 +19,7 @@ const PopoverBase: FC<PropsWithChildren<PopoverBaseProps>> = ({
   <div
     ref={setTooltipRef}
     {...tooltipProps({
-      className: clsx(classNames, 'z-10', {
+      className: clsx(classNames, 'z-12', {
         'tooltip-container': withTooltipStyles,
       }),
     })}


melyndav (dismissed):
Nice work with this one @warrencolliver. Works as expected and the issue looks to be resolved. Thanks!

jakubcolony (changes requested):
Nice one, the issue is fixed, but rather than bumping the z-index of a specific component I would question the validity of `.skeleton` having a z-index of 10 in the first place. Is something like a z-index of 1 not enough?

warrencolliver:
> Nice one, the issue is fixed, but rather than bumping the z-index of a specific component I would question the validity of `.skeleton` having a z-index of 10 in the first place. Is something like a z-index of 1 not enough?

Hi,

I have made these changes (skeleton z index to 1, and menu pop up z index back to 10)

jakubcolony (approved):
Thanks Warren, very nice 👍

bassgeta (approved):

melyndav (approved):

PR #1696, opened 2024-01-12
Increase create-data query max retries

jakubcolony:
## Description

I have noticed the `create-data` script failing due to not being able to get colony metadata in time. This PR increases the max retries for the query to match the ingestor's value.

### DIFF
--- a/scripts/create-data.js
+++ b/scripts/create-data.js
@@ -1,6 +1,13 @@
 require('cross-fetch/polyfill');
 
-const { utils, Wallet, providers, BigNumber, constants, Contract } = require('ethers');
+const {
+  utils,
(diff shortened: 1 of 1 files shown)


scripts/create-data.js
```diff
@@ -950,8 +1019,8 @@ const toggleRepMining = async () => {
 const tryFetchGraphqlQuery = async (
   queryOrMutation,
   variables,
-  maxRetries = 3,
-  blockTime = 5000
+  maxRetries = 10,
```
jakubcolony, line 1022: This is the only thing that changed, other changes are prettifying

bassgeta (approved):

PR #1702, opened 2024-01-12
Update domain ids on forms to numbers

### LINKED ISSUE #1426: Change action form to handle domain IDs as numbers
Currently, the action form stores domain IDs as strings which is different to the rest of the codebase and has already proven problematic (see issue fixed in #1395). They should be changed to use number type instead.

ArmandoGraterol:
Updated action sidebar forms to not convert domain ids to strings; instead, leave them as numbers like it is everywhere else in the codebase.

The forms should be working as they were before this change. These changes are "dev facing" only, they are made to keep the data type consistent with the rest of the CDapp.

Resolves #1426


bassgeta (approved):

jakubcolony (dismissed):
So much better! 👍

jakubcolony (approved):

bassgeta (approved):

PR #1708, opened 2024-01-15
Remove unused code

### LINKED ISSUE #1703: Remove dead/unused code
This issue is to clean up the repository from any "dead code", which may include:
- "v4" components (as mentioned previously, we can always look up GitHub history/Dapp for reference)
- Unused files: running `npx unimported` (https://www.npmjs.com/package/unimported) suggests a large number (~260) of files that aren't used anywhere 
- Unused dependencies: again, `npx unimported` can help with identifying those

jakubcolony:
## Description

This PR will focus on removing the "v4/dapp" components that are no longer used in the app.

I apologise for the size of this PR; clearly I've underestimated the amount of code we had in the previous version of the app. The vast majority of the changes are simple deletions, and I've kept other modifications to a minimum.

The cleanup is not finished - there's a potential for a more involved refactoring that would remove many instances of v5 depending on the old code but as mentioned, I wanted to focus this PR on removing old code only.

Resolves #1703


src/components/v5/common/ActionSidebar/partials/forms/AdvancedPaymentForm/hooks.ts
```diff
@@ -119,23 +116,9 @@ export const useAdvancedPayment = (
...
-            },
-            networkInverseFee,
-          );
-        }),
-      ),
-      [colony, networkInverseFee],
-    ),
+    transform: mapPayload((payload: AdvancedPaymentFormValues) => {
+      // @TODO: Add a helper function mapping form values to action payload
+      return payload;
```
bassgeta, line 121: hmmm will this work or is it OK since we don't have advanced payments in yet?

bassgeta (commented):
right so it builds, it lints and it works, so I think we're good :D nice one going the extra step and moving just the things we needed to the "new" app.
I dropped a comment for advanced payment form, but I realized we don't have it implemented yet, so LGTM

src/components/v5/common/ActionSidebar/partials/forms/AdvancedPaymentForm/hooks.ts
jakubcolony: It has never worked before, and as you pointed we don't have this action enabled yet. The problem here was that it was incorrectly depending on the old payment dialog helper, so the payload was wrong regardless.

ArmandoGraterol (dismissed):
LGTM 🎉 Good job with all of this! Always happy to see the CDapp getting cleaned up :)

chmanie (dismissed):
Tested locally and on on-demand, works perfectly! Thanks for the clean up

Issue #1577, opened 2023-12-18
Actions: Wallet address jumping in when a member is selected


AdrianDudko:
**Teamwork task:** https://tw.auroracreation.com/#/tasks/15640605
**Created by:** @melyndav 

**Description**
The wallet address shouldnt jump in when a member is selected. It also happens in the tables and breaks a smooth user experience. 

https://www.awesomescreenshot.com/video/21856562?key=c0619a9db609c01ca1bb7dcf74686061

AdrianDudko:
@melyndav @arrenv 

I close this Issue, fixed by https://github.com/JoinColony/colonyCDapp/pull/1711

PR #1711, opened 2024-01-16
fix: get user data from user select options fallback

joanna-pagepro:
https://github.com/JoinColony/colonyCDapp/issues/1577

Added a fallback to get user data from user select options array to avoid the jumping effect after selecting a user from the dropdown.


melyndav (approved):
Hey @joanna-pagepro thanks for picking up this issue. Looks great and is working as expected for me. Thank you!

ArmandoGraterol (approved):
Looking better now, good work!

PR #1716, opened 2024-01-16
Fix create data race condition on colony creation

area:
## Description

With the introduction of temporary colony metadata in the database on colony creation (#1513), we created a race condition. I believe this is present in the frontend as well, but this fixes it for the `create-data` script, at least, to avoid issues that both myself and Chris have been seeing intermittently (as you'd expect with a race condition) with the script - for example:  

```
2024-01-16T15:42:02.022+02:00    ingestor | 2024-01-16T13:42:01.342Z Processing block 315
2024-01-16T15:42:04.028+02:00    ingestor | 2024-01-16T13:42:01.856Z Block 943 added to the queue
2024-01-16T15:42:04.028+02:00    ingestor | 2024-01-16T13:42:03.948Z Could not find metadata for colony 0x1133560dB4AebBebC712d4273C8e3137f58c3A65. Skipping...
```

**New stuff** ✨

Previously, we broadcast the transaction, then added the metadata to the database. If the block ingestor picked up the transaction before the metadata was added to the database, we get the above message. This message happens after the ingestor has (meant to have) retried for a little while, so I _suspect_ something is going wrong with the default values in `tryFetchGraphqlQuery`.

Broadcast-then-get-transaction-hash is the natural way for it to have been implemented. Typically, we only think about the transaction hash after we have broadcast the transaction, when we query for the receipt or look on a block explorer to see if it has been mined yet.

**Changes** 🏗

This PR updates the `create-data` script so that it gets the transaction hash _before_ the transaction has been broadcast, adds the mutation to the database, and only then broadcasts the transaction to the chain. This guarantees the ethereal data is in the database when the ingestor goes looking for it.

Really, we should be doing things in that order in the front-end as well, but it remains an open question (to me) whether it's possible with metamask, or indeed, other wallets that we're looking to support in the future. Either way, fixing the 'retry' functionality should be, hopefully, good enough.

### DIFF
--- a/scripts/create-data.js
+++ b/scripts/create-data.js
@@ -409,7 +409,9 @@ const createColony = async (
 
   const currentNetworkVersion = await colonyNetwork.getCurrentColonyVersion();
 
-  const colonyDeployment = await colonyNetwork['createColonyForFrontend'](
+  const populatedTransaction = await colonyNetwork.populateTransaction[
+    'createColonyForFrontend'
+  ](
     constants.AddressZero,
     tokenName,
     tokenSymbol,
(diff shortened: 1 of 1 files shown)


rdig (approved):
Seems to be working fine for me, then again, I never had this issue. Regardless, than you for fixing this.

image

PR #1740, opened 2024-01-17
Fix validation of manage colony objective form

### LINKED ISSUE #1464: Manage Colony objective action type validation issues
## Description

The Manage Colony Objective active type has validation issues. As seen in the screenshot below, I have entered a objective title, objective description and a percentage value all in their respective fields. However, the error is showing as these inputs are missing. 

For the title and objective description, the actual error is that both fields exceed the character limits defined in the product specs. This correct validation error should show to users.

For the percentage, the user should not be shown the input is missing. The user should also be able to change other parts of the objective and leave an existing percentage value untouched or at 0%. 

image

ArmandoGraterol:
This PR fixes the validation of the manage colony objective form. The problem was that the form didn't have default/initial values for the title, description, and progress so when you submitted the form still required you to enter a value on those fields even if you didn't want to change them but that's because for the form they were empty.

I also added validation for the action title as it was missing.

Resolves #1464


bassgeta (approved):

jakubcolony (approved):
LGTM, nice fix 👍

PR #1741, opened 2024-01-18
Fix: Issues with upgrading colony version

iamsamgibbs:
## How to Test

Requires changes to your dev environment to have a Colony running on version 13. Please follow the steps in #1677 (You will also need to copy over this script into ColonyNetwork as it is missing from the `glwss4` branch: https://github.com/JoinColony/colonyNetwork/blob/develop/scripts/deployOldUpgradeableVersionTruffle.js)

## Description

If already on the latest colony version, the action sidebar will now show an info banner and disable the submit button when the 'upgrade colony version' action type is selected. Metadata has also been adjusted to: `{Upgrade Colony version to v{newVersion} by {initiator}}` and "latest version" in the table, as discussed with Arren.

Screenshot 2024-01-18 at 13 53 41

On the advanced settings page, the current colony version is now on the left and latest contract version on the right. Text has been updated from "Last version" to "Latest version".

Screenshot 2024-01-18 at 13 48 25

"Upgrade now" button now opens the action sidebar with the upgrade colony action selected.

Screenshot 2024-01-18 at 13 48 56

"Upgrade Colony version" link in the admin menu should remain as discussed with Arren.

**Changes** 🏗

* If already on the latest colony version, action sidebar now shows an info banner
* If already on the latest colony version, action sidebar submit button is disabled
* Action sidebar metadata adjusted to `{Upgrade Colony version to v{newVersion} by {initiator}}`
* In the action sidebar form: "New version" updated to "Latest version"
* On the advanced settings page, current colony version and latest contract version now in the correct positions, and "Last version" adjusted to "Latest version".
* "Upgrade now" button on advanced settings page now opens action sidebar

Resolves #1677

PR #1753, opened 2024-01-18
Token Activation Modal pending transaction button state

### LINKED ISSUE #1548: Slow transaction trigger and gas limit issue when Activating tokens
## Steps to reproduce
- On production, try to activate some tokens.
- Enter an amount and click the 'Activate tokens' button.
- The button will remain in a clickable state, until the spending cap in Metamask window is loaded.
 - Clicking the button over and over again will cause a too many requests issue.
- Once the it appears, continue with activating tokens.
- It will fail due to a gas limit issue.

## Expected behaviour
- The 'Activate tokens' button should disable and show a loading state on click.
- Activation tokens should not fail due to gas limit issues.
- This issue should fail more gracefully, so, that you can try again when staking.

## Actual behaviour
- The 'Activate tokens' button is continuously clickable despite already being triggered.
- Activating tokens in production is failing due to a gas limit issue.
- When staking on a motion, this error also occurs, which results in the staking button being stuck.

image

rdig:
This PR fixes part of the related issue, which is to add a loading state to the button that generates, and sends out, the required transactions to activate the user's tokens.

Screenshot from 2024-01-19 01-16-59

Screenshot from 2024-01-19 01-18-31

**Note** that I've used the same pattern for the button loading state as the one found in the Motions flow: https://github.com/JoinColony/colonyCDapp/blob/3d7878984fbe2c60f672b74d1d952ba2311f66c1/src/components/v5/common/ActionSidebar/partials/Motions/steps/FinalizeStep/FinalizeStep.tsx#L142-L157

The second part of this issue cannot be fixed by the client, since it requires a contract upgrade _(In order to have the Token Authority allow users to activate tokens, while the token is still locked)_. For a quick "fix" for existing colonies, unlocking the token allows users to activate their tokens as expected.

Note however that the above contract fix still requires #1741 to be merged before we can attempt new contract deployments.

Resolves #1548


iamsamgibbs (approved):
Looks good to me!

jakubcolony:
@rdig could you rebase to include Chris's fix?

rdig:
Done

jakubcolony (approved):
LGTM

arrenv (approved):
Looks good to me, thank you @rdig!

PR #1757, opened 2024-01-19
Remove scrollbar in pending transactions list

### LINKED ISSUE #1687: Prevent horizontal scroll when transactions pending
## Steps to reproduce
- Create a new action transaction.
- Click on the pending button to bring up the Transactions tab in the UserHub.
- Notice the horizontal scroll bar flashing in and out.

## Expected behavior
- There should be no horizontal scroll bar in the transactions tab.

## Actual behaviour
- There is a flashing horizontal scroll bar in the transactions tab.

pending-transactions

warrencolliver:
## Description

This PR removes the scrollbar used in the pending transactions list. this was done by removing the scrollbar styles and code

**New stuff** ✨
scrollbar is no longer visible or flashing.
Screenshot 2024-01-19 at 13 01 59

Resolves #1687


src/components/common/Extensions/UserHub/partials/TransactionsTab/TransactionsTab.tsx
```diff
@@ -60,20 +58,15 @@ const TransactionsTab: FC<TransactionsProps> = () =>
         <p className="heading-5 mb-2.5 sm:mb-0.5">
           {formatText({ id: 'transactions' })}
         </p>
-        <div
-          id="transactionsListContainer"
```
ArmandoGraterol, line 64: Are we sure this id is not being used anywhere else? Can we safely remove this? I see that there's a hook that uses it

src/components/common/Extensions/UserHub/partials/TransactionsTab/TransactionsTab.tsx
```diff
@@ -8,8 +8,6 @@ import { useTransactionsListObserver } from './hooks';
 import TransactionList from './partials/TransactionList';
 import { TransactionsProps } from './types';
 
-import styles from './TransactionsTab.css';
```
ArmandoGraterol, line 11: if we are not going to use the styles provided by the stylesheet, then we should remove it as well seeing that this is the only component that uses it.

ArmandoGraterol (changes requested):
Nice work, the scrollbar is no longer flashing, however I have a few comments/questions regarding the removed parts

src/components/common/Extensions/UserHub/partials/TransactionsTab/TransactionsTab.tsx
warrencolliver: I saw that hook which uses it, TransactionsTab.css contains transactionsListContainer. but apart from that, I can't find anywhere else it's being used

arrenv (approved):
Scrollbar issue looks resolved to me, thank you @warrencolliver

src/components/common/Extensions/UserHub/partials/TransactionsTab/TransactionsTab.tsx
rdig: Correct. Remove that import, and if you so happen to need the styles declared in there, they should be transitioned over to tailwind

rdig (approved):
All good from my end, but please address Armando's concerns before merging.

[Screencast from 2024-01-19 17-24-24.webm](https://github.com/JoinColony/colonyCDapp/assets/1193222/837f7961-bd87-428e-997e-86c6a2ff9f29)

arrenv (approved):
Still good on my end.

src/components/common/Extensions/UserHub/partials/TransactionsTab/TransactionsTab.tsx
warrencolliver: Removed the css import as that was bringing in the styles for the scrollbar, then added tailwind css for the container which meant the css file was no longer needed. issue has been resolved and files have been cleaned up.

jakubcolony: @warrencolliver it sounds like it is being used elsewhere then, are you sure it can be safely removed? If the answer is yes then we need to remove everything that previously relied on this id, including the hook you mentioned.

src/components/common/Extensions/UserHub/partials/TransactionsTab/TransactionsTab.css
```diff
@@ -1,35 +0,0 @@
-@value query700 from "~styles/queries.css";
```
jakubcolony, line 1: Is deleting the entire file intentional? There are other styles here for the scrollbar (vertical), which might be needed.

jakubcolony (changes requested):

src/components/common/Extensions/UserHub/partials/TransactionsTab/TransactionsTab.css
jakubcolony: In fact, scrollbar styles might still apply from `main.global.css`.

src/components/common/Extensions/UserHub/partials/TransactionsTab/TransactionsTab.tsx
warrencolliver: > @warrencolliver it sounds like it is being used elsewhere then, are you sure it can be safely removed? If the answer is yes then we need to remove everything that previously relied on this id, including the hook you mentioned.

@jakubcolony The hook cannot be safely removed. Without this hook the transactions tab will not show more than 10 transactions at a time. this wasn't picked up previously as it wasn't tested with more than 10 transactions. because of this I've kept the id in there so the hook is still functional

jakubcolony (approved):
All good now, thanks @warrencolliver 👍

PR #705, opened 2023-06-28
Port Safe Control feature

ArmandoGraterol:
This is the feature branch for the Safe Control feature.

Issues related to this feature can be found here: https://github.com/orgs/JoinColony/projects/25

## Resources
- Docs [here](https://docs.colony.io/use/managing-funds/gnosis-safe-control/)
- Flow chart [here](https://lucid.app/lucidspark/43230a2b-4fa9-46de-bcd9-be8cd4031276/edit?viewport_loc=-600%2C-1075%2C9000%2C4438%2C0_0&invitationId=inv_bc948a06-9365-4966-bfcf-c4c73d62b430)
- Figma [here](https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=115%3A7477&mode=dev)

## Safes for testing
- If you want to be co-owner of any of these safes in order to execute transactions/snoop around/debug, please notify me.
- Otherwise, you can easily create your own safe via https://app.safe.global/welcome

### Ethereum

**Safe**: `0x3F107AFF0342Cfb5519A68B3241565F6FC9BAC1e` (It should have some funds and 1 NFT)
**Module**: `0x8dfABd117Db8BC64829Ce7C595F20998b054cDC2`

### Binance Smart Chain

**Safe**: `0xe759f388c97674D5B8a60406b254aB59f194163d`
**Module**: `0xCC199aCAb2B38C6DC5f7304B29BDEb3a83E7d212`

## Testing
1. Checkout this branch
2. A new amplify function and package have been added so run `npm i`.
3. `npm run dev` as usual. The difference is that now a `safe` docker container should be created and be up once the dev env is completely set up.
4. Once the safe docker container is successfully up, it will print out in the console the values you need for the .env file (Check the `.env.example` for a reference of what env vars are needed.)
5. That's it! After running `npm run webpack` as usual, you should be able to add and remove safes, and send cross-chain safe transactions through the UAC (Advanced > Safe Control).

Clarifications:
- Once you create and send transactions, the `safe` docker container will print a message out in the console notifying you that it saw a message pass through the bridge. (In the case of motions, this message will appear AFTER a successful motion has been finalized.)
- All the transactions should appear as `Completed` immediately after making them because all of the transactions are executed automatically when they are sent in the case of the dev env. (Not the case with the QA/Prod env, in there the user needs to manually execute the transactions in the other chain.)
- You need `Root` permission to add, remove, and make safe transactions.
[...]

PR #1760, opened 2024-01-19
Fix `fetchTokenFromChain` lambda rpc env variables

### LINKED ISSUE #1501: Correct temp fix for `fetchTokenFromChain` Lambda
## Description

We need to fix the temporarily hardcoded RPC URL for `fetchTokenFromChain` Lambda that is required by Safe Control, but causing issues with fetching tokens.

This is impacting the Colony creation process.

rdig:
This quick PR fixes a value I hard-coded during the first production deployment rush, in an effort to quickly fix the "token not found in db" issue that we were having.

The error was introduced initially by the [Safe Control](https://github.com/JoinColony/colonyCDapp/pull/705) feature which didn't account for other chains besides the bridges, in a non-dev environment.

The fix was to simply add a new default case for the `getRpcUrlParamName` util, which will default to gnosis in a `qa` or `prod` environments.

While I was at it, I also removed all references to the test `sc` environment, which no longer exists in our deployment infrastructure.

**Notes on testing**
You can't really test my fix while running this locally, as it only affects `qa` and `prod` environments which all run on the `gnosis` chain.

The only thing you can test locally is to check that my fix didn't break the local dev environment for this lambda.

Alternatively you can co-opt Bogdan when doing the testing and have him deploy this to QA and testing the `fetchTokenFromChain` lambda from the AWS console manually.
 
Resolves #1501

### DIFF
--- a/amplify/backend/function/createColonyEtherealMetadata/src/index.js
+++ b/amplify/backend/function/createColonyEtherealMetadata/src/index.js
@@ -17,7 +17,7 @@ let graphqlURL = 'http://localhost:20002/graphql';
 
 const setEnvVariables = async () => {
   const ENV = process.env.ENV;
-  if (ENV === 'qa' || ENV === 'sc' || ENV === 'prod') {
+  if (ENV === 'qa' || ENV === 'prod') {
     const { getParams } = require('/opt/nodejs/getParams');
     [apiKey, graphqlURL] = await getParams([
       'appsyncApiKey',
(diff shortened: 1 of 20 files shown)


jakubcolony (approved):
Hard to test as you pointed, but looks reasonable 👍

bassgeta (approved):

ArmandoGraterol (commented):
Everything working as expected as far as I can tell

PR #1771, opened 2024-01-22
Fix motion timeouts query data

rdig:
Currently, once creating a motion, a spinner will show up next to the staking tab, signifying that it can't load the timeout until the staking is finished.

This is because the `fetMotionTimeoutPeriods` lambda breaks, and returns all values as `-1` which triggers a infinite loading state in the component.

The issue at fault here is that the call to `getBlockTime` has the params inversed _(`blockHash` and `provider` vs. `provider` and `blockHash`)_

This issue fixes this, making motions work once again.

~~Note that if you try to create a motion with a description _(annotation)_, it will still crash, but that is a separate issue, that was highlighted here: https://discord.com/channels/562263648173555742/998629903740182580/1198967494564646932~~

While I was at it, I also fixed the above issue, with creating a motion with an annotation. The actual fix was on the `colonyJS` side, but I brought the fix in by updating `@colony/events` in both the CDapp repo and the lambda functions

**Testing**

- Make sure to install packages both in the CDapp and in the lambdas
- Create a new motion without annotations
- Create a new motion with annotations

### DIFF
--- a/amplify/backend/function/createUniqueColony/src/package-lock.json
+++ b/amplify/backend/function/createUniqueColony/src/package-lock.json
@@ -10,6 +10,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@colony/colony-js": "^7.0.3",
+        "@colony/events": "^1.0.2",
         "cross-fetch": "^4.0.0",
         "ethers": "^5.7.1"
       },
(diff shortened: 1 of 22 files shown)


iamsamgibbs (dismissed):
Tested and can confirm this resolves the staking spinner and getBlockTime issue.

iamsamgibbs (approved):
Approved! Description now works too.

Think this is an unrelated issue, but raising here just in case:

On initial submission, both title and description are missing.
Screenshot 2024-01-22 at 14 59 10

On refresh, appears as expected.
Screenshot 2024-01-22 at 15 00 20

arrenv:
> Approved! Description now works too.
> 
> Think this is an unrelated issue, but raising here just in case:

> On refresh, appears as expected. 

I experienced this in @bassgeta's PR https://github.com/JoinColony/colonyCDapp/pull/1608, but he resolved it I believe.

jakubcolony (approved):
Works like a charm! 

image

Issue #1764, opened 2024-01-21
Clean up folder structure


jakubcolony:
1. Agree on a revised folder structure.
2. Clean up codebase to align with the new structure.
3. Update https://github.com/JoinColony/ColonyFrontEndLivingStandard with documentation regarding folder structure.

jakubcolony:
Blocked by #1748

iamsamgibbs:
Just making note of this here so it doesn't get lost. @ArmandoGraterol suggested https://github.com/alan2207/bulletproof-react/blob/master/docs/project-structure.md

PR #1772, opened 2024-01-22
refactor: create new table component, rework balance page table

### LINKED ISSUE #1707: Refactor / remove existing Table react library
Currently we have 4 different table components (`Table`, `TableWithActionsHeader`, `TableWithMeatballMenu`, `TableWithHeaderAndMeatballMenu`, all with a pretty complicated API.

We should aim to simplify all of this and remove the `@tanstack/react-table` library, since it's too complicated for our simple use cases.

### What should we do?

Remove the library and build a simple to use table. 
The table component should just apply styles via a Table element and render all the children.
So the way we use it could be something like

```JSX
<Table>
  <thead>
    <tr>
      <th>Heading 1</th>
      <th>Heading 2</th>
      <th>Heading 3</th>
    </tr>
  </thead>
  <tbody>
    {rows.map(row => (
      <tr>
        <td>{row.name}</td>
        <td>{row.someProperty}</td>
        <td><MeatBallMenu ... /></td>
      </tr>
    ))}
  </tbody>
</Table>
````
We will also need reusable/customizable pagination components to go along with the table too.

We can add in some complex `data grid` libraries when we actually get complex data to load.

### PR DESCRIPTION
**Pull request by bassgeta**
## Description

Implemented a really basic table that just styles the rows.
I've used a CSS module to style children, because tailwind selectors are really bad for this type of CSS.
They provided an escape hatch with [`theme`](https://github.com/JoinColony/colonyCDapp/pull/1743).
I've also refactored the balance page table, to provide an example of usage.
Sorting also works via the `useTableSort` hook :sunglasses: 
The table looks a bit differently regarding column sizing, but it's really minor.

**New stuff** ✨

* `Table` component which is really basic and just applies styles to its cells
* `useTableSort` hook which changes a sort object, returns it, and takes in a generic type of enums to type the fields
* `TokenMeatballMenu` is just the previous hook extracted into a component

**Changes** 🏗

* Renamed existing `Table` component to `TableOld`
* reworked the balance page table with new sorting and everything needed

**Deletions** ⚰️

* just the column helper for table completely deleted

Resolves #1707

### CODE REVIEW

**Review by melyndav — CHANGES_REQUESTED** [review]
Hey @bassgeta - great start on this one! As discussed in Discord:

1. Mobile needs to be covered since PagePro have developed mobile variations as per the design for all table variations. Please also check tablet breakpoints too.

Fig link- https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-v5?type=design&node-id=5226-352480&mode=design&t=B8nCiyAMEuW9XWdz-4

Every table will have a mobile variation designed and they are all in Figma. So be sure to check these are some have custom interactions as they have to handle the content differently. 

Also be mindful that not all tables are built the same. Incoming funds is different from our stand table design such as activity feed etc. Figma covered every table and it's unique functionality. 

2. Pagination also needs to be handled for all table variations. All tables need to handle pagination and show once 10+ results are shown. Mobile also need to be covered with specific mobile pagination interactions.

Fig designs with annotations for handling pagination: 

https://www.figma.com/file/moME7kMfr5pdt6PYrfDtm6/Members?type=design&node-id=201-9440&mode=design&t=HYHAVRzAHQLK1nMw-4

Feel free to ping for a call and I can walk you through this.

File: src/components/frame/v5/pages/BalancePage/partials/BalanceTable/BalanceTable.tsx [hunk 1]
```diff
@@ -1,124 +1,162 @@
...
+                  onClick={() => {
+                    handleSortFieldClick(BalanceTableSortFields.BALANCE);
+                  }}
+                >
+                  {formatText({ id: 'table.row.balance' })}
+                  {getSortArrow(BalanceTableSortFields.BALANCE)}
+                </button>
+              </th>
+
+              <th> </th>
```
ArmandoGraterol, line 104: Is this needed?

src/components/frame/v5/pages/BalancePage/partials/BalanceTable/BalanceTable.tsx
```diff
@@ -1,124 +1,162 @@
...
+                      />
+                    </td>
+                    <td className="hidden sm:table-cell">{token.symbol}</td>
+                    <td className="hidden sm:table-cell">
+                      {isTokenNative && (
+                        <TokenTypeBadge tokenType={TOKEN_TYPE.native}>
+                          {formatText({ id: 'token.type.native' })}
+                        </TokenTypeBadge>
+                      )}
+                    </td>
```
ArmandoGraterol, line 127: If `isTokenNative` is `false`, do we want to render an empty `td` anyways?

ArmandoGraterol (changes requested):
Good work! I've noticed that there's warning on the console when I'm on the balance page:
Screenshot 2024-01-23 at 13 42 54

src/components/v5/common/Table/Table.tsx
```diff
@@ -1,332 +1,23 @@
-import {
-  flexRender,
-  useReactTable,
-  getCoreRowModel as libGetCoreRowModel,
-} from '@tanstack/react-table';
 import clsx from 'clsx';
```
jakubcolony, line 1: We will clarify this as part of #1764 but a reusable table should presumably live in `shared` instead? What do you think?

jakubcolony (approved):
Looks reasonable, hopefully we can also get rid of the old table at some point 👍 

The only thing I spotted is the rounded corners are a bit off on the balance table (the border is rounded but the background isn't):
image

src/components/frame/v5/pages/BalancePage/partials/BalanceTable/BalanceTable.tsx
bassgeta: yeah for proper formatting, I could maybe colspan the last header

bassgeta:
Thank you for all the reviews, but we found out this one is way too much work with this approach.
The tables have a specific UX which is already covered in our current `Table` component, reimplementing sorting, pagination and everything neccessary would take me 2 weeks :D  
Closing it and we will do a quality of life improvement via #1784

PR #1773, opened 2024-01-22
[Verified members] feature: Implement verified members

bassgeta:
## Description

This is the feature branch for the verified members functionality


rdig:
@bassgeta do you need reviewes here to merge it? 

Because otherwise, we've more or less reviewed the various PRs that make up this over time

iamsamgibbs (changes requested):
Tested and this is all still working great! Just one tiny issue with the activity feed remains:

Screenshot 2024-03-06 at 08 09 57

src/i18n/en-actions.ts
```diff
@@ -78,8 +80,8 @@ const actionsMessageDescriptors = {
       ${ColonyActionType.EmitDomainReputationReward} {Award}
       ${ColonyActionType.EmitDomainReputationRewardMotion} {Award}
       ${ColonyActionType.CreateDecisionMotion} {Decision}
-      ${ColonyActionType.AddVerifiedMembers} {Manage verified members}
-      ${ColonyActionType.RemoveVerifiedMembers} {Manage verified members}
+      ${ColonyActionType.AddVerifiedMembersMotion} {Manage verified members}
+      ${ColonyActionType.RemoveVerifiedMembersMotion} {Manage verified members}
```
iamsamgibbs, line 84: Think this was meant to be an addition rather than a change!

iamsamgibbs (changes requested):
Thanks for the fix - left a comment regarding the action type.

And the 'createdIn' field for both motions should be readonly to ensure it can only be created in 'General'.

Screenshot 2024-03-07 at 08 51 16

src/i18n/en-actions.ts
bassgeta: how did this even happen, mea culpa :facepalm: but nice catch!

iamsamgibbs (dismissed):
Looks all good now! 🎉

melyndav (changes requested):
Hey @bassgeta rockstar effort with this PR.

I've noticed a few issues. Should these be raised as separate issues or can they be handled in the one PR? 

* Editing this since all non-critical UI issues have been moved to a new issue #2029

1. The verified icon shouldn't show in the metadata for this or any action type.

image


2. The action should be assigned to the general team and display the general team in the activity feed. The pill only shows for motions. 
image
 
 
 3. The user pop-over cards aren't being updated with the verified icon when the verified user status is added to a user until a refresh. Can this happen without refreshing to match how the members/contributor cards update once the transaction is complete?
 
image

It also happens the other way when the user has their verification removed
image

4. When I remove verification from leela who is making the transaction, the transaction overview still shows Leela as verified in the motion widget. 

image


5. On mobile, the header is not the correct action header component and is displaying the inside page version across all action types on mobile. It's not in production. 
image

6. I removed all verification's for all members. However when I went to check again, users were showing up in the member dropdown in the remove variation. It resolved itself after a page reload. 


8. I received the following error with the app crashing. it is repeatable by triggering validation on an empty member table then selecting any two members from within the members dropdown. 
image

 9. When adding a member who is verified via their wallet address, it doesn't pick up their wallet address as a member and also does not detect that they are already verified. In my screenshot, I'm using Leela's wallet address and you can see on the left she's already verified on the members page behind the side panel.  However, the username does get picked up in the completed state and you can successful re-verify her. 
 
 When I use her address for removing, it detects her correctly in the members dropdown. 
 
image

10. In the completed state, the action type is showing as Generic. This is happening in permissions completed state. 
image

12. When I trigger the user popover in the completed state, it breaks the UI.
image

13. When I used reputation and went to a motion, the title shows as 'Generic' for 5 seconds and then changed to the custom title I had added.

14. The remove members text is incorrectly showing however I performed an add action as you can see by the metadata.
image

arrenv:
@bassgeta Adding record here regarding being able to add addresses that are not members, which don't work at the moment.

We still need this functionality, as a Colony needs to be able to pay addresses that are not members. However

For now though, let's validate to not allow when an address is added manually that is not an existing member. We can show the following validation error:

> "Address not found in the members list."

melyndav (dismissed):
Approving this so it can be merged since the UI issues identified in this PR were resolved in a separate PR @bassgeta.

Nice work on this whole feature!

iamsamgibbs (approved):
Witnessed this working on QA

rdig (approved):

PR #1775, opened 2024-01-23
Fix: disconnected wallet currency conversion

### LINKED ISSUE #1717: Currency conversion: Disconnected wallet state
## Description

Currently, the currency conversion feature is showing in the disconnected wallet state. Hover, it should not show at all during beta because every user with access to a Colony during the beta phase needs their wallet connected in order to view a Colony. So this feature does nothing.

However, it needs to show in the public version to allow users to freely change the currency of a Colony in a disconnected wallet state. However, in it's current implementation, it's not working on the disconnect state. Nothing saves when another currency is selected. 

## PR Scope

1. Fix the disconnect state to allow the saving of the user selections within the currency conversion section of the hamburger menu.
2. After fixing it point 1, hide the currency conversion is a disconnected state now for beta users.

## Steps to reproduce

1. Disconnect your wallet
2. Select the hamburger menu top right
3. Attempt to change the currency, it does not save the selection

iamsamgibbs:
## Description

In the public version of the app, users should be able to switch currencies without having a wallet connected. Previously, when no wallet was connected, the currency switcher did not function properly. This PR makes changes to the CurrencyContext so that the preferred currency can be set even when no wallet is connected. This is saved to local storage to persist the selection and will fallback to the defaultCurrency if nothing is found in local storage. Once a wallet is connected, user.profile.preferredCurrency will take priority over the selection in local storage.

However, whilst the app is in beta, the currency switcher should be hidden unless a wallet is connected as users need a connected wallet in order to access a colony. This PR hides the currency switcher when no wallet is connected.

**New stuff** ✨

* CurrencyContext now stores preferred currency in local storage and uses this selection when no wallet is connected

**Changes** 🏗

* CurrencyContext now updates currency state even when no wallet is connected
* Currency switcher is hidden from user menu when no wallet is connected

Screenshot 2024-01-23 at 12 54 47

Resolves #1717

### DIFF
--- a/src/components/common/Extensions/UserNavigation/partials/UserMenu/UserMenu.tsx
+++ b/src/components/common/Extensions/UserNavigation/partials/UserMenu/UserMenu.tsx
@@ -140,24 +140,26 @@ const UserMenu: FC<UserMenuProps> = ({
               </li>
             ))}
 
-            <li className="mb-0 hover:bg-gray-50 rounded -ml-4 w-[calc(100%+2rem)]">
-              <button
-                type="button"
-                className="navigation-link"
-                onClick={() => setActiveSubmenu(UserMenuItemName.CURRENCY)}
-                aria-expanded={activeSubmenu === UserMenuItemName.CURRENCY}
-                aria-controls="actionsWithVisibility"
-              >
-                <span className="flex items-center shrink-0 mr-2 sm:mr-0 flex-grow">
-                  <Icon
-                    name={currencyIconTitles[currency]}
-                    appearance={{ size: iconSize }}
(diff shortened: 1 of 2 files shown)


melyndav (approved):
Hey @iamsamgibbs nice work! Thank you for picking up my issue.

I can't test the _CurrencyContext now updates currency state even when no wallet is connected_ scenario given it's hidden during beta, but the disconnected state is working as expected.

jakubcolony (approved):
Nice work @iamsamgibbs!

warrencolliver (approved):
Great job

PR #1781, opened 2024-01-24
fix token blockies not showing consistently

### LINKED ISSUE #1712: Global: Blockies are not showing consistent
## Description

When navigating around the user interface, the blockies are showing inconsistently for the same element. Here in my screenshot below, you can see that the popover for the same token, shows a different blockie. This shouldn't be happening to avoid user confusion. This is not just for tokens, but for members etc.

image

## Expected behaviour

* Example: Blockies should display consistently to avoid user confusion and help users connect blockies to members and other UI elements.

warrencolliver:
## Description

This PR fixes the issue where blockies for tokens were showing inconsistently, when in the colony balance token list, each token has it's own thumbnail, if this thumbnail / image is a default blockie it'll appear as a different blockie when the user clicks on the token for more info. This is because currently a new blockie is being generated specifically for this token info image instead of using the tokens original blockie default

**New stuff** ✨
The same blockie is now shown in the token info image
Screenshot 2024-01-24 at 12 43 40

User blockies in the user drop down menu now show consistently with the user's blockie
Screenshot 2024-01-25 at 15 31 54

**Changes** 🏗

Changed <avatar> within TokenInfo.tsx to <TokenIcon> 
Removed the <Avatar> import 
Added <TokenIcon> import
Removed { avatar, thumbnail } as they are no longer used.

Resolves #1712

### DIFF
--- a/src/components/shared/TokenInfoPopover/TokenInfo.tsx
+++ b/src/components/shared/TokenInfoPopover/TokenInfo.tsx
@@ -4,10 +4,10 @@ import { defineMessages, useIntl } from 'react-intl';
 
 import { DEFAULT_NETWORK_INFO } from '~constants';
 import Icon from '~shared/Icon';
+import TokenIcon from '~shared/TokenIcon';
 import { Token } from '~types';
 import { getBlockExplorerLink } from '~utils/external';
 import PillsBase from '~v5/common/Pills/PillsBase';
-import Avatar from '~v5/shared/Avatar';
 import CopyableAddress from '~v5/shared/CopyableAddress';
 
 const displayName = 'TokenInfoPopover.TokenInfo';
@@ -30,7 +30,7 @@ const MSG = defineMessages({
 });
 
 const TokenInfo = ({ token, isTokenNative, className }: Props) => {
-  const { avatar, name, symbol, tokenAddress, thumbnail } = token;
+  const { name, symbol, tokenAddress } = token;
   const { formatMessage } = useIntl();
 
   return (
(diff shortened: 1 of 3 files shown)


rdig (dismissed):
Looks good. Thank you for your fixing this.

Screenshot from 2024-01-24 16-21-53

melyndav (changes requested):
Hey @warrencolliver - as mentioned in Discord, this issue needs to be resolved globally and not just for the balance tokens as per my orgional issue raised. 

Here you can see EchoTango's blockie on the left does not match the same blockie within the user dropdown. 

image

ArmandoGraterol (commented):
Looking good! The only place I could find with a different blockie was in the dropdowns of the action side panel like Mel already pointed out. I'll re-review once that's addressed

melyndav:
> @melyndav if have a list of locations where this issue can be encountered, I'm sure it will help out @warrencolliver

It's globally. So if it's a member, the icons will the different on the members page then from a members dropdown in actions. It's also the members pop-over etc.

Anywhere where a blockie can be show needs to align with what is assigned to the user profile or the token on creation.

ArmandoGraterol (approved):
I've looked into every place that I know that has a blockie and they seem to be consistent 👍 Good job

melyndav (approved):
@warrencolliver you've gone and smashed this out of the park! Thank you for fixing an issue that had bothered me for so long with the lack of the one avatr showing consistently for users. 💯 🥇 THANK YOU!

bassgeta (approved):

PR #1785, opened 2024-01-24
Remove even more unused components

chmanie:
## Description

While looking at the old `Button` component as well as the `getMainClasses` method that we used in a lot of the old components, I found many files with code that isn't used anymore.

I deleted all of that and kept around a few utils and types that are used in new components as well.


rdig (approved):
I'm good with all the files remove here. Nice cleanup Chris!

I can confirm the bundle builds correctly
Screenshot from 2024-01-25 11-26-41

No regressions introduced in the interface _(as far as I can see, anyway)_
Screenshot from 2024-01-25 11-25-56

However, I do want to note, that now _(not necessarily because of the files here though)_ the old interface which is accessible via the `/old/` route is now broken

Screenshot from 2024-01-25 11-27-08

I think we should just get rid of the old interface at this point as it serves no purpose anymore, however I am not suggesting that you that here, but do keep it in mind, and maybe open a issue to track / remove it.

chmanie:
> I think we should just get rid of the old interface at this point as it serves no purpose anymore, however I am not suggesting that you that here, but do keep it in mind, and maybe open a issue to track / remove it.

Thank you for the thorough review! I will open an issue (as I didn't even know that it was there).

bassgeta (approved):
looks legit, lints, builds, works!

PR #1786, opened 2024-01-24
Consolidate Table components

### LINKED ISSUE #1784: Refactor: consolidate table components
### Story

We are using 4 different table components, `Table`, `TableWithMeatballMenu`, `TableWithHeaderAndMeatballMenu` and `TableWithHeader`.
We would like to just be using the `Table` component.

### Description

Three of these table components are just wrappers around `Table` which inject menu props or headers.
We can skip the extra steps.
1. `TableWithHeader` -> just a `Table` and add the header in each component
2. `TableWithMeatballMenu` -> we can manually add the last column to be a `MeatballMenu` for each table using this
3. `TableWithHeaderWithMeatballMenu` -> a combination of the above

We have only 1 table component, and all the tables still function as expected.

ArmandoGraterol:
This PR aims to combine the Table component and the variations that we have of it. Like `TableWithMeatballMenu`, `TableWithActionHeader`, etc...

Now there should only exist 1 `Table`, of course, there's some wrappers that are specific to the page that the table is being used in but other than that, most of the other, more generic, wrappers have been integrated into `Table`.

All of the tables should still work and look as we would expect them to, there shouldn't be any changes in behavior or appearance in comparison with `master`.

Let me know if you spot any places where the components can be further simplified, this issue was technically only about consolidating the various components but I'm keen on continuing to make them as simple as possible in other ways.

Resolves #1784


src/components/v5/common/Table/Table.tsx
```diff
@@ -62,269 +101,288 @@ const Table = <T,>({
   const shouldShowEmptyContent = emptyContent && data.length === 0;
 
   return (
-    <div
-      className={clsx(
-        className,
-        'border border-gray-200 rounded-lg overflow-hidden',
+    <>
+      {title && (
```
bassgeta, line 105: could we maybe just use the `TableHeader` component wherever we want the title? it's a fragment so it's not tied to any layout things, we could get rid of 3 more props here

src/components/v5/common/Table/types.ts
```diff
@@ -29,4 +31,11 @@ export interface TableProps<T>
   renderCellWrapper?: RenderCellWrapper<T>;
   additionalPaginationButtonsContent?: React.ReactNode;
   emptyContent?: React.ReactNode;
+  getMenuProps?: (row: Row<T>) => MeatBallMenuProps | undefined;
```
bassgeta, line 34: not sure if this would be beneficial, but maybe we could get rid of this and just pass a rendered meatball menu. Maybe we could get rid of both sizes too, if it's static we set the column size there, etc.
But feel free to ignore this if it would be useless

bassgeta (changes requested):
This is good stuff, it looks much more sane with just one table component.
Left 2 comments which would scuff this even more, I hope it's not much more work :D

src/components/frame/v5/pages/BalancePage/partials/BalanceTable/BalanceTable.tsx
```diff
@@ -53,16 +82,94 @@ const BalanceTable: FC<BalanceTableProps> = ({ data }) => {
...
+                    [ACTION_TYPE_FIELD_NAME]: ACTION.MINT_TOKENS,
+                  });
+                },
+                label: formatMessage(MSG.labelMintToken),
+                icon: 'bank',
+              },
+            ]
+          : []),
+        {
+          key: 'transfer_funds',
```
rdig, line 141: Didn't we have a issue that fixed the "string" key names ?

rdig (changes requested):
I'm good with this, however please address Kerry's concerns first, after which I'll approve

src/components/v5/common/Table/Table.tsx
ArmandoGraterol: Hey man, what do you specifically mean by just using the TableHeader? We are already using it to render the title

src/components/frame/v5/pages/BalancePage/partials/BalanceTable/BalanceTable.tsx
ArmandoGraterol: I also remember at some point seeing an issue/PR for that but it seems like those changes are not on `master` yet

bassgeta (approved):
We'll find a way to get the menu thingy untangled down the line, for now, it's a great improvement and let's get this in before you pull out your hair due to merge conflicts :D!

rdig (approved):
Tested and looks good. Nice work.

Screenshot from 2024-02-06 18-20-55

PR #1787, opened 2024-01-25
Fix activity feed (Apollo) issues

### LINKED ISSUE #1755: Issues with Activity table
**Teamwork task:** [Dashboard - latest activity table](https://tw.auroracreation.com/#/tasks/15736415)

**Actual**

https://app.screencast.com/cS1lukvnpDUKc

https://qa-app.colony.io/pageproqa

When I am in the dashboard of a given colony and I change to the second page in the latest activity table, the data does not load and I cannot go to the first page.

**Expected**

Pagination works correctly, I can change the page forward and backward, data loads correctly.

jakubcolony:
## Description

Activity feed pagination stopped working properly after upgrading Apollo to `3.8.8` but it wasn't clear what was going on - the first 2/3 pages would usually load fine, but then it would get stuck in a loading state.

After a lengthy investigation, I don't feel much closer to the answer what is actually happening, but it was introduced in [this Apollo release](https://github.com/apollographql/apollo-client/releases/tag/v3.7.15). I rebuilt `useActivityFeed` from scratch and the issue I noticed was the paginated query losing all its data after a few `fetchMore` calls, and never being able to fetch any more. 

Unfortunately this still isn't fixed in `3.8.10` and I don't have reliable steps to reproduce to open an issue. I downgraded Apollo to `3.7.14` and introduced a custom `removeTypenameLink`, which was the reason for upgrade to a higher version in the first place.

**Changes** 🏗

* Renamed `~cache` to `~apollo` and moved all Apollo related code there 
* Changed `sortExternalLinks` not to mutate the passed array
* Some minor cleanup

Resolves #1755


bassgeta (approved):
looks good, can this perhaps also make us revert the changes I had to do in `cache/colony/tokens.ts` ? Since I had to lookup ref objects everywhere, maybe we can revert those changes too

package.json
```diff
@@ -160,7 +160,7 @@
     "webpack-dev-server": "^4.11.1"
   },
   "dependencies": {
-    "@apollo/client": "^3.8.8",
+    "@apollo/client": "^3.7.14",
```
rdig, line 163: If this is the version that works for us, then I suggest we pin it in place until further notice.

```suggestion
    "@apollo/client": "3.7.14",
```

rdig (approved):
Changes seem sensible, code looks fine. It's just weird to me that we had to downgrade in order to fix this.

Tested and can confirm it's working as expected. 

[Screencast from 2024-01-25 11-19-58.webm](https://github.com/JoinColony/colonyCDapp/assets/1193222/49391870-c07f-4cfa-ba70-94c110e9f32d)

Very nice work Jakub! :tumbler_glass:

jakubcolony:
@bassgeta 
> can this perhaps also make us revert the changes I had to do in cache/colony/tokens.ts ? Since I had to lookup ref objects everywhere, maybe we can revert those changes too

The changes you made seem to work with the downgraded version so I'd be leaning to keep them in as they're future-proof for next versions.

package.json
jakubcolony: good idea!

bassgeta (approved):

PR #1788, opened 2024-01-25
[Verified members] feat: add graphql entities for verified members

### LINKED ISSUE #1721: [Verified members] Feature: Implement the graphql backend and update the database with a new entity
### Story

When the block ingestor receives events for adding/removing verified members, it will need to modify that data in the database.
The client will also want to be able to fetch verified members.

### Description
We need a new entity which will store pairs of `colonyAddress` and `userAddress`. Additionally we need a gql query for fetching verified members and a gql mutation for adding them. The mutation cannot be called outside the block ingestor, so the auth proxy will need to be modified too.

bassgeta:
## Description

I've updated the graphql schema to include the `VerifiedMember` entity, and let Amplify autogenerate all the queries for us.
I've used a combination of primary and sort key for the `VerifiedMember` entity, this is basically a compound primary key.

**New stuff** ✨

* `VerifiedMember` entity
* the 3 dev wallets get automatically added as verified members

Resolves #1721

### DIFF
--- a/scripts/create-data.js
+++ b/scripts/create-data.js
@@ -130,6 +130,15 @@ const createColonyEtherealMetadata = /* GraphQL */ `
   }
 `;
 
+const addVerifiedMember = /* GraphQL */ `
+  mutation AddVerifiedMember($input: CreateVerifiedMemberInput!) {
+    createVerifiedMember(input: $input) {
+      userAddress
+      colonyAddress
+    }
+  }
+`;
+
 /*
  * Queries
  */
@@ -427,7 +436,8 @@ const createColony = async (
 
   const gas = await signerOrWallet.provider.estimateGas(populatedTransaction);
   populatedTransaction.gasLimit = gas;
-  const signedTransaction = await signerOrWallet.signTransaction(populatedTransaction);
+  const signedTransaction =
+    await signerOrWallet.signTransaction(populatedTransaction);
   const hash = utils.keccak256(signedTransaction);
 
   // create the colony
@@ -450,14 +460,18 @@ const createColony = async (
   );
 
   if (colonyQuery?.errors) {
-    console.log('COLONY ETHEREAL DATA COULD NOT BE CREATED.', colonyQuery.errors[0].message);
+    console.log(
+      'COLONY ETHEREAL DATA COULD NOT BE CREATED.',
+      colonyQuery.errors[0].message,
+    );
   } else {
     console.log(
       `Creating colony ethereal data { name: "${colonyName}", creationTransactionHash: "${hash}", version: "${currentNetworkVersion.toString()}" }`,
     );
   }
 
-  const colonyDeployment = await signerOrWallet.provider.sendTransaction(signedTransaction);
+  const colonyDeployment =
+    await signerOrWallet.provider.sendTransaction(signedTransaction);
   const colonyDeploymentTransaction = await colonyDeployment.wait();
 
   await delay();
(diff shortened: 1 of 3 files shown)


amplify/backend/function/removeVerifiedMembers/src/event.json
```diff
@@ -0,0 +1,5 @@
+{
```
jakubcolony, line 1: this file can be deleted

amplify/backend/function/addVerifiedMembers/src/graphql.js
```diff
@@ -0,0 +1,35 @@
+module.exports = {
+  createVerifiedMember: /* GraphQL */ `
+    mutation CreateVerifiedMember($input: CreateVerifiedMemberInput!) {
+      createVerifiedMember(input: $input) {
+        colonyAddress
+        userAddress
```
jakubcolony, line 6: In the schema, the return type for this function is `[VerifiedMember]`, but in reality it's whatever subset of fields you fetch in this mutation. And because one of the fields of `VerifiedMember` is of `User` type, you would also need to query all the fields of the `User` model, and keep updating them whenever they change - a bit of a maintenance hell.

This would fail:
```gql
mutation AddMember($input: AddVerifiedMembersInput!) {
  addVerifiedMembers(input: $input) {
    user {
      address
    }
  }
}
```

Try to avoid making lambdas dependent on GraphQL types as they can unpredictably change, maybe returning a boolean like the other mutation would be a better option?

jakubcolony (commented):
Hey @bassgeta, congrats on your first lambdas!

I have a few questions:
 - Have you considered authorization, ie. checking permissions of whoever is trying to add members?
 - Do you have a use case for those batch add/remove mutations? It's just not a pattern we've used in the past and I was wondering if the auto-generated mutations were not sufficient

Other than that, top notch job!

scripts/create-data.js
```diff
@@ -1164,14 +1178,13 @@ const createUserAndColonyData = async () => {
       }`,
     );
 
-    // verify users
+    // Add verified members to colony
```
rdig, line 1181: There's a issue with this, as the first three 3 dev accounts are no longer verified  with this configuration

Please account for them as well

rdig (changes requested):
Tested, and everything seems to be working as expected

Screenshot from 2024-01-29 23-55-51
Screenshot from 2024-01-29 23-59-19
Screenshot from 2024-01-29 23-59-33

I've only rejected this because of the creation script. Once that is handled, I'll approve it.

Nice work!

iamsamgibbs (approved):
Looks good to me! Ran it locally to confirm create-data script runs okay.

scripts/create-data.js
```diff
@@ -1165,20 +1179,20 @@ const createUserAndColonyData = async () => {
...
-          whitelistedAddresses: Object.keys(availableUsers.randomUsers).slice(
-            0,
-            noOfMembers,
-          ),
-        },
-      },
-      GRAPHQL_URI,
-      API_KEY,
+    await Promise.all(
+      Object.keys(availableUsers.walletUsers).map(async (userAddress) => {
```
rdig, line 1183: Just a heads up that I think `availableUsers` doesn't include the first dev wallet (leela), which means that the colony creator is not actually verified

Screenshot from 2024-01-30 14-35-06

This however only affects the creation script, and not the actual user / colony creation wizard in the CDapp

But it's nice to have it here as well :)

rdig (changes requested):
All good, except the verified members thing I've mentioned below

Nice work!

rdig (approved):
Re-tested this and same thing, the `leela` user shows up unverified.

Then I looked in the database (via graphql) and indeed `leela` is verified, however only a handful of users showed up as verified in the database, while the client showed all of them to be verified.

Screenshot from 2024-01-31 13-48-08
Screenshot from 2024-01-31 13-48-54

I think there is some UI nonsense that goes on here, but I think that's beyond the scope of this PR.

PR #1792, opened 2024-01-25
Fix additional colony invite codes failing

### LINKED ISSUE #1689: Additional Colony invite codes are failing
## Steps to reproduce
- Generate an invite code using the query below.
- Navigate to the app to test the invite code using the generated id `http://localhost:9091/invite/{COLONY_NAME}/{INVITE_CODE}`
- The invite code will show as valid.
- Clicking the Join button will fail.

```
mutation ColonyMemberInvite {
  createColonyMemberInvite(
    input: {colonyId: "{COLONY_ADDRESS}", valid: true, invitesRemaining: 4}
  ) {
    id
    invitesRemaining
    valid
  }
}
```

# Expected behavior
- Valid invite codes with invites remaining should allow members to join a Colony, even if not within the Colony metadata object.

# Current behavior
- When using invite codes that have been generated new and are not within the main colony metadata object, the code validates, but joining fails.

invite-code

jakubcolony:
## Description

The reason behind the issue described by Arren was that while it was technically possible to create multiple invite codes (entries of type `ColonyMemberInvite`), only the one generated at colony creation was considered valid.

Now the `validateUserInvite` lambda gets the invite object from the DB and checks for matching colony address. The existing 1 to 1 relation between `Colony > ColonyMemberInvite` might be confusing, so I put some comments to hopefully clear things up.

For testing instructions please refer to the original issue, make sure to exclude the `valid` field from the mutation like so:
```gql
mutation ColonyMemberInvite {
  createColonyMemberInvite(
    input: {colonyId: "{COLONY_ADDRESS}", valid: true, invitesRemaining: 4}
  ) {
    id
    invitesRemaining
  }
}
```

Resolves #1689


amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -1374,10 +1376,6 @@ type ColonyMemberInvite @model {
   Decrementing count of how many times this invite has been used
   """
   invitesRemaining: Int!
-  """
-  Whether the invite is still valid or not
-  """
-  valid: Boolean!
```
jakubcolony, line 1380: We can achieve the same functionality by setting `invitesRemaining` to 0

arrenv (approved):
@jakubcolony Very nice Jakub! Works like a charm.

warrencolliver (approved):
all looks good to me

iamsamgibbs (approved):
Works great! Tested with multiple invite codes and correctly checks invites remaining.

PR #1798, opened 2024-01-26
Fix: Verified and Non-verified users actions icon

### LINKED ISSUE #1780: Actions: Verified & non-verified users
## Description

When selecting a user from any dropdown field across supporting action types in master, there is a problem with the verification of users where both icons (the verified blue icon and the orange alert icon) are displaying on the front end at the same time.

image

This is a recent regression introduced to master. I am not able to reproduce this in production. 

The expected behavior is for only one icon to show matching the correct users validation type. 

## Correct UI behaviors

Example of correct unverified member:
image

Example of correct verified member: 
image

iamsamgibbs:
## Description

* Show verified / non verified icon correctly for transfer recipient
* (Also fixed transfer recipient thumbnail as there was some overlap)

**Changes** 🏗

* useUserSelect hook now returns isVerified property for each member
* Conditional UI changes now depend on selectedUser?.isVerified rather than unused showVerifiedUsers property
* useUserSelect hook now returns thumbnail property for each member

Verified user:
Screenshot 2024-01-26 at 17 04 53

Unverified user:
Screenshot 2024-01-26 at 17 05 03

Resolves #1780

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/UserSelect/UserSelect.tsx
+++ b/src/components/v5/common/ActionSidebar/partials/UserSelect/UserSelect.tsx
@@ -57,6 +57,7 @@ const UserSelect: FC<UserSelectProps> = ({ name }) => {
       ? {
           profile: {
             displayName: selectedUserOption?.label,
+            thumbnail: selectedUserOption?.thumbnail,
           },
           walletAddress: selectedUserOption?.walletAddress,
           isVerified: selectedUserOption?.isVerified,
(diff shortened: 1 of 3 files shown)


rdig (approved):
Code-wise I'm good with all the changes, and for functional, I've tested both cases and they seem to be working as expected

Screenshot from 2024-01-26 19-54-44
Screenshot from 2024-01-26 19-54-50

Thank you for fixing this Sam!

jakubcolony (approved):
LGTM, nice work @iamsamgibbs!

PR #1825, opened 2024-01-30
Fix incoming transfers: ingestor hash update

### LINKED ISSUE #1797: Incoming funds not working in production
## Description

This was experienced by a Beta user in the Deed3Labs colony `/deed3labs/` in production, where incoming funds were not picked up in the UI.

A payment of $1 in xDai and $10 in USDC was made to the Colony but not appearing in the Colony.

Both tokens are being managed by the Colony and and are visible in the Balance page. Filtering in the incoming funds table has no impact on the results shown.

- Balance can be seen here - https://gnosisscan.io/address/0xeB38458F4E88fCa8C69D39AC2D477009b5a3010D#internaltx
- xDAI transaction is here - https://gnosisscan.io/tx/0xcdb1752ec53b0f1dfb949579814d203a574f1fda3cedd9053350df55870b4ff1
- USDC transaction is here - https://gnosisscan.io/tx/0x0e914d5a5f0c622d8e1a9495a7ae141acb47f0c348e0eae9c07ee877874c8049

While, this can be claimed manually via a contract call I am holding off from doing this for now to give us a chance to see what the issue is and that it does not relate to the same issue experienced in the DApp.

jakubcolony:
## Description

This PR fixes a couple of issues discovered while investigating the incoming funds bug. For context, sending tokens to a colony triggers a `Transfer` event, which block-ingestor is supposed to pick up and process.

1. block-ingestor was not able to match the `Transfer` events with any of the existing event listeners and it was simply ignoring it.
2. We had a bug in setting up the listeners anyway, and it would only ever set up a single listener (presumably for the first colony it starts tracking).
3. When it finally made it to the correct handler, we failed to ensure the transferred token is in the DB, and it would make entire CDapp crash (BTW. This is another good arguments to break down the massive `GetFullColonyByName` query into smaller queries, since a single field failure will lead to a consistent 404). 

All the magic happens in [block-ingestor](https://github.com/JoinColony/block-ingestor/pull/165), this PR simply updates the docker image hash.

## Testing
You need a brand new token that you will transfer to a colony. To create a token, run `npm run truffle console` with the following commands line-by-line: 
```ts
t = await Token.new('Token name', 'TKN', 18)
t.unlock()
// mint some tokens (10 ether)
t.mint('10000000000000000000')
// transfer some tokens to your colony (0.5 ether)
t.transfer('<COLONY_ADDRESS>', '500000000000000000')
```

Head to the incoming funds page of your colony (make sure to refresh) and check "Unapproved" option in the filters:
image

The transfer should appear in the list.

Resolves #1797

### DIFF
--- a/amplify/backend/api/colonycdapp/schema.graphql
+++ b/amplify/backend/api/colonycdapp/schema.graphql
@@ -1263,9 +1263,9 @@ type Colony @model {
   # If you have a better idea, on how to merged them, I'll all ears..
   fundsClaims: [ColonyFundsClaim] @hasMany
   """
-  List of native chain token claims (e.g., Token 0x0000...0000: ETH, xDAI, etc.)
+  Native chain token claim (e.g., Token 0x0000...0000: ETH, xDAI, etc.)
+  This is not an array since only a single token type can be returned
   """
-  # This is not an array since only a single token type can be returned
   chainFundsClaim: ColonyChainFundsClaim
     @function(name: "fetchColonyNativeFundsClaim-${env}")
   """
--- a/docker/colony-cdapp-dev-env-block-ingestor
+++ b/docker/colony-cdapp-dev-env-block-ingestor
@@ -1,6 +1,6 @@
 FROM colony-cdapp-dev-env/base:latest
 
-ENV BLOCK_INGESTOR_HASH=8c49224e8b7553c13910c3db9acda3f294df875a
+ENV BLOCK_INGESTOR_HASH=53227363e3b768a48d65029b3c0f4eef9cb04dde
 
 # Declare volumes to set up metadata
 VOLUME [ "/colonyCDapp/amplify/mock-data" ]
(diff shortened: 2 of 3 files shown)


bassgeta (approved):

rdig (changes requested):
Although code-wise this is all good, and I was there when you made most of the changes, I can't actually get it to work locally. 
- I create tokens
- I transfer them to the colony
- Block ingestor picks up the transaction
- They never show in the UI

I'll test again tommorow with the ingestor running separately so I can investigate what's going on

ArmandoGraterol (commented):
I haven't been able to run this branch even after restarting :(

Screenshot 2024-01-31 at 18 18 31

I'll try again tomorrow tho just in case this is just me

jakubcolony:
@ArmandoGraterol my bad, that should be fixed now

@rdig If it persist let me know and we can debug it on your end. I've just tried on a clean env and it "works on my machine":
image

ArmandoGraterol (approved):
Thank you, Jakub! Working as expected for me 👌

rdig (approved):
Rebuilt everything and seems that now it works as expected

Screenshot from 2024-02-01 19-49-56

However, It seems that the UI does still have issues with tokens with less than 18 decimals.

Screenshot from 2024-02-01 19-50-23

Screenshot from 2024-02-01 19-50-09

But we can open a separate issue to deal with displaying those numbers. For now, I'll take that this now works as it should.

Very nice work Jakub!

PR #1829, opened 2024-01-31
fix: switch all sagas to use waitForTxResult

### LINKED ISSUE #1018: Switch all sagas to use `waitForTxResult`
There are some sagas (including expenditure sagas) that are using the following approach instead of `waitForTxResult` utility:
```js
yield takeFrom(createDomain.channel, ActionTypes.TRANSACTION_SUCCEEDED);

// put the success action
```

This means if there's an error or the transaction is cancelled, the error action will never be created, causing the UI to infinitely show loading state.

iamsamgibbs:
## Description

Switched all sagas from: `yield takeFrom(channel, ActionTypes.TRANSACTION_SUCCEEDED);` to: `yield waitForTxResult(channel);`

Previously, if there was an error or the transaction was cancelled, no error action was created so the UI would continually show a loading state. Now the waitForTxResult should properly throw an error in the event of an error or a cancelled transaction which should trigger an error state on the front end.

Resolves #1018

### DIFF
--- a/src/redux/sagas/actions/createDomain.ts
+++ b/src/redux/sagas/actions/createDomain.ts
@@ -28,6 +28,7 @@ import {
   createGroupTransaction,
   createTransactionChannels,
   getTxChannel,
+  waitForTxResult,
 } from '../transactions/index.ts';
 import {
   initiateTransaction,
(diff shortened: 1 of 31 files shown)


rdig (approved):
Tested most of the actions/motions that were touched here and everything seems to be working as expected.

Nice work Sam!

bassgeta (approved):
:ok_hand: lgtm

warrencolliver (approved):
Tested and can't see any issues. great work

Issue #1809, opened 2024-01-29
Refactor core Input component framework


jakubcolony:
Two components:
 - Input (low-level)
 - FormInput (tied to react-hook-form)

jakubcolony:
This issue should also fix the disabled state for `Input` component

PR #1831, opened 2024-01-31
Fix: permissions validation regression

### LINKED ISSUE #1778: Incorrect Decision Methods showing & validation regression
## Steps to reproduce

Tested in production.

1. Access a Colony as a member without any permissions.
2. Try to create an action using Permissions.
3. You will be able to select permission as decision method, even if you don't have permissions.
4. You are able to create the transaction, however, it is stuck in a pending state. Failing only in the console.

Note: there seems to be a regression of this PR https://github.com/JoinColony/colonyCDapp/pull/1395

## Expected behaviour

We want to be as pro-active as possible with showing what a user is able to do or not based on the permissions they have and the decision methods available. However, we don't want to have checks everywhere throughout the UI where action types are able to be triggered from. We also need to account for cases where users have different permission types in different teams.

Therefore, we would like to have a combination of approaches dependent on the user and whether or not they have permissions.

#### All users
- Remove permission check and blocking in the "Choose action type" dropdown:
image

#### Users with no permissions at all for the action
- It should show an error message that they don't have any permissions to take the action.
- All fields should be disabled.
https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-v5?type=design&node-id=3536-687292&mode=design&t=hJ7bZDuFrLX6bk0u-4
image

#### Users with at least one permission in any team related to the action
- No validation will appear on selecting the action.
- All fields will be active.
- As soon as enough information is available to determine permissions, it should validate on blur, such as after selecting the team and decision method.
- If the team has been changed, then it should reset the Decision method field.

#### Reputation weighted extension is enabled
- If other decision methods are active, for example the Reputation Weighted extension is installed, then the "Reputation" Decision method should be available as an option.
- Both users with and without permissions for the action will see all fields enabled.
- Users with no permissions at all can see only the Reputation decision method.
- Users with any permissions in the action can see both the Permissions and Reputation decision methods and validation happens for permission based on selections.
- Validation should still appear if there is no reputation in the team.

#### Non-connected users (may need a separate issue)
- Do same as "Users with no permissions at all for the action" but instead show non-logged in message.
- Message: "You are not signed in, please connect your wallet to create actions."
https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-v5?type=design&node-id=20469-232289&mode=design&t=hJ7bZDuFrLX6bk0u-4

## Actual behaviour

* You are able to select the Permission decision method, even if you don't have permission.
* You are able to click to trigger the action, although it fails, even without permissions.

image

image

### LINKED ISSUE #1866: Not able to Manage permissions using permissions
## Steps to reproduce

1. Go to create a Manage permissions action.
2. Fill in all the fields and select the "Permissions" decision type.
3. Click on "Update permissions" button.
4. You will get a validation error for a field that is hidden and should be required for this Decision method.

This essentially completely breaks Permissions, as I also have issues creating motions for Permission updates.

## Expected behaviour

* "Created In" should not be a required field for the permission decision method.

## Actual behaviour

* Validation is happening on a field that is hidden and should not be required for Permissions decision method.

image

jakubcolony:
## Description

This issue has had its scope changed twice so it's possible that this PR contains some unrelated stuff, but the main point is to re-introduce the validation logic in a new form as outlined in the issue.

As usual, there is a fair bit of refactoring and moving stuff around as well.

Please note point 4 from the issue is not included in this PR:
> Non-connected users (may need a separate issue)
> Do same as "Users with no permissions at all for the action" but instead show non-logged in message.
> Message: "You are not signed in, please connect your wallet to create actions."
> https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-v5?type=design&node-id=20469-232289&mode=design&t=hJ7bZDuFrLX6bk0u-4

@arrenv I believe this can only be done after the colony gating is over? Right now you can't access the action sidebar without being logged in.

## Testing

For testing, please refer to the original issue.

Resolves #1778 
Resolves #1866


iamsamgibbs (changes requested):
This is a great start. The new permissions helpers and hooks are very nice.

I’ve noticed just a couple of things.

From the original ticket, when the reputation weighted extension is enabled: `Users with no permissions at all can see only the Reputation decision method.`

In this screenshot, the user has no permissions at all but can still see both decision methods:

Screenshot 2024-02-08 at 10 32 46

The “You don’t have the right permissions to create this action” error message shows when viewing already completed actions.

Screenshot 2024-02-08 at 10 16 48

(Might be out of the scope of the ticket, but the fields on the “Manage colony objective” action don’t grey out when fields are disabled.)

Screenshot 2024-02-08 at 10 14 34

src/components/v5/common/ActionSidebar/hooks/permissions/helpers.ts
```diff
@@ -0,0 +1,101 @@
...
+        ? [ColonyRole.Root, ColonyRole.Architecture]
+        : [ColonyRole.Architecture];
+    }
+    case ACTION.EDIT_COLONY_DETAILS:
+    case ACTION.MANAGE_COLONY_OBJECTIVES:
+      return [ColonyRole.Root];
+    case ACTION.UPGRADE_COLONY_VERSION:
+      return [ColonyRole.Root];
+    case ACTION.ENTER_RECOVERY_MODE:
+      return [ColonyRole.Recovery];
```
ArmandoGraterol, line 49: We could group all of the cases that use `Root` here.
```
case ACTION.MINT_TOKENS:
case ACTION.UNLOCK_TOKEN:
case ACTION.MANAGE_TOKENS:
case ACTION.EDIT_COLONY_DETAILS:
case ACTION.MANAGE_COLONY_OBJECTIVES:
case ACTION.UPGRADE_COLONY_VERSION:
  return [ColonyRole.Root];
```

ArmandoGraterol (changes requested):
Looking good comrade Jakub! I've noticed what Sam already pointed out but besides that, I noticed this is also not happening:

>Users with at least one permission in any team related to the action
>- If the team has been changed, then it should reset the Decision method field.


EDIT: Also this small detail:

Screenshot 2024-02-08 at 10 46 38

The colony description is greyed out when disabled but the name isn't. I think the description shouldn't be, no? At least that's what I infer from looking at how the other disabled prefilled fields look.

jakubcolony:
Thank you for the reviews. I believe I fixed all the issues reported apart from:

> If the team has been changed, then it should reset the Decision method field.

I think we should drop this requirement as it doesn't impact validation but would require yet another layer of knowledge as to which fields are needed to determine required permissions (changing any of them, not just team, would require decision method reset to be consistent).

I will also not fix the non-grayed out fields in this PR, as it's an issue with the `InputBase` component not showing disabled state correctly - I added a note to #1809 to fix that.

iamsamgibbs (changes requested):
Thanks for making those changes, previous issues all appear resolved now.

I've only noticed one new issue. With a user who has only "payer" permissions, select the "Manage permissions" action. The invalid permissions error message should appear at the top straight away (as it does with the other actions), but instead it only appears once you select a team.

Screenshot 2024-02-09 at 09 07 51

Screenshot 2024-02-09 at 09 08 06

jakubcolony:
@iamsamgibbs great spot, I made changes to how `useHasNoDecisionMethods` work. Ready for a re-test!

src/components/v5/common/ActionSidebar/partials/ActionSidebarContent/partials/NoReputationError.tsx
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  return (
+    <div className="mt-6">
+      <NotificationBanner
+        status="warning"
+        icon="warning-circle"
+        description={formatMessage(MSG.noReputationError)}
+        callToAction={
+          <Link to="https://docs.colony.io/use/reputation">
```
rdig, line 39: This needs to be URL constant

src/components/v5/common/ActionSidebar/partials/ColonyDetailsFields/ColonyDetailsFields.tsx
```diff
@@ -9,82 +9,94 @@ import ActionFormRow from '~v5/common/ActionFormRow/index.ts';
...
+        isDisabled={hasNoDecisionMethods}
+      >
+        <div className="flex items-center">
+          <ColonyAvatarField
+            name="avatar"
+            fileOptions={{
+              fileFormat: ['.PNG', '.JPG', '.SVG'],
+              fileDimension: '250x250px',
+              fileSize: '1MB',
+            }}
```
rdig, line 67: This should be some sort of constant as well, as I feel weird having them declared specifically in the component.

src/components/v5/common/ActionSidebar/partials/forms/BatchPaymentForm/BatchPaymentForm.tsx
```diff
@@ -32,27 +29,7 @@ const BatchPaymentForm: FC<ActionFormBaseProps> = () => {
...
-        <FormCardSelect
-          name="decisionMethod"
-          placeholder={formatText({
-            id: 'actionSidebar.decisionMethod.placeholder',
-          })}
-          options={decisionMethods}
-          title={formatText({ id: 'actionSidebar.availableDecisions' })}
-        />
-      </ActionFormRow>
+      <DecisionMethodField />
```
rdig, line 32: Wasn't the decision method component extracted in #1882 ? Or is this just a case of needing a rebase ?

rdig (dismissed):
Left some minor comments that I picked up on, otherwise this is all good.

Nice work!

src/components/v5/common/ActionSidebar/partials/forms/BatchPaymentForm/BatchPaymentForm.tsx
jakubcolony: That PR is still open, so I guess whoever merges first?

src/components/v5/common/ActionSidebar/partials/ActionSidebarContent/partials/NoReputationError.tsx
jakubcolony: These aren't my changes and I want to avoid bloating this PR any more with unrelated fixes

rdig: In that case, just make sure these changes aren't lost, and we keep track of them

ArmandoGraterol (approved):
Looks good to me! Nice work Jakub 🎉

src/components/v5/common/ActionSidebar/hooks/permissions/helpers.ts
ArmandoGraterol: @jakubcolony Could it be that perhaps you missed this suggestion? :)

iamsamgibbs (changes requested):
The issue with the `manage permissions` is fixed now! But I've noticed a couple of other error validation issues when switching between action types.

https://github.com/JoinColony/colonyCDapp/assets/34915414/e1f7373a-f9f1-4899-87c4-cd5c88931de7

After I stopped recording, I switched to an action I do have permissions for, and the error validation messages still include fields from the previous action.

Screenshot 2024-02-12 at 11 08 19

arrenv (commented):
Great work on this @jakubcolony, other then the same issue as @iamsamgibbs, all the logic seems to work. I will test again when it is in a ready state though as there is a lot of validation happening all over the place.

jakubcolony:
@iamsamgibbs thanks for having another look, I went through the issues you highlighted and fixed some of the fields showing in red color while they're in both error and disabled states (disabled should take precedence). 

All the other issues such as incorrectly validating when switching action type and showing error messages seem to be present in master.

iamsamgibbs (dismissed):
Great job @jakubcolony - all seems resolved now 👍

rdig (approved):
Approving, nice work Jakub

ArmandoGraterol (dismissed):

arrenv (commented):
Great work @jakubcolony, all working great for every combination I could think of.

arrenv (dismissed):

iamsamgibbs (approved):

arrenv (approved):
Re-approving

PR #1837, opened 2024-02-01
[Verified members] Feat: Implement remove verified members permissions

iamsamgibbs:
## Description

- [] TODO write description

## Testing

* Step 1. E.g. add reputation
* Step 3. E.g. Create motion

## Diffs

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

* Changes go here

**Deletions** ⚰️

* Deletions go here

## TODO

- [ ] Add TODOs


src/redux/sagas/actions/removeVerifiedMembers.ts
```diff
@@ -0,0 +1,152 @@
...
+      meta,
+    });
+
+    if (colonyName && navigate) {
+      navigate(`/${colonyName}?tx=${txHash}`, {
+        state: { isRedirect: true },
+      });
+    }
+  } catch (error) {
+    return yield putError(ActionTypes.ACTION_DOMAIN_CREATE_ERROR, error, meta);
```
bassgeta, line 140: I had the same typo in the `add` saga, we need `ACTION_REMOVE_VERIFIED_MEMBERS_ERROR` here not the domain one :D

bassgeta (changes requested):
just an inherited typo, otherwise goooooooood job :muscle:

src/redux/sagas/actions/removeVerifiedMembers.ts
iamsamgibbs: Good catch, thank you! All fixed up.

bassgeta (approved):

ArmandoGraterol (commented):
Code seems good to me, however, even after following the steps I can't seem to unverify members. They still appear as verified when I use the query. I don't know if it's just me tho that I'm doing something wrong or my dev env is being quirky.

Besides that, the transaction title is missing in the gas station:
Screenshot 2024-02-02 at 13 26 59

The transaction is stuck pending for me btw, although that might be because of the annotation?

iamsamgibbs:
Thanks @ArmandoGraterol ! I've fixed the editColonyByDelta title and sanity checked the functionality, it's working for me showing the "pending" state correctly and removing verified members as expected. (Wouldn't hurt to have another person test it though)

https://github.com/JoinColony/colonyCDapp/assets/34915414/05c46b0a-55ec-4a55-b1f4-c68884286931

jakubcolony (changes requested):
Hey @iamsamgibbs, really nice work! I did some clicking around and all seems to be working well.

Please update the block ingestor commit hash as per Discord chat.

Also, I'm aware that Kerry's doing some potentially breaking refactoring so it might be a good idea to mark this as blocked as it will trigger conflicts with your block-ingestor PR.

rdig (approved):
Looks good codewise, and works as expected.

This is not really related to your PR, but I just want to raise attention on the fact that if you verify an address that doesn't have a user profile created _(which is a valid use case)_, the query will fail _(if you also fetch the user profile)_

Here's some screenshots I took while trying this out:

User w/o account created, but not fetching the User connection:
Screenshot from 2024-02-06 17-18-26

User w/o account created, fetching it's `User` model connection
Screenshot from 2024-02-06 17-18-44

Fetching all users, while one of them doesn't have an account created:
Screenshot from 2024-02-06 17-19-06

As you can see it does return data, but it also returns an error, and I can't quite remember how Apollo deals with this exactly.

Anyway, just wanted to raise attention to this, and as far as this PR is concerned, all good!

ArmandoGraterol (approved):
Nice work comrade! Working as expected for me

jakubcolony (approved):
Great work @iamsamgibbs!

Issue #1836, opened 2024-02-01
Pre-filling recipient not working on non-members


arrenv:
**Teamwork task:** [Members: Pre-filling recipient not working on non-members](https://tw.auroracreation.com/#/tasks/15771889)

## Steps to reproduce

1. Award permissions to an address in a Colony that is not a member.
2. Navigate to the members page where that member should appear as a contributor.
3. From the kebab menu, select the option to "Make payment".
4. Take note of the "Recipient" field, which will not have an address pre-filled.

## Expected behaviour

* Using the "Make payment" option in the user kebab menu should pre-fill the "Recipient" field in all cases.

## Actual behaviour

* Using the "Make payment" option in the user kebab menu is not pre-filling the "Recipient" field if the listed Contributor is not a member.

![payment-not-filling-nonmember](https://github.com/JoinColony/colonyCDapp/assets/33682027/05b1a5bd-6364-40cc-adc9-cbfc5f1c2701)

PR #1849, opened 2024-02-02
Fix: prefilling recipient for non members

KornelKwiatkowski:
## Description

Implement the fix, where selecting the "Make Payment" option from the user's kebab menu automatically fills in the "Recipient" field, ensuring this functionality works for both colony members and non-members.

## Testing

Automatic filling of the "Recipient" field for users who are not colony members

Instructions for testing this branch:

* Step 1. Award permissions to an address in a Colony that is not a member.
* Step 2. Navigate to the members page where that member should appear as a contributor.
* Step 3. From the kebab menu, select the option to "Make payment".
* Step 4. Take note of the "Recipient" field, which will not have an address pre-filled.

## Diffs

**Changes** 🏗

https://github.com/JoinColony/colonyCDapp/assets/90605528/6a285d37-7f9d-4347-9599-49a198dccfb6

Resolves: https://github.com/JoinColony/colonyCDapp/issues/1836

### DIFF
--- a/src/components/frame/v5/pages/MembersPage/partials/MembersTabContent/hooks.tsx
+++ b/src/components/frame/v5/pages/MembersPage/partials/MembersTabContent/hooks.tsx
@@ -29,8 +29,7 @@ export const useMembersTabContentItems = (
   return useMemo(
     () =>
       items.map(({ userAvatarProps, ...item }) => {
-        const { user } = userAvatarProps;
-        const { walletAddress } = user || {};
+        const { walletAddress } = userAvatarProps;
 
         return {
           ...item,


jakubcolony:
Good spot @ArmandoGraterol

> LGTM 👍

> I did notice something that is not really related to this PR, so it could be it's own issue if we don't want to address it here:

> Screenshot 2024-02-02 at 16 13 09

> After making the payment to a user that doesn't have a profile, this is how the sidepannel looks. The avatar is not clickable so there's no way of knowing who the recipient is except for the avatar.

@arrenv not sure if it's already tracked or needs an issue?

arrenv:
> @arrenv not sure if it's already tracked or needs an issue?

This is not being tracked, at least I am not aware of one or have been able to find it. Looks like it will need an issue.

melyndav (approved):
Nice work @KornelKwiatkowski! 👍

PR #1748, opened 2024-01-18
Replace webpack with Vite

chmanie:
## Description

I'm _really_ sorry this got so big again but I wanted to provide you with the best possible developer experience! Hope you can forgive me.

**Breaking changes**

* Vite uses the index.html as the entrypoint. I moved it to the root for the least hassle (Vite expects it there)
* To start the dev server / build the production bundle (formerly known as `npm run webpack[:prod]`) is now `npm run frontend` and `npm run build` respectively to be agnostic of the build system in the future for these commands
* `.tsx` files are now only allowed _one_ export which must be a component
* `process.env` is now `import.meta.env` - Furthermore we have access to `import.meta.env.(DEV|PROD)` and [other env vars](https://vitejs.dev/guide/env-and-mode#env-variables)
* New `.env` regime!! 
 
### **Please read this!**

The .env variables are now set in different files. We have an `.env.example` file which defines all available variables and gives descriptions and example values for it. Then there's a `.env.local` which defines all these variables for the local development environment. Furthermore there is `.env.development` which contains the environment variables that do not exist on production. Types for all the production env variables are generated automatically. There is NO `.env` anymore! Please delete it locally!!

tl;dr:
- Delete your local `.env` file
- Variables that are needed in production need to go into `.env.example` and `.env.local`

**New stuff** ✨

* Vite is now in charge of creating the frontend builds. It's _fast_ :zap: and reduced the bundle size by more than 25%
* Faster production builds (more than 3x faster!)

**Changes** 🏗

* Use Vite for storybook as well

**Deletions** ⚰️

* Webpack :yawning_face:
* Unused modules and css cleanup
* I removed all svgo plugins as we do not need them anymore (we barely have custom svgs in there after the move to phosphor)
* Removed the different tsconfig and webpack configs
* Removed loads of dev dependencies

Resolves #1804

PR #1856, opened 2024-02-05
Maintenance: Remove all css.d.ts files

### LINKED ISSUE #1767: Remove `css.d.ts` files and research CSS modules <> Tailwind alternatives
We have been ignoring `css.d.ts` files using `.gitignore` for a while now and they should be properly disabled/deleted. 

@chmanie raised concerns over using CSS modules with Tailwind. The other part of this issue is to research alternatives to this approach (and either implement it or open a separate issue for it).

iamsamgibbs:
## Description

We have been ignoring css.d.ts files using .gitignore for a while now and any remaining files needed deleting. This PR removes all .css.d.ts files.

## Discussion

@chmanie raised concerns over using CSS modules with Tailwind. The other part of this issue is to research alternatives to this approach (and either implement it or open a separate issue for it).

The main issue with how we are currently use CSS modules with Tailwind is the use of ‘@apply’. There are quite a few instances where we are using @apply just to avoid having long lines of tailwind directly in the components. We should avoid doing this where possible as it will result in the CSS bundle size increasing: https://tailwindcss.com/docs/reusing-styles#avoiding-premature-abstraction

---

[src/components/v5/common/Pills/PillsBase.module.css](https://github.com/JoinColony/colonyCDapp/blob/7fcc807642c9de671db6cfd4be9c6086607bfe7c/src/components/v5/common/Pills/PillsBase.module.css)

^ This is an example of where we are using @apply badly. This should just be used inline in the component file instead.

---

[src/components/common/Extensions/UserHub/partials/ReputationTab/partials/PendingReputation/PendingReputation.tsx](https://github.com/JoinColony/colonyCDapp/blob/7fcc807642c9de671db6cfd4be9c6086607bfe7c/src/components/common/Extensions/UserHub/partials/ReputationTab/partials/PendingReputation/PendingReputation.tsx#L39)

```
<div className={styles.row}>
  <p className={styles.rowName}>
```

^ This is an example of where we are using @apply simply because the same class is used multiple times in the component. (Both styles.row and styles.rowName are used multiple names in the same file). This should just be a variable in the component instead.

---

[src/components/shared/Avatar/Avatar.module.css](https://github.com/JoinColony/colonyCDapp/blob/7fcc807642c9de671db6cfd4be9c6086607bfe7c/src/components/shared/Avatar/Avatar.module.css)

^ Here we have a bunch of different size options, which is used in the component file like this:

`const class = styles[‘xs’];`

We should instead have a ‘sizeClasses’ variable in the component file which is simply an object containing the different tailwind classes:

```
const sizeClasses = {  
  xxxs: 'relative w-[1rem] h-[1rem] rounded-full',
  xxs: 'relative w-[1.125rem] h-[1.125rem] rounded-full',
  …
}
```

And would use it like:

`const class = sizeClasses[‘xs’];`

It’s a small but important change as it means we get the tailwind classes added directly rather than a css module class.

We could keep these in their own .ts files rather than directly in the component if it bloats the file too much.

---

[src/components/shared/Extensions/Toast/Toast.module.css](https://github.com/JoinColony/colonyCDapp/blob/7fcc807642c9de671db6cfd4be9c6086607bfe7c/src/components/shared/Extensions/Toast/Toast.module.css)

^ This seems like a good example of how we should be using @apply, where a third party package is adding components with set classnames and we want to apply our own styles.

In this instance we should keep the file as is.

---

After reviewing this I would suggest we remove as many uses of @apply as possible by either using the tailwind style directly in the component, or creating a classes object. There are 226 uses of @apply across the project so it may be a big task to refactor all these uses.

We would need to decide whether it is acceptable to move class objects into their own .ts files or if they should always remain part of the component file.

We would need to decide how to handle situations where a module.css file is imported in multiple places. If we simply moved the tailwind to a variable in each component then if any changes needed to be made to the styling it would need to be updated in multiple untracked places so this is not a suitable solution. We could instead replace the module.css file with a .ts file which simply exports variables with the tailwind styles.

We should discuss this further and make a decision on whether this is the best approach.

## Diffs

**Deletions** ⚰️

* All .css.d.ts files

Resolves #1767


chmanie (approved):
Really nice investigation! I agree with your findings and conclusions and they make sense to me. Have you looked into whether we could use the `@layer` directive to help us out somehow?

iamsamgibbs:
Thank you!

I don’t think we would get much use from @layer. It lets you tell Tailwind which “bucket” a set of custom styles belong to, either `base`, `components` or `utilities`. It applies these buckets in order of specificity, so a class in `utilities` would override `components` which would override `base`. So it’s just a way of giving certain custom classes more specificity, you still hit the same issues as with @apply if you’re simply using tailwind classes in an @layer as it increases the bundle size.

iamsamgibbs:
Somewhat related, but this package is highly suggested to use with tailwind and prettier and I don't believe we are currently using it: https://tailwindcss.com/blog/automatic-class-sorting-with-prettier would want to check it plays nicely with clsx.

rdig (approved):
Ran, built and tested this locally. Builds fine, as well as I can't really spot any UI regressions. So that that means this is good to go.

Screenshot from 2024-02-05 19-00-13
Screenshot from 2024-02-05 19-00-27

I do like the idea of having automatic sorting via the `prettier-plugin-tailwindcss`  plugin, but that should be done via  separate issue, as not bloat this more than it is.

ArmandoGraterol (dismissed):
Nice job 🎉 Cleaning the CDapp one step at a time :)

chmanie (changes requested):
Actually I have one more thing. While you deleted all the d.ts files, they are still being re-generated.
Please see this commit to see how to disable that in webpack: https://github.com/JoinColony/colonyCDapp/pull/1748/commits/553b095afbb61b34e6f41a01a91d1cf5e0d8cb7e#diff-26b9fc42ec181eee49b1258fab5ad305a630d25e505c747c22ed26f667d7084dR78

Please also remove d.ts file from the .gitignore. Thanks!

iamsamgibbs:
Good catch! I've removed the @teamsupercell/typings-for-css-modules-loader package, updated the webpack files and gitignore.

rdig (approved):
Re-approved

chmanie (approved):
Great, thank you!

ArmandoGraterol (approved):
LGTM 👍

PR #1862, opened 2024-02-05
Delete more components. It never stops!

chmanie:
## Description

I found even more components that could be deleted.

## Testing

Test if app still builds (if that's fine there's no reason why it shouldn't work)

## Diffs

**Deletions** ⚰️

* Many components as well as the `relativetimeformat` polyfill (as this is supported by all browsers now)


iamsamgibbs (approved):
Seems good to me! Tested and everything still builds okay.

PR #1864, opened 2024-02-06
Use ref directly in forms instead of forwardRef

chmanie:
## Description

So this is an interesting one. We were overwriting the default react types to accommodate  generics in `forwardRef` components (probably taken from [this article](https://fettblog.eu/typescript-react-generic-forward-refs/). The problem here is that some third party libraries actually depend on the original react type definition (which makes sense to me). Generally, I think overwriting existing types (especially react ones) is a terrible idea for that exact reason (plus it always has to be adapted to newer versions, etc.).\

## Reviewing

This PR is successful if the forms still function properly.


rdig (changes requested):
I'm getting the following error when opening the action sidebar _(which renders the `ActionForm` component)_:
```
ActionForm.tsx:58 Warning: Form.ActionForm: `ref` is not a prop. Trying to access it will result in `undefined` being returned. If you need to access the same value within the child component, you should pass it as a different prop.
```

Screenshot from 2024-02-06 18-23-58
Screenshot from 2024-02-06 18-24-10
Screenshot from 2024-02-06 18-24-16

chmanie:
Ah a shame. Yeah I will have to rename them. Sorry, I'll fix this asap

chmanie:
Okay, I'm pretty sure I fixed it now. Sorry for the round-trip.

rdig (approved):
All good now Chris. Nice work!

PR #1865, opened 2024-02-06
Maintenance: Replace window.location.host with env variable

### LINKED ISSUE #1766: Replace `window.location.host`
There should be a new environment variable added to store the env-specific app URL. It should replace all `window.location.host` calls in the code.

iamsamgibbs:
## Description

Added new env variable for URL to replace the use of window.location.host and window.location.origin

New APP_URL global const variable added as URL constructor, so `window.location.host` and `window.location.origin` can be replaced with `APP_URL.host` / `APP_URL.origin`

Updates the places they were used and added the variable to the docker file.

## Testing

Add URL variable to your .env file and build.

```
URL="http://localhost:9091"
```

## Diffs

**Changes** 🏗

* window.location.host and window.location.origin replaced with URL env variable
* New APP_URL global const variable added

Resolves #1766


src/redux/sagas/utils/getNetworkClient.ts
```diff
@@ -25,7 +25,7 @@ const getNetworkClient = async () => {
 
   const reputationOracleUrl = process.env.REPUTATION_ORACLE_ENDPOINT
     ? new URL(process.env.REPUTATION_ORACLE_ENDPOINT)
-    : new URL(`/reputation`, window.location.origin);
```
rdig, line 28: This actually not a good change, since we were using `window.location.origin` as a fallback, which is still valid.

You still have to pass in a env variable `REPUTATION_ORACLE_ENDPOINT` to it, so if one's not set, chances are, the second is not set either.

Also, this is just a dev only use

src/routes/routeConstants.ts
```diff
@@ -56,4 +56,4 @@ export const LANDING_PAGE_ROUTE = '/';
 export const TX_SEARCH_PARAM = 'tx';
 export const TEAM_SEARCH_PARAM = 'team';
 
-export const HOST = window.location.host;
+export const HOST = process.env.HOST || 'localhost:9091';
```
rdig, line 59: This should be removed entirely, and just use `process.env.HOST` directly, not set inside a `const` and without a fallback

src/auth/index.ts
```diff
@@ -4,7 +4,9 @@ import { ContextModule, getContext } from '~context/index.ts';
 import { isFullWallet } from '~types/wallet.ts';
 
 const authProxyRequest = async (urlPartial: string, options?: RequestInit) => {
-  const { host, origin } = window.location;
+  const host = process.env.HOST || 'localhost:9091';
+  const origin = process.env.ORIGIN || 'http://localhost:9091';
+
```
rdig, line 9: Same in this file as well, please remove the fallbacks as they're not needed anymore

.env.example
```diff
@@ -5,6 +5,9 @@ NETWORK=ganache
 # DOES NOT NEED TO BE SET WHEN DEVELOPING LOCALLY
 NETWORK_CONTRACT_ADDRESS=0x0000000000000000000000000000000000000000
 
+HOST="localhost:9091"
+ORIGIN="http://localhost:9091"
```
rdig, line 9: I don't think that the origin is actually needed. I think we can just get away with using just the host.

The communication protocol should still come from the client, `window.location.protocol`, to which we just append the host

rdig (changes requested):
Mostly good, however I left some notes which I want you to address first.

Thank you for putting this in

iamsamgibbs:
Thank you for the review! I've fixed all those requests. :)

jakubcolony (changes requested):
It seems like the majority of use cases require a full URL which leads us to having to combine the protocol and the host. 

How about storing app URL in env variable? 
- It can be used in links directly
- It's dead easy to get host from it:
```ts
const host = new URL(process.env.URL).host
```

What are your thoughts? This is also a question to @rdig I suppose

rdig:
Actually yeah, that's a really smart idea! :100: 

Screenshot from 2024-02-07 13-58-01

Just have the `URL` as an env variable _(`http://localhost:9091` for local dev, `https://app.colony.io` for prod)_

Then put: `const appUrl = new URL(process.env.URL)` inside the global `constants` file, and just import it whenever you need it.

You can get whatever you need out it. Origin, Host, Protocolo, etc

Really good pickup @jakubcolony

iamsamgibbs:
Great solution @jakubcolony - I've made these changes

rdig (approved):
All good now. Thank you Sam for putting up with all the back and fourth here!

Screenshot from 2024-02-07 20-34-06

One thing though, before merging, or rather just before pressing the merge button, make sure to do an announcement in the dev channel, to both our dev team, pagepro and the product team (since they run the app locally as well) to tell them that they need to update the `env` file

So as to not get a lot of people complaining about their local environment being broken

Another thought. After you merge this, QA will be broken until we introduce the new var to the online deployments, which we can't really do until @bogdan-1337 comes back.

So maybe wait until next week to merge this? Maybe?

jakubcolony (dismissed):
LGTM 👍

jakubcolony (approved):

rdig:
@iamsamgibbs heads up that @bogdan-1337 has now introduced the `URL` env variable on both QA and production, so that means that this can now be merged.

Do notify everyone when merging this, about updating they're respective `env` files

rdig (approved):

PR #1887, opened 2024-02-09
Fix: Completed action rich text description formatting

### LINKED ISSUE #1776: Rich text description formatting not working
## Steps to reproduce

1. Write a rich text description in an action.
2. Create a Simple payment action with a description.
3. Complete the action and view the description in the completed state.

Note: I believe this may be due to the difference between HTML and JSON handling of the rich text.

## Expected behaviour

* Rich text should be formatted correctly in both the creation and completed state.

## Actual behaviour

* Formatting is being stripped out in the completed transaction state.

image

iamsamgibbs:
## Description

Rich text should be formatted correctly in both the creation and completed state.

Previously the HTML from the description input field was being stripped out as part of the error validation, as such it was not being properly submitted. This has been replaced with a custom yup test.

Formatted the completed action description to match the action description input.

## Testing

1. Write a rich text description in an action.
2. Create a Simple payment action with a description.
3. Complete the action and view the description in the completed state.

## Diffs

**Changes** 🏗

* `ACTION_BASE_VALIDATION_SCHEMA` description max length validation replaced with custom yup test.
* Applied the same styles from the action description input to the completed action description

Screenshot 2024-02-09 at 16 25 58

Resolves #1776

### DIFF
--- a/src/components/v5/common/ActionSidebar/consts.tsx
+++ b/src/components/v5/common/ActionSidebar/consts.tsx
@@ -42,15 +42,10 @@ export const ACTION_BASE_VALIDATION_SCHEMA = object()
     description: string()
       .test(
         'isValidDescriptionLength',
-        ({ value }) =>
-          formatText(
-            { id: 'errors.title.maxLength' },
-            {
-              maxLength: MAX_ANNOTATION_LENGTH,
-              currentLength:
-                value !== undefined ? stripHTMLFromText(value).length : 0,
-            },
-          ),
+        formatText(
+          { id: 'errors.description.maxLength' },
+          { maxLength: MAX_ANNOTATION_LENGTH },
+        ),
         isValidDescriptionLength,
       )
       .notRequired(),
--- a/src/components/v5/common/ActionSidebar/consts.tsx
+++ b/src/components/v5/common/ActionSidebar/consts.tsx
@@ -22,6 +22,11 @@ export const actionSidebarAnimation: Variants = {
   },
 };
 
+function isValidDescriptionLength(description: string) {
+  const strippedDescription = stripHTMLFromText(description);
+  return strippedDescription.length <= MAX_ANNOTATION_LENGTH;
+}
+
 export const ACTION_BASE_VALIDATION_SCHEMA = object()
   .shape({
     title: string()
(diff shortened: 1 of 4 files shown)


arrenv (changes requested):
Nice one @iamsamgibbs, this is looking much better.

I am still not able to get Reputation decision method working locally, so could not test them.

The one thing I noticed is that empty lines seem to get ignored. Is there something where we can limit it to 1, but not allow more then 2 empty lines?

Having one seems to make it easier to ready, but allowing infinite means that people can just submit big empty descriptions.

### Before
image

### After
image

iamsamgibbs:
Thanks for testing @arrenv and picking up the spacing issue, I've pushed up a fix for this so it now matches the action sidebar description editor.

The way the editor handles new lines is to just make new paragraph tags, so I think if we want to restrict to just 2 lines of empty space, then it would be better to add this restriction on the action sidebar editor rather than on the completed view. This will ensure the styling on the completed action accurately reflects how the description was entered in the editor.

Screenshot 2024-02-12 at 09 42 15

Can also confirm the permissions completed action uses the same description component and looks the same.

rdig (approved):
Working as expected. Nice job!

Screenshot from 2024-02-12 21-46-24

Not related to this specifically, but I did found that if you try to write a code block, you do run into some "weird" interactions

[Screencast from 2024-02-12 21-48-30.webm](https://github.com/JoinColony/colonyCDapp/assets/1193222/d02499de-6a31-4d31-92c2-4f014096a3fc)

arrenv (changes requested):
The spacing works much better, thanks for that @iamsamgibbs.

> The way the editor handles new lines is to just make new paragraph tags, so I think if we want to restrict to just 2 lines of empty space, then it would be better to add this restriction on the action sidebar editor rather than on the completed view. This will ensure the styling on the completed action accurately reflects how the description was entered in the editor.

Do you have an example of this?

This is what we are trying to avoid.

#### Creation
image

#### Completed default (collapsed)
image

#### Completed (expanded)
image

iamsamgibbs:
@arrenv Ah, I was suggesting having the editor handle it as a possible solution rather than work I’ve actually implemented.

I’ve had a bit of a look today at how feasible it would be to limit the editor input to only allow a maximum of 2 lines of empty space and I think it would be doable and still the best solution but it would be best to create a new issue for this.

Issue #1617, opened 2023-12-22
Teams: As a user I can sort and search teams in TEAMS pages to quickly find information that interests me


AdrianDudko:
**Teamwork task:** https://tw.auroracreation.com/#/tasks/15716167
**Created by:** @AdrianDudko 

**Notes**
- check annotations in Figma
- check mobile/tablet design in Figma

**Description**
As a user, I can sort and search teams in TEAMS pages to quickly find information that interests me. 

There's no Teams page in the old UI. 

**Acceptance criteria**

General requirements:
- UI should match the Figma design

Search teams:
- searching for phrases in team names and descriptions
	
Sort Teams by Team funds and Reputation percentage:
- Highest to lowest
- Lowest to highest

**Out of scope**
- Teams page itself

**Design**

Filtering:
https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-V5?type=design&node-id=8532-446373&mode=design&t=Q903NY6Ejo3qCPgB-4
![image.png](https://s3.dualstack.eu-west-1.amazonaws.com/tw-eu-inline/437576/0/0/a6e3ff0490613bbaae3c52d810200b1a.png)
![image.png](https://s3.dualstack.eu-west-1.amazonaws.com/tw-eu-inline/437576/0/0/a6e6427fcb5a6363b761082cb33415a3.png)
![image.png](https://s3.dualstack.eu-west-1.amazonaws.com/tw-eu-inline/437576/0/0/a6ed1f3cb1b065402ff01160ba22f584.png)

PR #1889, opened 2024-02-12
feat: added teams page filter

CzarekDryl:
https://tw.auroracreation.com/app/tasks/15716167

## Description

- Added filters to teams page

Resolves [1617](https://github.com/JoinColony/colonyCDapp/issues/1617)

### DIFF
--- a/src/components/frame/v5/pages/TeamsPage/hooks.tsx
+++ b/src/components/frame/v5/pages/TeamsPage/hooks.tsx
@@ -3,16 +3,20 @@ import {
   // @todo: uncomment when decisions page will be ready
   // Handshake,
   Layout,
+  Star,
   Pencil,
   PresentationChart,
   Users,
+  Cardholder,
 } from '@phosphor-icons/react';
-import React, { useCallback } from 'react';
+import React, { useCallback, useState } from 'react';
 
 import { Action } from '~constants/actions.ts';
 import { useActionSidebarContext } from '~context/ActionSidebarContext/index.tsx';
 import { useColonyContext } from '~context/ColonyContext.tsx';
 import { useMemberContext } from '~context/MemberContext.tsx';
+import { ModelSortDirection } from '~gql';
+import { useMobile } from '~hooks';
 import { useActivityData } from '~hooks/useActivityData.ts';
 import useGetSelectedDomainFilter from '~hooks/useGetSelectedDomainFilter.tsx';
 import {
@@ -24,18 +28,20 @@ import {
   TEAM_SEARCH_PARAM,
 } from '~routes/index.ts';
 import Numeral from '~shared/Numeral/index.ts';
+import { convertToDecimal } from '~utils/convertToDecimal.ts';
 import { formatText } from '~utils/intl.ts';
 import { getBalanceForTokenAndDomain } from '~utils/tokens.ts';
 import { ACTION_TYPE_FIELD_NAME } from '~v5/common/ActionSidebar/consts.tsx';
-import {
-  type TeamCardListItem,
-  type TeamCardListProps,
-} from '~v5/common/TeamCardList/types.ts';
+import { type TeamCardListItem } from '~v5/common/TeamCardList/types.ts';
 import Link from '~v5/shared/Link/index.ts';
 
 import { getMembersList } from '../MembersPage/utils.ts';
 
-export const useTeams = (): TeamCardListProps['items'] => {
+import { type TeamsPageFilterProps } from './partials/TeamsPageFilter/types.ts';
+import { TeamsPageFiltersField, type TeamsPageFilters } from './types.ts';
+
+export const useTeams = () => {
+  const isMobile = useMobile();
   const { colony } = useColonyContext();
   const selectedDomain = useGetSelectedDomainFilter();
   const { domains, balances, nativeToken, name: colonyName } = colony || {};
@@ -106,6 +112,14 @@ export const useTeams = (): TeamCardListProps['items'] => {
           members: getDomainMembers(nativeId),
           isMembersListLoading: membersLoading,
           reputation: reputationPercentage ? Number(reputationPercentage) : 0,
+          balanceValue: convertToDecimal(
+            getBalanceForTokenAndDomain(
+              balances,
+              nativeToken?.tokenAddress || '',
+              nativeId,
+            ) || 0,
+            nativeToken?.decimals || 0,
+          )?.toNumber(),
           balance: (
             <Numeral
               value={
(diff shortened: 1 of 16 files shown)


melyndav (changes requested):
Hey again @CzarekDryl thanks for working on those issues! A couple more:

1. Search heading on mobile should be capitals. 

image

2. I have the icon in the filter button as being globally 18x18 px to match the secondary buttons.
image

https://www.figma.com/file/l1dOM5qiQYwF0ElvKDqqjg/Design-System---Colony-v3?type=design&node-id=1782-13112&mode=design&t=blzmU3Z0fqsjGXuj-4

iamsamgibbs (changes requested):
Nice job! This is looking good to me. The only thing that I've noticed, is there isn't a "No results" state when there are no matches. I couldn't find anything in the design for this though, so may require some input from @melyndav 

Screenshot 2024-02-13 at 09 43 25

src/components/frame/v5/pages/TeamsPage/hooks.tsx
```diff
@@ -238,5 +252,110 @@ export const useTeams = (): TeamCardListProps['items'] => {
...
+      }
+      if (reputation?.descending) {
+        return b.reputation - a.reputation;
+      }
+    }
+
+    return 0;
+  };
+
+  const sortedTeams = teams.sort(sortTeam);
```
jakubcolony, line 292: `.sort` modifies the original array so please make sure you're using spread operator like so to avoid unexpected behaviour down the line: `[...teams].sort()`

jakubcolony (changes requested):
Thanks for your work @CzarekDryl, functionally all seems to be working well apart from the empty state mentioned by @iamsamgibbs.

Code-wise, I'm just wondering if the sorting logic could be made a bit simpler and extendable (if we want to add more fields to sort on). I think the state could be refactored to closer resemble the actual requirements (for example, only a single field can be sorted upon).

Do you think it's possible for the `useTeams` hook, instead of keeping track of this potentially ever-growing object:
```
funds: {
  ascending: false,
  descending: false,
},
reputation: {
  ascending: false,
  descending: false,
},
```
to simply keep track of the field it's sorted upon (funds/reputation) and the direction? That way the state will never grow even if you keep adding new fields to sort upon:
```
const sortState = { 
  field: 'funds',
  direction: 'asc',
}
```

You wouldn't need the `sortTeams` function to consider each field individually, instead it could be reduced to something like this:
```
if(sortState.direction === 'asc')
  return a[sortState.field] - b[sortState.field];
else 
  return b[sortState.field] - a[sortState.field];
```

This would potentially remove the need for the `allowSingleSelect` param too, as by modelling your state in this way, you inherently disallow having more than one field selected:
```
// changing from reputation asc to funds desc:
sortState = {
  field: 'funds',
  direction: 'desc'
}
// no need to "clear" reputation asc sorting!
```

The data we're keeping in state or passing around should be as lean as possible (because it can be mapped into whatever more complex object the UI needs in the UI layer). 

Those are some potential ideas (and they might be flawed), but please consider making the logic sorting specific rather than trying to fit it in within the filters. Thanks!

melyndav:
> Nice job! This is looking good to me. The only thing that I've noticed, is there isn't a "No results" state when there are no matches. I couldn't find anything in the design for this though, so may require some input from @melyndav
> 
> Screenshot 2024-02-13 at 09 43 25

@iamsamgibbs @CzarekDryl there will always be a team 'general' and the filters are actually sorting, so there should always be a result available.

melyndav:
> Nice job! This is looking good to me. The only thing that I've noticed, is there isn't a "No results" state when there are no matches. I couldn't find anything in the design for this though, so may require some input from @melyndav
> 
> Screenshot 2024-02-13 at 09 43 25

@CzarekDryl the empty state for searching as picked up by @iamsamgibbs is here - https://www.figma.com/file/PtHSWX6tMcNC3hO604tKMR/Teams?type=design&node-id=2486-7833&mode=design&t=5ae8oEhpqJKiqcsr-4

image

CzarekDryl:
@jakubcolony Thanks for checking. I did as you wrote, logic is simplified but I had to add a new component to the teams page.

iamsamgibbs (approved):
This looks great, thanks for adding the "no results" state.

I agree with jakub that `allowSingleSelection` is misleading and should be renamed.

melyndav (approved):
Another one smashed out of the park. Thanks @CzarekDryl! Approved.

melyndav (changes requested):
Sorry @CzarekDryl I just picked up that the active filters are missing. When a user has any type of filter, sort or search active it should show on desktop and mobile that a filter is active. 

image

src/components/frame/v5/pages/TeamsPage/types.ts
```diff
@@ -0,0 +1,11 @@
+import { type ModelSortDirection } from '~gql';
+
+export enum TeamsPageFiltersEnum {
```
jakubcolony, line 3: Being nit-picky but a small suggestion: `TeamsPageFiltersField`?

jakubcolony (dismissed):
Thanks for making changes @CzarekDryl, code-wise all looks good, nice work 👍

iamsamgibbs (changes requested):
Thanks for the changes, but there are some changes needed to how the active filter works.

On page load, the default "Funds: Highest to Lowest" filter is already selected in the dropdown but doesn't appear as the active filter.

Screenshot 2024-02-16 at 08 27 49

The 'x' button on the active filter is misleading as the default filter gets applied when one is removed, then the 'x' button does nothing on the default filter.

https://github.com/JoinColony/colonyCDapp/assets/34915414/1d9f8dd2-d3f8-43c8-8512-d9789e24a1c1

May need some guidance from @melyndav on how this is supposed to work.

melyndav (changes requested):
Hey Cezary @CzarekDryl thanks for all your continued work on this PR. 👍 

Just a minor change request, can we make sure the text in the blue active tabs that show when a sort has been selected have the same grammar as inside the dropdown menu.

As you can see, the H is capitalised in the blue tab but not the dropdown. 
image

Also the x icon should in the blue pill should be 12x12px as per Figma with 8px padding between the icon and the text.

image

melyndav:
@iamsamgibbs all pages with filters will have a default sort applied (rep, funds, member type etc), but this is not shown to the user. Once they change the sort properties, the active filter/sort pill is then shown.

iamsamgibbs (changes requested):
Thanks for the changes on this. Only one tiny thing, the padding between the text and the X should be 8px and it is currently 4px.

Screenshot 2024-02-19 at 08 56 43

src/components/frame/v5/pages/TeamsPage/types.ts
jakubcolony: This is still outstanding @CzarekDryl

jakubcolony (approved):
Re-approving

iamsamgibbs (dismissed):

iamsamgibbs (dismissed):

melyndav (dismissed):
Hey @CzarekDryl - looks like the desktop filters have carried over to mobile since these were added.

image

iamsamgibbs (approved):

src/components/frame/v5/pages/TeamsPage/partials/TeamsPageFilter/TeamsPageFilter.tsx
```diff
@@ -0,0 +1,260 @@
...
+import Header from '~v5/shared/SubNavigationItem/partials/Header.tsx';
+
+import { type TeamsPageFiltersField } from '../../types.ts';
+
+import {
+  type TeamsPageFilterRootProps,
+  type TeamsPageFilterProps,
+} from './types.ts';
+
+const RootFilter: FC<TeamsPageFilterRootProps> = ({
```
ArmandoGraterol, line 26: This component could be in its own file

ArmandoGraterol (changes requested):
The highest to lowest rep filter is active default, yet the label doesn't appear when you initially load the page, it only appears if I select that filter again.
Screenshot 2024-02-23 at 14 23 36

Whenever I open a submenu of the filter, it shows a warning, somewhere in that code there are elements that need keys:
Screenshot 2024-02-23 at 14 23 48

CzarekDryl:
@ArmandoGraterol Here is Melynda comment about this pill visibility - https://github.com/JoinColony/colonyCDapp/pull/1889#issuecomment-1948118991

ArmandoGraterol (approved):
> Here is Melynda comment about this pill visibility

My bad! I didn't see that comment. Looks good to me 👍 Thank you for the changes.

warrencolliver (approved):
looking good. Nothing to add from me

rdig (approved):
All good now

PR #1891, opened 2024-02-12
[Verified members] Implement remove verified members motion saga

### LINKED ISSUE #1732: [Verified members] Feature: implement the removed verified members motion saga
### Story

The manage verified members form should call this saga if the action is `REMOVE`, decision method is `Reputation` and initiate the transaction.

### Description

The saga receives an array of members to remove from verified members and the colony from which they need to be removed.
It should update the loading, success and error states accordingly.
We can use `manageVerifiedRecipients` as a reference, but the underlying logic will probably be very different.

## Implementation

The saga receives the following payload
```JSX
{
	colonyAddress: Address;
	colonyName: string;
	members: string[];
	annotationMessage?: string;
	customActionTitle?: string;
}
```
And builds the necessary transactions (TBD when we have more info after block ingestor work is completed).
Naturally, display loading/success/error states and on success redirect to `/${colonyName}?tx=${txHash}`.

- [ ] The saga receives the payload
- [ ] The transaction is shown as pending until it's complete
- [ ] The transaction is shown as completed and is shown as a `CompletedAction`
- [ ] If an error happened it is displayed to the user

iamsamgibbs:
## Description

This implements the option to unverify members via a motion.
Most of the heavy lifting has been done by `block-ingestor` in [this PR](https://github.com/JoinColony/block-ingestor/pull/171).

## Testing

* Step 1. Use the block-ingestor branch from above if the commit hash doesn't work for you
* Step 2. get already existing verified members in GraphiQL via 
```
getVerifiedMembersByColony(colonyAddress: "0x6D3cA18f13ED52f484344425C2e1680367C87778") {
    items {
      colonyAddress
      userAddress
      user {
        profile {
          displayName
        }
      }
    }
  }
```
* Step 3. Unverify a verified member by going to the colony dashboard (`localhost:9091/planex`), input their address and click "Unverify via motion"

* Step 4. Wait for the motion to pop up, reject it or fully support it
* Step 5. Run `curl -X POST --data '{"jsonrpc":"2.0","method":"evm_increaseTime","params":[300],"id":1}' localhost:8545` to advance the time by 5 minutes
* Step 6. Finalize the motion, if you supported it, run the query above, the user you've entered, should now be returned

## Diffs

**New stuff** ✨

* `removeVerifiedMembers` motion saga

Resolves #1732


src/redux/sagas/motions/removeVerifiedMembers.ts
```diff
@@ -0,0 +1,186 @@
...
+    const {
+      payload: { hash: txHash },
+    } = yield takeFrom(
+      createMotion.channel,
+      ActionTypes.TRANSACTION_HASH_RECEIVED,
+    );
+
+    setTxHash?.(txHash);
+
+    yield takeFrom(createMotion.channel, ActionTypes.TRANSACTION_SUCCEEDED);
```
rdig, line 146: Shouldn't this use `waitForTxResult` here as per #1829 ?

rdig (approved):
All good except the one comment I've left

ArmandoGraterol (approved):
Good work 👍

src/redux/sagas/motions/removeVerifiedMembers.ts
iamsamgibbs: Good catch! Made that change on both add and remove.

PR #1892, opened 2024-02-12
Add Google verification for Google Search Console 

chmanie:
## Description

We want the app to be properly indexed by Google as some people are finding it already anyways. For further questions please refer to @naynovi.

### DIFF
--- a/src/templates/index.html
+++ b/src/templates/index.html
@@ -7,6 +7,7 @@
       content="width=device-width, initial-scale=1.0, maximum-scale=1"
     />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    <meta name="google-site-verification" content="l1ibLfGChyPkho3mtmGABRKUe7ADfDpLl4iyj9iAnAI" />
     <title>Colony App</title>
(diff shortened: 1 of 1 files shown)


naynovi:
Hi there you are all amazing for picking up this thing for me. 
As @chmanie indicates, this is for verification with Google Search console. It helps us know how people are finding our app and measure that.

We can verify several different ways. HTML tag is done here, and if that's suitable we can be on our way. 

If prefer an alternative to the snippet we can: 

- Upload a small html file to the site (GitHub does not allow me to include here-- interesting) @rdig said @bogdan-1337 could help if this was the way. 
- Verify through DNS which would need to be done by @rdig through Search Console itself which would be kind of annoying to do, as I'd need to set him up on that. I'd prefer he wouldn't have to deal with it.

If you would prefer either of the above let me know and I'll let you know what you need for that. 
Thanks for your help with this team.

chmanie:
> I'm more favorable to verifying our assets through DNS. Let's do that.

Great, I didn't know that was an option! Closing this.

Issue #1629, opened 2024-01-01
Actions: Paragraph input fields within action types 


melyndav:
**Teamwork task:** [Actions: Paragraph input fields within action types](https://tw.auroracreation.com/#/tasks/15782295)

## Descriptions

The objective description works as exactly the same as the Colony description in the 'Edit Colony details' action. I.e the content should wrap and the chevron should be ignored.

## Figma

Link to example action type: https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-V5?type=design&node-id=3487-329307&mode=design&t=V9WB8knufbS8936p-4

KornelKwiatkowski:
hi @melyndav I'm a bit confused about this task :D

If we want to add a chevron to "Colony description", shouldn't this field behave like the "Description" field? I think as a user I would expect the fields that have chevron to work the same way. What do you think?

Just like in the video below:

https://github.com/JoinColony/colonyCDapp/assets/90605528/38ea6b20-cc61-4b5b-8e05-1c2134d8a094

arrenv:
> If we want to add a chevron to "Colony description", shouldn't this field behave like the "Description" field? I think as a user I would expect the fields that have chevron to work the same way. What do you think?

Let's go with the same as you have demonstrated in the video. We should still have the max length of 200 characters though i.e. `MAX_OBJECTIVE_DESCRIPTION_LENGTH`.

KornelKwiatkowski:
@arrenv At this moment, the maximum number of characters for the objective description field is 200. Is there a chance that the API will return a default value for this field with more characters? If so, should the text be cut to 200 characters?

arrenv:
> @arrenv At this moment, the maximum number of characters for the objective description field is 200. Is there a chance that the API will return a default value for this field with more characters? If so, should the text be cut to 200 characters?

Thank you for confirming, 200 should be fine, I will update the spec.

There is a chance that the API can return more characters, but it is hopefully unlikely. But in terms of the UI, showing more characters, it will break parts of the UI. So, yes please, perhaps the frontend can limit to a max of 250 characters?

KornelKwiatkowski:
@arrenv  So, should I change the `MAX_OBJECTIVE_DESCRIPTION_LENGTH` from 200 to 250?

KornelKwiatkowski:
I opened PR. Let me know if you want to change the maximum number of characters you wrote about :) I can do it on a separate branch

[...]

PR #1907, opened 2024-02-14
Fix/1629 paragraph input fields within action types

KornelKwiatkowski:
## Description

"Objective description" and "Colony description" fields in the action form are now unified.

## Testing

The tests consist in checking whether the "Objective description" and "Colony description" fields work in the same way.

Instructions for testing this branch:

* Step 1. Create "Edit Colony details" or "Manage Colony objective" action
* Step 2. Check if "Objective description" and "Colony description" fields work in the same way

## Diffs

**Changes** 🏗

- From now on, the text does not overflow outside the component.

Resolves: https://github.com/JoinColony/colonyCDapp/issues/1629

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/ColonyObjectiveFields/ColonyObjectiveFields.tsx
+++ b/src/components/v5/common/ActionSidebar/partials/ColonyObjectiveFields/ColonyObjectiveFields.tsx
@@ -2,10 +2,12 @@ import { Article, FileText, Percent } from '@phosphor-icons/react';
 import React, { type FC } from 'react';
 import { defineMessages } from 'react-intl';
 
+import { MAX_OBJECTIVE_DESCRIPTION_LENGTH } from '~constants';
 import { useColonyContext } from '~context/ColonyContext.tsx';
 import { formatText } from '~utils/intl.ts';
 import ActionFormRow from '~v5/common/ActionFormRow/index.ts';
 import FormInputBase from '~v5/common/Fields/InputBase/FormInputBase.tsx';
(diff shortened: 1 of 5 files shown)


warrencolliver (changes requested):
Thanks for your work on this, However I did pick up on 2 small things

1. when submitting a action, the description field isn't displayed in the action summary once submitted (this happens on both tasks)
Screenshot 2024-02-14 at 16 52 20

2.  If I use one long single word (no breaks) and contract the field the same issue occurs where the text doesn't wrap and continues off screen.
no wrap on no break

melyndav (changes requested):
Hey @KornelKwiatkowski this PR is introducing a lot of unintended issues including changing the main description field, adding elements that aren't intended to be in this field etc. 

What we want is:

1. The objective description works as exactly the same as the Colony description in the 'Edit Colony details' action. I.e the content should wrap and the chevron should be ignored. 

All other changes should be reverted.

KornelKwiatkowski:
@melyndav done :)
If you can, please update the task: https://github.com/JoinColony/colonyCDapp/issues/1629

ArmandoGraterol (approved):
Looking good! They both work the same

iamsamgibbs (approved):
This looks good to me! (Although I couldn't test how it looks in the motion state as the edit colony details motion isn't pulling through the data properly)

melyndav (approved):
Thanks @KornelKwiatkowski for making the changes to the scope, much appreciated. Works much better now. 👍

warrencolliver (approved):
Good work on the changes

Issue #1904, opened 2024-02-14
Joined Colonies not showing in production


arrenv:
## Steps to reproduce

This seems to be a me only issue in production, others seems to be able to see their Colonies, but it may be worth looking into anyway as to the cause.
In production Joined colonies are not showing at all for me, this is the same for newly created and newly joined colonies. I have not been able to test leaving and rejoining due to this issue - https://github.com/JoinColony/colonyCDapp/issues/1902

## Expected behaviour

* Joined colonies should be showing.

## Actual behaviour

* No joined colonies are showing.

Running the following query:
```
query MyQuery {
  getProfile(id: "0x1FdA4692515dfd6f3c957d32825A8eFc08d6123c") {
    id
  }
  getContributorsByAddress(contributorAddress: "0x1FdA4692515dfd6f3c957d32825A8eFc08d6123c") {
    items {
      colonyAddress
      colony {
        name
      }
    }
  }
}
```

Results in this response:

```
{
  "data": {
    "getProfile": {
      "id": "0x1FdA4692515dfd6f3c957d32825A8eFc08d6123c"
    },
    "getContributorsByAddress": {
      "items": [
        {
          "colonyAddress": "0x3d529059baA06b792fB2CaCfC7E85A468Ab777e4",
          "colony": {
            "name": "animal"
          }
        },
        null,
        {
          "colonyAddress": "0xE19Ea092dc06f35c46cAAbeE2BAf1458513FA85e",
          "colony": {
            "name": "privaterc5"
          }
        },
        null,
        {
          "colonyAddress": "0x51547E3F1A895c4b6a7b408564BbE96F64844c87",
          "colony": {
            "name": "spacetravel"
          }
        },
        {
          "colonyAddress": "0xeB38458F4E88fCa8C69D39AC2D477009b5a3010D",
          "colony": {
            "name": "deed3labs"
          }
        },
        {
          "colonyAddress": "0x6c49Fe2beA3A0661587146E118878777A3510829",
          "colony": {
            "name": "beta"
          }
        }
      ]
    }
  },
  "errors": [
    {
      "path": [
        "getContributorsByAddress",
        "items",
        1,
        "colony"
      ],
      "locations": null,
      "message": "Cannot return null for non-nullable type: 'Colony' within parent 'ColonyContributor' (/getContributorsByAddress/items[1]/colony)"
    },
    {
      "path": [
        "getContributorsByAddress",
        "items",
        3,
        "colony"
      ],
[...]

PR #1916, opened 2024-02-15
Fix: Unable to leave a colony

### LINKED ISSUE #1902: Not able to leave a Colony
## Steps to reproduce

1. Join a Colony.
2. Try to leave the colony.

image

## Expected behaviour

* You should be able to leave a Colony.

## Actual behaviour

* Trying to leave a Colony results in an Apollo 400 error.

image

iamsamgibbs:
## Description

Leave this colony button results in a 400 error.

Leaving a colony should also remove it from the 'joined colonies' list in the colony switcher and the 'join' button in the top right should also reappear.

## Testing

1. Join a Colony.
2. Try to leave the colony.
3. 'Joined colonies' list in the colony switcher should show correct list of colonies
4. 'Join' button should show after leaving the colony.

image

## Diffs

**Changes** 🏗

* 'Leave this colony' button no longer results in a 400 error (caused by not passing ColonyFragment to useColonySubscription)
* `GetContributorsByAddress` query now accepts an `isWatching` filter to ensure the joined colonies list only shows colonies a user is watching
* `getContributorsByAddress` and `getColonyContributor` caches are now cleared when watching / unwatching a colony. This ensures the 'joined colonies' list updates correctly and the 'join' button shows correctly.

https://github.com/JoinColony/colonyCDapp/assets/34915414/4f245c0d-6385-4177-8058-9fc539d09846

Resolves #1902

### DIFF
--- a/src/components/v5/frame/ColonyHome/partials/LeaveColonyModal/LeaveColonyModal.tsx
+++ b/src/components/v5/frame/ColonyHome/partials/LeaveColonyModal/LeaveColonyModal.tsx
@@ -2,7 +2,6 @@ import { WarningCircle } from '@phosphor-icons/react';
 import React from 'react';
 import { defineMessages } from 'react-intl';
 
-import { apolloClient } from '~apollo';
 import { useColonyContext } from '~context/ColonyContext.tsx';
 import { useColonyDashboardContext } from '~context/ColonyDashboardContext.tsx';
 import useColonySubscription from '~hooks/useColonySubscription.ts';
@@ -40,7 +39,6 @@ const LeaveColonyModal = () => {
       onConfirm={() => {
         closeLeaveColonyModal();
         handleUnwatch();
-        apolloClient.cache.evict({ fieldName: 'getContributorsByAddress' });
       }}
       icon={WarningCircle}
       buttonMode="primarySolid"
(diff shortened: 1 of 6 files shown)


iamsamgibbs:
May also resolve #1904 ? I've not been able to recreate the issue locally.

bassgeta (approved):
works as intended :ship:

warrencolliver (approved):
great job. everything is working as it should.

melyndav (commented):
Hey @iamsamgibbs is there anyway we can get the join button to show immediately in the UI without needing the user to refresh? There's no immediate feedback in the UI when the user leaves the colony until a refresh is forced.

iamsamgibbs:
> Hey @iamsamgibbs is there anyway we can get the join button to show immediately in the UI without needing the user to refresh? There's no immediate feedback in the UI when the user leaves the colony until a refresh is forced.

@melyndav I've just given this a test and it does show immediately the first time you leave the colony. However, it does seem that if you leave a colony, rejoin it, then leave again, in that scenario the join button won't appear without a refresh - but it is so unlikely a user would actually run into this situation. (makes sense why it would show up in testing though!)


https://github.com/JoinColony/colonyCDapp/assets/34915414/be474ef7-e19b-4d6b-ac27-3caba0385b22

melyndav:
> @melyndav I've just given this a test and it does show immediately the first time you leave the colony. However, it does seem that if you leave a colony, rejoin it, then leave again, in that scenario the join button won't appear without a refresh - but it is so unlikely a user would actually run into this situation. (makes sense why it would show up in testing though!)
> Screen.Recording.2024-02-19.at.10.21.36.mov

Nice! Thanks for confirming this happens @iamsamgibbs. Appreciate it. 👍

melyndav (approved):

PR #1918, opened 2024-02-15
Colony balance should include funds currently in expenditures' pots

### LINKED ISSUE #938: Colony balance should include funds currently in expenditures' pots
As per the title, we should include all the unclaimed balance in all the expenditures' pots when calculating colony total balance. 

This issue will be relevant during the refreshed UI wiring stage.

At the moment the expenditure balance is fetched using a lambda, so fetching the balances of all existing colony expenditures using this lambda will be pretty slow. The assignee should consider creating a new lambda to return a single balance for all expenditures or re-consider the approach of getting the balance via lambda in the first place (and instead, keep track of the expenditure's funding pot balance by listening to relevant events).

jakubcolony:
## Description

Expenditures are funded by moving funds from a domain's funding pot to an expenditure's funding pot. Because of this, the balances shown in the balance table do not include any funds that are in expenditures' funding pots and have not yet been claimed. 

As part of Advanced Payments project, we want to include funds held in expenditures in colony balances, and provide a way of getting the expenditures balances separately from the total balances.

This PR addresses those requirements by:
1. Refactoring the way expenditure balances are stored: Balances used to be fetched via a lambda which wasn't suitable for getting multiple expenditures at the same time. Instead, there are block-ingestor handlers to listen for events that modify the balance and update it in the DB accordingly.
2. Including unclaimed expenditure funds in the balances table.
3. Introducing `useColonyExpenditureBalances` hook that returns a map between token addresses and total balance.

## Testing

1. Create a new colony (you can use `scripts/create-colony-url.js`)
2. Mint some tokens.
3. Use the testing UI to create expenditure. The ID is incremental and starts at 1. (you can also run `listExpenditures` query to get the native ID of an expenditure). **Note**: This is not the same as expenditure database ID
4. Lock and fund the expenditure.
5. Navigate to the Balances page. The native token balance should not decrease.
6. Run `listExpenditures` query (making sure to include `balances` field) to confirm the balance has been updated. Note: The balance is stored in wei and includes the expected network fee.
7. Finalize the expenditure using the testing UI. Your wallet balance should increase by 10 ETH of the colony's native token.
8. The balance on the Balances page should decrease by 10 ETH + network fee.
9. The expenditure balance should be "0".

Resolves #938


bassgeta (approved):
Tried it out, it works, nice one!

PR #1288, opened 2023-11-09
Refactor routes

willm30:
## Description

Refactoring routes per #1261. 

When testing, please check all existing routes and flows still work, and that document titles are updated correctly.

The route `user/:username` is broken on `master`, and new routes such `streaming-payments` are yet to be wired in. All other routes I'm expecting to work and not be redirected to 404. 

I have tested all nested colony routes, transactions, account nested routes, create-colony, 404, and create-profile (although this doesn't appear to have been wired in to new UI).

**Changes** 🏗

- Refactored route paths
- Swapped hardcoded strings for constants

**Deletions** ⚰️

- Deleted ColonyHome and moved route components up into `Routes`, and logic into `ColonyHomeContext`

Contributes to #1261

PR #1921, opened 2024-02-15
Fix lost colony name validation rules

### LINKED ISSUE #1919: Colony Creation Accessibility
### tl:dr

We want to give users a choice on their colony creation URL. We should implement a few safeguards to protect our domain authority.

### Description

Users may choose nearly any URL they wish in the Colony creation process. Colony, nor the Users themselves, cannot change these later. We should prevent some of the worst terms being used to preserve our site authority and limit opportunities for malicious actions/spoofing attempts. 

### SEO reliability
Google downgrades site discoverability when domains are connected to URLs that look suspicious or may lead to harmful content. Robots have a basic vocabulary and are scared easily. Poor user-created URLs can impact the visibility and discovery of the application as a whole, which we will be unable to improve later. If we introduce a few filters it will prevent some of the worst terms being used and attached to our sitemap.

*Bad Words* Google doesn't like bad words image

*Underscores* 
image
https://developers.google.com/search/docs/crawling-indexing/url-structure

*Reserved text*
image
image
image
image
image

^these might not be issues but might be good to exclude

### Spoofing: 

An issue that has been raised by users previously
image

image

## How we could do it

- Bad words/Underscores: Add the following terms (https://github.com/coffee-and-fun/google-profanity-words/tree/main ) as well as _ to colony name regex 

So when 

```
const isNameTaken = createYupTestFromQuery({
  query: CheckColonyNameExistsDocument,
  circuitBreaker: isValidName,
});
```
is run, it returns `'error.colonyURL': 'This is not a valid colony URL'`

https://github.com/JoinColony/colonyCDapp/blob/0ef19a797b5f813cd56196b2bc09b4f5474eeb28/src/components/common/Onboarding/wizardSteps/CreateColony/validation.ts#L132 

- Spoofing: As above, add list of Country names (maybe from here https://github.com/annexare/Countries) to colony name regex to at prevent delicious opportunity to impersonate a government authority. Maybe include some of the capital cities or prominent locations in that list as well. 

- Reserved text: maybe we have a list of things we'd like to not allow, 404 and index seem like obvious candidates.

- I'm not sure how we can get around corporate impersonation unless we add a similar list of names. 

- [ ] Users receive an error notice when they enter prohibited text in the URL generation field.

rdig:
We introduced colony name validation rules during the implementation of #1288

But apparently, either a bad merge, or oversight removed the checks for the "protected" names, when ##1396 was merged _(it removed the `RESERVED_ROUTES` check from the `isValidName` validation schema helper.

Subsequently, during a routine maintenance PR, seeing as the `RESERVED_ROUTES` values were not used at all, anywhere, they were purged from the codebase via #1708

This PR restores both the list of protected names, as well as the validation logic when creating a new colony

### Current Colony Name Enforcement Spec

A colony name should be:
- alpha numeric, including uppercase
- up to 20 characters in length
- can include underscores `_` but **not** any other special character
- not contain any of the "protected" names: https://github.com/JoinColony/colonyCDapp/blob/cf42ede316a3452363f4249708b350eaa9a840f9/src/routes/routeConstants.ts#L62-L198

### Special Thanks

Shout out to @jakubcolony for helping track this down

---

Resolves #1919


iamsamgibbs (approved):
Looks good! All working as expected.

Screenshot 2024-02-16 at 08 17 55

warrencolliver (approved):
Looks good to me!

jakubcolony (approved):
👍

melyndav (approved):
Hey @rdig looks good from my end! Approved :)

Issue #1840, opened 2024-02-01
Agreements: As a user I can filter search, and sort AGREEMENTS 


arrenv:
**Teamwork task:** https://tw.auroracreation.com/app/projects/562144/tasks/15577568
**Created by:** @AdrianDudko 
**Accepted by:** @arrenv 

**Notes**
- check annotations in Figma
- check mobile/tablet design in Figma
- This functionality exists in OLD APP, but differs in design and flow.

## Description
A list view of all of the Agreements a Colony has made. With the ability to search, filter, and sort agreements to easily find the agreement you are looking for. Agreements should be easy to access with one click to open the agreement and view it in the Action panel.

## User Stories
- As a User, I want to be able search and filter agreements in a Colony to be able to easily find an agreement I am looking for.

## User Flows

- Filtering & Searching Agreements
    - User is able to click on the “Filter” button to open the filter menu.
    - User can select the filters or enter the search they want to apply and it will update the list.
    - Active filters will be visible at the top of the table.
    - Clicking the “x” on active filters and searches will reset the table to the default.

## Acceptance Criteria

- **General requirements**
    - Should be in alignment with desktop/tablet/mobile design from Designs.
    - “Decisions” is now called “Agreements”.
    - Should use existing components and styles where possible for consistency.
    - Should be visible regardless of whether the ‘Reputation weighted’ extension is installed and enabled or not. This ensures past agreements are visible if the extension is disabled.

- **Breadcrumb & Team filtering**
    - Teams filter should be available in the header breadcrumb.
    - Selecting a team should filter the table and show Agreements made in this Team.

- **List**
    - Filtering
        - Status (refer to Motion process [[Motion Process](https://www.notion.so/Motion-Process-e756d351dfbe4576ac63a6dd8fb103d1?pvs=21)](https://www.notion.so/Motion-Process-e756d351dfbe4576ac63a6dd8fb103d1?pvs=21) for more details on each.)
            - Staking
            - Supported
            - Opposed
            - Voting
            - Reveal
            - Passed
            - Failed
            - Escalated
        - Created date
            - Allow filtering by a date range using a date picker for the date range. Filtering by the created date of the action.
            - Date picker should allow month and year select, and have month day calendar view for selecting range.
    - Search
        - Allow simple string matching search.
        - Search by matching fields:
            - Title
            - Description

    - Pagination
        - Show up to 10 agreements on load.
        - Lazy loading on scroll.
            - Loading using skeleton loaders.
[...]

Issue #1876, opened 2024-02-08
Not able to "Finalize" agreements


arrenv:
## Steps to reproduce

1. Create a new Agreement.
2. Go through the whole process of supporting the Agreement.
3. Click on the "Finalize" button.
4. You won't be able to finalize or return stakes, and there will be an error in the console.

Note: There is no transaction that is being performed or needing to be performed other then the returning of stakes. The usage of the "Finalize" text here is keep it consistent with other actions.

## Expected behaviour

* When in the "Finalize" step, clicking the "Finalize" button should only return all user stakes.
* When all stakes have been returned, there should not be a finalize button anymore. 
* This should be the same process when the Agreement has failed. i.e. stakes can be returned when clicking "Finalize".

## Actual behaviour

* You are not able to finalize the agreement and stakes are not able to be returned.
* If you claim back stakes via the Userhub stakes tab and return to the finalize step of the agreement, the finalize button will still be visible despite there being no stakes left to be returned.

rdig:
So this needs some discussions, b/c as far as I have knowledge of, Agreements have no support in the ingestor at all. I have no idea why they're even enabled in the UI

This might actually explain why finalize is not working, since it's a different way of handling that event.

So if this is needed, we'll consider it a new feature, and we can estimate it and schedule it accordingly.

arrenv:
> So this needs some discussions, b/c as far as I have knowledge of, Agreements have no support in the ingestor at all. I have no idea why they're even enabled in the UI
> 
> This might actually explain why finalize is not working, since it's a different way of handling that event.
> 
> So if this is needed, we'll consider it a new feature, and we can estimate it and schedule it accordingly.

Just to clarify, this is the renaming of "Simple decisions". Was this not migrated to CDApp?

rdig:
Not as far as I know of, at least not on the ingestor side. I'll look into it to know for sure.

arrenv:
> @arrenv Upon investigation it looks like it's not possible to finalise agreements due to the contracts - it would be a wasted transaction so we just skip it.
> 
> We need to find a different approach to handle this

Yes, that is correct. There are no transactions that needs to be finalized. The expected behaviour is just to return stakes in the finalize step. It could change the button text to "Return stakes", but if that is not built in, then the finalize is fine.

warrencolliver:
[...]

PR #1927, opened 2024-02-16
feat: added agreements filters

CzarekDryl:
https://tw.auroracreation.com/app/tasks/15759957

## Description

- Added agreements page filtering

Resolves [1840](https://github.com/JoinColony/colonyCDapp/issues/1840)


melyndav (changes requested):
Hey @CzarekDryl nice work on this one!

1. When I filter by failed agreements, it does not show any results despite me having three failed agreements already (as per the screenshot)
image
 
 2. Also when I filter by the date (as you can see I have three from this month in the screenshot) it returns no values.
 
 
image
 
 3. When I filter by date, there are no active pills showing. But the status shows the active pills.
 
image

4. There are some problems with the number of active filters on mobile in the button. The alignment is off. 

image
 
 5. Mobile filter is missing the heading 'Filter & sort' as per Figma example below.
 
image

6. No search result states are missing for both desktop and mobile. Here is the screenshot from Figma:

image
 
 
 7. The search button has different heights on mobile compared to the other two buttons next to it.
 
image

melyndav (changes requested):
Hey @CzarekDryl appreciate all the work on the updates. Some more - 

1. I have 2 active filters but the mobile pill is showing 3

Here you can see two selected: 
image

But when the filter is closed, it shows 3
image
 
 2. This is very minor, but the outter light blue on the search box has a gap on the border radius
 
 
image

Thanks for the other fixes.

melyndav (dismissed):
Thanks for those further updates @CzarekDryl, this is approved from my side of things. 👍

ArmandoGraterol (changes requested):
Good chunk of work! Noticed some weird behavior:

- When I select a range, with the end range being the date of the only agreement I have the agreement doesn't appear:
Screenshot 2024-02-21 at 14 06 50
Screenshot 2024-02-21 at 14 07 00

- I created a motion and proceeded throughout the flow and the status filters were working as expected until I arrived to the finalize step:
Screenshot 2024-02-21 at 14 12 12
Screenshot 2024-02-21 at 14 12 31
Screenshot 2024-02-21 at 14 13 09

  The motion is finalizable but it's not appearing when the filter `Finalizable` is active. It does appear when the filter includes `Passed` status

CzarekDryl:
@ArmandoGraterol Here is data for finalize agreement. I think this needs to be handled somewhere else.
image

src/components/frame/v5/pages/AgreementsPage/FilterContext/FilterContext.tsx
```diff
@@ -0,0 +1,188 @@
...
+  handleDateFilterChange: (event: React.ChangeEvent<HTMLInputElement>) => void;
+  handleCustomDateFilterChange: (date: [Date | null, Date | null]) => void;
+  handleResetFilters: (category: string) => void;
+}
+
+export const FiltersContext = createContext<FiltersContextValue | undefined>(
+  undefined,
+);
+
+export const FiltersContextProvider: FC<PropsWithChildren> = ({ children }) => {
```
jakubcolony, line 39: The name of this file is missing the `s` in `Filters`

src/components/frame/v5/pages/AgreementsPage/FilterContext/FilterContext.tsx
```diff
@@ -0,0 +1,188 @@
...
+          pastMonth: false,
+          pastYear: false,
+          custom: undefined,
+        });
+      }
+      default: {
+        return undefined;
+      }
+    }
+  }, []);
```
jakubcolony, line 144: The category should be an enum on its own, we shouldn't rely on translated strings as identifiers. 

You can format the enum value to a translated string right when you are displaying them.

src/components/frame/v5/pages/AgreementsPage/FilterContext/FilterContext.tsx
```diff
@@ -0,0 +1,188 @@
...
+  });
+
+  const handleMotionStatesFilterChange = useCallback(
+    (event: React.ChangeEvent<HTMLInputElement>) => {
+      const isChecked = event.target.checked;
+      const name = event.target.name as MotionState;
+      const failedStatuses = [
+        'Failed',
+        'FailedNotFinalizable',
+      ] as MotionState[];
```
jakubcolony, line 58: I feel like this should be a matter for filters config, e.g. if you could set up a mapping between motion states and arrays of states they should select, it would be much easier to see what value is actually being set without digging through multiple layers of code. Getting rid of code handling specific edge cases is always nice.

```tsx
const MOTION_FILTERS = [
  [MotionState.Failed]: [MotionState.Failed, MotionState.FailedNotFinalizable],
  [MotionState.Passed]: [MotionState.Passed]
  // ...
]
```

jakubcolony (changes requested):
Nice work so far @CzarekDryl, just left you a suggestion about using i18n strings as keys

iamsamgibbs (changes requested):
Great work on this so far! Only a few issues I've noticed.

The filter count is wrong when "Failed" is the only option selected, should only show 1 instead of 2. (Possibly miscounting as it can be Failed / FailedNotFinalizable?)

Screenshot 2024-02-23 at 09 45 20

There is a loading spinner here. Recreate this by staking both sides, then voting for the motion to fail, reveal your vote then refresh the page.

Screenshot 2024-02-23 at 09 58 09

No results show when the "finalizable" filter is selected. (issue #1876 is related to finalizing agreements, so it could be that agreements are getting an incorrect motion state currently).

https://github.com/JoinColony/colonyCDapp/assets/34915414/7f7815ee-7ded-406c-b796-97792cb86d17

---

Two slightly related things I noticed whilst testing. Currently if you're on the agreements page and you create a new agreement, or progress an existing one, the agreements page does not update until there is a refresh. It would make for a better user experience if it updated on save. This would require a change to the wiring and refetching the agreements query.

Secondly, the same html issue that was previously being experienced here is on the dashboard agreements widget. We can open a new issue for this unless you can slip it into this PR.

Screenshot 2024-02-23 at 09 42 13

CzarekDryl:
@iamsamgibbs For refetching agreements I've added new [issue](https://github.com/JoinColony/colonyCDapp/issues/1989). Html issue is being fixed by @joanna-pagepro in other issue. 

I don't think that we can do something with finalizable state for now.

ArmandoGraterol:
> Motion state is set as Passed. I think this needs to be handled somewhere else.
> I don't think that we can do something with finalizable state for now.

The `motionState` is indeed the same for finalizable and finalized motions (`MotionState.Passed`) like you pointed out. Still, there is a way to differentiate the 2 already available that could be used by the filters:

Screenshot 2024-02-27 at 12 53 30

If `motionState === MotionState.Passed` and `!motionData.isFinalized` then that means the motion is `Finalizable`, otherwise if `isFinalized` is `true` then the motion has `Passed`

jakubcolony:
@ArmandoGraterol you're uncovering a deeper problem which is motion state trying to do too many things at once - describe both **state** and **outcome** - logically motion can be both finalizable and passed.

`getMotionState` util in `utils/colonyMotions.ts` would never return `Finalizable` as a state, instead it replaces `Finalizable` network state with either Passed or Failed.

As it stands now, the best option is probably to remove `Finalizable` state from `MotionState` enum and the filters.

jakubcolony (approved):
Nice work on this PR @CzarekDryl, I'm pre-approving for now but will need to come back regardless once you do a rebase.

iamsamgibbs (approved):
This is looking good now, thanks for resolving my previously raised issues.

I still get 0 results with the 'Finalized' filter so may be best to remove for now as Jakub suggested.

ArmandoGraterol (approved):
Looks good to me! 👌 I'm okay with removing `Finalizable` if it's not gonna work at all, at least for now.

PR #1914, opened 2024-02-15
feat: activity table filters

joanna-pagepro:
https://tw.auroracreation.com/app/tasks/15669197

- filters and search for activity table on activity page

**Acceptance criteria**
- Filtering
  - Search activity:
    - The search sound query both the action title (in dark bold) and the metadata information including the user who created the action (in bold blue).
    - All of the results shown in the example either have the title, metadata or user within the metadata in the search result.
    - After typing user should be abe to use X button to reset search
  - Filter by:
    - Action type:
      - all availabe action types
      - scroll on desktop/tablet
    - Status:
      - Staked, Objected, Escalated, Voting, Reveal, Finalizable, Passed, Failed
    - Date:
      - past hour, 24 hours, week, month, year
      - Datapicker for custom range
    - Decision method:
      - Multi-sig, Permission, Reputation
  - General requirements:
    - hide pagination if not needed
    - show empty state if theres no results to show

PR #1929, opened 2024-02-16
Feat/15669197 datepicker

joanna-pagepro:
I've created a separate PR for Datepicker to merge it before https://github.com/JoinColony/colonyCDapp/pull/1914 and not to block other tasks that require this component.

To check the component, please use Storybook.

This PR includes:

- `Datepicker` component with requested changes from #1914 
- tailwind package update
- `tailwind.config.js` file now includes `base`, `utilities`, and `components` from `main.global.css` to make `Storybook` work and to avoid errors related with global classes (https://github.com/tailwindlabs/tailwindcss/issues/5989#issuecomment-962048436)


src/components/v5/common/Fields/Datepicker/partials/DatepickerCustomRangeInput/DatepickerCustomRangeInput.tsx
```diff
@@ -0,0 +1,37 @@
...
+      <span className="flex justify-center items-center flex-shrink-0 text-gray-900 text-2 w-6 text-center">
+        -
+      </span>
+      <InputBase
+        placeholder={formatText({ id: 'calendar.to' })}
+        readOnly
+        onClick={onClick}
+        onFocus={onFocus}
+        value={endDate ? format(endDate, dateFormat) : ''}
+      />
```
rdig, line 32: Can't really find a spec for this, so maybe @melyndav can chime in, but should the inputs allow users to change the values manually, as to me they looks more for display purposes only 

Screenshot from 2024-02-16 18-56-56
Screenshot from 2024-02-16 18-57-03

This is especially relevant since the date format is quite specific, making it nearly impossible for the layman user to actually write this by hand

So what I'm saying is that we still keep the inputs, we just disable them

src/components/v5/common/Fields/Datepicker/Datepicker.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+  const resetValues = useCallback(() => {
+    if (selectsRange) {
+      setStartDate(startDateProp || null);
+      setEndDate(endDateProp || null);
+    } else {
+      setStartDate(selectedProp || null);
+      setEndDate(null);
+    }
+  }, [endDateProp, selectedProp, selectsRange, startDateProp]);
```
rdig, line 69: The props passed to these callbacks are _by designed_ made to be changed every single time the specific function is called, so memorizing them makes no sense in this context.

They should just be simple functions, as they don't require all that overhead.

src/components/v5/common/Fields/Datepicker/Datepicker.tsx
```diff
@@ -0,0 +1,180 @@
...
+          />
+        ) : (
+          <DatepickerCustomInput />
+        )
+      }
+      dateFormat={dateFormat}
+      popperClassName={clsx(popperClassName, 'max-w-[20.5rem] !z-[1000]')}
+      renderDayContents={(day) => (
+        <div className="react-datepicker__day-content">{day}</div>
+      )}
```
rdig, line 121: I'm really not a fan of mixing global classes and tailwind classes

src/components/v5/common/Fields/Datepicker/Datepicker.tsx
```diff
@@ -0,0 +1,180 @@
...
+        (!selectsRange && startDate)) && (
+        <div className="p-4 border-t border-t-gray-200 w-full flex items-center justify-between gap-2">
+          <Button
+            {...cancelButtonProps}
+            mode={cancelButtonMode || ('primaryOutline' as ButtonMode)}
+            onClick={(e) => {
+              resetValues();
+              calendarRef.current?.setOpen(false);
+              cancelButtonOnClick?.(e);
+            }}
```
rdig, line 150: Please rename this variable to variable to `event` and add the proper synthetic event type _(I believe it's `MouseEvent`)_

src/components/v5/common/Fields/Datepicker/Datepicker.tsx
```diff
@@ -0,0 +1,180 @@
...
+            mode={applyButtonMode || ('primarySolid' as ButtonMode)}
+            onClick={(e) => {
+              const onChangeArg = (
+                selectsRange ? [startDate, endDate] : startDate
+              ) as Parameters<typeof onChange>[0];
+
+              onChange(onChangeArg, undefined);
+              calendarRef.current?.setOpen(false);
+              applyButtonOnClick?.(e);
+            }}
```
rdig, line 170: Same here

src/components/v5/common/Fields/Datepicker/Datepicker.tsx
```diff
@@ -0,0 +1,180 @@
...
+            disabled={
+              !startDate ||
+              (!endDate && !!selectsRange) ||
+              (!!endDate && !!startDate && isAfter(startDate, endDate))
+            }
+            mode={applyButtonMode || ('primarySolid' as ButtonMode)}
+            onClick={(e) => {
+              const onChangeArg = (
+                selectsRange ? [startDate, endDate] : startDate
+              ) as Parameters<typeof onChange>[0];
```
rdig, line 165: If `selectsRange` is `false`, then `onChangeArg` will not be a tuple, so why are we accessing it's first index value in that case ?

Screenshot from 2024-02-16 19-13-42

src/components/v5/common/Fields/Datepicker/partials/DatepickerCustomHeader/DatepickerCustomHeader.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+import { formatText } from '~utils/intl.ts';
+import { range } from '~utils/lodash.ts';
+import InputBase from '~v5/common/Fields/InputBase/index.ts';
+
+import DatepickerYearDropdown from '../DatepickerYearDropdown/index.ts';
+
+import { type DatepickerCustomHeaderProps } from './types.ts';
+
+const years = range(1990, getYear(new Date()) + 1, 1);
```
rdig, line 22: The min and max year values should be configurable via either props or constants _(same as `DEFAULT_DATE_FORMAT`)_ with a fallback to the values you provided above

src/components/v5/common/Fields/Datepicker/partials/DatepickerCustomHeader/DatepickerCustomHeader.tsx
```diff
@@ -0,0 +1,141 @@
...
+      if (!isMatch(value, dateFormat)) {
+        return;
+      }
+
+      const parsedDate = parse(value, dateFormat, new Date());
+
+      setDateValue(parsedDate);
+    };
+
+  useEffect(() => {
```
rdig, line 63: This should actually use `useLayoutEffect`, as even though most of the time `useEffect` is preferable, in this case it's actually opposite as the changes you are making inside it are observable

src/components/v5/common/Fields/Datepicker/partials/DatepickerCustomHeader/DatepickerCustomHeader.tsx
```diff
@@ -0,0 +1,141 @@
...
+          onClick={decreaseMonth}
+          disabled={prevMonthButtonDisabled}
+          className={navigationButtonClassName}
+        >
+          <CaretLeft size={16} className="text-inherit" />
+        </button>
+        <DatepickerYearDropdown
+          years={years}
+          selectedYear={getYear(date)}
+          selectedMonth={monthDate.toLocaleString('en-US', {
```
rdig, line 98: To some degree this should be tied into `i18n` as internationalization and languages should also change the numbers, time and date formats to they're local regional equivalent.

But I guess that's a larger discussion, so might want to hold off for now.

src/i18n/en.json
```diff
@@ -1573,5 +1573,9 @@
     "teamsPage.menu.viewTeamDashboard": "View team dashboard",
     "teamsPage.links.agreements": "Agreements",
     "teamsPage.links.activity": "Activity",
-    "view.extension": "View extension"
-}
\ No newline at end of file
+    "view.extension": "View extension",
+    "button.apply": "Apply",
```
rdig, line 1577: This should be placed next to the other `button` message descriptors

rdig (changes requested):
Overall it's mostly there, however it still needs some changes before it can be merged. 

But on the bright side, a lot of good work went in, so nice job! :tada: 

Please be mindful of the comments as well as the Discord discussion about splitting this up: https://discord.com/channels/562263648173555742/998629903740182580/1208063419681603696

Better yet, get in touch with @bassgeta and figure out the proper way to do this.

package.json
```diff
@@ -143,7 +143,7 @@
     "svg-sprite-loader": "^4.2.7",
     "svgo": "^1.3.2",
     "svgo-loader": "^2.1.0",
-    "tailwindcss": "^3.3.1",
+    "tailwindcss": "^3.4.1",
```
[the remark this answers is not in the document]
joanna-pagepro: I don't see anything that can break the UI. The changes presented in this release include some utilities, that were requested by some of you in other PRs, like using `min-h-*` className with size variable. In many cases, tailwind's documentation is updated to reflect the changes from this update.
I won't upgrade anything else without asking you first. Sorry about this one.

tailwind.config.js
```diff
@@ -148,6 +148,138 @@ module.exports = {
   plugins: [
     require('@tailwindcss/typography'),
     require('@tailwindcss/container-queries'),
+    // https://github.com/tailwindlabs/tailwindcss/issues/5989#issuecomment-962048436
+    function ({ addUtilities, addComponents, addBase, theme }) {
+      addBase({
```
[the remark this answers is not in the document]
joanna-pagepro: Sometimes there was an error in the console, when you were running the project, that some class does not exists and to add it using the `@layer` directive and here's the explanation from the [issue that I've pasted in line 151](https://github.com/tailwindlabs/tailwindcss/issues/5989#issuecomment-962048436):
```
Again, what is happening is that your build tool is effectively running all of these processes in parallel:

tailwindcss -i global.css
tailwindcss -i module-one.module.css
tailwindcss -i module-two.module.css
tailwindcss -i module-three.module.css
When you realize this it is a bit less surprising that code in module-one.module.css can't "see" code in global.css. But if they share a config file and the code is in the config file, you are all good.
```
By moving everything that was using this `@layer` directive from `main.global.css` to `tailwind.config.js` I have resolved the problem described above. Because of this issue, I wasn't able to run Storybook and it would be hard to test Datepicker component, because it's not used anywhere in the app right now (in this PR).

src/components/v5/common/Fields/Datepicker/Datepicker.tsx
joanna-pagepro: I'm using this class with CSS Modules. With tailwind classes, it would be pretty hard to modify some elements based on other classes added to the Datepicker by default.

joanna-pagepro:
There are two separate components now.

@rdig I've changed the inputs that are added to datepicker header to be read-only - I think that you're right and that they shouldn't be editable.

bassgeta (approved):
looks good now, thanks for splitting them up!

rdig (approved):
Looks good, and I like the way you split up the datepicker / range datepicker components as it makes it more readable.

Good to go.

Screenshot from 2024-02-19 21-37-00

Issue #1903, opened 2024-02-14
Mobile QA: Members


melyndav:
**Teamwork task:** [Mobile QA: Members](https://tw.auroracreation.com/#/tasks/15789352)

## Description

This covers all issues identified during mobile QA on production for the members page (including contributors and all members tabs)

Link to mobile members design: https://www.figma.com/file/moME7kMfr5pdt6PYrfDtm6/Members?type=design&node-id=201-8572&mode=design&t=BIrjNY0DBzBjCegT-4 

**Issues**

**Contributors tab**

1. Kebab menu next to contributors title exceeds content container width
<img width="482" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/70daf666-a09e-49d5-a8ec-dac4982a4eda">

2. Individual kebab menus should also be the width of the main container.

<img width="404" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/2f4ba0c1-a7df-4fdc-941b-ac88558f3543">

3. Spacing between each card should be 1.5rem
<img width="406" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/cdd7a04c-68bf-47f6-adda-668d4b94ceff">

4. Content clips into the header upon scrolling 
<img width="478" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/f26bcd4a-26e0-4618-b34f-5dfe4c4ba8a4">

**All members tab**

1. Kebab menu next to all members title on the contributors page exceeds the content container width

<img width="556" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/05356a15-2d92-4e83-9cc4-f74e91acb39f">

2. Individual kebab menus should be width of main container as per every other menu similar across the app.

<img width="409" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/8c473b33-31f1-49d3-8f75-6793b64cd771">

### Both tabs:

1. Filter menus should be expanded by default upon opening each time unless the user has closed them. 
<img width="410" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/d5bb75ee-7486-4541-86b8-200e152c8e25">
 
2. Chevron is not correct colour to match Figma. Also size of the X and chevron are different and should match Figma.
[...]

PR #1937, opened 2024-02-19
fix: ui on members page

CzarekDryl:
https://tw.auroracreation.com/app/tasks/15789352

## Description

- Few changes to members page on mobile

Resolves [1903](https://github.com/JoinColony/colonyCDapp/issues/1903)


melyndav (changes requested):
Hey @CzarekDryl thanks for picking up these mobile issues.

1. I still see the kebab menu exceeding the container width on both member type tabs.

image
 
 
 2. Content is still clipping into the top header nav bar across both tabs
 
image
image


3. Reputation is missing 'sort by' in the title as per Figma. 

4. X icon in the search input isn't correct size.
image

5. When I search for a user on the contributors tab and there is no user that matches the search, the same empty search block should show on the contributors tab. This should also happen naturally for desktop as well (replacing the invite members dotted box when a false search is active)
image

melyndav (changes requested):
Hey again @CzarekDryl appreciate the updates from my last round of checks.

When I search and get no results, the UI is showing empty on both tabs. 

image

They should match Figma here across all sizes: https://www.figma.com/file/moME7kMfr5pdt6PYrfDtm6/Members?type=design&node-id=201-9520&mode=design&t=GJMtoLQ3GtUchGy4-4

ArmandoGraterol (commented):
Nice work man! Thank you for the changes, I only noticed that when I apply a custom filter and I go back to the filter menu, the options aren't checked as if the filter wasn't applied:

Screenshot 2024-02-21 at 14 41 19

This happens in the mobile version too. When I click on it, then it does recognize the filter is active and shows `Root` checked along with `Custom`

melyndav (changes requested):
Hey @CzarekDryl 

When I open the filter, it breaks the whole page due to the alignment of the filter menu. As you can see in the screenshot, content is pushed off to the left of the browser window. 

image

iamsamgibbs (approved):
Looks good to me! Only one thing (and it might be more of a personal preference thing) but I would expect clicking custom here to remove all custom filters, currently it does nothing.

https://github.com/JoinColony/colonyCDapp/assets/34915414/69e35c2a-6303-4b3c-932a-8f5ffae72dde

warrencolliver (approved):
tested the changes made and can't find any issues + I agree with Sam's idea about the custom filter - it provides better user experience

ArmandoGraterol (dismissed):
Looking good to me 👌 Thank you for the changes!

ArmandoGraterol (dismissed):

warrencolliver (approved):

rdig (dismissed):
Approving so we can get this merged

iamsamgibbs (approved):

warrencolliver (approved):

PR #1947, opened 2024-02-19
Fix non-members address not appearing on payment side pannel

### LINKED ISSUE #1858: Non-members not showing on completed transactions
Initially raised here - https://github.com/JoinColony/colonyCDapp/pull/1849 - thanks to @ArmandoGraterol 

## Steps to reproduce

1. Make a payment to an address that is not a member in the Colony.
2. Execute the action and view the completed action screen.
3. You will notice that the recipient address is not visible and there is no MemberPopover.

## Expected behaviour

* The "Recipient" should populate with the correct recipient address.
* The recipient address should be clickable, showing the `MemberPopover` with the correct details.

## Actual behaviour

* The address is not showing on the completed action screen.

image

ArmandoGraterol:
## Description

- This PR fixes the complete action side pannel not showing the address of a user with no profile. With the changes, now the shortened address should appear on the description and the recipient row.

Screenshot 2024-02-19 at 18 44 36

## Testing

- Just create a simple payment action with a user without a profile and check that we are showing the masked address.
- The action should continue to show the other cases correctly, with no regressions.

Resolves #1858

### DIFF
--- a/src/components/v5/common/CompletedAction/partials/SimplePayment/SimplePayment.tsx
+++ b/src/components/v5/common/CompletedAction/partials/SimplePayment/SimplePayment.tsx
@@ -2,9 +2,11 @@ import { UserFocus } from '@phosphor-icons/react';
 import React from 'react';
 import { defineMessages } from 'react-intl';
 
+import { ADDRESS_ZERO } from '~constants';
 import Tooltip from '~shared/Extensions/Tooltip/index.ts';
 import { type ColonyAction } from '~types/graphql.ts';
 import { formatText } from '~utils/intl.ts';
+import { splitWalletAddress } from '~utils/splitWalletAddress.ts';
 import UserAvatar from '~v5/shared/UserAvatar/index.ts';
 import UserPopover from '~v5/shared/UserPopover/index.ts';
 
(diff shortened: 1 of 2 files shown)


iamsamgibbs (changes requested):
Good start on this issue! The masked address is showing correctly in the completed action, but it doesn't seem to be clickable to get the MemberPopover to show.

src/components/v5/common/CompletedAction/partials/SimplePayment/SimplePayment.tsx
```diff
@@ -57,14 +60,15 @@ const SimplePayment = ({ action }: SimplePaymentProps) => {
         {formatText(MSG.subtitle, {
           amount: formattedAmount,
           token: action.token?.symbol,
-          recipient: recipientUser ? (
+          recipient: recipientAddress ? (
             <UserPopover
-              userName={recipientUser.profile?.displayName}
-              walletAddress={recipientUser.walletAddress}
+              userName={recipientUser?.profile?.displayName}
```
bassgeta, line 65: do we need to change the `userName` to be the address if the profile doesn't exist?

bassgeta (approved):
Nice :+1: just left a comment that I'm not sure is completely relevant

src/components/v5/common/CompletedAction/partials/SimplePayment/SimplePayment.tsx
ArmandoGraterol: Correct! If the user hasn't registered a profile yet, then we should show the address, masked or not. This applies to every other place in the CDapp.

iamsamgibbs (approved):
Thanks for the changes, this works great now! 🎉

I noticed this unrelated UI issue where there should be spacing between the avatar and the warning notice, but this can be a new issue.

Screenshot 2024-02-23 at 08 55 39

Screenshot 2024-02-23 at 08 56 18

PR #1953, opened 2024-02-20
[Advanced Payments] fix expenditure saga selected token decimals

### LINKED ISSUE #1905: Expenditure sagas should take into account selected token's decimals
The sagas listed below incorrectly use `DEFAULT_TOKEN_DECIMALS` instead of retrieving the decimals of the token the payout is in (note: this can be different from colony's native token):
 
- `createExpenditure`
- `editExpenditure`
- `createStakedExpenditure`

The issue is to refactor the sagas to accept token decimals for the selected token and use that when converting between units.

warrencolliver:
## Description

This PR fixes the bug that set the decimal value of every token to 18 by default, now the user can enter a decimal value when creating an expenditure and the amount will reflect the entered decimal. 

## Testing

A newly created token that doesn't have a decimal value of 18 shouldn't have a default decimal of 18 when creating an expenditure, instead it should be the same as the token decimal when the token is first created. this functionality needs wiring to get the token decimal without user input. The logic changes for `createExpenditure`, `createStakedExpenditure` and `editExpenditure` are the same so we only need to test `createExpenditure` to ensure the logic works.

* Step 1. Set up Truffle console
* Step 2. Create a token:
```ts
t = await Token.new('Token name', 'TKN', 12)
t.unlock()
// mint some tokens (10 ether)
t.mint('10000000000000000000')
// transfer some tokens to your colony (0.5 ether)
t.transfer('<COLONY_ADDRESS>', '500000000000000000')
```
* Step 3. Go to dashboard and enter the token address, token decimals and transaction amount and create the transaction
* Step 4. Go to GraphQl db and use this query, search for your token address or for the ID using the ID counter '_XX' there should be 32 transactions by default. this number increases by 1 with each new expenditure. so after the first expenditure is created, search for '_33'

```
query listExpenditures {
  listExpenditures {
    items {
      id
      isStaked
      slots {
        payouts {
          amount
          isClaimed
          tokenAddress
        }
      }
    }
  }
}
```

The amount seen here should reflect the amount you entered + the amount of decimals you entered. If you enter an invalid decimal amount e.g. "efjk", the decimal value should default to 18

## Diffs

**New stuff** ✨

Input fields and create expenditure button added to TmpAdvancedPayments

**Changes** 🏗

tokenDecimals added to each saga 

Resolves #1905


src/components/v5/payments/TmpAdvancedPayments.tsx
```diff
@@ -23,6 +23,11 @@ const TmpAdvancedPayments = () => {
   const { user } = useAppContext();
   const { networkInverseFee } = useNetworkInverseFee();
 
+  const [tokenId, settokenId] = useState('');
+  const [decimalAmount, setdecimalAmount] = useState('');
+  const [transactionAmount, settransactionAmount] = useState('');
```
jakubcolony, line 28: Setter functions should also be in camelCase.

src/components/v5/payments/TmpAdvancedPayments.tsx
```diff
@@ -66,6 +71,7 @@ const TmpAdvancedPayments = () => {
     createdInDomain: rootDomain,
     fundFromDomainId: 1,
     networkInverseFee: networkInverseFee ?? '0',
+    tokenDecimals: tokenDecimalAmount,
```
jakubcolony, line 74: This will only ever work if you have all payouts in one token. As we discussed, there can multiple payouts in multiple tokens (potentially with different decimals) so the decimals need to be a property of a payout, not payload.

jakubcolony (changes requested):
Nice work so far @warrencolliver, but since there can be multiple token decimals within an expenditure, passing a single value won't work. Please see my in-code comment for details.

src/graphql/generated.ts
```diff
@@ -8405,7 +8405,7 @@ export type DomainFragment = { __typename?: 'Domain', id: string, nativeId: numb
```
jakubcolony, line 8405: There are no schema or *.graphql files changes, this file shouldn't have been modified

src/redux/types/actions/expenditures.ts
```diff
@@ -89,6 +89,7 @@ export type ExpendituresActionTypes =
         expenditure: Expenditure;
         payouts: ExpenditurePayoutFieldValue[];
         networkInverseFee: string;
+        tokenDecimals: number;
```
jakubcolony, line 92: This is not needed here, right?

jakubcolony (changes requested):
Thanks for making the changes @warrencolliver, looks better but I still spotted some things that need addressing.

warrencolliver:
> Has this been pushed with `--no-verify`? It contains many build and linting errors.
> 
> While resolving conflicts, please ensure every single commit you fix has no errors.

no, a normal push was used (pushed within VSCode). I'll rebuild and go through these errors now

PR #1958, opened 2024-02-20
Store expenditure stake in the DB

### LINKED ISSUE #1031: Store unclaimed Staked Expenditure stakes in the database
Similarly to motion stakes, we should store the Staked Expenditures with unclaimed stakes in the DB.

In the new UI both motion and expenditure stakes will be displayed in the same list, so it makes sense for them to be a single model.

Since creating this issue, a new model was created to allow for this - `UserStake`. 

This issue is to:
- On expenditure staking, create a new `UserStake` DB entry.
- When stake is reclaimed, update the DB entry accordingly.
- If expenditure is cancelled and the creator punished, update the DB entry accordingly.
- Ensure the stake is shown in the User Hub together with a correct status.

jakubcolony:
## Description

This PR addresses the requirement of tracking the expenditures user staked on. It's achieved by creating and modifying `UserStake` entities in the database. 

Since testing this without Staked Expenditure extension is super tricky, it's been made available to install in the UI. There's also a bunch of new buttons allowing you to interact with the expenditure for testing purposes.

The stakes are showing up in User Hub stakes tab but they haven't been wired properly, so you won't see any info regarding the expenditure that was staked and the status will only switch between `Unknown` (not claimed) or `Claimed`. 

block-ingestor PR: https://github.com/JoinColony/block-ingestor/pull/177

## Testing

There are a few requirements to test, which are listed below. 

### Prerequisite
Install and enable Staked Expenditure extension. Activate all your tokens to have enough for staking.

### Creating `UserStake` in the DB
1. Click on the "Create staked expenditure" button. Note: once the saga completes, you will find the native ID of the newly created expenditure in the console.
2. Run `listUserStakes` query to confirm the stake has been created. Neither `isClaimed` nor `isForfeited` should be set to `true`. Depending on the reputation amount in your colony, the staking amount should be around 4 ethers.
```gql
query ListStakes {
  listUserStakes {
    items {
      id
      isClaimed
      isForfeited
      amount
    }
  }
}
```
The stake will also show in User Hub, with `Unknown` status.

### Updating the `isClaimed` property when stake is reclaimed
1. Grab native ID of a staked expenditure. Using the testing UI, lock, fund and finalize the expenditure, then reclaim stake.
2. Using the query above, confirm `isClaimed` has been set to `true` for the expenditure stake. The stake will also show in User Hub as `Claimed`.

### Updating the `isForfeited` property when expenditure is cancelled (and staker punished)
A bit of context first: A staked expenditure can be cancelled by someone with permissions/via a motion, and a decision can be made to punish the creator by forfeiting their stake. Whether this happened is indicated by the `isForfeited` property on `UserStake`.
1. After creating and locking a staked expenditure, click on the "Click and punish" button. Verify the `isForfeited` property has been updated.

Resolves #1031


src/hooks/useExpenditureStaking.ts
```diff
@@ -0,0 +1,82 @@
...
+  const { walletAddress = '' } = user ?? {};
+
+  const [stakeAmount, setStakeAmount] = useState<string>();
+
+  const { data, loading: reputationLoading } = useGetUserReputationQuery({
+    fetchPolicy: 'network-only',
+    variables: {
+      input: {
+        colonyAddress: colony.colonyAddress,
+        walletAddress: ADDRESS_ZERO,
```
bassgeta, line 25: intentional or do we want `walletAddress` here?

src/hooks/useExpenditureStaking.ts
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    const { stakeFraction } = extensionData.params?.stakedExpenditure || {};
+    if (!stakeFraction) {
+      return;
+    }
+
+    const requiredAmount = BigNumber.from(stakeFraction)
+      .mul(totalDomainReputation)
+      .div(BigNumber.from(10).pow(18))
```
bassgeta, line 60: is the pow(18) gonna be always a thing or do we want the staking token decimals here?

bassgeta (approved):
image
Left 2 comments that may or may not be relevant :D 
Additionally, should we maybe set `isForfeited` to `false` by default or is null fine?

Other than that, everything works, gooooood stuff :sunglasses:

src/hooks/useExpenditureStaking.ts
jakubcolony: Super non-obvious but passing `ADDRESS_ZERO` as walletAddress to this query will return reputation for the entire domain. If you passed an address of a particular user, you'd get their reputation amount.

src/hooks/useExpenditureStaking.ts
jakubcolony: great catch, it should be native token decimals

src/hooks/useExpenditureStaking.ts
bassgeta: makes sense, gotcha :+1:

iamsamgibbs (approved):
This is great, all works exactly as expected.

Issue #1933, opened 2024-02-16
Mobile QA: Balances


melyndav:
**Teamwork task:** [Mobile QA: Balances](https://tw.auroracreation.com/#/tasks/15795751)

## Description

This covers all issues identified during mobile QA on production for the balances page.

Link to mobile balance design: https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-v5?type=design&node-id=5226-352539&mode=design&t=T1eUuWXosqyNx3lt-4

Components: https://www.figma.com/file/l1dOM5qiQYwF0ElvKDqqjg/Design-System---Colony-v3?type=design&node-id=1782-33992&mode=design&t=XhbkiQKGmndBtCBQ-4

### Issues:

1. The mobile version of the balance table is not implemented.

Current:
<img width="398" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/ea0d8c52-234c-4b4d-af92-3a00153322ec">

Figma:
<img width="697" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/5e1b1ea0-10ef-4721-b836-320d8d7f1fac">

2. Buttons under the table title should align to the left. 

<img width="221" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/4542fdab-804d-4e17-9ddf-3a682f225b71">

3. Meatball menu triggers from table rows should extend the full width of the container as per the FIgma spec
<img width="411" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/b0a7c181-7cf0-4823-bb0b-e9c4d593e1a8">

Figma: https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-v5?type=design&node-id=5384-368520&mode=design&t=T1eUuWXosqyNx3lt-4

PR #1962, opened 2024-02-21
fix: balances table styles update

joanna-pagepro:
https://github.com/JoinColony/colonyCDapp/issues/1933

Updated mobile view:

- table styles
- spacing next to `Add funds to the Colony` removed
image

- meatball menu is full width now
image


warrencolliver (dismissed):
Great work, looks so much better now.

ArmandoGraterol (changes requested):
Good job! The only thing I could point out right now is that the locked icon seems to be really far away from the name of the token when it's long and goes to a new line when in the designs it is side by side:
Screenshot 2024-02-21 at 15 26 13

src/components/frame/v5/pages/BalancePage/partials/BalanceTable/hooks.tsx
```diff
@@ -15,22 +20,70 @@ import TokenAvatar from '../TokenAvatar/index.ts';
...
+  const columns: ColumnDef<BalanceTableFieldModel, string>[] = useMemo(() => {
+    const columnHelper = createColumnHelper<BalanceTableFieldModel>();
 
-  const columns: ColumnDef<BalanceTableFieldModel, string>[] = useMemo(
-    () => [
+    return [
       columnHelper.display({
         id: 'asset',
         header: () => formatText({ id: 'table.row.asset' }),
+        headCellClassName: isMobile ? 'pr-2' : undefined,
```
bassgeta, line 85: would it be possible to put these classNames inside of the `header` function? so we render a span with  this className or something

src/components/frame/v5/pages/BalancePage/partials/BalanceTable/hooks.tsx
```diff
@@ -15,22 +20,70 @@ import TokenAvatar from '../TokenAvatar/index.ts';
...
+        cellContentWrapperClassName: isMobile ? 'pr-2' : undefined,
```
bassgeta, line 86: would it be possible to put these classNames inside of the `cell` function? so we render a span with  this className or something

bassgeta (changes requested):
Looks good, if my 2 comments are relevant, I think readability would be nicer, feel free to lemme know if it's not doable

melyndav (dismissed):
Outside of Armando's feedback re: the lock icon, everything looks good. Please resolve this before merging.

joanna-pagepro:
@ArmandoGraterol @melyndav  I can't make the textarea smaller so that the lock icon will be closer to the text. Also, on a design it looks the same as implemented: 
image

src/components/frame/v5/pages/BalancePage/partials/BalanceTable/hooks.tsx
joanna-pagepro: You can return a string only 
image
Also, the value that I want to change is set on `th` element, so a `span` wouldn't change anything in this case.

src/components/frame/v5/pages/BalancePage/partials/BalanceTable/hooks.tsx
joanna-pagepro: I can't find anything in the docs. I used a function that we already have implemented to update the spacings on the wrapper.

bassgeta: can we try it out? it's probably not a huge thing, just add the className to the `cell` attribute like we already do in some cases
image

joanna-pagepro: But this is an additional element inside the cell's wrapper, so it won't work the way I need it to work

bassgeta (approved):

iamsamgibbs (approved):
Looks good to me!

warrencolliver (approved):

ArmandoGraterol (approved):
I didn't see that screen! Looks good to me then 👍

PR #1964, opened 2024-02-21
fix: agreement description, transfer funds field name

CzarekDryl:
https://tw.auroracreation.com/app/tasks/15800786

## Description

- Changed agreement action description length to 4000 and made it required
- Changed transfer funds field name to `Amount` from `Field`

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/forms/CreateDecisionForm/consts.ts
+++ b/src/components/v5/common/ActionSidebar/partials/forms/CreateDecisionForm/consts.ts
@@ -1,6 +1,6 @@
 import { type InferType, object, string, number } from 'yup';
 
-import { MAX_OBJECTIVE_DESCRIPTION_LENGTH } from '~constants/index.ts';
+import { MAX_ANNOTATION_LENGTH } from '~constants/index.ts';
 import { ACTION_BASE_VALIDATION_SCHEMA } from '~v5/common/ActionSidebar/consts.tsx';
 
 export const validationSchema = object()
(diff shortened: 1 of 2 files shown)


warrencolliver (changes requested):
Good work so far. The amount field validation now works, as seen here: 
Screenshot 2024-02-21 at 15 16 33

I've found two potential issues:

1. the agreement action description max length is now 4000 characters, however, this is the same for every action type. Should the max length be 4000 characters for every action type or just 'Create Agreement'?

2. it is also possible to create an agreement with an empty description by inputting text into the field, then removing and submitting with the empty desc field
emptydesctest

CzarekDryl:
@warrencolliver Melynda said that each action should have 4000 characters.

warrencolliver (changes requested):
I'm still able to submit an agreement with an empty description by clicking in the desc field and submitting

iamsamgibbs (changes requested):
Can confirm it is still possible to create an agreement with an empty description.

`/src/components/v5/common/ActionSidebar/consts` in `ACTION_BASE_VALIDATION_SCHEMA` description is set as `notRequired` which I assume is overriding the required option you've added.

CzarekDryl:
@iamsamgibbs @warrencolliver @ArmandoGraterol Thanks for checking, please test it once again. It's working for me now - 
image
image

warrencolliver (dismissed):
I'm unable to submit an empty description on agreement actions now, good work. 

This issue can be replicated on other action types but I don't think it was introduced within this PR so I will raise a separate issue.

iamsamgibbs (dismissed):
All good now! Thank you!

ArmandoGraterol (commented):
Thank you for the changes! Found another case worth covering.
Screenshot 2024-02-23 at 15 38 47
Screenshot 2024-02-23 at 15 39 05

What I did was add a space and it allowed me to submit, which results in a technically empty description, even if it's not really empty. We should remove all trailing spaces from the description to avoid this from happening.

warrencolliver (approved):
Looks like the issue raised by Armando has been resolved. good work

PR #1965, opened 2024-02-21
fix: currency conversion uses token decimals

### LINKED ISSUE #1936: Issue with currency conversion
## Description

Currency conversion balances are not showing correctly on balances.

This was noticed in the Deed3Labs colony in production (https://app.colony.io/deed3labs/balances). You will need to be a member to access it, or, it can be reproduced by transferring xDAI (0x0000...0000) or USDC to a colony.

image

Thoughts:
- I have the assumption the USDC conversion is just not set to abbreviate correctly and is manually showing the "M" in the value.
- The xDAI could be related to the API query for the contract address to the CoinGecko API.

## Expected behaviour

* You should see accurate balances in the user dominated currency.

## Actual behaviour

* Balances are not showing accurately in  the dominated currency.

bassgeta:
## Description

Currently when trying to fetch prices, we didn't move decimal points, so the entire value was way too high.
Now the `CurrencyConversion` component accepts `tokenDecimals` and modifies the number accordingly.

## Testing

* Step 1. E.g. add reputation
* Step 3. E.g. Create motion

## Diffs

**Changes** 🏗

* `CurrencyConversion` accepts `tokenDecimals` and multiplies the conversion rate with moved decimals

Resolves #1936

### DIFF
--- a/src/components/frame/v5/pages/BalancePage/partials/BalanceTable/hooks.tsx
+++ b/src/components/frame/v5/pages/BalancePage/partials/BalanceTable/hooks.tsx
@@ -92,6 +92,11 @@ export const useBalanceTableColumns = (
               />
               <CurrencyConversion
                 tokenBalance={currentTokenBalance}
+                tokenDecimals={
+                  row.original.token
+                    ? getTokenDecimalsWithFallback(row.original.token.decimals)
+                    : 0
+                }
                 contractAddress={row.original.token?.tokenAddress ?? ''}
                 className="text-gray-600 !text-sm"
               />
--- a/src/components/shared/CurrencyConversion/CurrencyConversion.tsx
+++ b/src/components/shared/CurrencyConversion/CurrencyConversion.tsx
@@ -1,5 +1,6 @@
 import Decimal from 'decimal.js';
 import { type BigNumber } from 'ethers';
+import moveDecimal from 'move-decimal-point';
 import React from 'react';
 
 import { useCurrencyContext } from '~context/CurrencyContext.tsx';
@@ -32,6 +33,9 @@ interface Props extends Omit<NumeralProps, 'value'> {
 
   /** Balance of token to be converted */
   tokenBalance: BigNumber;
+
+  /** Decimals of the token */
+  tokenDecimals: number;
 }
 
 const CurrencyConversion = ({
@@ -39,6 +43,7 @@ const CurrencyConversion = ({
   showSuffix = true,
   placeholder = '-',
   tokenBalance,
+  tokenDecimals,
   contractAddress,
   chainId,
   className,
@@ -59,7 +64,9 @@ const CurrencyConversion = ({
       value={
         conversionRate == null
           ? placeholder
-          : new Decimal(tokenBalance.toString()).mul(conversionRate)
+          : new Decimal(
+              moveDecimal(tokenBalance.toString(), -tokenDecimals),
+            ).mul(conversionRate)
       }
       {...rest}
     />


src/components/frame/v5/pages/BalancePage/partials/BalanceTable/hooks.tsx
```diff
@@ -92,6 +92,11 @@ export const useBalanceTableColumns = (
               />
               <CurrencyConversion
                 tokenBalance={currentTokenBalance}
+                tokenDecimals={
+                  row.original.token
+                    ? getTokenDecimalsWithFallback(row.original.token.decimals)
+                    : 0
+                }
```
rdig, line 99: This should just be 
```suggestion
                tokenDecimals={
                  getTokenDecimalsWithFallback(row.original.token.decimals)
                }
```

Since `getTokenDecimalsWithFallback` with fall back to `18` even if the first argument is `null`/`undefined`.

And in any case, `0` is bad value for decimals

src/utils/currency/contractPrice.ts
```diff
@@ -0,0 +1,99 @@
...
+    return 0;
+  }
+};
+
+const isContractPriceSuccessResponse = (
+  data: TokenContractPriceResponse,
+): data is TokenContractPriceSuccessResponse => {
+  if (
+    typeof data === 'undefined' || // undefined
+    data == null || // null
```
rdig, line 57: ?
```suggestion
    data === null || // null
```

src/utils/currency/contractPrice.ts
```diff
@@ -0,0 +1,99 @@
...
+    (typeof data == 'object' && Object.keys(data).length === 0) // empty object
```
rdig, line 58: Same here
```suggestion
    (typeof data === 'object' && Object.keys(data).length === 0) // empty object
```

src/utils/currency/contractPrice.ts
```diff
@@ -0,0 +1,99 @@
...
+  });
+};
+
+const extractContractPriceFromResponse = (
+  data: TokenContractPriceSuccessResponse,
+  contractAddress: string,
+  conversionDenomination: CoinGeckoSupportedCurrencies,
+) => {
+  try {
+    return data[contractAddress.toLowerCase()][
```
rdig, line 40: Why `toLowerCase` and not checksumming? Is it coming as a lowercase value from coingecko ?

rdig (changes requested):
Codewise looks good, however I noticed that the native chain token has disappeared

Screenshot from 2024-02-22 18-40-51
 from the balances list:

That list should include: Native Colony Token (CREDS), Native Chain Token (ETH) and any other tokens added to the colony (STK and GGG in this case)

It should look similar to this:
Screenshot from 2024-02-22 18-46-50

I'm mentioning this in this PR since I couldn't replicate this on master.

src/utils/currency/contractPrice.ts
bassgeta: I just copied this over, and double checked that yes, CoinGecko does actually return everything in lower case
should we rather do checksumming here?

rdig: Nope. If they return lowercase, then you've handled it here correctly.

src/utils/currency/currency.ts
```diff
@@ -1,115 +1,57 @@
...
-import {
-  buildAPIEndpoint,
-  convertTokenToCLNY,
-  fetchData,
-  mapToAPIFormat,
-} from './utils.ts';
+import { convertTokenToCLNY } from './utils.ts';
 
 // The functions defined in this file assume something about the shape of the api response.
 // If that changes, or if we change the api, these functions will need to be updated.
```
jakubcolony, line 17: Is this comment still relevant here?

src/utils/currency/contractPrice.ts
jakubcolony: Can probably also get rid of the comments, kind of self-explanatory

src/utils/currency/currency.ts
```diff
@@ -1,115 +1,57 @@
...
+      tokenName: networkToken,
+      conversionDenomination,
+    });
+  }
+
+  return fetchContractPrice({
+    contractAddress,
+    chainId,
+    conversionDenomination,
+  });
```
jakubcolony, line 46: Gotta nit-pick on naming here, so they're both fetching token price, but one is doing so by name, other by address, right? Could the naming reflect that?

src/apollo/cache/colony/tokens.ts
```diff
@@ -19,7 +19,7 @@ const tokensFieldCache = {
           token: {
             __typename: 'Token',
             ...DEFAULT_NETWORK_TOKEN,
-            id: ADDRESS_ZERO,
+            tokenAddress: ADDRESS_ZERO,
```
jakubcolony, line 22: I'm not too sure what having `id` instead of `tokenAddress` broke, but it does look intentional especially if you consider the comment next to the `UnaliasedColonyTokensItem` type:

image

Could you address that as a whole, and either remove the type and the comment or update it?

jakubcolony (changes requested):
Nice work @bassgeta, this whole currency fetching seems like a pretty tangled piece of code so well done tackling this. 

I'm also noticing chain token being gone from the balances table

src/apollo/cache/colony/tokens.ts
bassgeta: I will revert this for now so I don't inflict a lot of potential breaking changes :D
The cache is also already hardwired to work with the `id` field.

rdig (approved):
All good now, and working as expected _(I've tested native chain and USDC on Gnosis)_

Screenshot from 2024-02-28 20-22-54

Nicely done! :100:

jakubcolony (approved):
Looks good, nice work @bassgeta!

image

PR #1966, opened 2024-02-21
Fix: Inconsistent versions of colony network and extensions

### LINKED ISSUE #1838: Inconsistent versions of Colony network and extensions
## Steps to reproduce

There are various inconsistencies with Colony network versions and extension versions that are shown or are being picked up in the UI.

## Expected behaviour

* Versions should be consistent and accurate to what is installed or available to upgrade to.

## Actual behaviour

* Versions are inconsistent or are hardcoded and not correct.

### Showing version 14 is available when 13 was available, and it says you are on the latest version
image

### Showing version 6 and 10
image

### Showing version 9 when it shows 10 on the previous screen
image

### Showing version 5 when it shows 6 on the previous screen
image

iamsamgibbs:
## Description

There are various inconsistencies with Colony network versions and extension versions that are shown or are being picked up in the UI.

Previously, the action sidebar new version was hardcoded to be +1 from the current version. This PR gets the correct version from the context.

Previously, the extensions page would show the latest available version even if an older version was installed, this would result in inconsistencies with the extension details page which shows the current version. This PR shows the currentVersion if the extension is enabled.

Please someone suggest a better approach to this as it feels bad. extension.isEnabled feels like the better thing to check against, but typescript will complain that extension.currentVersion does not exist on AnyExtensionData as it only exists on the InstalledExtensionData and not on InstallableExtensionData.
```
                  version={
                    'currentVersion' in extension
                      ? extension.currentVersion
                      : extension.availableVersion
                  }

```

## Testing

This is a bit hard to properly test as it would require installing older versions of the colony network and extensions, it should be possible to verify by looking at the code.

Create an action to upgrade the colony, the new version row and action description should both show the correct latest version (14).

Extensions page will now show current version if installed rather than the latest available version

## Diffs

**Changes** 🏗

* Action sidebar now correctly shows latest colony contract version
* Extensions page now shows current version if installed rather than available version

Screenshot 2024-02-21 at 15 37 00

Resolves #1838

### DIFF
--- a/src/components/frame/Extensions/pages/ExtensionsPage/ExtensionsPage.tsx
+++ b/src/components/frame/Extensions/pages/ExtensionsPage/ExtensionsPage.tsx
@@ -4,6 +4,7 @@ import ExtensionItem from '~common/Extensions/ExtensionItem/index.ts';
 import { useSetPageHeadingTitle } from '~context/PageHeadingContext/hooks.ts';
 import useExtensionsData from '~hooks/useExtensionsData.ts';
 import { type AnyExtensionData } from '~types/extensions.ts';
+import { isInstalledExtensionData } from '~utils/extensions.ts';
 import { formatText } from '~utils/intl.ts';
 
 const displayName = 'frame.Extensions.pages.ExtensionsPage';
@@ -48,7 +49,7 @@ const ExtensionsPage: FC = () => {
                   title={extension.name}
                   description={extension.descriptionShort}
                   version={
-                    'currentVersion' in extension
+                    isInstalledExtensionData(extension)
                       ? extension.currentVersion
                       : extension.availableVersion
                   }
--- a/src/components/frame/Extensions/pages/ExtensionsPage/ExtensionsPage.tsx
+++ b/src/components/frame/Extensions/pages/ExtensionsPage/ExtensionsPage.tsx
@@ -47,7 +47,11 @@ const ExtensionsPage: FC = () => {
                 <ExtensionItem
                   title={extension.name}
                   description={extension.descriptionShort}
-                  version={extension.availableVersion}
+                  version={
+                    'currentVersion' in extension
+                      ? extension.currentVersion
+                      : extension.availableVersion
+                  }
                   icon={extension.icon}
                   extensionId={extension.extensionId}
                 />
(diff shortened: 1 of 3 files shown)


bassgeta (dismissed):
Code looks like it makes sense, if you want to rewrite the version value, you can extract it into a function `getExtensionVersion` which would check if extension is installed and if available version exists, but I don't have any qualms with the approach you've used here :D

src/components/frame/Extensions/pages/ExtensionsPage/ExtensionsPage.tsx
```diff
@@ -47,7 +47,11 @@ const ExtensionsPage: FC = () => {
                 <ExtensionItem
                   title={extension.name}
                   description={extension.descriptionShort}
-                  version={extension.availableVersion}
+                  version={
+                    'currentVersion' in extension
+                      ? extension.currentVersion
+                      : extension.availableVersion
+                  }
```
jakubcolony, line 54: `isInstalledExtensionData` type guard will help you here

jakubcolony (dismissed):
Great work @iamsamgibbs, looks like that'll fix all the issues, I left you a comment on how you could improve the `'currentVersion' in extension` check

bassgeta (approved):
nice one :sunglasses:

rdig (changes requested):
Versions show up correctly in the UI, both colony and extensions

Screenshot from 2024-02-23 17-24-04
Screenshot from 2024-02-23 17-24-13
Screenshot from 2024-02-23 17-25-08
Screenshot from 2024-02-23 17-25-24
Screenshot from 2024-02-23 17-25-30
Screenshot from 2024-02-23 17-29-32
Screenshot from 2024-02-23 17-26-05
Screenshot from 2024-02-23 17-26-12


However, I cannot / am not allowed to upgrade extensions at all. Upgrade button disabled by default.

Screenshot from 2024-02-23 17-26-57

So this needs further investigation.

iamsamgibbs:
@rdig I think it is not possible to upgrade to version 7 of OneTxPayment here because the disabled button check calls `isExtensionCompatible` from `@colony/colony-js` which checks if the version is included in the [oneTxPaymentIncompatibilityMap](https://github.com/JoinColony/colonyJS/blob/22179035ed7b28e0604dc5d06863755f248d97e5/packages/core/src/versions/OneTxPayment.ts#L15) If I've understood correctly, upgrading from version 5 to 6 should be possible, so I don't believe this to be a bug.

rdig:
@iamsamgibbs I'm incline to believe you, since I did indeed "fake" the version number.

I am however remembering this not working correctly in production, so I'll test again just to be sure.

rdig (approved):
You were indeed correct Sam! :100: 

If I use a "proper" version number the upgrade button seems to be working / behaving as expected

Screenshot from 2024-02-26 22-48-51

Nice work!

iamsamgibbs:
Thanks for looking into that and confirming!

PR #1973, opened 2024-02-22
Fix: Changes to files inside amplify folder are not being picked up by amplify container

### LINKED ISSUE #1972: Changes to files inside `amplify` folder are not being picked up by `amplify` container
## Steps to reproduce

1. With the dev env running, make some changes to `schema.graphql`, e.g. add a new field to one of the models.

## Expected behaviour

`amplify` container regenerates models.

## Actual behaviour

Nothing happens. This is because `amplify` container no longer uses mounted volume, so it doesn't have access to user-edited files after it's started.

The solution is to introduce a file watcher service that will copy any files changed in the `amplify` directory across to the docker container.

jakubcolony:
## Description

Please see issue for description.

## Testing

Run `npm run dev` and wait for the `amplify` container to start.

Make some changes in the `amplify` directory, e.g.: 
 - modify the schema
 - create a new lambda (`npm run amplify function add`)
 - add a new file

The files you add/update should be copied over to `amplify` container. You'll see a relevant message in the console. If you are Docker Desktop user, you can go to the container > Files and check they've been copied:
Screenshot 2024-02-23 at 16 37 11

For CLI people, I think you can use `docker exec -it amplify /bin/sh` to enter the container's shell and run a combination of `cd` and `ls` to see the copied files.

Resolves #1972

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -32,6 +32,14 @@ _Note: at the end of the install there's a post-install script that will recursi
 
 Copy `.env.example` and rename it to `.env`. You should not need to change any of the values to get CDapp running.
 
+### (Linux only) Increase the limit of watched files
+
+On Linux, you'll need to run the following commands to increase the default limit of watched files. Otherwise, the `watchAmplifyFiles` script will not work properly.
+
+```bash
+echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
+```
+
 ## Running the dev environment
 
 ```bash
--- a/package-lock.json
+++ b/package-lock.json
@@ -51,6 +51,7 @@
         "axios": "^1.4.0",
         "bn.js": "^5.1.1",
         "camelcase": "^6.0.0",
+        "chokidar": "^3.6.0",
         "classnames": "^2.2.6",
         "cleave.js": "1.4.10",
         "clsx": "^1.2.1",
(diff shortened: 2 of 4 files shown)


bassgeta (approved):
Worked for me, good quality of life upgrade <3

ArmandoGraterol (dismissed):
Working great for me 👌

rdig (changes requested):
Seems that this is running into the same issue that @chmanie ran into during #1748 as on linux machines that number of file watchers is restricted by default, meaning this script will crash and not start at all.

Screenshot from 2024-02-26 23-01-07

By default, my system has `65536` max file watchers
Screenshot from 2024-02-26 23-02-17

I had to quadruple that value in order for this work _(just doubling it didn't work)_
Screenshot from 2024-02-26 23-04-57

What's more, this is for the current number of files inside the `amplify` folder. Once that number increases by either adding new lambdas, or new services, we'll run into this again, so we need to make this dynamic somehow

I'd say there's two way about this:
- combine your fix with Chris's, since both vite and this script need the same thing
- or we just try/catch the script, put a warning if it crashes, and just let linux users fend for themselves :)

But at the very least this needs to be added to the readme.

Here's a demo of it working, after I've increased that value:
[Screencast from 2024-02-26 23-08-36.webm](https://github.com/JoinColony/colonyCDapp/assets/1193222/889d0929-fa65-4894-be39-7d44bd7b019d)

rdig (approved):
Tested with you today, and can confirm it's working as expected _(with the changes listed in the readme)_

bassgeta (approved):
Works on Linux for me :+1:

ArmandoGraterol (approved):
Still working as expected for me 👌

PR #1980, opened 2024-02-23
Add auto-opening functionality to user hub after tx error

### LINKED ISSUE #1807: Ensure Transaction List from User Hub is showing in case of a transaction error
The transactions list is not being properly displayed in case of an error.

Example:

https://github.com/JoinColony/colonyCDapp/assets/34915414/3b1b2a38-7907-4e77-9387-d448f86a01c0

## Steps to reproduce

There are probably better examples as this can occur on all motions, but one example:

1. Create two motions to upgrade the colony, and move them both to the 'execution' step.
2. Finalize the first motion
3. Attempt to finalize the second motion, it will appear as though nothing happens, but if you check the gas station (userhub / transactions) the failed transaction will be listed.

(Initially observed when testing https://github.com/JoinColony/colonyCDapp/pull/1741 - will need to follow setup steps in https://github.com/JoinColony/colonyCDapp/issues/1677 to have a colony which needs upgrading)

## Expected behaviour

* When a transaction fails, the transactions list should display.

## Actual behaviour

* When a transaction fails there is no feedback to the user.

ArmandoGraterol:
## Description

Implemented logic to open the user hub when a transaction fails. Inspired by the logic we had in the Dapp/Old UI.

## Testing

- We need to make a transaction fail and check if the user hub opens automatically and shows the error.
- One way I could think of is to transfer 1 token to a newly created team, and then try to make a simple payment from that team to any user for 1 token. This transaction should fail because you don't have enough tokens to cover the network fees.
- Another one is to try to transfer all of the tokens from one team to another through a motion, and before finalizing it, create an action and transfer any amount from the same team so that now the transfer motion can't be finalized because the funds aren't enough. (You are gonna have to go to `FinalizeStep` and remove the `!isFinalizable` from the submit button)

Resolves #1807

### DIFF
--- a/src/components/common/Extensions/UserHubButton/UserHubButton.tsx
+++ b/src/components/common/Extensions/UserHubButton/UserHubButton.tsx
@@ -40,9 +40,6 @@ const UserHubButton: FC<UserHubButtonProps> = ({
   const [prevGroupStatus, setPrevGroupStatus] = useState<
     TransactionStatus | undefined
   >();
-  const [groupStatus, setGroupStatus] = useState<
-    TransactionStatus | undefined
-  >();
 
   const { trackEvent } = useAnalyticsContext();
   const walletAddress = wallet?.address;
@@ -95,6 +92,10 @@ const UserHubButton: FC<UserHubButtonProps> = ({
     toggleOff();
   };
 
+  const groupStatus = getGroupStatus(
+    findNewestGroup(transactionAndMessageGroups),
+  );
+
   useEffect(() => {
     if (
       groupStatus === TransactionStatus.Failed &&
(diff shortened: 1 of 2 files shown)


iamsamgibbs (changes requested):
Thanks for working on this, really nice implementation. Only one minor issue I've noticed, the user hub button should have the blue border when open.

Screenshot 2024-02-26 at 09 27 37

(Unrelated issues I noticed whilst testing - the "pending" button just spins forever if the action fails and the error message formatting is incorrect, there is nowhere to view the full error message, and the error message itself isn't actually helpful to the user.)

iamsamgibbs (approved):
This works great! 🎉

bassgeta (approved):
Works and looks good, lgtm :shipit:

jakubcolony (approved):
Nice work @ArmandoGraterol 👍

PR #1983, opened 2024-02-26
correct expenditure sagas to accept tokenDecimals

### LINKED ISSUE #1905: Expenditure sagas should take into account selected token's decimals

warrencolliver:
## Description

This PR fixes the bug that set the decimal value of every token to 18 by default, now the user can enter a decimal value when creating an expenditure and the amount will reflect the entered decimal. tokenDecimals has been added as a param in payouts due to different tokens being used in multiple payouts

## Testing

A newly created token that doesn't have a decimal value of 18 shouldn't have a default decimal of 18 when creating an expenditure. This PR allows you to enter a custom decimal value for any token to text if different tokenDecimal values can be passed through these expenditures. this functionality needs wiring to get the token decimal without user input. The logic changes for `createExpenditure`, `createStakedExpenditure` and `editExpenditure` are the same so we only need to test `createExpenditure` to ensure the logic works.

There are two ways to test this - ultimately they both have the same result and method 1 is much faster:

Method 1: 

* Step 1. Get any token address (from token list in the colony)
* Step 2. go to dashboard and enter the token address, token amount and token decimals. (The amount & decimals can be any value)
* Step 3. Check the db using the query seen below (in step 4) to get the expenditure details

Method 2:

* Step 1. Set up Truffle console
* Step 2. Create a token:
```ts
t = await Token.new('Token name', 'TKN', 12)
```
* Step 3. Go to dashboard and enter the token address (you can get this address from the console when you create the token), token decimals and transaction amount and create the transaction
* Step 4. Go to GraphQl db and use this query, search for your token address or for the ID using the ID counter '_XX' there should be 32 transactions by default. this number increases by 1 with each new expenditure. so after the first expenditure is created, search for '_33'

```
query listExpenditures {
  listExpenditures {
    items {
      id
      isStaked
      slots {
        payouts {
          amount
          isClaimed
          tokenAddress
        }
      }
    }
  }
}
```

The amount seen here should reflect the amount you entered + the amount of decimals you entered. If you enter an invalid decimal amount e.g. "efjk", the decimal value should default to 18

## Diffs

**New stuff** ✨

Input fields and create expenditure button added to TmpAdvancedPayments

**Changes** 🏗

tokenDecimals added to payouts within the sagas 

Resolves #1905

### DIFF
--- a/src/redux/types/actions/expenditures.ts
+++ b/src/redux/types/actions/expenditures.ts
@@ -89,7 +89,6 @@ export type ExpendituresActionTypes =
         expenditure: Expenditure;
         payouts: ExpenditurePayoutFieldValue[];
         networkInverseFee: string;
-        tokenDecimals: number;
       },
       object
     >
--- a/src/redux/types/actions/expenditures.ts
+++ b/src/redux/types/actions/expenditures.ts
@@ -135,7 +135,6 @@ export type ExpendituresActionTypes =
         stages?: ExpenditureStageFieldValue[];
         networkInverseFee: string;
         decisionMethod: ExpenditureDecisionMethod;
-        tokenDecimals: number;
       },
       MetaWithSetter<object>
     >
--- a/src/redux/types/actions/expenditures.ts
+++ b/src/redux/types/actions/expenditures.ts
@@ -48,7 +48,6 @@ export type ExpendituresActionTypes =
         stages?: ExpenditureStageFieldValue[];
         networkInverseFee: string;
         decisionMethod: ExpenditureDecisionMethod;
-        tokenDecimals: number;
       },
       MetaWithSetter<object>
     >
--- a/src/redux/types/actions/expenditures.ts
+++ b/src/redux/types/actions/expenditures.ts
@@ -48,6 +48,7 @@ export type ExpendituresActionTypes =
         stages?: ExpenditureStageFieldValue[];
         networkInverseFee: string;
         decisionMethod: ExpenditureDecisionMethod;
+        tokenDecimals: number;
       },
       MetaWithSetter<object>
     >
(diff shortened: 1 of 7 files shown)


warrencolliver:
related to #1953 new PR created due to issues with rebasing etc

src/redux/types/actions/expenditures.ts
```diff
@@ -135,6 +135,7 @@ export type ExpendituresActionTypes =
         stages?: ExpenditureStageFieldValue[];
         networkInverseFee: string;
         decisionMethod: ExpenditureDecisionMethod;
+        tokenDecimals: number;
```
jakubcolony, line 138: This shouldn't be here

jakubcolony (changes requested):
Looks like we're almost there, just spotted a leftover from your previous implementation

jakubcolony (approved):
Tested the changes with different decimals and all looks good, nice work @warrencolliver!

It'd be great to clean up the description a bit as it seems outdated (e.g. sagas no longer accept token decimals). Also the testing instructions feel a bit incomplete, after creating a token it's not clear how I'm supposed to get its address. Thinking about it tho, we could get away without creating tokens at all - with the new inputs you've added one can "lie" about the decimals of any token and set it to whatever they want for testing purposes.

rdig (approved):
Don't really know too much about the history of this issue, but the changes make sense, and if Jakub already approved it, means that you guys figured out a proper way to repair this branch

PR #1986, opened 2024-02-26
Fix: change not enough reputation error message

### LINKED ISSUE #1940: Improve not enough reputation in team error message
## Steps to reproduce

1. Make sure you have a team where you don't have reputation.
2. Create an action in a team e.g. `Simple payment` using the `Reputation` Decision method.
3. Correctly, you won't be able to stake on the action and you will get an error message.

## Description
The issue here is that this has proved to be confusing to users as the message is too generic, especially when they can see that they have reputation.

> The current message is: "You did not have enough reputation at the time this action was created to stake on this action."

However, we would like this message to be more specific including the team name.

> New message: "When this action was created, your reputation in the {team_name} team was too low."

### Design
image

warrencolliver:
## Description

This PR changes the error message a user gets when they don't have enough reputation to contribute to a stake decision

## Testing

1. Create a motion in a team using the Reputation Decision method (I created an agreement).
2. View the decision from a user without any reputation in the team
3. Correctly, you won't be able to stake on the action and you will get an error message.

## Diffs

**Changes** 🏗

New error message can be seen here:
Screenshot 2024-02-26 at 21 37 44

I've added some code to get the user's team / domain name to display it in the error message, however it feels messy, if someone has a better suggestion on how to get the team name please let me know.

Resolves #1940

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/Motions/steps/StakingStep/StakingStep.tsx
+++ b/src/components/v5/common/ActionSidebar/partials/Motions/steps/StakingStep/StakingStep.tsx
@@ -4,7 +4,6 @@ import { formatRelative } from 'date-fns';
 import React, { type FC } from 'react';
 
 import { useAppContext } from '~context/AppContext.tsx';
-import { useColonyContext } from '~context/ColonyContext.tsx';
 import useToggle from '~hooks/useToggle/index.ts';
 import { SpinnerLoader } from '~shared/Preloaders/index.ts';
 import { SystemMessages } from '~types/actions.ts';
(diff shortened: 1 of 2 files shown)


iamsamgibbs (changes requested):
Looks like this will always return `General` as the team regardless of which team the motion was created in. You should be able to get the correct team from `motionAction.motionData.motionDomain.metadata?.name`

ArmandoGraterol (approved):
LGTM 👍 Nice job Warren!

iamsamgibbs (approved):
Looks good to me now! 🎉 Changes for the change

Screenshot 2024-02-28 at 11 04 37

rdig (approved):
All good.

Sorry I'm adding a third review on a PR that has enough already, but I actually ran out of things to review :)

PR #1991, opened 2024-02-27
feat: sanitize HTML before sending description to saga

### LINKED ISSUE #1028: Investigate rich text security
In a demo call we had about Decisions, Alex mentioned that rich text potentially presents a security vulnerability (script injections) if it's not sanitised correctly.

This issue is to try and inject a script into our rich text form to (in)validate this hypothesis. We're not so worried about TipTap itself, more about somebody editing the html in localstorage with something malicious like an iframe and then publishing the decision, so now when anybody accesses it they can see the malicious content.

bassgeta:
## Description

So turns out tiptap handles sanitization, and I wasn't able to inject a hacky iframe.
Nonetheless, I've created a helper for sanitizing HTML which we now call before sending annotation data to the saga.

## Testing
You can use this hacky iframe `<iframe src="javascript:alert(1)" />`

1. Install governance extension
2. Create new decision, save draft.
3. Go to localstorage to see where decision is being saved. Edit the html string with that iframe.
4. Refresh decision preview to see how it renders.
5. Publish decision to see how it renders.

You can also try setting the description to the iframe, it shouldn't alert you when pasting it or creating any action.

## Diffs

**Changes** 🏗

* Before submitting data to the saga, we sanitize the HTML

Resolves #1028

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/forms/CreateDecisionForm/hooks.ts
+++ b/src/components/v5/common/ActionSidebar/partials/forms/CreateDecisionForm/hooks.ts
@@ -8,6 +8,7 @@ import { createDecisionAction } from '~redux/actionCreators/index.ts';
 import { ActionTypes } from '~redux/index.ts';
 import { mapPayload, pipe } from '~utils/actions.ts';
 import { type DecisionDraft } from '~utils/decisions.ts';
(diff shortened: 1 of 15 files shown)


ArmandoGraterol (approved):
Working as expected for me 👍 Description is looking quite sanitized!

PR #1992, opened 2024-02-27
Changes to expenditures after locking using permissions

### LINKED ISSUE #1975: Advanced Payments API: Changes to expenditures after locking (using permissions)
Related user story from main spec (#1956):
> As a USER, I would like to be able to make changes to the advanced payment during the process. However, I want it to require collective input after an initial review step.

Changes to expenditures after locking can be made using `setExpenditureState` contracts function. This issue is to implement a `editLockedExpenditure` saga (or modify the existing `editExpenditure` saga) that will accept an array of updated payouts and call the contracts function to modify the expenditure. It will also require creating a new block-ingestor handler.

jakubcolony:
## Description

This PR will enable making changes to locked expenditures using permissions.

block-ingestor PR: https://github.com/JoinColony/block-ingestor/pull/181

## Testing

Test the following scenario with draft and locked expenditure:

1. Create an expenditure. Run the following query and inspect the values of the slots (recipient address, claim delay, payouts):
```gql
query ListExpenditures {
  listExpenditures(filter: {nativeId: {gt: 32}}) {
    items {
      nativeId
      slots {
        id
        recipientAddress
        claimDelay
        payouts {
          amount
          tokenAddress
        }
      }
    }
  }
}
```

2. Enter the expenditure ID in the field and click on "Edit expenditure". The hardcoded payload should change the existing slot's recipient to the colony address, change its token address to the native token and set the amount to `23.45`. It should also add a second slot with colony's native token, user address set as recipient, amount of `67.89` and claim delay of `300`. Feel free to customise those values.

3. Run the query again to verify the values have updated accordingly.

Bonus points: When creating expenditure, use a token address that's not the colony's native (e.g. `0x0000000000000000000000000000000000000000`) to test that deleting tokens from payouts work. After editing, the payout amount for this token should be set to 0.

Resolves #1975

### DIFF
--- a/amplify/backend/function/fetchVoterRewards/src/utils.js
+++ b/amplify/backend/function/fetchVoterRewards/src/utils.js
@@ -88,7 +88,7 @@ const getVoterReward = async (colonyAddress, motionId, userReputation) => {
       BigNumber.from(userReputation),
     );
     return reward;
-  } catch {
+  } catch (e) {
     console.error('Error getting voter reward', e);
     return undefined;
   }
--- a/docker/colony-cdapp-dev-env-block-ingestor
+++ b/docker/colony-cdapp-dev-env-block-ingestor
@@ -1,6 +1,6 @@
 FROM colony-cdapp-dev-env/base:latest
 
-ENV BLOCK_INGESTOR_HASH=9dfb3e8f058f5ea8c4343df166f3165ff37bd5b4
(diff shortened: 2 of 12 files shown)


bassgeta (approved):
Worked with the native token
image
worked with test ETH
image
also the balance was returned to the colony
image
(i had 4 ETH to begin with). So gj, LGTFB :D

iamsamgibbs (approved):
Looks good to me! All tested and working as expected!

PR #1994, opened 2024-02-28
fix: added permissions page to prod route

CzarekDryl:
## Description

- Removed `! WEBPACK_IS_PRODUCTION` from permission page route so it should be visible on QA


iamsamgibbs (approved):
Looks good to me!

warrencolliver (approved):
This looks good to me. LGTM

PR #1999, opened 2024-02-29
Feat: Remove user from colony whitelist when they leave a colony

### LINKED ISSUE #1934: When user leaves a colony they should be removed from the colony whitelist
## Specification

During the beta, when a user leaves a colony, they should be removed from the colony whitelist and should be redirected away from the colony.

This should only happen during the beta and be disabled once the beta has ended.

Currently the `validateUserInvite` lambda adds users to the whitelist when they accept an invite. There is no function for removing a user from a whitelist, so one should be added following a similar logic to `validateUserInvite` lambda.

~~Blocked by #1902~~

iamsamgibbs:
## Description

Related auth proxy PR: https://github.com/JoinColony/colonyCDappAuthProxy/pull/14

During the beta, when a user leaves a colony, they should be removed from the colony whitelist and should be redirected away from the colony.

This should only happen during the beta and be disabled once the beta has ended.

This PR adds a new 'RemoveMemberFromColonyWhitelist' lambda which is called when a user leaves a colony. There is a related PR on auth proxy which ensures only the authenticated user can remove themselves from the whitelist to prevent users removing other others from the whitelist.

## Testing

Switch to either the Planex or Wayne colony and leave the colony. You should be redirected and the colony should no longer show in the colony switcher. You can confirm you have been removed from the whitelist with this query:

```
query MyQuery {
  listColonies {
    items {
      id
      name
      whitelist
    }
  }
}

```

## Diffs

**New stuff** ✨

* 'RemoveMemberFromColonyWhitelist' lambda

**Changes** 🏗

* Leaving a colony now removes you from the colony whitelist

Resolves #1934

### DIFF
--- a/amplify/backend/function/removeMemberFromColonyWhitelist/src/event.json
+++ b/amplify/backend/function/removeMemberFromColonyWhitelist/src/event.json
@@ -1,6 +1,0 @@
-{
-  "arguments": {
-      "colonyAddress": "",
-      "userAddress": ""
-  }
-}


amplify/backend/function/removeMemberFromColonyWhitelist/removeMemberFromColonyWhitelist-cloudformation-template.json
```diff
@@ -0,0 +1,223 @@
...
+            }
+          }
+        },
+        "Role": {
+          "Fn::GetAtt": [
+            "LambdaExecutionRole",
+            "Arn"
+          ]
+        },
+        "Runtime": "nodejs14.x",
```
bassgeta, line 85: I wonder why it generated with 14 as the version, we are using 20 and other cloudformation templates have `nodejs20.x` as the value here

amplify/backend/function/removeMemberFromColonyWhitelist/src/index.js
```diff
@@ -0,0 +1,89 @@
...
+  const updatedWhitelist = whitelist.filter(
+    (whitelistedAddress) => whitelistedAddress !== userAddress,
+  );
+
+  const colonyMutation = await graphqlRequest(
+    updateColony,
+    {
+      input: {
+        id: colonyAddress,
+        whitelist: [...updatedWhitelist],
```
bassgeta, line 76: i think we don't need to spread here

amplify/backend/function/removeMemberFromColonyWhitelist/src/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "removememberfromcolonywhitelist",
+  "version": "2.0.0",
+  "description": "Lambda function generated by Amplify",
+  "main": "index.js",
+  "license": "Apache-2.0",
+  "devDependencies": {
```
bassgeta, line 7: dropping a comment just in case, all the other lambdas also have `cross-fetch` as the dependency and the `index.js` file has `require('cross-fetch/polyfill');` as the first line. Not sure if it's necessary to include it here too

bassgeta (changes requested):
Works! Just dropped a few comments regarding `cross-fetch` :D

ArmandoGraterol (dismissed):
Working as expected for me 👌 Nice job

amplify/backend/function/removeMemberFromColonyWhitelist/src/event.json
```diff
@@ -0,0 +1,6 @@
+{
```
jakubcolony, line 1: this file can be deleted

amplify/backend/function/removeMemberFromColonyWhitelist/src/package.json
jakubcolony: It's only really necessary if you're fetching reputation using `colony-js`

jakubcolony (approved):
LGTM, nice work Sam!

ArmandoGraterol (approved):

rdig (approved):
Seems to be working for me. Nice job!

bassgeta (approved):
Works the same after the changes, nice one! :muscle:

PR #2006, opened 2024-02-29
fix: handle transaction rejected from wallet error 

### LINKED ISSUE #1939: Handle rejecting transactions when creating actions
## Steps to reproduce

1. Create a new action filling in all the details.
2. Click on the create button to initiate the action.
3. In your wallet, "Reject" the transaction.
4. Return to the action creation screen and try to initiate the action again.

## Expected behaviour
* After rejecting a transaction, you should be able to initiate the transaction again.
* (Enhancement) If there are multiple transactions required, and one of the subsequent transactions were rejected, the UI should be able to detect it and update the button to be "Complete creation" and an error message with the text "Creation did not finish. Complete the transactions using the Complete creation button."

image

## Actual behaviour
* After rejecting a transaction, the pending button for creation stays pending.

reject-transactions

bassgeta:
## Description

Turns out, our way of doing sagas failed if the user rejected a transaction when it was first created.
This was due to us trying to get the `txHash` but not listening to events for a transaction failing.
I've unified that to just always get the result and use the `txHash` from there.
Additionally, the helper for sending transactions emits a transaction failed error if the signing failed.

## Testing

My friend, I am very sorry.... The testing scope for this is to test **ALL** the actions.

1. Start up your dev environment, enable the Reputation extension
2. go to `http://localhost:3006/ganache-accounts.json` and copy the private key for "leela" - the wallet address is `0xb77d57f4959eafa0339424b83fcfaf9c15407461`
3. Go to Metamask and add an account by importing the private key
4. Maybe you will need to refresh the page after disconnecting your wallet, but open Metamask again and you should see this:
image
Click `Connect` to connect with your second account (if this is your only account you can probably just immediately connect)
5. Login with your imported account and now you should have all the tokens and permissions the colony creator has (and the test ETH which is **very** important for trying all of this out
6. Test all the actions in the UI for the following cases
- Create with permissions -> Reject in Metamask -> The button should not be pending and you can click it again
- Create with reputation -> Reject in Metamask -> The button should not be pending and you can click it again
- Create with permissions -> Sign in Metamask -> The transaction should go through and the completed action view should show
- Create with reputation -> Sign in Metamask -> The transaction should go through and the completed action view should show with the motion sidebar to the right. You don't need to test motions going through technically since that part wasn't touched

## Diffs

**Changes** 🏗

* the `sendTransaction` helper emits a transaction failed error when it fails (previously it returned null and that was it)
* all the sagas applicable were realigned, using the `txHash` from `waitForTxResult`'s response instead of listening for that specific event
* all the sagas that didn't yield the error in the catch block now do, thus correctly refreshing the UI
* the `setTxHash` helper gets called after all the transactions are completed, so we can cancel any transaction in between and retry. Additionally the title now doesn't jank in, but gets displayed with the action

Resolves #1939

### DIFF
--- a/src/components/shared/Fields/Form/ActionForm.tsx
+++ b/src/components/shared/Fields/Form/ActionForm.tsx
@@ -69,7 +69,7 @@ const ActionForm = <V extends Record<string, any>>({
       const res = await asyncFunction(values);
       onSuccess?.(values, formHelpers, res);
     } catch (e) {
-      console.error(e);
+      console.error('Error while submitting form', e);
       onError?.(e, formHelpers);
     }
   };
--- a/src/redux/sagas/actions/createDomain.ts
+++ b/src/redux/sagas/actions/createDomain.ts
@@ -176,11 +176,10 @@ function* createDomainAction({
       });
     }
   } catch (error) {
-    return yield putError(ActionTypes.ACTION_DOMAIN_CREATE_ERROR, error, meta);
+    yield putError(ActionTypes.ACTION_DOMAIN_CREATE_ERROR, error, meta);
   } finally {
     txChannel.close();
   }
-  return null;
 }
 
 export default function* createDomainActionSaga() {
--- a/src/redux/sagas/actions/createDomain.ts
+++ b/src/redux/sagas/actions/createDomain.ts
@@ -132,8 +132,6 @@ function* createDomainAction({
       },
     } = yield waitForTxResult(createDomain.channel);
 
-    setTxHash?.(txHash);
-
     const { domainId } = eventData?.DomainAdded || {};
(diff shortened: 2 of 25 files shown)


rdig (approved):
I do like the approach here as it makes it quite strait forward to wrap your head around. 

Screenshot from 2024-02-29 22-15-16

Screenshot from 2024-02-29 22-15-22

Thanks for fixing this!

iamsamgibbs (approved):
Great job, tested (almost) all actions with permissions and reputation. I noticed a few issues with the actions sidebar, but none related to this PR. The only action I couldn’t test was upgrading a colony, but will assume it works as all others are working as expected. Thanks for the detailed testing instructions.

Issue #2004, opened 2024-02-29
Advanced Payments - Action: Payment builder


AdrianDudko:
**Teamwork task:** [Advanced Payment - Action: Payment builder](https://tw.auroracreation.com/#/tasks/15804107)
**Created by:** @AdrianDudko 
**Accepted by:** @arrenv 

## Related sub-issues:
https://github.com/JoinColony/colonyCDapp/issues/2044

## Acceptance criteria

**Action - Creation - Form**

* User can select the "Payment Builder" option from the "Action type" options.

  * Without the Open payments extension enabled and without Lazy Consensus extension enabled The user requires the "Admin" permission or higher (Administration on the contracts) to create the payment.

    * If they don’t have permissions, they will receive the disabled action state and an error message.

* Fund from

  * Select the team to fund the payment from.

  * Uses the team drop down.

  * Required

* Decision method

  * Required

* Created in

  * Hide if Permission Decision method is chosen

  * Required for other Decision Methods

* Use exisiting validation rules

**Action - Creation - Table**

* Payments table

  * Select a recipient

    * User can select a recipient from the select member column.

    * User is able to manually enter an address as a recipient.

    * List of members should include all members and be searchable.

    * Verified members should be shown with their icon.

    * If an address is entered that is not in the verified list, then a warning should be shown.

  * Token and amount

    * User can select from the available tokens in the Colony to make the payment in.

    * User can enter a value for the amount of tokens to pay the recipient.

    * Clicking the token selection should include available funds for that token in the fund from team.

    * Entering multiple amounts should sum for the same token and validate available funds.

  * Claim delay

    * User can enter a claim delay value for each recipient.

    * User can select the period for the claim delay.

    * The claim delay input only allows a positive input value.

    * There is a max of 5 digits possible for the claim delay.

  * Kebab menu

    * Click on the kebab icon provides a dropdown with the following options:

      * Add token

        * Allows the user to add another token to the recipient.

        * This adds a duplicate row with the recipient selected and the same claim delay but an empty amount field.

      * Duplicate row

        * Allow the user to duplicate the recipient row, adding another row with the exact same details.

      * Delete row

        * Allow the user to delete the enter row.

  * Add recipient

    * A button to be able to add another row to the table. No details other then token should be pre-fields.

  * Multiple tokens

    * You can pay in multiple different tokens. From a contract perspective, these are grouped per recipient, however, from a UI perspective we are treating each row separately.

**Action - Preview - Form**

[...]

PR #2011, opened 2024-03-01
feat: payment builder form update

joanna-pagepro:
https://github.com/JoinColony/colonyCDapp/issues/2004

Create expenditure, without preview mode


src/components/v5/common/ActionSidebar/partials/forms/PaymentBuilderForm/utils.tsx
```diff
@@ -74,3 +74,37 @@ export const advancedPaymentDescriptionMetadataGetter: DescriptionMetadataGetter
...
+      : null,
+    fundFromDomainId: values.from,
+    isStaged: false,
+    networkInverseFee,
+    annotationMessage: values.description,
+    payouts: values.payments.map((payment) => ({
+      recipientAddress: payment.recipient,
+      tokenAddress: payment.amount.tokenAddress,
+      amount: payment.amount.amount.toString(),
+      claimDelay: payment.delay,
```
jakubcolony, line 103: that needs be converted to seconds

jakubcolony (changes requested):
Looks good @joanna-pagepro, thanks for your work on this PR.

A few things that I spotted:

1. Not sure about this one, but advanced payment can be created in a specific domain, so we might need to show `Created In` field at all times. Then with reputation decision method, I presume we could use the same domain for both the motion and the payment. @arrenv could you confirm?

image

2. Typing in big numbers causes some overlaps:

image

3. I have about 91M tokens. The validation checks correctly that the single payout amount doesn't exceed the balance (althought the error with the path looks odd):
image

But it doesn't check that the total amount of all payouts doesn't exceed the balance:
image

4. It crashes when doing the following (on an empty form, with just action type selected):
   1. Set the amount to something exceeding the balance.
   2. Click on Add transaction.
   
https://github.com/JoinColony/colonyCDapp/assets/112586815/488a65f8-fb55-480d-946f-a0bb4ce7c0e7

5. The heading of the table should be `Payments` instead of `Additional Payments` according to Figma.

But overall, the payments are being created, so nice work! 👍 
image

image

AdrianDudko:
> 1. Not sure about this one, but advanced payment can be created in a specific domain, so we might need to show `Created In` field at all times. Then with reputation decision method, I presume we could use the same domain for both the motion and the payment. @arrenv could you confirm?
> 
> image


I talked about the Created In field with @arrenv. From what I understand, the field should be hidden. But you're right, it will be good to get confirmation from @arrenv again. Nevertheless, this is a rather minor change and if it does not affect the creation of the action itself (the action passes), I think we can merge this PR after fixing the rest of the errors - so as not to block further work on the Adv.Payment process.

I'll make a note to come back to this when @arrenv  is available.
Are you ok with that?

jakubcolony:
Hey @AdrianDudko, yes I agree that this isn't a blocker for merging this PR.

joanna-pagepro:
@jakubcolony please review again. There are no changes with `createdIn` field for now.

AdrianDudko:
> @jakubcolony please review again. There are no changes with `createdIn` field for now.

Agreed with @arrenv to hide `Created In` for `Decision Method: Permission`

jakubcolony (changes requested):
Thanks for making changes @joanna-pagepro. 

A couple of things I noticed:

- The validation for amount happens immediately after selecting a recipient, even though the amount field has not been touched. It only seems to happen if you fill in other fields beforehand:

https://github.com/JoinColony/colonyCDapp/assets/112586815/935c4cc3-2315-4d17-8546-28ba1419f137


- If you click on the warning icon next to recipient, the popover gets clipped by the table:
image


- Clicking on "Add token" throws an error:

https://github.com/JoinColony/colonyCDapp/assets/112586815/5bf7f0c8-051a-4291-808b-305270b2d018

melyndav (changes requested):
Hey @joanna-pagepro! Nice start on the payment builder!

Some feedback for this one:

1. The enter amount text should not show and should show a 0 as per Figma:
image

https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-v5?type=design&node-id=3889-462623&mode=design&t=BNXdv8DRozb21TVR-4
 
 2. When I select a different token that has no funds, the validation shows displaying a message that the amount should be greater than 0 but there are no funds.
 
image
 
 Then when I select back to the token that has funds available, the validation stays and does not hide. This is a confusing user experience. 
image

4. The column widths of the table can be match better to Figma which should prevent data from being displayed in a small area
https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-v5?type=design&node-id=3889-462623&mode=design&t=BNXdv8DRozb21TVR-4 
 
 
 5. When I select 'add token' it breaks
image

6. The claim delay breaks when a section row is added
image

7. The claim delay placeholder is not styled correctly to match Figma
image

9. The styling of the alert and validated user icons int he table are two big and the padding is incorrect 
image

Figma: https://www.figma.com/file/l1dOM5qiQYwF0ElvKDqqjg/Design-System---Colony-v3?type=design&node-id=1782-151741&mode=design&t=FacBc4ZDz7L2AA0Q-4

10. Button should be 'add payment' not 'add transaction' 

11. When I have 1 row with an error validation, as soon I select to add more rows, the error validation continues into those new rows added. The validation should clear for the new rows and stay for the wrong row only
image

12. Number formatting is not consistent int he table with two rows treating numbers differently 
image

13. On mobile, the Hours font changes
image

14. On mobile, there should be no HR's between all individual rows.

joanna-pagepro:
@jakubcolony @melyndav form updated, please review again.
@melyndav I don't know what you mean in point 12. Can you explain more, what do you expect?

src/components/v5/common/ActionSidebar/partials/forms/PaymentBuilderForm/partials/PaymentBuilderTokensTotal/types.ts
```diff
@@ -0,0 +1,9 @@
+import { type Token } from '~types/graphql.ts';
+
+export interface PaymentBuilderTokenItem extends Token {
+  amount: string;
+}
```
jakubcolony, line 5: The domain term for those would be "payouts", I think it's worth updating it here to avoid confusion with the `Token` type (which is a property rather than a parent of a payout)

src/utils/validation/amountGreaterThanZeroValidation.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  const colonyTokens =
+    colony?.tokens?.items
+      .filter(notNull)
+      .map((colonyToken) => colonyToken.token) || [];
+
+  const selectedToken = colonyTokens.find(
+    ({ tokenAddress: selectedTokenAddress }) =>
+      selectedTokenAddress === tokenAddressFieldValue || tokenAddress,
+  );
```
jakubcolony, line 30: There's a util for that in `~utils/tokens.ts`: `getSelectedToken`

src/components/v5/common/Table/utils.tsx
```diff
@@ -31,19 +30,13 @@ export const makeMenuColumn = <T,>(
...
-              props.contentWrapperClassName,
-              '!z-[60]',
-            )}
-          />
-        </div>
+        <MeatBallMenu
+          {...props}
+          contentWrapperClassName={clsx(
+            props.contentWrapperClassName,
+            '!z-[65] md:!z-[60]',
```
jakubcolony, line 37: Setting different z-indexes based on the screen size makes it look like we've got a bigger problem presumably with another component depending on this? Is it possible to come up with a single value?

jakubcolony (changes requested):
Thanks for another round of fixes @joanna-pagepro. All the previous issues seems to be resolved 👍 

I left a few comments and noticed two more things:

- The error message is not human-friendly when claim delay is too set too large:
image

- The totals aren't calculated until you select a domain (which I don't think is necessary to calculate just the total amount of payments):
image

joanna-pagepro:
Some features stopped working after the rebase with the parent branch. Please hold off on Code Review until it's fixed.

melyndav:
@melyndav I don't know what you mean in point 12. Can you explain more, what do you expect?

Hey @joanna-pagepro I've discussed this with Arren and please ignore this point for now since it's a wider decision on number formatting. Thanks!

src/components/v5/common/ActionSidebar/partials/forms/PaymentBuilderForm/partials/PaymentBuilderRecipientsField/hooks.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+export const useRecipientsFieldTableColumns = (
+  name: string,
+  data: PaymentBuilderRecipientsFieldModel[],
+): ColumnDef<PaymentBuilderRecipientsTableModel, string>[] => {
+  const {
+    colony: { expendituresGlobalClaimDelay, tokens },
+  } = useColonyContext();
+
+  const dataRef = useWrapWithRef(data);
```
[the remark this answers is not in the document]
joanna-pagepro: In this case, we will have to add eslint ignore rule and because of that we won't see if anything is missing from the dependency array.

src/components/v5/common/ActionSidebar/partials/forms/PaymentBuilderForm/partials/PaymentBuilderTokensTotal/types.ts
joanna-pagepro: Do you mean that I should replace `amount` with `payouts` or something else?

src/components/v5/common/Table/utils.tsx
joanna-pagepro: It's not needed here anymore. I forgot to remove it.

joanna-pagepro:
@jakubcolony @melyndav it's ready to review again :)

src/components/v5/common/ActionSidebar/partials/forms/PaymentBuilderForm/partials/PaymentBuilderRecipientsField/hooks.tsx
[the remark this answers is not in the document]
jakubcolony: Sure, I think that's cleaner as you know straight away what's going on

melyndav (changes requested):
Hey @joanna-pagepro! Great work on these updates. Some more feedback.

1. The hours input should be styled the same as the amount in the default state.
image

2. members don't show with verified icons in the select dropdown. It shows only in the table. 
image

3. I spoke with @arrenv regarding the claim day max. Right now we have it set to a max of 4 (9999 hrs). We'd like to increase this to 5 (99 999hrs). Could you update this? 

4. In the kebab menu, can you please change the following text 'Add token' to 'Add row'. The functionality remains the same.

joanna-pagepro:
@melyndav all updated, please check if it's ok now

src/components/v5/common/ActionSidebar/ActionSidebar.tsx
```diff
@@ -117,7 +117,8 @@ const ActionSidebar: FC<PropsWithChildren<ActionSidebarProps>> = ({
           border-gray-200
           rounded-l-lg
           shadow-default
-          z-[60]
+          z-[65]
+          md:z-[60]
```
jakubcolony, line 121: If possible, it would be good not to change z-index depending on the screen width

src/components/v5/common/ActionSidebar/partials/forms/PaymentBuilderForm/partials/PaymentBuilderPayoutsTotal/PaymentBuilderPayoutsTotal.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const sortedPayouts =
+    useMemo(
+      () =>
+        payouts?.sort((a, b) => {
+          if (
+            !a.amount ||
+            a.amount === '' ||
+            !b.amount ||
+            b.amount === '' ||
+            BigNumber.from(a.amount || '0').eq(BigNumber.from(b.amount || '0'))
```
jakubcolony, line 30: If `!a.amount` is false, `a.amount === ''` will also always be false, so we only need one

jakubcolony (approved):
Nice work @joanna-pagepro, thanks for addressing all the feedback 👍

jakubcolony (changes requested):
Apologies @joanna-pagepro, I've just noticed this while testing #2039:

https://github.com/JoinColony/colonyCDapp/assets/112586815/21770a58-32f4-40fb-b6ce-8cc3d468d207

It seems to happen when you include a dot in the amount and then add a new row.

joanna-pagepro:
@jakubcolony good you've noticed that! It should be ok now. Can you check again?

melyndav (approved):
Superstar effort @joanna-pagepro! Thank you for fixing all the updates, great work! 💯

jakubcolony (approved):
Nice work, the issue seems fixed! There's just a couple of comments outstanding from the previous reviews if you could have a look at those.

src/components/v5/common/ActionSidebar/ActionSidebar.tsx
joanna-pagepro: If I remember it right, @CzarekDryl was working on it in a separate PR. The solution I added is temporary, only to make it possible to see the changes on mobile right now. I will remove it when Czarek's changes will be merged.

src/components/v5/common/ActionSidebar/partials/forms/PaymentBuilderForm/partials/PaymentBuilderPayoutsTotal/PaymentBuilderPayoutsTotal.tsx
joanna-pagepro: Thanks for noticing, removed

PR #2013, opened 2024-03-01
feat: finalize expenditure via permissions as a non owner

### LINKED ISSUE #1977: Advanced Payments API: Finalizing expenditure as a non-owner (using permissions)
Related specs from the main issue (#1956):

> (Stage) Release
> - Opens the Release payment modal with the users possible decision methods.
> - If funded, the owner can release the funds for the payment, starting the claim delay.
> - If any claim delays are 0, then those payments will happen immediately on release and also release the creators stake if present.
> - Alternatively, users can use available decision methods, including “Admin” tiered permissions or higher (Arbitration on the contracts), or use the Reputation decision method.

This issue is to make changes to the `finalizeExpenditure` saga to allow non-owners to finalize ("release" in UI) an expenditure if they have arbitration permissions in the domain the expenditure has been created in. This can be achieved using `setExpenditureState` contracts function.

bassgeta:
## Description

Implemented a saga which either finalizes the expenditure if the user is the creator, otherwise it gets the caller's arbitration permissions and tries to finish a transaction.

## Testing

You will need to operate with 2 users here, so we can test the permissions part properly.

1. Install the Staked expenditure extension
2. Give the user `fry` or `amy` Payer permissions from `leela` 
3.  create an expenditure with something like 100 as the transaction amount. Check the console for its `id`
4. fund the expenditure as leela (can also do it with the Payer user)
5. On a different browser, login as `fry` (dev wallet 3) or `amy` (dev wallet 2). NOTE: you may need to get the invite code and have them join. Just run
```
getColony(id: "0xa6Ee101fe6D2C220e373B2D1DE70A905Aa03bf41") {
    colonyMemberInviteCode
  }
```
And then go to `/invite/planex/{colonyMemberInviteCode}`
6. Enter the expenditure id and click finalize.
7. Check if the expenditure status is set to finalized by running 
```
query ListExpenditures {
  listExpenditures(filter: {nativeId: {eq: YOUR_EXPENDITURE_ID }}) {
    items {
      nativeId
      status
      finalizedAt
      ownerAddress
      balances {
        tokenAddress
        amount
      }
    }
  }
}
```
Also try to finalize a draft expenditure which is not funded as a non owner. It should stay in `DRAFT` state.
## Diffs

**Changes** 🏗

* `finalizeExpenditure` saga split into two parts, one runs for owners, the other for users with permissions

Resolves #1977


jakubcolony (approved):
Great work! Tested finalizing as both owner and non-owner. 

In case anyone else spots it, i18n messages are missing for finalize transactions:
image

But that's fixed in `feat/advanced-payments` so just needs rebasing.

iamsamgibbs (approved):
This all looks great and works as expected when following the testing steps! 🎉 

One thing I've noticed (and not sure if it is an issue or not) when you attempt to finalize an expenditure you have created, it cannot be finalized unless it has previously been locked and funded (as expected). However, if you attempt to finalize the expenditure as another user, it will not check if it has been locked or funded first and will allow you to finalize the expenditure.

rdig (approved):
Codewise it looks all good.

I intended to test it, but GH is down right now, can't build the new block ingestor image :(

Screenshot from 2024-03-05 22-05-11

jakubcolony (approved):

src/redux/sagas/expenditures/finalizeExpenditure.ts
```diff
@@ -95,12 +131,118 @@ function* finalizeExpenditureAction({
...
   );
+}
+
+type FinalizeExpenditureWithPermissions = {
+  colonyAddress: Address;
+  expenditure: Expenditure;
+  meta: Action<ActionTypes.EXPENDITURE_FINALIZE>['meta'];
+  annotationMessage?: string;
+  userAddress: Address;
+};
```
jakubcolony, line 147: nit-pick - this could extend the other type or use a single type for both

src/redux/sagas/expenditures/finalizeExpenditure.ts
```diff
@@ -95,12 +131,118 @@ function* finalizeExpenditureAction({
...
+      type: ActionTypes.EXPENDITURE_FINALIZE_SUCCESS,
+      payload: {},
+      meta,
+    });
+  } catch (error) {
+    return yield putError(ActionTypes.EXPENDITURE_FINALIZE_ERROR, error, meta);
+  }
+  [finalizeExpenditure, annotateFinalizeExpenditure].forEach((channel) =>
+    channel.channel.close(),
+  );
```
jakubcolony, line 245: It feels like we've ended up with way more code (and duplications) by splitting it into helpers (all because of annotations) - maybe having an if inside the main saga instead would make it a bit more readable?

jakubcolony (approved):
Great work, gave it another round of testing with the new contract method and it works like a charm! 👍

iamsamgibbs (approved):
All tested and working as expected, good job! 🎉  Thanks for resolving the finalizing draft as non-owner issue. Also tested as a user without payer permissions and was unable to finalize the expenditure which is intended.

jakubcolony (approved):
All good, nice work!

PR #2016, opened 2024-03-04
Fix: not able to load extension pages directly in production

### LINKED ISSUE #1995: Not able to load extension pages directly in production
## Steps to reproduce

1. Go to a Colony in production.
2. Go to an extension page and reload the page, you can also load URL directly.
3. The page redirects to a 404.

## Expected behaviour

* The extension pages should load directly.

## Actual behaviour

* Loading extension pages redirect to a 404.

warrencolliver:
## Description

This PR fixes the issue where an extension directs to a 404 page when the direct URL is used, or if the page is refreshed on the extension details page. This happens because extensionData is not available fast enough so NotFoundRoute is rendered.

## Testing

Use a direct URL like `http://localhost:9091/planex/extensions/OneTxPayment` and see if the extension details page loads

AND

go to an extension detail page e.g. Voting reputation or One tx payment and refresh the page

The expected behaviour is that the extension details page is loaded without issues and does not redirect to a 404.

## Diffs

**Changes** 🏗

!loading added to the if check which renders NotFoundRoute

Resolves #1995

### DIFF
--- a/src/components/frame/Extensions/pages/ExtensionDetailsPage/ExtensionDetailsPage.tsx
+++ b/src/components/frame/Extensions/pages/ExtensionDetailsPage/ExtensionDetailsPage.tsx
@@ -39,7 +39,7 @@ const ExtensionDetailsPage: FC = () => {
     refetchColony,
   } = useColonyContext();
   const navigate = useNavigate();
-  const { extensionData, refetchExtensionData } = useExtensionData(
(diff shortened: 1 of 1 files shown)


jakubcolony (approved):
LGTM 👍 

https://github.com/JoinColony/colonyCDapp/assets/112586815/7a281226-3b7d-4305-aa29-f26a1b759aac

bassgeta (approved):
LGTM :sunglasses:

PR #2018, opened 2024-03-04
Hide created in field when it's read only on actions

### LINKED ISSUE #2009: "Created in" field shows when "Permissions" decision method is selected on mint tokens action
## Steps to reproduce

Create a new action and select the 'mint tokens' action.

## Expected behaviour

"Created in" field should not exist by default.
When "Reputation" decision method is selected, "Created in" field should show and have "General" pre-selected.

## Actual behaviour

"Created in" field shows straight away and regardless of selected decision method.

Screenshot 2024-03-01 at 11 44 03

ArmandoGraterol:
Not reported on the issue but the same was happening in the "Unlock token" action.

Screenshot 2024-03-04 at 18 22 30

Screenshot 2024-03-04 at 18 22 41

Resolves #2009


bassgeta (approved):
Tried it out and it works, quite nice that it turned out to be a oneliner! :tada:

warrencolliver (approved):
Looks good to me!

PR #2019, opened 2024-03-05
fix: popover position, loading, copy, menu item

CzarekDryl:
https://tw.auroracreation.com/app/tasks/15763226
https://tw.auroracreation.com/app/tasks/15577569

**Changes** 🏗

* Member popover position changed to left instead of center
* Changed copy
* Moved permissions link to members menu
* Permission page loading

### DIFF
--- a/src/components/frame/Extensions/layouts/consts.ts
+++ b/src/components/frame/Extensions/layouts/consts.ts
@@ -63,6 +63,12 @@ export const membersMenu: NavigationSidebarLinksListProps['items'] = [
   //   icon: SealCheck,
   // },
   {
+    key: '4',
+    label: formatText({ id: 'navigation.members.permissions' }),
+    to: COLONY_PERMISSIONS_ROUTE,
+    icon: Signature,
+  },
+  {
     key: '5',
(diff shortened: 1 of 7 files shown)


iamsamgibbs (dismissed):
All looks good to me! 🎉

warrencolliver (dismissed):
Can't find any issues with this. Good work

melyndav (changes requested):
Hey @CzarekDryl!Great start on this.

With the skeleton loader for all sizes (desktop, mobile etc) can the loading elements match the same size as the actual content? Currently the avatar and other skeleton elements is bigger which causes a change to the UI when the actual content loads in.

iamsamgibbs (changes requested):
The loading skeleton is looking a lot better now.

It is possible to hide the "No results available" message whilst loading without breaking anything? It briefly shows when first loading the page.


https://github.com/JoinColony/colonyCDapp/assets/34915414/dee046e6-9ec1-43eb-b5da-f7c78bc079e8

iamsamgibbs (approved):
Nice one, thank you!

PR #2020, opened 2024-03-05
Cancelling non-staked expenditure as non-owner (using motions)

### LINKED ISSUE #1998: Cancelling non-staked expenditure as non-owner (using motions)
Implement a new saga to allow creating motions to cancel expenditures that are either locked or which the user is not an owner of.  This can be achieved by calling `setExpenditureState` contracts function.

This will also require updating block-ingestor motions handlers.

jakubcolony:
## Description

This PR introduces new saga, `cancelExpenditureMotion`, that allows creating a motion to cancel an expenditure as a non-owner.

Related block-ingestor PR: https://github.com/JoinColony/block-ingestor/pull/191

## Testing

1. Install and enable voting reputation extension.
2. Create an expenditure.
3. At any point of the flow (before finalization), create a motion to cancel using the testing UI.
4. Check the expenditure in the DB. Under `motions`, there should be a new motion linked to an action of type `CANCEL_EXPENDITURE_MOTION`. From the action, it should also be possible to access the linked expenditure:
```gql
query ListExpenditures {
    items {
      nativeId
      status
      motions {
        items {
          id
          action {
            type
            expenditure {
              nativeId
            }
          }
        }
      }
    }
  }
}
```
5. Progress the motion to finalization. Upon finalizing, the status of the expenditure should change to cancelled.

Resolves #1998


iamsamgibbs (approved):
Tested and working perfectly 🎉

ArmandoGraterol (approved):
Working as expected for me too 👌

warrencolliver (approved):
Looks good to me 👍

src/redux/sagas/motions/expenditures/cancelExpenditureMotion.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  try {
+    if (
+      !colony ||
+      !expenditure ||
+      !votingReputationAddress ||
+      !motionDomainId
+    ) {
+      throw new Error('Invalid payload');
+    }
```
rdig, line 40: We have this pattern in most sagas. I guess it would be a good idea to actually convert it across all of them to some sort of util.

But yeah, that's just some more for the maintenance pile for which we don't really have time right now.

PR #2021, opened 2024-03-05
Feat: Implement edit locked expenditure motion saga

iamsamgibbs:
## Description

[Related block-ingestor PR](https://github.com/JoinColony/block-ingestor/pull/189)

Changes to expenditures after locking can be made using `setExpenditureState` contracts function. This PR implements an `editLockedExpenditureMotion` saga that accepts an array of updated payouts and creates a motion which, upon finalizing, calls the contracts function and updates the expenditure accordingly.

## Testing

1. Install the reputation weighting extension
2. Create a new expenditure (The expenditure ID will be logged to the console)
3. Enter the expenditure ID and lock the expenditure.
4. Run this query to verify the expenditure has been created with no motions attached, and one payout.

```
query MyQuery {
  listExpenditures(filter: {nativeId: {ge: 33}}) {
    items {
      nativeId
      motions {
        items {
          motionStateHistory {
            hasPassed
          }
          editedExpenditureSlots {
            claimDelay
            id
            payoutModifier
            recipientAddress
            payouts {
              amount
              isClaimed
              networkFee
              tokenAddress
            }
          }
        }
      }
      slots {
        payouts {
          amount
          isClaimed
          networkFee
          tokenAddress
        }
        claimDelay
        recipientAddress
        id
        payoutModifier
      }
      status
    }
  }
}

```

5. Click the 'Edit via motion' button and follow the link logged in the console
6. Run the query again to verify that the 'editedExpenditureSlots' field of the attached motion is correctly storing all of the proposed changes
7. Stake both sides and vote to support the motion. Then finalize the motion.
8. Run the query a third time to verify that the motion has passed, and the payouts have updated.

## Diffs

**New stuff** ✨

* EditLockedExpenditureMotion saga

Resolves #1976

PR #2022, opened 2024-03-05
Fix `editExpenditure`, `releaseExpenditureStageMotion` sagas 

### LINKED ISSUE #2014: Advanced Payments API: Fix `releaseExpenditureStageMotion` saga
Instead of dealing with all the complexities of `setExpenditureState`, the motion to release a payment stage can simply execute `releaseStagedPaymentViaArbitration` function call (https://github.com/JoinColony/colonyNetwork/blob/8f4968cfa56675c49ebca3d7468d09a936585ab3/contracts/extensions/StagedExpenditure.sol#L126).

This issue is to update the saga and block-ingestor handlers accordingly.

jakubcolony:
## Description

Please see issue for details. In short, we were using incorrect storage slot when calling `setExpenditureState` which broke editing locked expenditure as well as releasing staged payment. It also appeared in Sam's #2021.

While investigating, I realised releasing staged payment via motion has never worked but we couldn't tell because we missed an important step while creating staged payments - setting the claim delays to max possible value. Releasing a stage simply means resetting that delay to 0. 

This subsequently revealed an issue with our schema which was storing claim delays as `Int`, whereas in the contracts they're represented as `uint128`. Therefore setting the claim delays to their maximum possible value was causing a crash of block-ingestor. The schema has been updated and here's the block-ingestor PR: https://github.com/JoinColony/block-ingestor/pull/188 

### Edit locked expenditure
Create and lock an expenditure. You might need to click "Refetch" once you lock, after which click on "Edit". Testing UI is using the following payload so run a query to verify all the values (recipient, amount, token address, claim delay) are being set correctly:
```ts
[
        {
          amount: '23.45',
          tokenAddress: colony.nativeToken.tokenAddress,
          recipientAddress: colony.colonyAddress,
          claimDelay: '0',
        },
        {
          amount: '67.89',
          tokenAddress: colony.nativeToken.tokenAddress,
          recipientAddress: user?.walletAddress ?? '',
          claimDelay: '300',
        },
      ]
```

### Releasing expenditure stage
Install voting reputation and staged expenditure extensions. Create a staged expenditure. Confirm the claim delay is set to 2**128 - 1 (`340282366920938463463374607431768211455`).

Create a motion to release a stage. Stake and finalize it. Apart from the released stage's `isReleased` property being set to true, you should also see the claim delay for the slot being set to 0.

Resolves #2014


iamsamgibbs (changes requested):
Editing a locked expenditure tested and working as expected!

Creating a staged expenditure correctly sets the claim delay, however I was unable to fund the expenditure after locking it so was unable to test the release stage motion.

bassgeta (approved):
I didn't get the issues above, I was able to fund a locked staged expenditure
image
and then finalize it
image
and when the motion passed, the payout was claimed and `claimDelay` set to 0, so nice one!
image

jakubcolony:
@iamsamgibbs would you be able to re-test considering both me and Kerry are unable to reproduce? If it still happens, can you give me a call to debug this live?

iamsamgibbs (approved):
Tested again today and it's all good 🎉  (leaving the transaction amount in the form at 0 prevents it from being funded, but works fine with an actual value)

PR #2026, opened 2024-03-06
Update `cleave.js` to `cleave-zen`

### LINKED ISSUE #2001: Cleave.js causes componentWillReceiveProps warning in the console
## Steps to reproduce

1. Create a new action which uses the 'Amount' field (Simple Payment / Transfer funds / Mint tokens).
2. Look for the 'componentWillReceiveProps' warning in the console.

## Suggestions

This is being caused by an old version of the `cleave.js` package which should be upgraded as newer versions of the package do not use componentWillReceiveProps.

It was deliberately downgraded to an older version in [this commit](https://github.com/JoinColony/colonyCDapp/pull/288/commits/1a998f6ddf4b0b99965c571d5ac14f541d84f7bf) although no reasoning behind this decision is given in the PR. Check where else cleave.js is being used in the project and check there are no breaking changes caused by upgrading.

ArmandoGraterol:
So this took longer than anticipated.

The author deprecated cleave.js: https://github.com/nosir/cleave.js/issues/723 and a new library has been created in its place: https://github.com/nosir/cleave-zen

The approach to how the library works changed almost completely, so I had to make some extra modifications to the forms (also to simplify the code as it was hard to understand what the heck was happening). I think the new library is easier to use and integrate.

Hopefully, everything still works as expected but now without the annoying warning of `componentWillReceiveProps`. The forms most affected by the changes are mint tokens, transfer funds, and simple payments.

Resolves #2001


rdig (changes requested):
Could reinstall npm packages, but couldn't get the dev environment running, due to the block ingestor not bulding.

Screenshot from 2024-03-07 23-37-49

Does perhaps this needs to be rebased ?

_(codewise looks all good)_

chmanie:
> Could reinstall npm packages, but couldn't get the dev environment running, due to the block ingestor not bulding.

`npm` had some hiccups yesterday, maybe it was that:

https://github.com/JoinColony/colonyCDapp/actions/runs/8192265293/job/22403383516?pr=2026#step:4:120

`npm run dev` running fine for me today

chmanie (approved):
This looks good code wise and functionality wise!

I was wondering whether we can do something about the cursor jumping when changing digits in the middle of the number. I know this was a problem before, I just thought it might be fixed by updating this library...

ArmandoGraterol:
@rdig It might be what Chris mentions but I've rebased it just in case

rdig (commented):
All good now. You guys were right, might have been a npm issue after all

rdig (approved):

Issue #170, opened 2023-01-11
Wire in "Edit" decision


willm30:
- [ ] Clicking "edit" should re-open the rich text editor.
- [ ] Clicking "Update" should persist changes to /preview

![edit-decision](https://user-images.githubusercontent.com/64402732/211741420-3ea3c3d5-f4b2-4cf4-8621-c2b5b958eb0c.png)

willm30:
Blocked by #168

PR #2031, opened 2024-03-08
[Verified members] Fix: various fixes

### LINKED ISSUE #2032: [Verified members] Verified members feature bugs
Tracking the bugs from [this comment](https://github.com/JoinColony/colonyCDapp/pull/1773#pullrequestreview-1922605395)

bassgeta:
## Description

Basically created a separate PR to solve issues from [this comment](https://github.com/JoinColony/colonyCDapp/pull/1773#pullrequestreview-1922605395).
I had to do A LOT of query refetching with Apollo, but I've tried to do it in the background, so I don't block the saga execution.
The schema needs to account if you're trying to verify already verified members by inputting the address and the opposite for removing.
This is the first instance of using field arrays in our forms, so our `useFlatFormErrors` didn't expect the format.
Basically if you have 3 fields, 2 with errors, and the  third one with an error you get back:
`[undefined, undefined, Error]`.

Related block-ingestor commit [#170](https://github.com/JoinColony/block-ingestor/pull/170/commits/875e1db23943a6a6ff810ba4bae8b310ad09fedd)

## Testing

Not much to say except going over the cases point by point and trying it out.

## Diffs

- a ton of query refetching
- move getting the verified members form validation schema to a helper function
- the `SelectedMembers` table now doesn't display loading rows if there is existing data (so when we invalidate, the table doesn't change).

Resolves #2032


ArmandoGraterol (changes requested):
Nice job! There are still some issues I was able to replicate:

- After removing verification and verifying once more, the check didn't appear:
Screenshot 2024-03-12 at 14 56 32

In general, I have noticed that the first time you make an action/motion, the relevant data refreshes without reloading but the second time it stops working for me.

- After removing 2 members, the meta description shows as if I was adding them:
Screenshot 2024-03-12 at 15 19 26

- In the activity feed, the actions are shown without a team.

melyndav (changes requested):
Hey @bassgeta rock star effort on this PR!

Some feedback on the updates that were missed: 

1. The general pill still does not show for permission type actions in the activity feed(s).
image

2. I removed verification from 4 members, however one still shows int he completed state withthe verification icon:
image

But it's removed from the members ui
image


3. When I trigger the user pop over from the table in the completed state it no longer breaks the UI (thanks for fixing!) but the title of the username changes to a different shade of gray?
image

However when triggered from the motion sidepanel in the same transaction window, the username is showing the correct gray colour token.

image

4. The issue of removing showing when add is selected in motions is still showing (#14 from previous issue).
image

Sing out if you want a call :)

src/utils/members.ts
```diff
@@ -22,3 +27,38 @@ export const getColonyContributorId = (
...
+
+      const contributorAddress =
+        getColonyContributorAddressFromId(colonyContributorId);
+      const isUserAffected = userAddresses.indexOf(contributorAddress) !== -1;
+
+      // refetch only if user's verified status was changed
+      return isUserAffected;
+    },
+  });
+};
```
jakubcolony, line 64: Can you explain why we need to call refetchQueries twice, preferably as an in-code comment?

jakubcolony (commented):
Hey @bassgeta, I left a couple of comments.

One thing I noticed is that `contributor` and `member` are kind of used interchangeably, do you think there's a case for deciding on one and making it consistent or do they actually need to be separate in some cases? Not sure this is one for this PR by the way, just spotted it here

src/components/v5/common/ActionSidebar/partials/Motions/steps/FinalizeStep/FinalizeStep.tsx
```diff
@@ -41,6 +40,7 @@ const FinalizeStep: FC<FinalizeStepProps> = ({
 }) => {
   const { canInteract } = useAppContext();
   const [isPolling, setIsPolling] = useState(false);
+  const hasFinalizedHandlerRun = useRef(false);
```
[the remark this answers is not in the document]
bassgeta: it's just standard for storing `meta` or i dunno how to put it :D other libraries with similar use cases do it via refs too like [rooks does](https://github.com/imbhargav5/rooks/blob/main/packages/rooks/src/hooks/useDidUpdate.ts) or [mantine](https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/hooks/src/use-did-update/use-did-update.ts) (it's not the same hook but quite similar)

src/utils/members.ts
bassgeta: I'll see if I can join them into one, but essentially the second one uses the `onQueryUpdated` callback, which should trigger just for the `GetColonyContributor` query and not any other ones. Maybe I can do a name check in the first line, nice catch

melyndav (changes requested):
Okay @bassgeta I got the pill to show after doing my 3rd restart ;) You're the best! 

A couple of issues:

1. In the completed state the verification icon is still hit or miss if it shows for add member type:
image
 
 2. In the table when you add a member in the remove type, the verified icon shows bigger in the table row than it does in the completed state. 

3. When I open a default verified members action, the metadata shows 'remove'. it should be managed until a type is selected.
image

ArmandoGraterol (commented):
All the issues I found last time were solved, nice work :) I checked if the marks were being updated, even verifying 20 members at the same time it seems to be working properly. However, for users without profile it seems like the mark is not showing:
Screenshot 2024-03-14 at 14 19 11

melyndav:
> All the issues I found last time were solved, nice work :) I checked if the marks were being updated, even verifying 20 members at the same time it seems to be working properly. However, for users without profile it seems like the mark is not showing: Screenshot 2024-03-14 at 14 19 11

@bassgeta here's how to handle addresses in tables that are not verified - https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-v5?type=design&node-id=3643-417416&mode=design&t=GU7PGCjsxu5u5B0A-4 

image

melyndav (approved):
Hey @bassgeta - super effort on continuing to update the PR.

I have good news! I have tried and tried but I can't replicate the issue where the verification icon showed sometimes. It's working every time I try to break it ;)  

The other two issues look to also be resolved after testing. Niceeeeeeee!

warrencolliver (approved):
Great work on this! the previous issues have been fixed and I can't find any new ones.

jakubcolony (approved):
LGTM, nice work @bassgeta!

PR #2034, opened 2024-03-11
Fix: Action button isLoading issue

iamsamgibbs:
## Description

The actionButton component currently does not actually use the isLoading property that can be passed to it. This is causing a console error on the extension installation page (http://localhost:9091/planex/extensions/VotingReputation). This PR ensures isLoading is actually used and resolves the console error.

## Testing

Go to the extension installation page (http://localhost:9091/planex/extensions/VotingReputation) ensure the 'Install' button works correctly and no console errors are logged.

## Diffs

**Changes** 🏗

* ActionButton now uses isLoading property


rdig (approved):
:+1:

PR #2035, opened 2024-03-11
Add reputation change percentage to completed action table

### LINKED ISSUE #2033: [Manage Reputation] Get old reputation and calculate change percentage on action side panel
The manage reputation actions (Smite/Award reputation) show a percentage of how much of the user's reputation changed with the action:

Image
Image

Currently, there's no way of calculating this percentage as for that we would need the reputation of the user by the time the action was made. Therefore, I suggest that we should move the `rootHash` prop we currently have on `motionData` to the action model and always store the `rootHash` by the time the action/motion was created. This way we should be able to get the previous reputation not only on motions, but on actions too.

Once the `rootHash` is available we can use `useUserReputation`, for example, to get the reputation and then calculate the percentage based on that.

ArmandoGraterol:
## Description

Added root hash to the action model, this way we can get the old user reputation, before the smiting/awarding and use that to calculate the percentage the rep changed, like how it works on the form version of the table.

## Testing

- Turn on the reputation monitor, except when you are going through the motion unless you want to accelerate the process.
- Create awarding/smiting actions and motions.
- Check that the percentage shown on the table is the same, or close. It may vary a little bit due to the reputation monitor advancing time and the reputation decaying by the time the block number of the action is mined.

Screenshot 2024-03-11 at 17 01 41
Screenshot 2024-03-11 at 17 01 56

Block ingestor PR: https://github.com/JoinColony/block-ingestor/pull/190

Resolves #2033


iamsamgibbs (approved):
Tested and all working as expected!

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -2861,6 +2856,10 @@ type ColonyAction @model @searchable {
   Metadata associated with the action (Eg. Custom action title)
   """
   metadata: ColonyActionMetadata @hasOne(fields: ["id"])
+  """
+  The reputation root hash at the time of the creation of the action
+  """
+  rootHash: String!
```
rdig, line 2862: Uff. I just realized. This needs a migration script, otherwise, once deployed to production, it will instantly break all existing motions.

rdig (approved):
A lot of changes in this one PR, which makes it harder to review, so most likely I'll try to do another pass at it.

From what I saw though, everything seems to check out, so congrats on some nice work.

As I mentioned, the schema change will break the live databases, both production and qa since they all have motions either passed or active.

It can stay like this until you finish the smite feature, but when you actually get to merging the feature branch into master, please ping me, as we'll then need to both write and the run a migration script on the database prior to merging the smite changes.

ArmandoGraterol:
@rdig 
> A lot of changes in this one PR, which makes it harder to review, so most likely I'll try to do another pass at it.

Yeah, I had to rebase as I think the parent branch was updated and it was including a commit that wasn't mine. It should be much more manageable now :)

> It can stay like this until you finish the smite feature, but when you actually get to merging the feature branch into master, please ping me, as we'll then need to both write and the run a migration script on the database prior to merging the smite changes.

You got it 👌

amplify/backend/api/colonycdapp/schema.graphql
jakubcolony: @rdig @ArmandoGraterol I think this needs to be done as a two-step migration to avoid downtime, the first step would be to add this field as non-nullable, migrate the existing data, then make it non-nullable in the schema.

PR #2036, opened 2024-03-12
Action sidebar preview for integrations

### LINKED ISSUE #1805: Refactor: prepare relevant components for integrations
Make sure `ActionSidebar` can be compiled as a separate entry point. This includes:

- refactoring all of the used components to use tailwind instead of CSS modules
- creating a new entry point that provides the required contexts (and uses stub contexts for info that is not present)

Eventually we might want to move these components to a separate repo or maybe npm module to be able to preview this without compiling all of the CDapp (let's see if this is even a problem in the future) - for this #1423 will need to be done first.

chmanie:
## Description

**NOTE:** this PR is based on #1748, which sets the foundation for rapid integration development (merge that first before reviewing this).

This main goal of this PR was to add an entrypoint to storybook which shows the ActionSidebar, without context in the CDapp. This should make it possible to test designs and functionality for the sidebar and containing actions without needing to run the whole dev environment.

Furthermore a few things were added/fixed:

- Upgraded Storybook to version 8
- Stories that previously required the colony context to work (e.g. Member Card) now actually work again 

## Testing

```
npm run storybook
```

Keep in mind that full functionality of the ActionSidebar is not part of this PR. There will be an entire refactor targeting this, making it possible to preview almost all of the actions without running the actual CDapp. 

Closes #1805


src/integrations/stories/ActionSidebar.stories.tsx
```diff
@@ -0,0 +1,15 @@
+import { type Meta, type StoryObj } from '@storybook/react';
```
rdig, line 1: What's the top level `integrations` folder going to be used for? Just the stories?

rdig (approved):
All good here, nice work.

The brunt of the changes here seem to be for removal / refactoring of barrel files, but that is not to say that the storybook additions are less important, or that they were "easy" :)

Screenshot from 2024-03-18 21-46-31

jakubcolony (approved):
LGTM, shows up in Storybook and doesn't introduce any obvious regressions to the CDapp's sidebar. Nice one 👍 

image

src/integrations/stories/ActionSidebar.stories.tsx
chmanie: Yeah, for now this is the spot where I think the actual integrations will end up that people write. Eventually it'd be good to maybe even have an own repository for that but for now I think this is the easiest way. Maybe this will even change when I refactor the ActionSidebar and/or write the first integration.

PR #2038, opened 2024-03-12
[Verified members] Fix core UI issues

### LINKED ISSUE #2029: Verified members action: Core UI issues
## Description

In an attempt to not roadblock #1773 all related/non-critical to the user experience UI issues have been added into one separate PR. 

1. Hover states in the table row meatballs are off in Firefox. They have a fixed width, so the hover state breaks the UI.
image

2. Sizing of the verified icons is wrong in the table. The icon should be 14x14px with 4px padding between the username and the icon as per the Figma component (linked below)
image

https://www.figma.com/file/l1dOM5qiQYwF0ElvKDqqjg/Design-System---Colony-v3?type=design&node-id=1782-151741&mode=design&t=sLQYtauTDdjBIAcm-4

warrencolliver:
## Description

This PR fixes two UI issues:

Verified Icon in the verified members action table was 20px and had 8px left padding - this PR updates the sizing of the icon to 14px and the left padding to 4px aligning with figma

Table row meatball item in a hover state breaks the UI but this was only occuring in firefox. 

## Testing

* Go to verified members action
*  select 'Remove members'
* Add a member to the selected members table
* Verify the verified icon size is 14px with 4px left padding

Firefox is required to test the meatball item issue
* select the table row meatball
* hover over 'Remove Row' and verify that it's hover state doesn't break the UI

**Changes** 🏗

Hovering over the 'remove account' item in firefox:
Screenshot 2024-03-12 at 11 40 08

Icon sizing updated:
Screenshot 2024-03-12 at 11 40 38

Resolves #2029

### DIFF
--- a/amplify/backend/api/colonycdapp/schema.graphql
+++ b/amplify/backend/api/colonycdapp/schema.graphql
@@ -3443,8 +3443,12 @@ type VerifiedMember @model {
   """
   A colony address - user address pair that tells us if a user is verified.
   """
-  colonyAddress: ID! @primaryKey(sortKeyFields: ["userAddress"])
-    @index(name: "verifiedMembersByColony", queryField: "getVerifiedMembersByColony")
+  colonyAddress: ID!
+    @primaryKey(sortKeyFields: ["userAddress"])
+    @index(
+      name: "verifiedMembersByColony"
+      queryField: "getVerifiedMembersByColony"
+    )
   userAddress: ID!
   user: User! @hasOne(fields: ["userAddress"])
   createdAt: AWSDateTime!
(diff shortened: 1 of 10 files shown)


bassgeta (changes requested):
Code looks good, but you may need to tweak the MeatBallMenu css a bit more, so on master currently the meatball menu items are all full width

image

and the same menu on your branch:
image

has them cutoff at their content width. Maybe we need some more flex-grow/whatever things :D I tested this on the members page, using the dropdown from the cards

jakubcolony (changes requested):
Nice work, seeing the same issue as Kerry

melyndav (commented):
Hey @warrencolliver nice work on this! Outside the other comments, the small UI updates look great. Thanks for picking up the issue. 👍

melyndav (approved):
Nice work on that fix @warrencolliver everything else looks good to go! 👍

bassgeta (approved):
Works like a charm, thanks for the update!

Issue #2005, opened 2024-02-29
Advanced Payments - Process: Permission Decision Method


AdrianDudko:
Teamwork task: [Advanced Payment - Process: Permission Decision Method](https://tw.auroracreation.com/#/tasks/15804108)
Created by: @AdrianDudko
Accepted by: @arrenv

## Related sub-issues/PR's

- [x] https://github.com/JoinColony/colonyCDapp/pull/2081
- [x] https://github.com/JoinColony/colonyCDapp/pull/2085 -> https://github.com/JoinColony/colonyCDapp/issues/2084
- [x] https://github.com/JoinColony/colonyCDapp/pull/2090 -> https://github.com/JoinColony/colonyCDapp/issues/2096
- [x] https://github.com/JoinColony/colonyCDapp/pull/2102
- [x] https://github.com/JoinColony/colonyCDapp/pull/2103
- [x] https://github.com/JoinColony/colonyCDapp/pull/2101
- [x] https://github.com/JoinColony/colonyCDapp/pull/2110
- [x] https://github.com/JoinColony/colonyCDapp/pull/2123
- [x] https://github.com/JoinColony/colonyCDapp/pull/2182

## User flows
![adv-payment-flow](https://github.com/JoinColony/colonyCDapp/assets/33682027/b2e13c20-706d-495c-a5ec-b35976a2c897)
## Acceptance criteria

**Stepper - Stage - Create**
- Clicking back on the Create stage pill will show the permission decision method was used.	

**Stepper - Stage - Review**
- First stage after creating the payment, it makes the payment public for others to review.
- Owner has the option to confirm details, locking the payment details and moving it to the Funding stage.
	- User who didn't create an action should not see a button/button should be diabled (TBD) 
- Clicking back on the Review stage pill will show the permission decision method was used.
- "Review" pill appears at the top of the action panel
	- Show tooltip when mouseover
		- "Payment is currently in review. The payment creator can make changes freely until details are confirmed."
						
**Stepper - Stage - Funding**
- Step to fund the payment with the required funds to make the payment.
- Opens the Funding modal.
- Funding opens a modal that allows the user see the total amount of what is going to get funded for each token and the network fee. The user also will need to select from one of the decision methods that is available to them.
- Funding with Permissions decision method:
    - Immediately moves the process to the Release stage.
	- Clicking back on the funding stage pill will show the permission decision method was used.

**Modals - Funding modal**
[...]

PR #1969, opened 2024-02-21
Connect expenditures with actions model

jakubcolony:
## Description

Advanced payments feature requires displaying payments in the activity feed. This PR introduces relations between expenditures and actions model so that there's a corresponding action created for each expenditure.

Additionally, it refactors the relation between Expenditure and UserStake models, making the latter available as a field on the expenditure.

I have also pulled in the commits from #1973 to make them available on `feat/advanced-payments-2` after merge - it's not the last time we modify the schema on this feature branch and manual copying has become a bit annoying.

block-ingestor PR: https://github.com/JoinColony/block-ingestor/pull/179

## Testing

1. Create an expenditure using testing UI.
2. Run a query to verify an action of type `CREATE_EXPENDITURE` has been created and that it has a linked expenditure to it, for example:
```gql
query ListActions {
  listColonyActions {
    items {
      id
      type
      expenditure {
        id
        nativeId
      }
    }
  }
}
```
3. Run a query to verify the expenditure also has an action linked to it:
```gql
query ListExpenditures {
  listExpenditures {
    items {
      nativeId
      action {
        id
        type 
      }
    }
  }
}
```
4. Test Simple payments still function and are stored in the database as `PAYMENT` action as opposed to `CREATE_EXPENDITURE`. On the contracts level they emit the same events so we need extra checks in block-ingestor to distinguish between the two.

Resolves #1102

PR #2039, opened 2024-03-12
Feat/15828803 action fields

CzarekDryl:
https://tw.auroracreation.com/app/tasks/15828803

**New stuff** ✨

* Added fields to created action - draft mode and completed action

## TODO

- Custom action title

## Out of scope

- Right side of action sidebar

Resolves [https://github.com/JoinColony/colonyCDapp/issues/2044](https://github.com/JoinColony/colonyCDapp/issues/2044)


jakubcolony:
@arrenv @AdrianDudko are we missing a corresponding Github issue? I'm not sure of the scope of this ticket and can't access the aurora creation link.

CzarekDryl:
@jakubcolony We do have issue but it applies to the whole task, not just part of it, because we make the subtasks ourselves.[https://github.com/JoinColony/colonyCDapp/issues/2005](https://github.com/JoinColony/colonyCDapp/issues/2005)

arrenv:
> @jakubcolony We do have issue but it applies to the whole task, not just part of it, because we make the subtasks ourselves.#2005

@CzarekDryl It would be good to copy the relevant acceptance criteria into it's own issue that can be referenced.

CzarekDryl:
@arrenv @jakubcolony New issue added

jakubcolony (changes requested):
Thanks for adding the related issue @CzarekDryl.

It looks like this PR needs rebasing but I ran some high-level functional testing. A few things I noticed:

- Action type is `Generic` instead of `Payment builder`:
Screenshot 2024-03-14 at 22 30 03

- It is possible to crash the app when you create a payment with some specific amounts (e.g. `5342534`) which is most likely due to JS float precision, we need to be able to handle it though:


https://github.com/JoinColony/colonyCDapp/assets/112586815/2e369ba7-b6da-4084-a0a7-adf03fd67d12

CzarekDryl:
@jakubcolony I think that I've fixed that issue with decimals, please check it again.

src/components/v5/common/ActionSidebar/hooks/useGetActionData.ts
```diff
@@ -232,6 +241,36 @@ export const useGetActionData = (transactionId: string | undefined) => {
...
+          payments: slots?.map((slot) => {
+            if (!slot) {
+              return undefined;
+            }
+
+            const currentToken = tokens?.items.find(
+              (slotToken) =>
+                slotToken?.token.tokenAddress ===
+                slot?.payouts?.[0].tokenAddress,
+            );
```
jakubcolony, line 257: Can also use the `getSelectedToken` here

src/components/v5/common/ActionSidebar/ActionSidebar.tsx
```diff
@@ -38,6 +40,9 @@ const ActionSidebar: FC<PropsWithChildren<ActionSidebarProps>> = ({
 }) => {
   const { action, defaultValues, loadingAction, isMotion, motionState } =
     useGetActionData(transactionId);
+  const { expenditure, loadingExpenditure } = useGetExpenditureData(
+    action?.expenditureId,
+  );
```
jakubcolony, line 45: This hook seems to be called in `useGetActionData` as well, so maybe whatever data you need here could be exported from that hook instead?

src/components/v5/common/ActionSidebar/hooks/useGetExpenditureData.ts
```diff
@@ -0,0 +1,16 @@
+import { useGetExpenditureQuery } from '~gql';
+
+export const useGetExpenditureData = (
+  expenditureId: string | null | undefined,
+) => {
+  const { data, loading } = useGetExpenditureQuery({
+    variables: {
+      expenditureId: expenditureId || '',
+    },
+  });
```
jakubcolony, line 10: The query shouldn't just run with an empty id, please make sure to specify `skip` option to avoid unnecessary requests

jakubcolony (changes requested):
Thanks for your work @CzarekDryl, left a couple of comments in code.

melyndav (changes requested):
Hey @CzarekDryl thanks for your work on this PR.

Every time I try a new transaction to test I just get caught in this loading state. I've tried opening it again using the transaction URL and get the same loading state. It was using permissions.

image

If I try using reputation, I can't get past the pending state. 

image

Maybe I'm not doing something correctly here?

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -753,6 +753,7 @@ enum ColonyActionType {
   An action related to creating expenditure (dvanced payment) 
   """
   CREATE_EXPENDITURE
+  TEMP_ADVANCED_PAYMENT
```
jakubcolony, line 756: Looks like you've pulled my changes from #1969 - that means `TEMP_ADVANCED_PAYMENT` isn't needed anymore

jakubcolony (changes requested):
Looking very solid @CzarekDryl, thanks for your changes.

I left some comments in code, and also noticed the following:

- The table breaks quite a bit shortly before going mobile, also notice how the description breaks into multiple lines but `leela` stays aligned to the right:
image

- Overlapping issue on larger screen width with longer amounts:
image

- Neither the recipients nor the tokens are clickable which feels inconsistent with every other instance of user/token link that opens a popover. But I'm not sure what the requirements say about this so one for @melyndav to confirm maybe? 🙂 

- I also thought I saw the payments table load without any rows but I wasn't able to replicate that - noting it here though in case anyone else spots it.

melyndav (changes requested):
Hey @CzarekDryl thanks for helping me fix my issue.

I got it to work and have some feedback.

1. When I'm only paying 1 user, the metadata is showing 'multiple recipients and tokens' which isn't accurate. It should show the number here when only 1 member is being paid.
image

2. The table looses it's styling in the completed state.
image

There should be NO horizontal rules between members at any stage (during the action creation and completed states) but only between the last row and the total field as per Figma below: 

image


3. The mobile total field at the bottom of the table has padding issues:
image
 
 Figma showing correct padding:
 
image

4. Unverified members are missing their alert icon showing in the table.
image


5. Size of the verified icon and the padding after the username should match Figma on desktop and mobile - it should match the same size as the building state before the CTA is selected.
image

Link to example: https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-v5?type=design&node-id=3643-417416&mode=design&t=GU7PGCjsxu5u5B0A-4

iamsamgibbs (changes requested):
Thanks for your work so far! I've noticed a couple of issues:

It isn't possible to change the token type on mobile.

https://github.com/JoinColony/colonyCDapp/assets/34915414/3556a530-8c49-488a-a825-fa809de6bc96

On the completed view, the horizontal rules should continue the entire way across. The total number of tokens should also be right aligned.

Screenshot 2024-03-19 at 10 36 24

Figma:

Screenshot 2024-03-19 at 10 36 36

CzarekDryl:
@iamsamgibbs First one is out of scope because in this issue I'm taking care of completed action.

jakubcolony (changes requested):
Codewise it looks good 👍 

I'm still able to cause some overlay issues however, and there is a weird horizontal scrollbar in the user popover (which doesn't happen elsewhere this component is used):

Screenshot 2024-03-19 at 18 57 10

image

CzarekDryl:
@jakubcolony Currently, I have no idea how to improve the table, it would be best if Melynda came up with something. Can we merge it into a feature branch and then fix it, because we are seriously blocked by this task.

arrenv:
Adding a comment here regarding the table at table view, the solution to this can just be to toggle the mobile version of the table at this size.

I think this is worth it's own issue though, so, I have created an issue here https://github.com/JoinColony/colonyCDapp/issues/2072 and this PR, so that this PR can be completed without including this fix.

melyndav (changes requested):
@CzarekDryl all looks good to go, I've approved it. :)

jakubcolony (approved):
LGTM code-wise, good to merge once Mel's comments are addressed 👍

iamsamgibbs (approved):

PR #2040, opened 2024-03-12
fix: Not able to finalize agreements

### LINKED ISSUE #1876: Not able to "Finalize" agreements

warrencolliver:
## Description

This PR fixes the issue which was occuring when trying to finalize an agreement / decision action. this was done by adding a condition so agreements used `MOTION_CLAIM` instead of `MOTION_FINALIZE`. By doing this the stakes can be returned to the user and the agreement is closed. When the agreement is at the 'Finalize' step, the button should also be labelled 'Return Stakes' rather than 'Claim' like other actions. 

## Testing

* Install reputation extension
* Create agreement motion and go through the steps
* Get to finalize step, and return the stakes

Expected result: you should not see any errors and the stakes are returned to the user:
finalizeagreements

You can also repeat these steps but instead of claiming via the motion 'Return stakes' claim via the user hub and check if the decision finalizes without any issues.

## Diffs

**Changes** 🏗
Decision motion check added into finalizeStep to use MOTION_CLAIM to theoretically 'finalize' the motion  

Resolves #1876

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/Motions/steps/FinalizeStep/FinalizeStep.tsx
+++ b/src/components/v5/common/ActionSidebar/partials/Motions/steps/FinalizeStep/FinalizeStep.tsx
@@ -199,13 +199,11 @@ const FinalizeStep: FC<FinalizeStepProps> = ({
                         mode="primarySolid"
                         disabled={!canClaimStakes || wrongMotionState}
                         isFullSize
-                        text={
-                          isMotionAgreement
-                            ? formatText({
-                                id: 'motion.finalizeStep.returnStakes',
-                              })
-                            : formatText({ id: buttonTextId })
-                        }
+                        text={formatText({
+                          id: isMotionAgreement
+                            ? 'motion.finalizeStep.returnStakes'
+                            : buttonTextId,
+                        })}
                         type="submit"
                       />
                     )}
--- a/src/components/v5/common/ActionSidebar/partials/Motions/steps/FinalizeStep/FinalizeStep.tsx
+++ b/src/components/v5/common/ActionSidebar/partials/Motions/steps/FinalizeStep/FinalizeStep.tsx
@@ -56,6 +56,14 @@ const FinalizeStep: FC<FinalizeStepProps> = ({
     canClaimStakes,
   } = useClaimConfig(actionData, startPollingAction, refetchAction);
 
+  const isMotionFinalized = actionData.motionData.isFinalized;
+  const isMotionFailedNotFinalizable =
+    actionData.motionData.motionStateHistory.hasFailedNotFinalizable;
+  const isMotionAgreement =
+    actionData.type === ColonyActionType.CreateDecisionMotion;
+  const isMotionClaimable =
+    isMotionFinalized || isMotionFailedNotFinalizable || isMotionAgreement;
+
   const handleSuccess = () => {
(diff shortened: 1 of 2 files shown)


iamsamgibbs (changes requested):
This looks good and all works as expected! Just one small change: would be better if `formatText` took in an id from `src/il8n/en.json`

iamsamgibbs (approved):

src/components/v5/common/ActionSidebar/partials/Motions/steps/FinalizeStep/FinalizeStep.tsx
```diff
@@ -84,7 +85,8 @@ const FinalizeStep: FC<FinalizeStepProps> = ({
   };
   if (
     actionData.motionData.isFinalized ||
-    actionData.motionData.motionStateHistory.hasFailedNotFinalizable
+    actionData.motionData.motionStateHistory.hasFailedNotFinalizable ||
+    actionData.type === ColonyActionType.CreateDecisionMotion
```
bassgeta, line 89: maybe we could put this (perhaps some other too) into boolean variables for improved readability

bassgeta (changes requested):
Works like a charm, just dropped a comment if we may want to wrap some if statements into variables for more readability. Not necessary tho, just some QoL improvements :)

src/components/v5/common/ActionSidebar/partials/Motions/steps/FinalizeStep/FinalizeStep.tsx
```diff
@@ -175,28 +180,32 @@ const FinalizeStep: FC<FinalizeStepProps> = ({
...
-                        text={formatText({
-                          id: buttonTextId,
-                        })}
+                        text={
+                          isMotionAgreement
+                            ? formatText({
+                                id: 'motion.finalizeStep.returnStakes',
+                              })
+                            : formatText({ id: buttonTextId })
+                        }
```
ArmandoGraterol, line 208: A small nitpick but this way we avoid repetition:
```
formatText({ id: isMotionAgreement ? 'motion.finalizeStep.returnStakes' : buttonTextId })
```

ArmandoGraterol (changes requested):
Nice job Warren! Just left a nitpick but after that this should be good to go from my side

ArmandoGraterol (dismissed):
LGTM 👍 Thank you for the changes

iamsamgibbs (approved):
Looks good to me!

PR #2045, opened 2024-03-13
Feat: Reduce use of @apply from module.css files

### LINKED ISSUE #1885: Refactor: Remove reduce use of @apply from .module.css files
The way we are currently using `@apply` in `.module.css` files is adding unnecessary bloat to the CSS bundle.

This issue involves refactoring all `.module.css` files to remove `@apply` where possible.

—

1. Using `@apply` just to avoid having long lines of tailwind directly in the components. 

```
.pills {
  @apply inline-flex shrink-0 items-center text-center px-3 py-1 rounded-3xl capitalize;
}
```

This should just be used inline in the component file instead.

`<span className={clsx(“inline-flex shrink-0 items-center text-center px-3 py-1 rounded-3xl capitalize”, …`

—

2. Using `@apply` because the same class is used multiple times in the component.

```
<div className={styles.row}>
  <p className={styles.rowName}>
…
<div className={styles.row}>
  <p className={styles.rowName}>

```

This should be a variable in the component instead.

```
const rowClass = “…”
Const rowNameClass = “...”
```

—

3. Using `@apply` classes like an object to apply certain styles depending on a value.

```
.xxs {
  @apply relative w-[1.125rem] h-[1.125rem] rounded-full;
}

.xs {
  @apply h-[1.625rem] w-[1.625rem] relative rounded-full;
}

.s {
  @apply h-[2.125rem] w-[2.125rem] relative rounded-full;
}
…
```

Which is used like this: `const class = styles[‘xs’];`

We should instead have a variable in the component file which is an object containing the different tailwind classes.

```
const sizeClasses = {
  xxs: ‘relative w-[1.125rem] h-[1.125rem] rounded-full’,
  xs: ‘h-[1.625rem] w-[1.625rem] relative rounded-full’,
  s: ‘h-[2.125rem] w-[2.125rem] relative rounded-full’,
  …
} 
```

Which is used like this: `const class = sizeClasses[‘xs’];`

It’s a small but important change as it means we get the tailwind classes added directly rather than a css module class.

We could keep these in their own `styles.ts` files rather than directly in the component if it bloats the file too much. If we end up doing this, we would need to establish a naming convention for these files.

—

4. Using `@apply` to apply our own styles where a third party package is adding components with set classnames.

```
:global :local(.toast-notification) .Toastify__toast-body {
  @apply p-0;
}

:global :local(.toast-notification) .Toastify__toast-icon {
  @apply hidden;
}
```

This is a good example of how we should be using `@apply` and should remain unchanged.

—

Care should be taken in situations where a `.module.css` file is imported in multiple places. A `styles.ts` file containing the new tailwind variables should be used and imported as required.

—

Part of this issue is to decide on a naming convention for any new variables required.

Adding `tailwindCSS.experimental.classRegex` to the `/vscode/settings.json` file enables tailwind autocomplete to variables wherever the regex matches. For example, this will add autocomplete to any variable ending in `Classes`.

```
  "tailwindCSS.experimental.classRegex": [
    ["Classes \\=([^;]*);", "'([^']*)'"],
    ["Classes \\=([^;]*);", "\"([^\"]*)\""],
    ["Classes \\=([^;]*);", "\\`([^\\`]*)\\`"]
  ]
```

This regex could be expanded to include `Class` or `Style`.

We could also add a comment `// @tw` wherever tailwind classes are being used.

(Note the /vscode folder is currently ignored in the .gitignore file)

—

Each of the scenarios above should be tested and ensure they work as expected before making changes throughout the project.

This issue should remove as many uses of `@apply` and `.module.css` files as possible without any changes to styling.

iamsamgibbs:
## Description

The way we were using @apply in .module.css files was adding unnecessary bloat to the CSS bundle.

This PR removes as many instances of @apply as possible. See the issue for reasoning behind this decision: https://github.com/JoinColony/colonyCDapp/issues/1885

This PR adds a new `tw` utility function intended for use when tailwind classes are being assigned to a variable. It works in conjunction with `prettier-plugin-tailwindcss` to automatically sort the classes in template literal strings tagged with 'tw'.

**Remaining uses of @apply**

Only the following .module.css files with uses of @apply remain:

```
./src/components/shared/Extensions/Toast/Toast.module.css
./src/components/v5/shared/MembersSelect/MembersSelect.module.css
```

No change necessary as we are applying our own styles where a third party package is adding components with set classnames.

```
./src/components/v5/shared/RichTextDisplay/RichTextDisplay.module.css
./src/components/v5/shared/RichText/partials/RichTextContent/RichTextContent.module.css
./src/components/v5/common/Fields/Select/Select.module.css
```

 Target too many specific children elements to convert to tailwind.

```
./src/components/v5/common/MemberCardList/partials/MemberCardPlaceholder/MemberCardPlaceholder.module.css
```

Cannot properly apply to background-image url with tailwind

```
./src/components/v5/common/SubNavigation/partials/SubNavigationItem/SubNavigationItem.module.css
```

Cannot hide scrollbar with tailwind

After rebasing following the vite merge, there are a few renamed .css -> .module.css files, none of these contain any uses of @apply and are just css so I have left them untouched. We may want to consider refactoring these to use tailwind as a separate issue.

## Testing

Check the app builds okay and there are no styling regressions.

## Diffs

**New stuff** ✨

* New `tw` utility function - when assigning classes to a variable, they should be wrapped in a template literal tagged with 'tw' to provide automatic sorting and intellisense.

**Changes** 🏗

* Removal of @apply where possible

Resolves #1885


chmanie (dismissed):
This is great!

Love the tw function. Good work @iamsamgibbs!

warrencolliver (changes requested):
Hey, great work on this! Builds fine for me but I did notice one small thing in the UI - it seems the text on some button hover states is blue instead of white. I found these regressions on: 

the balance page (add funds to colony)
Screenshot 2024-03-14 at 18 02 52

Extensions page
Screenshot 2024-03-14 at 18 04 17

I couldn't find any other examples of this, but I may have missed some.

ArmandoGraterol (approved):
Other than what Warren pointed out, this looks good so I'm approving 👌

ArmandoGraterol (approved):
Looking good to me 👍 The issue Warren pointed out seems to be fixed.

warrencolliver (dismissed):
Thanks for the change can't see any other issues, LGTM

melyndav (approved):
Everything seems to be hunky-dory on the front-end side of things @iamsamgibbs. Can't spot anything out of the ordinary both in and out of a Colony context. So another smashed out of the park! 💯

warrencolliver (approved):

Issue #2055, opened 2024-03-18
Actions: Staking widget UI Issues


melyndav:
Teamwork task: [Actions: Staking widget UI Issues](https://tw.auroracreation.com/#/tasks/15839355)

## Description

Across all action types, the staking widget has the support and oppose button colours are incorrect for all states (default, selected etc). This issue should be resolved and tested across all responsive types including mobile.

The issue shows across all the staking steps including when a choice is selected by the user and the change state is available. 

<img width="341" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/5d8598f2-21db-4f89-9320-b2e34caec9df">
 
**As per Figma:**
 
<img width="572" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/1c121b89-a281-41fe-813d-5a6fc0707dd0">

## Figma
Figma link for SOT: https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-v5?type=design&node-id=3154-357025&mode=design&t=GU7PGCjsxu5u5B0A-4

AdrianDudko:
Closed by https://github.com/JoinColony/colonyCDapp/pull/2060

Issue #2056, opened 2024-03-18
Motions: Default outcome state 


melyndav:
Teamwork task: [Motions: Default outcome state](https://tw.auroracreation.com/#/tasks/15839356)

## Description

Across all responsive sizes including mobile, when a motion is created the default outcome pill shows as red. This is incorrect and should match Figma during all stages until a outcome is received during the motion flow where the pill should update accordingly. 

<img width="366" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/34e8f6d0-5cd5-4bff-b186-62e1a5bd9d60">

Figma showing the pill in it's default state:
<img width="470" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/4075589e-8b22-4b96-820c-f90704bb0efe">

## Figma
Step navigation states: https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-v5?type=design&node-id=3280-369096&mode=design&t=GU7PGCjsxu5u5B0A-4

PR #2060, opened 2024-03-18
fix: staking widget radio buttons colors

joanna-pagepro:
https://github.com/JoinColony/colonyCDapp/issues/2055

Updated colors for radio buttons

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/Motions/consts.ts
+++ b/src/components/v5/common/ActionSidebar/partials/Motions/consts.ts
@@ -1,4 +1,5 @@
 import { ThumbsDown, ThumbsUp } from '@phosphor-icons/react';
+import clsx from 'clsx';
 
 import { MotionVote } from '~utils/colonyMotions.ts';
 import { formatText } from '~utils/intl.ts';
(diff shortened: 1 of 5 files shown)


warrencolliver (changes requested):
The buttons are the correct colours, so good work. However I noticed the outcome step widget in figma is black, whereas in the app it's red, should this be changed too?
Screenshot 2024-03-18 at 13 47 26

joanna-pagepro:
@warrencolliver there's a separate task for that and it's already updated in a different PR
https://github.com/JoinColony/colonyCDapp/issues/2056

warrencolliver (approved):
Apologies, good work!

bassgeta:
could we maybe utilize native CSS for this?
https://developer.mozilla.org/en-US/docs/Web/CSS/:checked
https://developer.mozilla.org/en-US/docs/Web/CSS/:disabled

Tailwind also has support for these [selectors](https://tailwindcss.com/docs/hover-focus-and-other-states#disabled)

joanna-pagepro:
@bassgeta I'm not sure if it will work correctly because the disabled state is added to the input and not the label

bassgeta:
> @bassgeta I'm not sure if it will work correctly because the disabled state is added to the input and not the label

I mean since the label is actually the next sibling of the input we could use the `+` css selector. Maybe worth a shot, since className as a function feels unconventional

joanna-pagepro:
@bassgeta there's another wrapper inside the label that is getting the classes passed with className prop. I think that this solution is cleaner and you can still pass just a string.

melyndav (approved):
I checked both oppose and support flows and all looks to be corrected. Thanks @joanna-pagepro

warrencolliver (approved):

PR #2069, opened 2024-03-19
Fix amplify watcher script

### LINKED ISSUE #2063: Amplify watcher script crashes on startup
Following the merge of #1748, the amplify watcher script crashes when running `npm run dev`:

image

## Steps to reproduce

`npm run dev`

jakubcolony:
## Description

Please see issue for description.

I tried a few different configuration changes but I could not get the original TS script to work, and after consulting with Chris, I decided to change it to JS which now runs without issues. 

I also removed the `voyager` script, don't think it was used that much and the built-in playground improved quite a bit since I added the script. If you found it useful, you can still use their web version: https://graphql-kit.com/graphql-voyager/.

## Testing

Run `npm run dev`, you should see no errors from the watcher script and, once the `amplify` container starts, the copying files over should work as normal.

Resolves #2063


bassgeta (approved):
works for me, I've added a property to the last entity in the file, `cat` the file from docker
image
and it works for me, LGTM!

ArmandoGraterol (approved):
LGTM 👌

PR #2070, opened 2024-03-20
Fix leading decimal points crashing CDapp on manage rep form

### LINKED ISSUE #2067: Fix decimal point crashing CDapp on manage reputation form
## Steps to reproduce

1. Open the action side panel.
2. Select the manage reputation action.
3. Fill the required fields to enable the amount field.
4. Input a decimal point in the amount input first thing and watch the CDapp crash.

## Expected behaviour

* Inputting a `.` should count as a `0` to the validation

## Actual behaviour

* The CDapp crashes

ArmandoGraterol:
Check out the issue linked to this PR for info on how this bug was reproduced, but basically, inputting a `.` first thing in the amount field was crashing CDapp due to that char reaching the `new Decimal()` inside `formatReputationChange`. With `toFinite` this shouldn't be a problem any longer and it should be treated as a `0`.

Screenshot 2024-03-19 at 21 25 35

Resolves #2067


bassgeta (approved):
image
works, LGTM :ship: or LGTFB I guess :D

iamsamgibbs (approved):
Tested and working as expected!

Issue #2054, opened 2024-03-18
Global: User popovers


melyndav:
Teamwork task: [Global: User popovers](https://tw.auroracreation.com/#/tasks/15839237)

## Description

In production, there are a number of UI issues with the user popover cards. The user popovers can be triggered by selecting a user names across various points of the CDapp. 

All issues below should be resolved and tested on all responsive sizes including mobile.

**Issue 1: Usernames that are wallets have broken padding inside the popover compared to pre-set usernames:**

Wallet address username with broken padding:
<img width="349" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/d9068e08-0b1c-43b3-b852-4114e8026626">

Username with correct padding:
<img width="349" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/a5cc619a-a1bc-434a-814e-98b892a6c156">
 
 **Issue 2: Usernames with a wallet address are missing the full wallet address showing the in the alert status within the pop over. The full wallet address should show in this block:**
 
<img width="349" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/9c5eed52-3994-4f73-8881-a16bb6536c0f">

**Issue 3: Popovers that do not have a description break padding when the alert box is shown for non-verified members.** 

As you can see int he screenshot below, this user does not have a description and is tagged as the 'top' member type. The correct padding inside the card should be present if a user does or does not provide a custom description.

<img width="373" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/5dc9e8f4-e462-4a4e-8a75-b58461d11cba">

Correct padding when a description is shown:
<img width="341" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/d7372f3e-2469-4898-ad2d-526fe87b3ab4">

## Figma 
Figma component for reference for all issues:  https://www.figma.com/file/l1dOM5qiQYwF0ElvKDqqjg/Design-System---Colony-v3?type=design&node-id=1782-42685&mode=design&t=pjBfl9vi2n9SXfxZ-4

AdrianDudko:
According to @CzarekDryl  comment, reported issues have been fixed in https://github.com/JoinColony/colonyCDapp/pull/2066

PR #2071, opened 2024-03-20
fix: user popover ui fixes

CzarekDryl:
[tw](https://tw.auroracreation.com/app/tasks/15839237)

## Description

- Fixed few ui cases of user popover

Resolves [https://github.com/JoinColony/colonyCDapp/issues/2054](https://github.com/JoinColony/colonyCDapp/issues/2054)

### DIFF
--- a/src/components/v5/shared/UserPopover/partials/UserInfo.tsx
+++ b/src/components/v5/shared/UserPopover/partials/UserInfo.tsx
@@ -65,14 +65,15 @@ const UserInfo: FC<UserInfoProps> = ({
               text={formatText({ id: 'userInfo.top.contributor.in' })}
             />
             <div className="flex gap-1">
-              {domains?.slice(0, 3).map(({ domainName, domainId }) => (
-                <div key={domainId}>
+              {domains
+                ?.slice(0, 3)
+                .map(({ domainName, domainId }) => (
                   <UserStatus
+                    key={domainId}
                     mode="team"
                     text={multiLineTextEllipsis(domainName, 7)}
                   />
-                </div>
-              ))}
+                ))}
               {domains?.length > 3 && (
                 <UserStatus mode="team" className="!w-auto !max-w-none">
                   +{domains.length - 3}
--- a/src/components/v5/shared/UserPopover/partials/UserInfo.tsx
+++ b/src/components/v5/shared/UserPopover/partials/UserInfo.tsx
@@ -65,15 +65,14 @@ const UserInfo: FC<UserInfoProps> = ({
               text={formatText({ id: 'userInfo.top.contributor.in' })}
             />
             <div className="flex gap-1">
-              {domains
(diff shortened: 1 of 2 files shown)


melyndav (changes requested):
Hey @CzarekDryl thanks for picking up my issue! Appreciate it. :)

Just one bit of feedback as all the rest look to have been resolved. 

1. Can we make the padding inside the white area around the alert card 24px to match Figma. Currently you can see the top and bottom padding is uneven:
image
 
As per Figma:
 
image

This should always be even when a description is shown or not.

ArmandoGraterol (changes requested):
Looking good to me! We just need to fix the lint error related to the class name order

src/components/v5/shared/UserPopover/partials/UserInfo.tsx
```diff
@@ -65,15 +65,14 @@ const UserInfo: FC<UserInfoProps> = ({
               text={formatText({ id: 'userInfo.top.contributor.in' })}
             />
             <div className="flex gap-1">
-              {domains
-                ?.slice(0, 3)
-                .map(({ domainName, domainId }) => (
+              {domains?.slice(0, 3).map(({ domainName, domainId }) => (
+                <div key={domainId}>
```
bassgeta, line 69: any particular reason we wrap this in a div now?

bassgeta (changes requested):
Looks good other than the minor comment I dropped, nice one!

src/components/v5/shared/UserPopover/partials/UserInfo.tsx
CzarekDryl: I had a ts error, but now it's gone, so I'm reverting the old changes.

iamsamgibbs (dismissed):
Looks good to me!

bassgeta (dismissed):
Nice one!

ArmandoGraterol (commented):
I think there might be too much padding now:
Screenshot 2024-03-27 at 12 53 53

It is combining the margin of the top section plus the top padding of the warning

bassgeta (approved):

CzarekDryl:
@ArmandoGraterol, @iamsamgibbs Could you please take a look at this one?

iamsamgibbs (changes requested):
This is very close now, only one minor issue I noticed.

There is not enough padding above "about" - seems to be when a user is not verified:

Screenshot 2024-04-03 at 14 14 05

Compared to a verified user:
Screenshot 2024-04-03 at 14 14 35

ArmandoGraterol (commented):
Nice work! The issue I pointed out last time seems to be fixed. Only 1 thing left and that would be what Sam already metioned

bassgeta:
Can we check if this PR is applicable at all, since #2066 got merged and that took care of some paddings and margins on this component?

PR #2081, opened 2024-03-21
Payment builder widgets UI

joanna-pagepro:
Part of https://github.com/JoinColony/colonyCDapp/issues/2005

**Payment Builder widget with mock data. Right now you can go to the next steps, but no actual action will be triggered to proceed with the action. I also wasn't able to get the info about who clicked `Fund` or `Release` buttons, and because of that I wasn't able to display the box with the info about it when going back to these stages.** - there's an issue related to this: https://github.com/JoinColony/colonyCDapp/issues/2082

**We are focusing on one flow right now - with the permissions decision method only.**

**Stepper - Stage - Create**
- Clicking back on the Create stage pill will show the permission decision method was used.

**Stepper - Stage - Review**
- First stage after creating the payment, it makes the payment public for others to review.
- Owner has the option to confirm details, locking the payment details and moving it to the Funding stage.
  - User who didn't create an action should not see a button/button should be diabled (TBD)
- Clicking back on the Review stage pill will show the permission decision method was used.
- "Review" pill appears at the top of the action panel
  - Show tooltip when mouseover
    - "Payment is currently in review. The payment creator can make changes freely until details are confirmed."

**Stepper - Stage - Funding**
- Step to fund the payment with the required funds to make the payment.

Out of scope:
- Opens the Funding modal.
- Funding opens a modal that allows the user see the total amount of what is going to get funded for each token and the network fee. The user also will need to select from one of the decision methods that is available to them.
- Funding with Permissions decision method:
  - Immediately moves the process to the Release stage.
  - Clicking back on the funding stage pill will show the permission decision method was used.

**Stepper - Stage - Release**
Out of scope:
- Opens the Release payment modal with the users possible decision methods
- If funded, the owner can release the funds for the payment, starting the claim delay.
- If any claim delays are 0, then those payments will happen immediately on release and also release the creators stake if present.
- Alternatively, users can use available decision methods, including "Admin" tiered permissions or higher (Arbitration on the contracts)

**Stepper - Stage - Payment**
- Appears after the funds have been released.

Out of scope:
- Provides a break down of total funds to be paid, paid, and are ready to be paid.
- Shows a timer that counts down to the next available claim time.
- Button to "Make payment"
  - Button is disabled if there is nothing to pay out
  - Button is enabled as long as there is something to pay out. Anyone connected member is able to click and make the payment.
  - First payout will return the creators stake.
  - Button does not appear anymore if there are no more funds to be paid out and there are no claims to be released.

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/ExpenditureBadge/ExpenditureBadge.tsx
+++ b/src/components/v5/common/ActionSidebar/partials/ExpenditureBadge/ExpenditureBadge.tsx
@@ -2,6 +2,7 @@ import clsx from 'clsx';
 import React, { type FC } from 'react';
 
 import { ExpenditureStatus } from '~gql';
+import Tooltip from '~shared/Extensions/Tooltip/index.ts';
 import { formatText } from '~utils/intl.ts';
 import PillsBase from '~v5/common/Pills/PillsBase.tsx';
 
@@ -16,7 +17,7 @@ const ExpenditureBadge: FC<ExpenditureBadgeProps> = ({ status }) => {
     [ExpenditureStatus.Locked]: formatText({ id: 'expenditure.locked' }),
   };
 
-  return (
+  const pill = (
     <PillsBase
       className={clsx(
         EXPENDITURE_STATE_TO_CLASSNAME_MAP[status],
@@ -25,6 +26,16 @@ const ExpenditureBadge: FC<ExpenditureBadgeProps> = ({ status }) => {
     >
       {badgeTexts[status]}
     </PillsBase>
+  );
+
+  return status === ExpenditureStatus.Draft ? (
+    <Tooltip
+      tooltipContent={formatText({ id: 'expenditure.reviewStage.tooltip' })}
+    >
+      {pill}
+    </Tooltip>
+  ) : (
+    pill
   );
 };
 
--- a/src/components/v5/common/ActionSidebar/partials/forms/PaymentBuilderForm/partials/PaymentBuilderRecipientsField/hooks.tsx
+++ b/src/components/v5/common/ActionSidebar/partials/forms/PaymentBuilderForm/partials/PaymentBuilderRecipientsField/hooks.tsx
@@ -93,7 +93,6 @@ export const useRecipientsFieldTableColumns = (
                     return {
                       ...result,
                       [tokenAddress]: {
-                        ...tokenData,
                         amount: BigNumber.from(
                           !result[tokenAddress]?.amount ||
                             result[tokenAddress]?.amount === ''
--- a/src/components/v5/common/CompletedAction/CompletedAction.tsx
+++ b/src/components/v5/common/CompletedAction/CompletedAction.tsx
@@ -7,7 +7,6 @@ import { ExtendedColonyActionType } from '~types/actions.ts';
 import { type ColonyAction } from '~types/graphql.ts';
 import { getExtendedActionType } from '~utils/colonyActions.ts';
 
-import { useGetExpenditureData } from '../ActionSidebar/hooks/useGetExpenditureData.ts';
 import PermissionSidebar from '../ActionSidebar/partials/ActionSidebarContent/partials/PermissionSidebar.tsx';
 import Motions from '../ActionSidebar/partials/Motions/index.ts';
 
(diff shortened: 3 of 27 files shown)


iamsamgibbs (changes requested):
Good work on this so far! There are a few minor UI issues I've noticed:

When revisiting a previous step that has passed, the background should be light-blue-400.

Screenshot 2024-03-22 at 16 40 53

Figma:

Screenshot 2024-03-22 at 16 41 12

These icons should be light-gray-500.

Screenshot 2024-03-22 at 16 43 53

Screenshot 2024-03-22 at 16 46 27

joanna-pagepro:
@iamsamgibbs colors updated. I've also noticed that the Verified badge was duplicated in the UserSelect component after conflicts were resolved on the feature branch, so I also fixed it.

iamsamgibbs (changes requested):
Thanks for those changes and for fixing the verification icon too!

Just noticed the information icon on the 'Create' and 'Review' stages still need the colour change, but otherwise looking good!

joanna-pagepro:
@iamsamgibbs you're right! I missed it. It's updated now.

melyndav (changes requested):
Hey @joanna-pagepro 

For the next PR, can we include the test criteria in the PR? It's a lot of work to decipher the exact criteria when the Product specs are verbatim copied and the PR template is not utalized. Thanks! 

https://github.com/JoinColony/colonyCDapp/blob/master/.github/PULL_REQUEST_TEMPLATE.md

Feedback:

1. Can we align the icons in all widgets better with Figma. Currently the info icons site ajar from the text.

Currently:
image
 
 
 Figma: 
image

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/PaymentStageDetailsBlock/partials/PaymentItem/types.ts
```diff
@@ -0,0 +1,9 @@
+import { type Token } from '~types/graphql.ts';
+
+export interface PaymentItem extends Token {
+  amount: string;
+}
+
+export interface PaymentItemProps {
+  payouts: PaymentItem[];
```
jakubcolony, line 8: At some point this will need to work on the GraphQL types such as `ExpenditureSlot`, `ExpenditurePayouts`, etc. So maybe not something to change now, but to keep in mind when we get there.

jakubcolony (changes requested):
Nice work @joanna-pagepro, did some testing of the UI and everything's looking good.

After progressing through all stages I'm able to go back to funding/release and still see the action button - is that something that will be fixed when you have the required data?
image

The main reason I'm requesting changes is due to the possible confusion between expenditure stage being used in this PR to describe a step of the expenditure flow vs expenditure stage in the context of Staged Expenditure payments. Could you please rename "stage" to "step" in all instances?

joanna-pagepro:
@jakubcolony 
> After progressing through all stages I'm able to go back to funding/release and still see the action button - is that something that will be fixed when you have the required data?

It will be added in separate PRs, where actions for release and fund will be implemented.

@melyndav ok, the description for the next one will be created based on the template. I've changed the size of the icon in the box. I couldn't see any other differences there.

melyndav (changes requested):
Hi @joanna-pagepro thanks for the update witht he icon.

In the payment overview widget a couple of fixes are needed.

1. The chevron colour should be gray to match Figma and all other chevrons across motion widgets.
Figma: 
image

2. The spacing between different tokens in each of the sections in this widget should total 4px as shown in Figma.
image

3. Can you add a hover state to these chevrons so it matches the same as the other motion widgets like voting and staking that have the same interaction?

joanna-pagepro:
@melyndav can you add more details to 3? I'm not sure what elements you're referring to. What effect is expected here?

melyndav:
> @melyndav can you add more details to 3? I'm not sure what elements you're referring to. What effect is expected here?

Hey @joanna-pagepro I just checked and we actually don't have any hover states for these small chevrons across motion widgets, so please ignore this request since it won't be added globally as part of this PR. I'll log it for an issue to do later across every type of motion widget. Thanks

joanna-pagepro:
@melyndav @iamsamgibbs @jakubcolony can you review this PR again?

jakubcolony (approved):
LGTM code-wise, thanks for addressing the feedback 👍

iamsamgibbs (approved):
This is all looking good to me!

melyndav (approved):
Thanks for the updates @joanna-pagepro! Nice work!

PR #2083, opened 2024-03-21
Get all fund expenditure actions

### LINKED ISSUE #2082: Advanced Payments API: Get all expenditure funding actions
## Specification

[The UI](https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-v5?type=design&node-id=21066-441669&mode=design&t=rLHtmzKX5vkLZ2W3-4) needs to know the initiator of the `MOVE_FUNDS` action that funded the expenditure.

Although it's not possible via UI, technically there could be multiple actions funding the expenditure.

This issue is to provide a way of getting all the actions that have funded the expenditure.

jakubcolony:
## Description

This PR enables the UI to get all the "Move Funds" actions that funded an expenditure. It refactors the existing relation between actions and expenditures to make it many-to-many, with the idea that you can use filters to select actions related to expenditure creation, funding or any other action type we add in the future.

I considered a few approaches to this and all shared the same issue when it comes to motions. When a motion is created, a corresponding action is created, and then on motion finalization the triggered transactions cause an almost identical action to be created, though with different tx hash, type (e.g. `CREATE_DOMAIN` vs `CREATE_DOMAIN_MOTION`) and, most importantly, action initiator address which in this case is the voting reputation extension.

Those actions triggered by motion finalization are not shown in the UI because we check whether action initiator or recipient is an extension and set `showInActionsList` field accordingly. Coincidentally, this field is also used to hide motions that do not met the 10% staking threshold and "Move Funds" actions funding expenditures. This meant we could not simply fetch the expenditures related actions without also fetching motion finalization ones. To deal with this, I added a new field to action model: `isMotionFinalization`. 

Related block-ingestor PR: https://github.com/JoinColony/block-ingestor/pull/192

## Testing

Test funding expenditures via actions (you can fund expenditure multiple times using testing UI) and motions, running the following query in-between:

```gql
query ListExpenditures {
    items {
      nativeId
      nativeFundingPotId
      fundingActions: actions(
        filter: {type: {eq: MOVE_FUNDS}, isMotionFinalization: {ne: true}}
      ) {
        items {
          id
          initiatorAddress # This should be your wallet address, not voting reputation
          type # There should be no `MOVE_FUNDS_MOTION` or `CREATE_EXPENDITURE` actions here
          amount
          fromPotId 
          toPotId # Should match `nativeFundingPotId`
        }
      }
    }
  }
}
```

Resolves #2082


amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -3237,11 +3253,10 @@ type Expenditure @model {
...
   type: ExpenditureType!
   """
-  Hash of the `ExpenditureAdded` event transaction
-  @TODO: Make this a non-nullable field once existing data is updated
+  Actions associated with the expenditure
   """
-  transactionHash: ID
-  action: ColonyAction @hasOne(fields: ["transactionHash"])
+  actions: [ColonyAction!]
+    @hasMany(indexName: "byExpenditureId", fields: ["id"])
```
rdig, line 3259: I have no idea how we need to handle this migration on the db side. But there's a good chance it's done automatically for us.

rdig (approved):
Seems to work as the PR description expected

Screenshot from 2024-03-22 12-44-29

_Although I mentioned this in our internal channels, I want to mention here that I had read trouble with starting with branch. Problems seem to not really be related to the changes here, however one think I want to note is that the create data script is now super sensitive to race conditions as if you run it "too fast" it tends to overwrite / use the same nonces._

amplify/backend/api/colonycdapp/schema.graphql
jakubcolony: Fortunately it's never been deployed. The original `action` field was only added in #2015.

warrencolliver (approved):
All looks good and working as expected, couldn't find any issues. Great work on this.

Issue #2084, opened 2024-03-25
Adv payments: funding modal


CzarekDryl:
### Story

- Modal to reviewing the amount to be funded and select the decision method to use.

- Network fee needs to be added on top of the entered amount to ensure the recipient receives their expected funds without the network fee being deducted from their expected payment.

- Cancel closes the modal and does nothing. You can also click out to close the modal.

- Clicking on "Fund payment" will start the funding process, with the process dependent on the decision method chosen.

- Decision methods

    - Options
    - Permissions
    - Visible if the user holds the Payer permissions or Funding custom permission in the team.
    - Permissions will immediately fund and move to the release step.
        - Selecting a decision method will reveal a callout to explain the next step that will happen based on their selection.

AdrianDudko:
Closed by https://github.com/JoinColony/colonyCDapp/pull/2085

PR #2085, opened 2024-03-25
feat: added funding modal

CzarekDryl:
Added funding modal to payment builder completed action - for now it's added as temp button because we don't have stepper.
image

## Description

- Modal to reviewing the amount to be funded and select the decision method to use.

- Network fee needs to be added on top of the entered amount to ensure the recipient receives their expected funds without the network fee being deducted from their expected payment.

- Cancel closes the modal and does nothing. You can also click out to close the modal.

- Clicking on "Fund payment" will start the funding process, with the process dependent on the decision method chosen.

- Decision methods

    - Options
    - Permissions
    - Visible if the user holds the Payer permissions or Funding custom permission in the team.
    - Permissions will immediately fund and move to the release step.
        - Selecting a decision method will reveal a callout to explain the next step that will happen based on their selection.

Resolves [https://github.com/JoinColony/colonyCDapp/issues/2084](https://github.com/JoinColony/colonyCDapp/issues/2084)


src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/DecisionMethodSelect/index.ts
iamsamgibbs, line 1: Export from here / delete empty file.

iamsamgibbs (commented):
Good start on this issue!

I couldn't find an example on the figma design, but my instinct would be that each token should only appear once in the list with the cumulative amount of all payments. @melyndav can you confirm?

Screenshot 2024-03-25 at 09 40 40

Is the "Reputation" decision method out of scope for this PR?

Screenshot 2024-03-25 at 09 36 50

AdrianDudko:
> Is the "Reputation" decision method out of scope for this PR?
> 
> Screenshot 2024-03-25 at 09 36 50

Hi @iamsamgibbs

Yes, for now, Reputation will be hidden in every step Decision method can be chosen - even if the Colony has an extension installed.
That's because we are focusing on delivering a fully functional "Permissions" path first.

melyndav:
> Good start on this issue!
> 
> I couldn't find an example on the figma design, but my instinct would be that each token should only appear once in the list with the cumulative amount of all payments. @melyndav can you confirm?
> Screenshot 2024-03-25 at 09 40 40
>

Hey @CzarekDryl - @iamsamgibbs is correct. The same token should be bundled together.

melyndav (changes requested):
Hey @CzarekDryl nice start on this one.

1. I'm seeing double verification icons show in the member dropdown. 
image

2. The table didn't carry the info across and the modal doesn't show any data. It worked after refreshing. 
image
 
image
 
 
 3. This select input should be 44px h as per Figma:
 
image

Figma:
image

4. The font styling of 'select method' should match Figma (Regular size)
image
 
 As per Figma:
 
image

5. When the input dropdown is triggered, the option shows in the hover state without it being on active hover. 
image

 6. The weights of the borders of the input and the options dropdown do not match and should be 1px for consistency with the Figma component.
image

7. I shouldn't be able to type in this input, it should be static. 
image

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/FundingModal/hooks.ts
```diff
@@ -0,0 +1,40 @@
...
+    amount: slot.payouts?.[0].amount || '0',
+    fee: slot.payouts?.[0].networkFee || '0',
+    tokenAddress: slot.payouts?.[0].tokenAddress || '',
+  }));
+
+  const fundExpenditure = useAsyncFunction({
+    submit: ActionTypes.EXPENDITURE_FUND,
+    error: ActionTypes.EXPENDITURE_FUND_ERROR,
+    success: ActionTypes.EXPENDITURE_FUND_SUCCESS,
+  });
```
jakubcolony, line 20: I think the preference would be to use `ActionButton` in `FundingModal`, this also allows you to handle loading states easier. This hook could then go away potentially.

jakubcolony (changes requested):
Nice start on the funding modal @CzarekDryl, I left some in-code comments where some improvements could potentially be made

CzarekDryl:
@melyndav 1. and 2. are out of scope, could you please add them as separate issues?

melyndav:
> @melyndav 1. and 2. are out of scope, could you please add them as separate issues?

Both these issues relate to the branch as 1 is not on master and 2 is specific to this PR. I believe both should be fixed in this PR scope to ensure a) the double verification isn't pushed to production and b) the data actually shows in the UI correctly.

CzarekDryl:
@melyndav This changes are not on master because we are working on feature branch while doing advanced payments. I think this issues were fixed by Joanna in her PR - [https://github.com/JoinColony/colonyCDapp/pull/2081](https://github.com/JoinColony/colonyCDapp/pull/2081)

CzarekDryl:
@melyndav @iamsamgibbs @jakubcolony Could you please check this once again?

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/FundingModal/FundingModal.tsx
```diff
@@ -0,0 +1,131 @@
...
+    ([tokenAddress, { amount, networkFee }]: [
+      string,
+      { amount: number; networkFee?: number },
+    ]) => ({
+      tokenAddress,
+      amount: amount.toString(),
+      networkFee: networkFee?.toString(),
+      key: tokenAddress,
+    }),
+  );
```
jakubcolony, line 60: This chaining looks very confusing. If you're already using reduce, you should be able to transform the data in such way that `Object.entries` and `map` can be omitted. Otherwise, please consider splitting the transformation into multiple steps so it's easier to follow.

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/FundingModal/FundingModal.tsx
```diff
@@ -0,0 +1,131 @@
...
+      if (!slot.payouts) return grouped;
+
+      const mergedPayouts = { ...grouped };
+
+      slot.payouts.forEach((payout) => {
+        const { tokenAddress, amount, networkFee } = payout;
+
+        mergedPayouts[tokenAddress] = {
+          amount:
+            (mergedPayouts[tokenAddress]?.amount || 0) + parseFloat(amount),
```
jakubcolony, line 41: Those operations should use `BigNumber` to avoid running into JS precision problems

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/FundingModal/FundingModal.tsx
```diff
@@ -0,0 +1,131 @@
...
+    }, {}),
+  ).map(
+    ([tokenAddress, { amount, networkFee }]: [
+      string,
+      { amount: number; networkFee?: number },
+    ]) => ({
+      tokenAddress,
+      amount: amount.toString(),
+      networkFee: networkFee?.toString(),
+      key: tokenAddress,
```
jakubcolony, line 58: What's the benefit in having the token address duplicated under `key`? Could you not use `tokenAddress` directly? That seems to also solve the need for two types for "token item" and "funding item"

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/FundingModal/types.ts
```diff
@@ -0,0 +1,13 @@
+import { type ExpenditurePayout } from '~gql';
+import { type Expenditure } from '~types/graphql.ts';
+import { type ModalProps } from '~v5/shared/Modal/types.ts';
+
+export type TokenItemProps = Omit<ExpenditurePayout, 'isClaimed'>;
+
+export interface FundingItem extends TokenItemProps {
+  key: number;
+}
```
jakubcolony, line 9: I might be missing something but "token item" and "funding item" seem to be the same thing - except the `key`. Could we possibly keep just one?

jakubcolony (changes requested):
Thanks for amending the types @CzarekDryl, I'm sure it will make things easier in the long term. I left a few more comment regarding those, after which I think we'll be good to merge.

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/FundingModal/types.ts
CzarekDryl: `FundingItem` wasn't even used, my mistake. It's removed.

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/FundingModal/TokenItem.tsx
```diff
@@ -0,0 +1,46 @@
...
+            {formatText({ id: 'fundingModal.fee' })}
+          </span>
+        </div>
+      )}
+      <div className="flex flex-col items-end gap-1">
+        <span className="text-3">
+          <Numeral value={amount} decimals={token?.decimals} /> {token?.symbol}
+        </span>
+        <span className="text-xs text-gray-600">
+          <Numeral value={toFinite(networkFee)} decimals={token?.decimals} />{' '}
```
jakubcolony, line 38: Why do we need `toFinite`?

jakubcolony (approved):
LGTM, just left one comment regarding the `toFinite` function

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/FundingModal/TokenItem.tsx
CzarekDryl: I saw one of Armando's PR recently and I thought it would also be needed here because of decimal numbers, but from what I see it also works well without it.

melyndav (changes requested):
Hey @CzarekDryl,

Thanks for the fixes. Mobile buttons should be stickied to the bottom of the screen on mobile. Everything else looks on track. 

image

CzarekDryl:
@jakubcolony @iamsamgibbs Same situation as here - [comment](https://github.com/JoinColony/colonyCDapp/pull/2090#issuecomment-2026825063).

melyndav (approved):
Nice work @CzarekDryl thanks for all the fixes. Approving it from my side of things. Appreciate all the effort on this one!

jakubcolony (approved):

Issue #2062, opened 2024-03-18
Motions: Various widget UI issues


melyndav:
Teamwork task: [Motions: Various widget UI issues](https://tw.auroracreation.com/#/tasks/15840824)

## Description

There are a number of UI issues occurring across various motion widgets for all responsive sizing including desktop and mobile. These issues should be resolved to ensure they match Figma SOT and create a consistent and strong level of UX across the CDapp. 

The issues should be tested and resolved across all responsive sizes unless stated. 

## Issues 

- **Issue 1: Motion steps all have individual tool tips. These are currently broken on desktop as shown below. The tooltip should not display off the screen and wrap to a new line where needed. This is for desktop only as mobile tool tips are being addressed in a separate PR**

<img width="1111" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/229f89b2-62d6-4114-b523-299dacd0d62d">
 
 
<img width="452" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/5b45b1a5-dd37-4ea5-a69f-7f756167b70e">

 

- **Issue 2: The outcome widget has a number of UI issues and does not match Figma including the colour of headings and paddings etc. These issues appear in both outcome states.**

 
<img width="399" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/cf61fbc6-b1eb-4a95-bd19-10d566ed967f">
 
 How it should look per Figma:
 
<img width="279" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/6c90c012-d5de-4304-8aea-97bc97ee1ab0">

 Link to Figma component for all responsive sizes and variations: https://www.figma.com/file/l1dOM5qiQYwF0ElvKDqqjg/Design-System---Colony-v3?type=design&node-id=1782-149418&mode=design&t=pjBfl9vi2n9SXfxZ-4 
 
 
 

-  **Issue 3: Font sizes across the various motion widgets in the top information header do not match Figma. Here in the screenshot the reveal example doesn't match Figma with two different font sizes being used. This issue occurs over a number of widgets where the progress bar has been used in motions. **

 
 Reveal example:
<img width="327" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/3040710c-1016-4f5d-a722-a84376e4f0e9">

Voting example: 
[...]

PR #2094, opened 2024-03-26
Maintenance: Refactor Popover/Tooltip components

### LINKED ISSUE #1812: Refactor Popover/Tooltip components
Originally posted by @chmanie in #1816:

We have a quite involved [Popover component](https://github.com/JoinColony/colonyCDapp/blob/master/src/components/shared/Popover/Popover.tsx) that is only used once. Maybe we can investigate whether it is still necessary to have this (also to keep things consistent). Maybe also investigate the Tooltip component (which is used twice) in the same directory.

iamsamgibbs:
## Description

This PR removes a [Popover component](https://github.com/JoinColony/colonyCDapp/blob/master/src/components/shared/Popover/Popover.tsx) that was only used once.

It also removes the tooltip component from the same directory.

As part of this I tidied up the NativeTokenPill component where the popover component was in use to add hover and active states, and made the token popup function as a modal on mobile to be consistent with the over token popups across the app.

## Testing

Check the NativeTokenPill on the dashboard and colony admin page to ensure they have not regressed.

Screenshot 2024-03-26 at 14 28 48

Screenshot 2024-03-26 at 14 28 59

I'm not sure how which transactions get the `MessageCard` component to show - for the sake of fixing this I adjusted the `TransactionList` component to manually get it to show.

Screenshot 2024-03-26 at 15 36 20

## Diffs

**Changes** 🏗

* NativeTokenPill no longer uses popover component
* MessageCardStatus no longer uses old tooltip component

**Deletions** ⚰️

* Popover component
* Tooltip component

Resolves #1812


src/components/v5/common/NativeTokenPill/NativeTokenPill.tsx
```diff
@@ -20,29 +26,81 @@ const NativeTokenPill = ({
...
+        >
+          {token.symbol}
+        </span>
+        {isLocked && (
+          <Tooltip
+            tooltipContent={
+              <span>{formatText({ id: 'tooltip.lockedToken' })}</span>
+            }
+          >
+            <LockKey size={11} className="ml-0.5" />
```
rdig, line 74: As I commented on @bassgeta 's #2066, I really wish we'd had `enums` for sizes

rdig (approved):
All good from my perspective, nice work! I've picked up a couple of minor things, which you should look into, but overall this is good to go.

The token popovers didn't seem to have regressed

Screenshot from 2024-03-26 21-05-04
Screenshot from 2024-03-26 21-05-28

However, the tooltips on the the motion pills are being cut out to the right of the view port _(although not sure if it's related to this or not)_

Screenshot from 2024-03-26 21-07-15

Regarding your comment

> I'm not sure how which transactions get the MessageCard component to show - for the sake of fixing this I adjusted the TransactionList component to manually get it to show.

Anything that does message signing: voting, revealing, metatxs _(although I actually think we've removed it from showing up at all)_

melyndav:
@rdig @iamsamgibbs just an FYI. I opened an issue for the desktop tooltips on motion pills [here.](https://github.com/JoinColony/colonyCDapp/issues/2062).

melyndav (approved):
Heya @iamsamgibbs nice work on this one!

Token popovers look like they should. We discussed the tooltip scope, so I'm happy to see this one go live! Thanks again. :)

PR #2095, opened 2024-03-26
Update Change Action modal logic

### LINKED ISSUE #2089: Actions: Triggering the close modal
## Description

Currently, when a user switches action types without entering any content into each of the action's default inputs, they are prompted to confirm the close action change via a modal.

This modal adds extra steps to the user's navigation within the action types and adds cognitive load to the user's overall experience. As part of this PR, removing this modal being triggered is needed. 

## PR scope:

The scope of this PR is to remove this modal from being triggered in the case where no information or data has been inputed or changed in any action on all responsive sizes allowing the user to freely switch being action types. 

The modal should only trigger if a user has changed any of the content in the action inputs first before changing. 

The title should not be impacted with this change. 

Modal in scope:
image

image

ArmandoGraterol:
With these changes, now the modal should show only when you edit any fields other than the action type and title. Otherwise, it should never show when you change from one action type to another, for example.

Resolves #2089

### DIFF
--- a/src/components/v5/common/ActionSidebar/ActionTypeSelect.tsx
+++ b/src/components/v5/common/ActionSidebar/ActionTypeSelect.tsx
@@ -12,10 +12,7 @@ import SearchSelect from '~v5/shared/SearchSelect/index.ts';
 
 import ActionFormRow from '../ActionFormRow/index.ts';
 
-import {
-  ACTION_TYPE_FIELD_NAME,
-  DECISION_METHOD_FIELD_NAME,
-} from './consts.ts';
+import { ACTION_TYPE_FIELD_NAME, TITLE_FIELD_NAME } from './consts.ts';
 import useActionsList from './hooks/useActionsList.ts';
 import { translateAction } from './utils.ts';
 
@@ -42,8 +39,9 @@ const ActionTypeSelect: FC<ActionTypeSelectProps> = ({ className }) => {
     HTMLButtonElement,
     HTMLDivElement
   >([isSelectVisible]);
-  const { formState, setValue } = useFormContext();
+  const { formState, setValue, reset, watch } = useFormContext();
   const { readonly } = useAdditionalFormOptionsContext();
+  const title = watch('title');
 
   return (
     <div className={className}>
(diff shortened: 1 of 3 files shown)


iamsamgibbs (changes requested):
Great work on this, so much nicer switching between action types now! I've only noticed one very minor issue with the amount field:

https://github.com/JoinColony/colonyCDapp/assets/34915414/75947571-a9ed-4de5-a237-2c8a8a2d7268

(Happens on any action type with the amount field, enter an amount, change action type, you will be prompted as expected, change again to any other action and you will be incorrectly prompted despite making no changes)

src/components/v5/common/ActionSidebar/ActionTypeSelect.tsx
```diff
@@ -42,8 +39,9 @@ const ActionTypeSelect: FC<ActionTypeSelectProps> = ({ className }) => {
     HTMLButtonElement,
     HTMLDivElement
   >([isSelectVisible]);
-  const { formState, setValue } = useFormContext();
+  const { formState, setValue, reset, watch } = useFormContext();
   const { readonly } = useAdditionalFormOptionsContext();
+  const title = watch('title');
```
jakubcolony, line 44: Could also use `TITLE_FIELD_NAME`

src/components/v5/common/ActionSidebar/ActionTypeSelect.tsx
```diff
@@ -103,16 +101,19 @@ const ActionTypeSelect: FC<ActionTypeSelectProps> = ({ className }) => {
...
-                    actionType
-                  ) {
+                  const hasMadeChanges = Object.keys(
+                    formState.dirtyFields,
+                  ).find(
+                    (fieldName) =>
+                      fieldName !== 'title' && fieldName !== 'actionType',
+                  );
+
+                  if (hasMadeChanges && actionType) {
```
jakubcolony, line 111: It's a bit manual now that we have to specify the "common" fields here, and then make sure we're also setting them in `onConfirm`. Could those be stored in a config somewhere that we could iterate over?

jakubcolony (changes requested):
Nice start @ArmandoGraterol, however I'm also seeing the issue with fields being "left over" from the previously selected actions.

iamsamgibbs (approved):
This is great, all working as expected, and I like the addition of NON_RESETTABLE_FIELDS.

No longer experiencing the previous issue 🎉

https://github.com/JoinColony/colonyCDapp/assets/34915414/721bcd64-3619-4765-8482-20e87955a99d

bassgeta (approved):
Works, nice one :muscle:

jakubcolony (approved):
LGTM, works like a charm now 👌

melyndav (approved):
OMG thank you for picking up this issue @ArmandoGraterol. It's a small but huge UX improvement. It was so frustrating to get that modal anytime I wanted to explore actions so thank you for picking up my issue.

PR #2097, opened 2024-03-27
Finalizing expenditure as non-owner (using motions)

### LINKED ISSUE #1978: Advanced Payments API: Finalizing expenditure as non-owner (using motions)
Related specs from the main issue (#1956):

> (Stage) Release
> - Opens the Release payment modal with the users possible decision methods.
> - If funded, the owner can release the funds for the payment, starting the claim delay.
> - If any claim delays are 0, then those payments will happen immediately on release and also release the creators stake if present.
> - Alternatively, users can use available decision methods, including “Admin” tiered permissions or higher (Arbitration on the contracts), or use the Reputation decision method.

This issue is to implement a `finalizeExpenditureMotion` saga allowing non-owners to create a motion which, upon finalizing, will call `setExpenditureState` to also set the expenditure as finalized.

jakubcolony:
## Description

This PR introduces a new saga - `finalizeExpenditureMotion` - which allows users to create a motion to finalize an expenditure by calling `finalizeExpenditureViaArbitration` contract method.

## Testing

- Enable voting reputation in your colony
- Create an expenditure, lock and fund it, then use the "Finalize via motion" button
- Progress the motion to finalization
- Run the following query to verify expenditure status has changed and the motion is linked to the expenditure:
```gql
query ListExpenditures {
    items {
      nativeId
      status
      motions {
        items {
          id
          action {
            type
          }
        }
      }
    }
  }
}
```

Resolves #1978


ArmandoGraterol (approved):
Working as expected for me 👌 
Screenshot 2024-03-27 at 13 56 16

src/redux/sagas/motions/expenditures/finalizeExpenditureMotion.ts
```diff
@@ -0,0 +1,141 @@
...
+    if (
+      !colony ||
+      !expenditure ||
+      !votingReputationAddress ||
+      !motionDomainId
+    ) {
+      throw new Error('Invalid payload');
+    }
+
+    if (colony.version < 15) {
```
rdig, line 45: Can we get the colony version from colonyJS ? I know they all have tags, so would be nice if we could import it.

ie:
```
if (colony.version < ColonyVersion.GLWSS) {
```

Or something similar...

iamsamgibbs (approved):
All tested and working great! Code looks great too. (Do we need to check colony version on the block-ingestor side too?)

Screenshot 2024-03-28 at 10 15 29

src/redux/sagas/motions/expenditures/finalizeExpenditureMotion.ts
```diff
@@ -0,0 +1,141 @@
...
+        encodedAction,
+        key,
+        value,
+        branchMask,
+        siblings,
+      ],
+      group: {
+        key: batchKey,
+        id: meta.id,
+        index: 1,
```
rdig, line 101: Nitpick, and I don't remember this clearly, but I remember something as `index` only being required if you wanted to overwrite it (for custom transaction display)

Otherwise, you can just leave it out and it will progress incrementally

rdig (approved):
All good here.

I've tested this locally and it works as expected _(ignore that I've created two motions to finalize with, but just used one)_

Screenshot from 2024-03-28 12-13-24

Screenshot from 2024-03-28 12-13-17

Screenshot from 2024-03-28 12-14-05
Screenshot from 2024-03-28 12-14-14

src/redux/sagas/motions/expenditures/finalizeExpenditureMotion.ts
jakubcolony: we're still in development of v15 so that release has no tag yet I'm afraid

rdig: Good point. In that case, ignore this :)

Issue #2093, opened 2024-03-26
Cannot stake less than maximum for support or oppose in motions


bassgeta:
1. Create any motion
2. Click support/oppose and enter a number less than the maximum
3. The `Stake` button stays disabled

![image](https://github.com/JoinColony/colonyCDapp/assets/23449297/b6d7d046-a089-44b9-b3da-96a63fd50f98)

Upon clicking `Max`, it enables.

![image](https://github.com/JoinColony/colonyCDapp/assets/23449297/72c67b2b-1a69-4a83-8f8c-df0cc344125a)

## Expected behaviour

* Any valid number of tokens should get staked

## Actual behaviour

* Stake button is disabled

rdig:
I think this and #2092 should be tackled together

bassgeta:
Closing manually, because it didn't automatically close with the PR

PR #2098, opened 2024-03-27
fix: two way binding of FormFormattedInput

### LINKED ISSUE #2092: Cannot activate less than maximum amount of tokens
1. Open the user hub
2. Click activate tokens and enter an amount less than the maximum
3. See how you get an error instead of it activating the tokens

image

## Expected behaviour

* Any valid number of tokens should activate

## Actual behaviour

* User gets an error

bassgeta:
## Description

This fixes 2 bugs, both appearing on the token activation modal
image

and the staking widget

image

The first bug was that if you entered a number manually, it didn't work, it just threw an error that you need to enter a value.
The second bug was that if you clicked `Max` and then tried to change the value, it broke and the console spat out an error.

## Testing

Try both of those cases to see if they were fixed :D 

## Diffs

**Changes** 🏗

* two way data binding for `FormFormattedInput` restored

Resolves #2092 and #2093 (maybe I should have foreseen them being the same issue :facepalm: )

### DIFF
--- a/src/components/v5/common/Fields/InputBase/FormFormattedInput.tsx
+++ b/src/components/v5/common/Fields/InputBase/FormFormattedInput.tsx
@@ -14,7 +14,7 @@ const FormFormattedInput: FC<FormFormattedInputProps> = ({
   ...rest
 }) => {
   const {
-    field: { value },
+    field: { value, onChange },
     fieldState: { invalid, error },
   } = useController({
     name,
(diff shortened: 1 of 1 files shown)


iamsamgibbs (approved):
All working as expected!

rdig (approved):
Working as expected. Nice job fixing it! :100: 

Screenshot from 2024-03-27 13-31-52
Screenshot from 2024-03-27 13-32-52

Issue #2086, opened 2024-03-25
Advanced Payments API: Return info about Release action


jakubcolony:
## Specification

Similarly to #2083, the UI requires info about the Release action, which is analogous to finalizing an expenditure on the contracts. However, currently we don't create a colony action for `ExpenditureFinalized` event, so this will need to be implemented first.

### This issue is complete when...

- [ ] There's a colony action created when expenditure is finalized
- [ ] There's `finalizingActions` field on the expenditure in the UI

PR #2101, opened 2024-03-28
feat: release overview content

CzarekDryl:
## Description

- Opens the Release payment modal with the users possible decision methods
- If funded, the owner can release the funds for the payment, starting the claim delay.
- If any claim delays are 0, then those payments will happen immediately on release and also release the creators stake if present.
- Alternatively, users can use available decision methods, including "Admin" tiered permissions or higher (Arbitration on the contracts)

### DIFF
--- a/src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/PaymentBuilderWidget/utils.ts
+++ b/src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/PaymentBuilderWidget/utils.ts
@@ -32,24 +32,31 @@ export const isExpenditureFullyFunded = (expenditure?: Expenditure | null) => {
     return false;
   }
 
-  const amountsByToken = expenditure.slots.map((slot) => {
-    const amounts = {};
+  const slotAmountsByToken = expenditure.slots.flatMap((slot) => {
+    const amounts: { tokenAddress: string; amount: BigNumber }[] = [];
 
     slot.payouts?.forEach((payout) => {
       if (!payout.isClaimed) {
-        amounts[payout.tokenAddress] = BigNumber.from(
-          amounts[payout.tokenAddress] ?? 0,
-        ).add(payout.amount);
+        const existingAmountIndex = amounts.findIndex(
+          (item) => item.tokenAddress === payout.tokenAddress,
+        );
+        if (existingAmountIndex !== -1) {
+          amounts[existingAmountIndex].amount = amounts[
+            existingAmountIndex
+          ].amount.add(payout.amount);
+        } else {
+          amounts.push({
+            tokenAddress: payout.tokenAddress,
+            amount: BigNumber.from(payout.amount),
+          });
+        }
       }
     });
 
     return amounts;
   });
 
-  return amountsByToken.every((token) => {
-    const [tokenAddress] = Object.keys(token);
-    const amount = token[tokenAddress];
-
+  return slotAmountsByToken.every(({ tokenAddress, amount }) => {
     const tokenBalance = expenditure.balances?.find(
       (balance) => balance.tokenAddress === tokenAddress,
     );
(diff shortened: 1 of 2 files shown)


AdrianDudko:
Waiting for https://github.com/JoinColony/colonyCDapp/issues/2086

jakubcolony:
@AdrianDudko #2086 has just been merged. On a side note, once a PR is up, the work dependent on it can be started as long as it's checked out from that branch, without waiting for a merge.

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/PaymentBuilderWidget/utils.ts
```diff
@@ -9,7 +12,47 @@ export const getExpenditureStep = (status?: ExpenditureStatus) => {
...
+          amounts[payout.tokenAddress] = BigNumber.from(
+            amounts[payout.tokenAddress] ?? 0,
+          ).add(payout.amount);
+        }
+      });
+
+      return amounts;
+    },
+    {},
+  );
```
rdig, line 49: This is improper use of `reduce`. There no case where you actually modify or increment the accumulator, you always return the the same object.

This is at best a use case for `map` rather.

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/PaymentBuilderWidget/utils.ts
```diff
@@ -9,7 +12,47 @@ export const getExpenditureStep = (status?: ExpenditureStatus) => {
...
+  if (!expenditure.balances) {
+    return false;
+  }
+
+  const amountsByToken = expenditure.slots.reduce<Record<string, BigNumber>>(
+    (amounts, slot) => {
+      slot.payouts?.forEach((payout) => {
+        if (!payout.isClaimed) {
+          // eslint-disable-next-line no-param-reassign
+          amounts[payout.tokenAddress] = BigNumber.from(
```
rdig, line 40: Please be mindful of our eslint rules

In this case a better approach would have been to just declare a new variable which would hold the value you need

rdig (changes requested):
Mostly good, however the code inside `isExpenditureFullyFunded` needs to be refactored some.

Screenshot from 2024-03-30 18-45-31

Also, I am guessing this PR doesn't actually bring in the content, just creates a placeholder for it _(based on what I'm seeing inside the Release step)

CzarekDryl:
@rdig I've fixed content inside release overview box. Please check it once again.

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/PaymentBuilderWidget/utils.ts
```diff
@@ -9,7 +12,48 @@ export const getExpenditureStep = (status?: ExpenditureStatus) => {
...
+export const isExpenditureFullyFunded = (expenditure?: Expenditure | null) => {
+  if (!expenditure) {
+    return false;
+  }
+
+  if (!expenditure.balances) {
+    return false;
+  }
+
+  const amountsByToken = expenditure.slots.map((slot) => {
```
bassgeta, line 35: nitpicky, but could we name this something like `slotAmountsByToken`? `amountsByToken` sounds like an object with token addresses as keys and amounts as values tbh :D

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/PaymentBuilderWidget/utils.ts
```diff
@@ -9,7 +12,48 @@ export const getExpenditureStep = (status?: ExpenditureStatus) => {
...
+          amounts[payout.tokenAddress] ?? 0,
+        ).add(payout.amount);
+      }
+    });
+
+    return amounts;
+  });
+
+  return amountsByToken.every((token) => {
+    const [tokenAddress] = Object.keys(token);
```
bassgeta, line 50: I am not sure I am a fan of this, could we construct the object differently so we don't have to take the first key and then checking the amount?

bassgeta (changes requested):
Nice start on this! I would like it if we somehow simplify the slot amount checking logic so it's more readable (if possible) :)

bassgeta (approved):
Reads much nicer now, thanks for the quick update!

rdig (approved):
Good from my side as well. Nice work!

jakubcolony (approved):
Looks good @CzarekDryl 👍 

image

PR #2102, opened 2024-03-28
feat: handle lock expenditure

joanna-pagepro:
## Description

- Added logic to lock the expenditure when in review mode

## Testing

* Step 1. Create Payment Builder Action
* Step 2. Click "Confirm details" button on the right

## Diffs

**New stuff** ✨

* handling lock expenditure functionality

**Changes** 🏗

* buttonProps from StepDetailsBlockProps extended with ActionButtonProps
* payouts array passed to PaymentBuilderTokensTotal updated to contain tokenAddress

**Deletions** ⚰️

* no deletions

Contributes https://github.com/JoinColony/colonyCDapp/issues/2005

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/forms/PaymentBuilderForm/partials/PaymentBuilderRecipientsField/hooks.tsx
+++ b/src/components/v5/common/ActionSidebar/partials/forms/PaymentBuilderForm/partials/PaymentBuilderRecipientsField/hooks.tsx
@@ -1,13 +1,10 @@
 import { createColumnHelper, type ColumnDef } from '@tanstack/react-table';
-import { BigNumber } from 'ethers';
-import moveDecimal from 'move-decimal-point';
 import React, { useMemo } from 'react';
 import { useFormContext } from 'react-hook-form';
 
 import { useColonyContext } from '~context/ColonyContext/ColonyContext.ts';
 import useWrapWithRef from '~hooks/useWrapWithRef.ts';
 import { formatText } from '~utils/intl.ts';
-import { getSelectedToken } from '~utils/tokens.ts';
 import AmountField from '~v5/common/ActionSidebar/partials/AmountField/index.ts';
 import UserSelect from '~v5/common/ActionSidebar/partials/UserSelect/index.ts';
 import FormInputBase from '~v5/common/Fields/InputBase/FormInputBase.tsx';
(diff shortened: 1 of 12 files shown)


src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/StepDetailsBlock/StepDetailsBlock.tsx
```diff
@@ -21,10 +21,19 @@ const StepDetailsBlock: FC<StepDetailsBlockProps> = ({ text, buttonProps }) => (
       {
         key: 'fund',
         content: (
-          <Button
-            {...buttonProps}
-            className={clsx(buttonProps.className, 'w-full')}
-          />
+          <div className="flex w-full justify-center">
+            {'actionType' in buttonProps ? (
```
bassgeta, line 25: could we please make this composable instead of checking props. All we're doing is rendering a button with the provided props, a perfect use case to just pass it already rendered to this component

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/PaymentBuilderWidget/PaymentBuilderWidget.tsx
```diff
@@ -28,6 +32,13 @@ const PaymentBuilderWidget: FC<PaymentBuilderWidgetProps> = ({ action }) => {
     getExpenditureStep(status),
   );
 
+  const lockExpenditurePayload: LockExpenditurePayload | null = expenditure
+    ? {
+        colonyAddress: colony.colonyAddress,
+        nativeExpenditureId: expenditure?.nativeId,
```
bassgeta, line 38: i think we can omit the `?` here since we check if expenditure is defined above

bassgeta (changes requested):
Looks pretty good, dropped a minor comment and a suggestion to improve code readability :)

iamsamgibbs (approved):
This is looking good! Tested and expenditures are properly locked and stepper progresses as expected in the UI. (Although please make bassgeta's requested changes)

Before confirming details:
Screenshot 2024-03-28 at 10 51 53

After confirming details:
Screenshot 2024-03-28 at 10 52 07

Should this pill update after confirming details?

Screenshot 2024-03-28 at 10 52 15

src/components/v5/common/ActionSidebar/partials/forms/PaymentBuilderForm/partials/PaymentBuilderRecipientsField/hooks.tsx
```diff
@@ -114,7 +115,13 @@ export const useRecipientsFieldTableColumns = (
...
                 );
 
-                const payouts = Object.values(summedPayouts);
+                const payouts: ExpenditurePayout[] = Object.entries(
+                  summedPayouts,
+                ).map(([tokenAddress, { amount }]) => ({
+                  tokenAddress,
+                  amount,
+                  isClaimed: false,
+                }));
```
jakubcolony, line 124: You are not using the results of the `reduce` anywhere else, so you could modify the `reduce` to do all the data transformation in one go, without separate `Object.entries` and `map`. 

It is out of scope of this PR but it's worth noting these aren't payouts anymore, you've transformed it to be a map of token address and the total amount of payouts for that token address, and then made it into an array of objects representing token addresses and their amounts.

Since it's then passed somewhere else, it creates confusion as someone could reasonably expect the `payouts` passed to `PaymentBuilderPayoutsTotal` to be payouts, and not this special purpose array of objects.

I think we need to avoid this intermediary data manipulation as much as possible. If you passed "real" payouts to `PaymentBuilderPayoutsTotal` it could perform the necessary transformation there, and that would help us achieve a few important points:
- separation of concerns (this hook doesn't need to know what `PaymentBuilderPayoutsTotal` does)
- removal of the confusing data structures
- reducing back and forth data manipulation (currently we convert ETH to WEI in this hook, then convert it back to ETH in `paymentBuilderPayoutsTotal`)

jakubcolony (approved):
All good functionally, nice work @joanna-pagepro. Please look at my comment regarding the data transformation, hopefully it will help us simplify the upcoming features.

bassgeta (approved):

Issue #2065, opened 2024-03-19
Payment builder - not possible to change the token type on mobile.


AdrianDudko:
Teamwork task: [Payment builder: Not possible to change the token type on mobile](https://tw.auroracreation.com/#/tasks/15843770)

It isn't possible to change the token type on mobile.

https://github.com/JoinColony/colonyCDapp/assets/34915414/3556a530-8c49-488a-a825-fa809de6bc96

_Originally posted by @iamsamgibbs in https://github.com/JoinColony/colonyCDapp/pull/2039#pullrequestreview-1945813110_

AdrianDudko:
Closed by https://github.com/JoinColony/colonyCDapp/pull/2118
![image](https://github.com/JoinColony/colonyCDapp/assets/112850454/788a0a9b-00f4-46b8-862e-c3138adf3821)

PR #2104, opened 2024-03-28
fix: changed portal elements z-index

CzarekDryl:
## Description

- It isn't possible to change the token type on mobile because of z-indexes on mobile

Resolves [https://github.com/JoinColony/colonyCDapp/issues/2065](https://github.com/JoinColony/colonyCDapp/issues/2065)


jakubcolony (commented):
@CzarekDryl the linked issue demonstrates the problem when "Payment builder" action is selected, which is not available on your branch. I could not replicate the same issue on `master` (with "Mint tokens", "Transfer funds" or "Simple payment" actions), so could you provide some testing instructions?

Issue #2072, opened 2024-03-20
Table display issues at tablet screen size


arrenv:
Teamwork task: [Payment builder: Table display issues at tablet screen size](https://tw.auroracreation.com/#/tasks/15843771)

Note: This issue is being introduced in this PR https://github.com/JoinColony/colonyCDapp/pull/2039

## Steps to reproduce

1. Open up the Payment builder action in the action sidepanel
2. Make sure you are viewing the app at the tablet screen size i.e. between 1040px and 480px.

## Expected behaviour

* Columns should not overlap each other at the tablet screen size
* Suggestion: For tables at the screen size, we can toggle the mobile view version of the table.

## Actual behaviour

* Column content is overlapping each other at tablet screen size.

![image](https://github.com/JoinColony/colonyCDapp/assets/33682027/e13cdc87-7a17-4cac-bacc-08de6bd31cf8)

Issue #2078, opened 2024-03-21
Mobile: Can't trigger actions from the mobile menu


melyndav:
Teamwork task: [Mobile: Can't trigger actions from the mobile menu](https://tw.auroracreation.com/#/tasks/15854538)
## Description

In production, when I try to open an action from the mobile menu, nothing happens when I select the 'Create action' button from the mobile menu or select any action type from any of the 'related actions' under the main menu types.

Main CTA that isn't triggering the default sidepanel: 

<img width="564" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/49cd70ba-ec25-4492-857d-7f8d021df7c2">

Related actions under all main menus that are also not triggering the related action when selected:
<img width="538" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/c41bc486-5813-4f2b-9947-422e1b3f8e11">

PR #2105, opened 2024-03-28
fix: payment builder table tablet layout

joanna-pagepro:
## Description

- Payment Builder table tablet view is the same as the mobile view

A bug with not showing the total values in the table was resolved in this PR: https://github.com/JoinColony/colonyCDapp/pull/2102

## Testing

* Step 1. Open Action Sidebar on tablet screen size i.e. between 1040px and 480px.
* Step 2. Select "Payment builder" action type
* Step 3. Add payment

## Diffs

**New stuff** ✨

* no new stuff

**Changes** 🏗

* additional `changeLayoutBreakpoint` prop added to the `Table` component so that we can pick if the mobile view should start from tablet or mobile breakpoint.

**Deletions** ⚰️

* no deletions

Resolves https://github.com/JoinColony/colonyCDapp/issues/2072

### DIFF
--- a/src/components/v5/common/Table/Table.tsx
+++ b/src/components/v5/common/Table/Table.tsx
@@ -9,7 +9,7 @@ import {
 import clsx from 'clsx';
 import React, { useMemo } from 'react';
 
-import { useMobile } from '~hooks/index.ts';
+import { useMobile, useTablet } from '~hooks/index.ts';
 import { formatText } from '~utils/intl.ts';
 import Button from '~v5/shared/Button/index.ts';
 import MeatBallMenu from '~v5/shared/MeatBallMenu/index.ts';
@@ -43,15 +43,20 @@ const Table = <T,>({
   columns,
   renderSubComponent,
   getRowCanExpand,
+  changeLayoutBreakpoint = 'mobile',
   ...rest
 }: TableProps<T>) => {
   const isMobile = useMobile();
+  const isTablet = useTablet();
+  const shouldShowMobileContent =
+    (changeLayoutBreakpoint === 'mobile' && isMobile) ||
+    (changeLayoutBreakpoint === 'tablet' && isTablet);
   const helper = useMemo(() => createColumnHelper<T>(), []);
 
   const columnsWithMenu = useMemo(
     () => [
       ...columns,
-      ...(getMenuProps && !(isMobile && verticalOnMobile)
+      ...(getMenuProps && !(shouldShowMobileContent && verticalOnMobile)
         ? [
             makeMenuColumn<T>(
               helper,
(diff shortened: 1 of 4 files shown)


src/components/v5/common/Table/Table.tsx
```diff
@@ -375,18 +386,42 @@ const Table = <T,>({
               {
                 'sm:grid-cols-[1fr_auto_auto]':
                   canGoToNextPage ||
-                  (additionalPaginationButtonsContent && isMobile),
+                  (additionalPaginationButtonsContent &&
```
bassgeta, line 389: This component already has way too many booleans, but should we wrap all of these into variables? it's getting a bit unreadable

bassgeta (changes requested):
Works as it should, so nice job with this! But the `Table` component is getting unreadable with this many properties and conditional rendering.
1. Do we ever want to display a table horizontally on a tablet and vertically on mobile? If not, then I think we should just make the necessary tables vertical. Dunno what the design is, but we could probably work with having mobile and tablet components the same.
2. Should we break up the `Table` component to have pagination as a separate component?

joanna-pagepro:
@bassgeta I moved the pagination component to a separate file and updated the props a bit so that we won't have to check the breakpoint to change the layout in the component. I think that it's more readable now.

rdig (changes requested):
When I'm on a smaller screen size (~700px in this case), I cannot open the "Decision Method" dropdown

I've recorded the interaction below

[Screencast from 2024-03-30 17-27-57.webm](https://github.com/JoinColony/colonyCDapp/assets/1193222/4534cbce-3ea0-40cb-9671-c9f59c8579d5)

joanna-pagepro:
@rdig I think that we have a separate issue added to fix that. I found one with the action sidebar: https://github.com/JoinColony/colonyCDapp/issues/2078. The problem is global

bassgeta (approved):
Nice job, the table feels lighter now :+1:

iamsamgibbs (changes requested):
Great work on this so far, only minor pick up from me is the button to open this menu I think should be aligned right, and the menu shouldn't span the entire width.

Screenshot 2024-04-04 at 10 57 04

joanna-pagepro:
@iamsamgibbs should be ok now

iamsamgibbs (approved):
Thanks for the change, all looks good now!

On tablet:
Screenshot 2024-04-05 at 13 45 52

And on mobile:
Screenshot 2024-04-05 at 13 46 01

joanna-pagepro:
@rdig can you take a look?

bassgeta (approved):
Still looks good, reapproving :+1:

PR #2110, opened 2024-03-29
fix: added expenditure refetch after funding/releasing

CzarekDryl:
## Description

- Refetching expenditure data after funding and releasing actions. This is needed to refresh the data in the stepper and move to the next step
- Going from funding step to release is handled in this [PR](https://github.com/JoinColony/colonyCDapp/pull/2101)


rdig (approved):
Can confirm that this is polling as advertised, and since under the hood it uses `useGetExpenditureData` which in turn makes us of `getSafePollingInterval` whe're as close as possible to properly controlling the polling.

Screenshot from 2024-03-30 18-55-53
Screenshot from 2024-03-30 18-56-03

bassgeta (approved):
works and looks good :sunglasses:

jakubcolony (approved):
Such manual refetching will only ever work in local dev (and if you're lucky enough) due to the delay in block-ingestor updating the DB, which is why we need polling. But I suppose there's no harm in having manual refetches too, for the lucky cases

PR #2115, opened 2024-03-31
Upgrade 3rd party storybook addons

chmanie:
## Description

While the addons worked for us in their previous versions, we were not really using all of their features. This upgrade should introduce full compatibility with `storybook@8`.

I monitored the progress of the plugins and here we are.

## Testing

`npm run storybook`

## Diffs

**Changes** 🏗

* Upgraded `storybook-addon-apollo-client` and `storybook-react-intl` to their newest versions


iamsamgibbs (approved):
Looks good to me! Tested and storybook is still working as intended.

Screenshot 2024-04-02 at 09 47 12

jakubcolony (approved):
Seems to be working fine, no warnings in the console, LGTM 👍 

image

Issue #2064, opened 2024-03-19
Actions: Member dropdown - Fix the issue when displaying address instead of username


AdrianDudko:
I do want to note, as this is something that doesn't really show up that often, that, in case you have users without a username registered, or one that cannot be loaded, the dropdown user experience is less pleasant:

[Screencast from 2024-03-18 21-38-45.webm](https://github.com/JoinColony/colonyCDapp/assets/1193222/a3b74181-c082-4dc7-a973-d523ea72f910)

_Originally posted by @rdig in https://github.com/JoinColony/colonyCDapp/pull/2058#pullrequestreview-1944027305_

AdrianDudko:
@melyndav 

Do we have a design for this?

AdrianDudko:
Hey @melyndav could you please check the comment above? :)

melyndav:
Hey @AdrianDudko they just get treated the same as how they display with truncation - 
<img width="587" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/0b70951b-87f4-49c3-9fbf-af4882b25cce">


https://www.figma.com/file/l1dOM5qiQYwF0ElvKDqqjg/Design-System---Colony-v3?type=design&node-id=1782-11884&mode=design&t=Wovs7EuDn6RRCEcy-4

PR #2125, opened 2024-04-03
fix: added address truncating to user selects

CzarekDryl:
## Description

- Added address truncating to user selects to prevent overflowing

Resolves [https://github.com/JoinColony/colonyCDapp/issues/2064](https://github.com/JoinColony/colonyCDapp/issues/2064)

### DIFF
--- a/src/hooks/members/utils.ts
+++ b/src/hooks/members/utils.ts
@@ -39,6 +39,10 @@ export const hasSomeRole = (
   return includesCustomPermissions || isInFilters;
 };
 
+export const truncateAddress = (name: string) => {
+  return `${name.slice(0, 6)}...${name.slice(-4)}`;
+};
+
 export const sortByReputationAscending = (items: ColonyContributor[]) => {
   return items.sort(
     (a, b) => a.colonyReputationPercentage - b.colonyReputationPercentage,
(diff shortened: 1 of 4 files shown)


src/hooks/members/utils.ts
```diff
@@ -39,6 +39,10 @@ export const hasSomeRole = (
   return includesCustomPermissions || isInFilters;
 };
 
+export const truncateAddress = (name: string) => {
+  return `${name.slice(0, 6)}...${name.slice(-4)}`;
+};
```
rdig, line 44: We already have a util for this, `splitAddress` which is more robust.

Please use that one.

rdig (changes requested):
All good, except please don't add new utils, if they already exist for the same purpose.

melyndav (changes requested):
Hey @CzarekDryl nice work on this one, a nice improvement to the dropdown.

One thing I picked up while testing and please let me know if it's not related to this branch, but 'simple payment' changes to 'payment' on competition.

Active:
image

Completed:
image

CzarekDryl:
@rdig I've used `splitAddress` function and simplified it by removing try/catch.
@melyndav Action type should be good now

melyndav (approved):
Thanks for that fix @CzarekDryl looking back to normal now. Nice work!

rdig (approved):
All good now. Nicely done.

Screenshot from 2024-04-05 16-18-24

PR #2130, opened 2024-04-03
Maintenance: Put z-index variables into tailwind config

### LINKED ISSUE #2106: [Maintenance] Put z-index variables into our tailwind theme
## Specification

Currently we do not use any variables for z-index variables, we just use arbitrary values via `z-[some number]`.
We should store them in variables, thus improving readability and stability.

iamsamgibbs:
## Description

Previously we were not using any variables for z-index variables, instead just using arbitrary values via z-[some number].

This PR adds new z-index variants to the tailwind config:

```
      zIndex: {
        base: '1',
        mid: '2',
        sidebar: '10',
        top: '1000',
      },
```

I've also tried to remove unnecessary uses of z-indexes, or change the value to match an existing value to reduce the number of different values. I've tried to be pretty thorough in checking this has not introduced any new issues, but there is a chance some z-index values may need adjusting, testing should focus on looking for any instances of this.

## Testing

Use the app as usual and check for any instances where the z-index value is set incorrectly.

## Diffs

**New stuff** ✨

* New z-index variants introduced

**Changes** 🏗

* Arbitrary uses of z-[] now use one of the new z-index variants

Resolves #2106


rdig (approved):
All good. Nice work going through all these files.

Didn't check quite every screen, but from the one's I've tested, none show any visual regressions

jakubcolony (approved):
Very nice! If we can combine it with `isolation: isolate` (https://developer.mozilla.org/en-US/docs/Web/CSS/isolation) wherever possible, then maybe z-indexes won't be such a pain after all!

melyndav (approved):
Hey @iamsamgibbs I had a good dig around the UI and couldn't find anything that looked out of the ordinary or wasn't already an issue, so nice work on this one!

bassgeta:
just a question, does this by any chance also resolve the issue this PR was fixing https://github.com/JoinColony/colonyCDapp/pull/2104 ? I think it does but we can double check and close that one in this case

iamsamgibbs:
> just a question, does this by any chance also resolve the issue this PR was fixing #2104 ? I think it does but we can double check and close that one in this case

#2104 seems to be based on the wrong branch and needs to be reworked, that original issue was only happening on `feat/advanced-payments-2` and was specific to that action type, I don't think those other tokens were an issue on master.

PR #2143, opened 2024-04-04
Fix: Pending reputation next update shows incorrect time

### LINKED ISSUE #2121: Userhub: Pending reputation "Next update" field shows incorrect time
## Steps to reproduce

1. Select a colony.
2. Open the User Hub.
3. Navigate to the "Your balance" tab.
4. The "Next reputation update" field will show how much time there is left untill the next reputation update.
5. The countdown is reached and the user does not refresh the page or open a new page.

## Expected behaviour

Once the "Next reputation update" countdown is complete, the field should show the next reputation update, i.e. it should be in the future "in xx hours xx minutes".

## Actual behaviour

The "Next reputation update" field will show the last time reputation was updated as "xx hours xx minutes ago" i.e. in the past.

image

iamsamgibbs:
## Description

When the "Next update" time is passed, it does not update - so it will show an incorrect time: "15 seconds ago".

This PR adds a check to see if the "next update" time has passed, and gets the next cycle date to update the time.

## Testing

1. Select a colony.
2. Open the User Hub.
3. Navigate to the "Your balance" tab.
4. The "Next reputation update" field will show how much time there is left untill the next reputation update.
5. The countdown is reached and the user does not refresh the page or open a new page.

Once the timer reaches 0, it should briefly show "Now" before refreshing to the new time "60 minutes".

## Diffs

**Changes** 🏗

* `PendingReputation` component now checks if the `nextMiningCycleDate` has passed and updates it if necessary

Resolves #2121

https://github.com/JoinColony/colonyCDapp/assets/34915414/da392ebc-70fd-4e6f-935c-20524ffcac64

### DIFF
--- a/src/components/common/Extensions/UserHub/partials/ReputationTab/partials/PendingReputation/PendingReputation.tsx
+++ b/src/components/common/Extensions/UserHub/partials/ReputationTab/partials/PendingReputation/PendingReputation.tsx
@@ -31,7 +31,7 @@ const PendingReputation: FC<PendingReputationProps> = ({
   const { data } = useGetReputationMiningCycleMetadataQuery();
   const { lastCompletedAt } = data?.getReputationMiningCycleMetadata ?? {};
 
-  const [nextMiningCycleDate, setNextMiningCyleDate] = useState(
+  const [nextMiningCycleDate, setNextMiningCycleDate] = useState(
     lastCompletedAt
       ? getNextMiningCycleDate(new Date(lastCompletedAt ?? ''))
       : null,
@@ -48,7 +48,7 @@ const PendingReputation: FC<PendingReputationProps> = ({
       );
 
       if (nextMiningCycleDate < halfIntervalAgo) {
-        setNextMiningCyleDate(getNextMiningCycleDate(halfIntervalAgo));
+        setNextMiningCycleDate(getNextMiningCycleDate(halfIntervalAgo));
       }
     }, UPDATE_INTERVAL * 1000);
 
--- a/src/components/common/Extensions/UserHub/partials/ReputationTab/partials/PendingReputation/PendingReputation.tsx
+++ b/src/components/common/Extensions/UserHub/partials/ReputationTab/partials/PendingReputation/PendingReputation.tsx
@@ -1,4 +1,4 @@
-import React, { type FC } from 'react';
+import React, { type FC, useEffect, useState } from 'react';
 import { useIntl } from 'react-intl';
 
 import { useGetReputationMiningCycleMetadataQuery } from '~gql';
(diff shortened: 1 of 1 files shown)


src/components/common/Extensions/UserHub/partials/ReputationTab/partials/PendingReputation/PendingReputation.tsx
```diff
@@ -17,6 +17,8 @@ import {
 const displayName =
   'common.Extensions.UserHub.partials.ReputationTab.partials.PendingReputation';
 
+const UPDATE_INTERVAL = 15;
```
rdig, line 20: I'd put this same as the block time (from constants default network) but I guess any value will work in the end, as the update estimate is just that, and estimate.

rdig (approved):
Nicely done! You are even cleared the interval when un-mounting as to not have memory leaks.

Wouldn't expect anything less from you :P

src/components/common/Extensions/UserHub/partials/ReputationTab/partials/PendingReputation/PendingReputation.tsx
```diff
@@ -28,9 +30,30 @@ const PendingReputation: FC<PendingReputationProps> = ({
...
+      if (!nextMiningCycleDate) {
+        return;
+      }
+
+      const halfIntervalAgo = new Date(
+        Date.now() - (UPDATE_INTERVAL / 2) * 1000,
+      );
+
+      if (nextMiningCycleDate < halfIntervalAgo) {
+        setNextMiningCyleDate(getNextMiningCycleDate(halfIntervalAgo));
```
jakubcolony, line 51: Should we use `halfIntervalAgo` for the next mining cycle as opposed to `nextMiningCycleDate`?

Or maybe we could refetch the query, if that would return the updated cycle date?

jakubcolony (approved):
Nice work on this PR @iamsamgibbs, everything works as expected.

src/components/common/Extensions/UserHub/partials/ReputationTab/partials/PendingReputation/PendingReputation.tsx
iamsamgibbs: Ah yeah, on reflection, makes more sense to use nextMiningCycleDate to set the next one.

We could refetch the query, but my worry is as this is just an estimate, when we refetch it is possible that the next mining cycle won't have actually ran yet and we'd have to keep calling the query until it has updated. I can look at doing that though if that is preferred?

jakubcolony: That's a good point, let's leave it like it is. Nice one!

jakubcolony (approved):

rdig (approved):
Re--approving so you can merge

PR #2152, opened 2024-04-05
Fix: New colony URL should not allow capital letters

### LINKED ISSUE #2145: New Colony URL should not allow capital letters
## Steps to reproduce

1. Create a new colony.
2. Enter a colony name
3. Proceed to the confirmation step.

The input will appear to only show lowercase characters in the input, however the actual value will accept uppercase and lowercase letters.

## Expected behaviour

The input should only accept lowercase letters.

## Actual behaviour

The input will appear to only show lowercase characters in the input, however the actual value will accept uppercase and lowercase letters.

(Related to #2144)

iamsamgibbs:
## Description

The colony URL input in the new colony creation wizard appears to only let you enter lowercase letters, however the actual value will accept uppercase and lowercase letters.

This PR adds new validation to the input to ensure only lowercase letters are allowed.

## Testing

1. Create a new colony.
2. Try entering capital letters to the colony URL field
3. Proceed with the creation wizard and create the colony.
4. Check in the database that the colony has been created with a lowercase name:

```
query MyQuery {
  listColonies {
    items {
      name
    }
  }
}
```

## Diffs

**Changes** 🏗

* Added lowercase validation to the new colony URL input

**Deletions** ⚰️

* Removed the now unnecessary .toLowerCase() on the confirmation screen

Resolves #2145

### DIFF
--- a/amplify/backend/function/createColonyEtherealMetadata/src/index.js
+++ b/amplify/backend/function/createColonyEtherealMetadata/src/index.js
@@ -114,7 +114,7 @@ exports.handler = async (event) => {
         id: `etherealcolonymetadata-${transactionHash}`,
         displayName: colonyDisplayName,
         etherealData: {
-          colonyName,
+          colonyName: colonyName.toLowerCase(),
           colonyDisplayName,
           colonyAvatar,
           colonyThumbnail,
--- a/src/components/common/Onboarding/wizardSteps/CreateColony/CreateColonyCardRow.tsx
+++ b/src/components/common/Onboarding/wizardSteps/CreateColony/CreateColonyCardRow.tsx
@@ -42,7 +42,7 @@ const CardRow = ({ updatedWizardValues, setStep }: CardRowProps) => {
     {
       title: 'navigation.admin.colonyDetails',
       text: colonyDisplayName,
-      subText: `${APP_URL.origin}/${colonyName.toLowerCase()}`,
+      subText: `${APP_URL.origin}/${colonyName}`,
       step: 0,
     },
     {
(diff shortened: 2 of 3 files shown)


rdig (changes requested):
Changes are good on the client side, however the request might get changed in-transit _(think man in the middle)_.

For that you'll need to also update the lambdas, and make sure it's always written in lower case.

There's two places you'll need to do this:
- create colony metadata: https://github.com/JoinColony/colonyCDapp/blob/ce250e4c7e255d51f7c697a640c669e05b6eda9a/amplify/backend/function/createColonyEtherealMetadata/src/index.js#L117
- create colony: https://github.com/JoinColony/colonyCDapp/blob/ce250e4c7e255d51f7c697a640c669e05b6eda9a/amplify/backend/function/createUniqueColony/src/index.js#L238

iamsamgibbs:
Thanks for the explanation! I did think as I was doing this that it should also check on the "server side" but wasn't sure how this is implemented. Is there a reason we don't do any other validation on the lambda? (such as the max character length which we do check on client side).

I also don't think it is necessary to add .toLowerCase() on the createUniqueColony lambda as `colonyName` here comes from the ethereal metadata stored in the previous lambda.

rdig:
> I also don't think it is necessary to add .toLowerCase() on the createUniqueColony lambda as colonyName here comes from the ethereal metadata stored in the previous lambda.

Correct, I just wanted to make "double" sure that the correct value was being saved :)

>  Is there a reason we don't do any other validation on the lambda? (such as the max character length which we do check on client side).

I don't have a good reason for this, other that limit only existing in the UI, as it's not a technical db limit.

But for the purposes of being "extra careful", I agree, we should validate in the lambda as well _(the ethereal metadata one)_

rdig (approved):
Nicely done!

ArmandoGraterol (approved):
Working as expected for me 👌

PR #2162, opened 2024-04-05
Fix: incorrect latest colony version in the DB

### LINKED ISSUE #2088: The Colony contract version is getting set incorrectly in the database
## Steps to reproduce

Create a new action to upgrade the colony version.

## Expected behaviour

The 'New version' should match the latest available version (14 at the time of writing)

## Actual behaviour

The 'New version' is showing the incorrect version (4)

## Suggestions

It is most likely that the `setCurrentVersion` lambda is not correctly setting the version in the first place.

Running the `getCurrentVersionByKey` query with the key "colony" is returning version 4.

`const currentNetworkVersion = await colonyNetwork.getCurrentColonyVersion();` in the create-data.js script is showing the right version.

jakubcolony:
## Description

My best guess as to what happened is that after seeding the DB with the correct latest version (`15`), ingestor picked up two `ColonyVersionAdded` "adding" versions 3 and 4. Because there was no mechanism ensuring we only store the highest version we've seen so far, the version was being incorrectly set to 4.

The fix is mostly on block-ingestor side: https://github.com/JoinColony/block-ingestor/pull/197. I deleted `setCurrentVersion` lambda since its functionality can easily live in ingestor instead.

## Testing

1. Start the dev environment. Wait until all containers are running.
2. Run the following query:
```gql
query ListVersions {
  listCurrentVersions {
    items {
      key
      version
    }
  }
}
```
Previously, entry with key `colony` had a version of `4` which should've now been changed to `15`.

As I came across this when trying to create a new colony, it might be worth running that flow as well.

## Diffs

**Deletions** ⚰️

* `setCurrentVersion` lambda

Resolves #2088


iamsamgibbs (approved):
Looks good to me! Thanks for picking up this issue!

Version is now getting correctly set in the database
Screenshot 2024-04-08 at 08 36 52

And displaying properly on the frontend
Screenshot 2024-04-08 at 08 37 00

ArmandoGraterol (approved):
Working as expected for me:
Screenshot 2024-04-08 at 15 51 11

PR #2099, opened 2024-03-27
Return info about expenditure Release action

jakubcolony:
## Description

This PR adds a new action type describing expenditure finalization and exposes all actions finalizing an expenditure under `finalizingActions` field (most of the time it will be just a single action). This enables the UI to show details of the action initiator.

block-ingestor PR: https://github.com/JoinColony/block-ingestor/pull/194

## Testing

- Create and finalize an expenditure
- Run the following query to see the linked action:
```gql
query ListExpenditures {
  listExpenditures(filter: {nativeId: {ge: 33}}) {
    items {
      nativeId
      status
      finalizingActions: actions(
        filter: {type: {eq: FINALIZE_EXPENDITURE}, isMotionFinalization: {ne: true}}
      ) {
        items {
          id
          type
          initiatorAddress
          initiatorUser {
            profile {
              displayName
            }
          }
          expenditure {
            nativeId
          }
        }
      }
    }
  }
}
```

Resolves #2086

PR #2167, opened 2024-04-08
Auto install Staked Expenditure extension at colony creation

### LINKED ISSUE #2074: Install and enable Staked Expenditure extension at colony creation
### Story

As a user creating colony, I want the Staked Expenditure extension to be installed and enabled automatically as part of the creation flow, so that I can make staked payments without having to install anything manually.

### Description

This should happen in the same way as OneTxPayment installation.

To save gas, we should bundle OneTxPayment and StakedExpenditure `installExtension` transactions using multicall, then proceed with initialisation when we know their addresses.

Staked Expenditure accepts a single parameter, stake fraction, which should be set to: `1%`

### Subtasks

- [ ] Consider the flow for existing colonies which did not have Staked Expenditure installed at creation
- [ ] #2200

jakubcolony:
## Description

This PR modifies the `colonyCreate` saga to include deployment of Staked Expenditure extension at colony creation. The saga will now install, set permissions and enable the extension with a default stake fraction of 1%. It utilises multicall to reduce the number of transactions in order to save on gas.

## Testing

Create a colony (you can use `node scripts/create-colony-url.js`). Navigate to extensions page to see Staked Expenditure installed and enabled. 

Mint some tokens and activate them in the user hub, then use testing UI to create a staked expenditure.

Finally, create a simple payment to check OneTx payment has been installed and its permissions set.

Resolves #2074


iamsamgibbs (changes requested):
Looks like something has gotten messed up along the way and needs a rebase!

I tried taking a look anyway, the 'Staked Expenditure' extension does install correctly when a new colony is created.
 
Screenshot 2024-04-09 at 09 49 17

I received a 403 error when making a simple payment, but the action seemed to go through okay anyway? Then was able to activate the tokens and create a staked expenditure. I screen recorded the entire process. (There was also a console warning from the completed action screen when minting tokens).

https://github.com/JoinColony/colonyCDapp/assets/34915414/90e3218f-5972-4fd9-ab55-bbf9c2a1827b

Colony's created via the create-data script don't have the extension automatically enabled, so this will need updating too.
Screenshot 2024-04-09 at 09 48 16

ArmandoGraterol (changes requested):
Looking good! I noticed the same exact issues Sam pointed out, namely the `create-data` script is not installing the extension and the simple payment action is throwing some errors (the payment is being made it seems as I still get the funds):

Screenshot 2024-04-09 at 14 37 02

jakubcolony:
Great spot @iamsamgibbs, the issue with simple payment was actually introduced in #2099, I fixed it in block-ingestor and updated the commit hash. 

`create-data` script has also been updated to install and enable staked expenditure.

iamsamgibbs (approved):
Thanks for the changes, all working as expected now!

Extension is installed when creating a new colony, and when running the create-data script.

Screenshot 2024-04-10 at 10 38 51

Screenshot 2024-04-10 at 10 40 17

Able to mint tokens, pay myself and then make a staked expenditure.

Screenshot 2024-04-10 at 10 41 24

ArmandoGraterol (approved):
Thank you for the changes! Working as expected for me now 👌

PR #2181, opened 2024-04-09
Add masked address to user popover

### LINKED ISSUE #2177: Actions: Truncation of recipient in completed action states
## Descriptions

A leftover issue from #2066 is the requirement to fix the recipient address and ensure they are truncated across all responsive states when an action is completed.

Issue from #2066 with the display of addresses shown below:

image

How they correctly show on production:
image

FYI @bassgeta

ArmandoGraterol:
Added masked address to the user popover, so now every place where the popover is used and there's no user profile, it should show the partial address instead like the other places where we show addresses.

Every recipient address in the completed action side panel should now be masked:
Screenshot 2024-04-09 at 15 57 34

Resolves #2177


melyndav (approved):
Works as expected! Thanks for the fix @ArmandoGraterol

jakubcolony (approved):
Tested with `Manage permissions`, works like a charm 👍 

image

PR #2225, opened 2024-04-11
Fix: Colony creation failure handling

### LINKED ISSUE #2128: Joined Colonies not showing in production (again)
## Steps to reproduce

This is hard to replicate at the moment, but this is what I did:
1. Create a Colony with an existing token.
2. First step of the create a colony transactions failed, due to too many requests error in console. (don't have screenshot)
3. Reloaded the app and created another colony.
4. Went through successfully.
5. Now I no longer can see joined colonies.

This is related to this issue - https://github.com/JoinColony/colonyCDapp/issues/1904

## Expected behaviour

* You should be able to still see colonies you have joined when a creation fails.
* We need to be able to handle colonies that are created but are not being saved in the database correctly.

## Actual behaviour

* Joined colonies no longer appear in list.
* 2 nulls are returned in the `getContributorsByAddress` request.

image

image

iamsamgibbs:
Most of the actual changes are in block-ingestor: https://github.com/JoinColony/block-ingestor/pull/201

## Description

This PR addresses an issue (#2128) - if something goes wrong when joining a colony, it will cause the joined colonies list to be empty.

The issue occurs because a `colonyContributor` is created in the database very early on in the create colony flow. If the `createUniqueColony` lambda fails then the `colony` doesn’t actually get created in the database. This means the ‘colonyContributor’ will be in the database but not actually linked to a colony, so the query to fetch the colonies for the colony switcher will fail.

This PR addresses this by moving the logic from the `createUniqueColony` lambda to block-ingestor (as it makes it easier to know when it has finished, and there was a TODO comment to do this at some point anyway). The `colonyContributor` is now created only after `createUniqueColony` is successful.

Moving the `createUniqueColony` lambda to block-ingestor has the added benefit that we no longer add listeners to colonies which are not correctly created.

## Testing

Check there are no unexpected issues when creating a new colony (run `node scripts/create-colony-url.js` to get the invite link).

In order to test the specific issue in #2128, we require an error to occur when the colony is being created. The easiest way to force this behaviour is to run block-ingestor locally and add a new `throw new Error()` anywhere in `createUniqueColony` file.

## Diffs

**Changes** 🏗

* Moved logic from `createUniqueColony` lambda to block-ingestor
* Moved `createColonyFounderInitialRoleEntry` to after the colony creation.

## TODO

- [x] Remove the now unused `createUniqueColony` lambda (waiting until after block-ingestor changes are approved to do this)

Resolves #2128

(Also includes colony version fixes courtesy of Jakub cherry picked from master-arbitrum)


rdig (changes requested):
All good, but since you moved the lambda logic into the block ingestor, the lambda itself has not purpose, so please remove it.

Otherwise, stellar work! :100:

ArmandoGraterol (changes requested):
I'm having problems creating colonies as I'm getting this error:

Screenshot 2024-04-15 at 13 57 44

Any thoughts?

iamsamgibbs:
> I'm having problems creating colonies as I'm getting this error:
> 
> Any thoughts?

I don't get this error when running block-ingestor separately, but I am getting this error when it is running in docker - I'm also getting the same error on master, so I don't think it is being introduced here - looking into it.

ArmandoGraterol (commented):
Yeah, it is still happening to me :( Also, it seems like a `.DS_Store` file was committed to the branch, is that intentional?

iamsamgibbs:
> Yeah, it is still happening to me :( Also, it seems like a `.DS_Store` file was committed to the branch, is that intentional?

Issue should be resolved now thanks to help from Jakub!

iamsamgibbs:
> Why is this also removing the `setCurrentVersion` lambda ?
> 
> Like I know Jakub removed it, but why does it show up in here? Does the branch need to be rebased or something ?

Jakub identified the colony version issue as the cause for the issue Armando and I have been experiencing with colony creation failing, so he helped cherry pick his commits into this branch (and relevant block-ingestor branch too) to apply the fix here and get things working.

.DS_Store
ArmandoGraterol, line 1: Is this file needed?

ArmandoGraterol (dismissed):
Thank you for the changes, Sam! LGTM 👌

.DS_Store
jakubcolony: interesting that the `.gitignore` didn't prevent it from being commited

rdig (dismissed):
Right. Well in that case, I'll approve it :)

.DS_Store
iamsamgibbs: Strange! It is not needed! Looks like it was accidentally added about 8 months ago and it let my accidental changes get committed. Will remove it here.

iamsamgibbs:
Just removed accidental .DS_Store commit

ArmandoGraterol (approved):

rdig (approved):

Issue #1570, opened 2023-12-18
Actions: As a user I can manage VERIFIED MEMBERS to keep the whitelist clean


AdrianDudko:
**Teamwork task:** https://tw.auroracreation.com/#/tasks/15700919
**Created by:** @AdrianDudko 
**Accepted by:** @arrenv 

**Description**
It's an existing funcionality in old UI. Go to Actions -> Manage team -> Manage Address book.

As a user I want to add and remove members to the list of verified members, representing known and trusted members.

**Acceptance criteria**

- Creating Action:
    - User can add a Custom Title to the action.
        - This is a required field.
    - Metadata updates when user makes changes in the form as of the provided pattern.
        - [https://www.notion.so/colony/Actions-Motions-775cb0e78eb44783a6f28702b7a9dc33?pvs=4#2694b1f6d93b453b90fe79af41d3fa54](https://www.notion.so/Actions-Motions-775cb0e78eb44783a6f28702b7a9dc33?pvs=21)
    - User can choose an "Add/remove" option.
        - Options:
            - Add members
                - Adds new members to the verified members list.
            - Remove members
                - Removes members from the verified members list.
        - This is a required field.
    - User can set "Decision Method".
        - This is a required field.
        - Options available are based on available decision methods.
    - User can set "Created In" field
        - This action can only be performed in ‘Root’
            - Hover on pill: display tooltip
                - Tooltip: \"The decision to Manage verified members can only be done in the Root domain.\"
        - This field is visible only if the Reputation decision method is selected.
    - User can add a rich text "Description".
        - Not required.
    - Selected members table
        - List of members that are being added or removed.
        - Appears only if the \"Add/remove\" option has been selected.
        - By default one row is already added
        - Member drop down selection:
            - If removing, it should only verified members.
            - If adding, it should only show non-verified members.
            - Should also allow search and an address to be added.
        - Clicking on a selected member in the creation state triggers the members selection dropdown.
        - This is a required field, at least one member should be added. Validation should be: \"Please add a member\".
        - Add member button adds more members that can be selected.
        - Should validate if no member has been selected in a row.
        - The same member cannot be added twice.
            - It should hide previously selected members from the dropdown.
            - It needs to also validate, in cases where an address has been manually added. Validation message: \"Duplicate members\"
        - Each added member row should have a kebab menu icon.
            - Options:
                - \"Remove row\"
                    - Removes the row and from the action.
- Submit button:
    - Button text \"Update verified\"
[...]

PR #2231, opened 2024-04-12
fix: changed decision method tooltip in verified action

CzarekDryl:
Small bug was found by our QA, in [issue](https://github.com/JoinColony/colonyCDapp/issues/1570) it says to change the tooltip content to `"The decision to Manage verified members can only be done in the Root domain."`

## Description

- Changed decision method field tooltip content in verified members action


ArmandoGraterol (approved):
I think it would look better with the "manage" all in lowercase but other than that, I'm good with this 👍

bassgeta (approved):
Nice and clean, ship it!

PR #2232, opened 2024-04-12
Fix rich text description markup and toolbar

### LINKED ISSUE #2203: Rich text description markup and toolbar not working correctly
## Steps to reproduce

1. Create a new action.
2. Add a description using both markup and the toolbar.
3. Try to create a heading using `#`, it won't work correctly.
4. Try to create a list using `-`, it won't format.

## Expected behaviour

* You should be able to create rich text using both markup and the rich text toolbar as it previously worked and as of the spec.

## Actual behaviour

* Rich text and the toolbar are not functioning correctly.
* Lists and quotes are not being created using markup or the toolbar.
* Headings, bold, italic are not working correctly using markup, but do work using the toolbar.

markup-not-working

davecreaser:
## Description

- Fix the rich text description editor.
- Ensure that markup works as expected
- Ensure that the toolbar list buttons work as expected

https://github.com/JoinColony/colonyCDapp/assets/38098203/ed4b9fab-c150-495e-9f5f-8d67cbb5df8a

## Testing

Basically, test out the rich text description editor when making a new action. It should work as you would expect.

* Create a new action.
* Add a description using both markup and the toolbar.
* Try to create a heading using #, it should work correctly.
* Try to create a list using -, it should format.
* Try to create an ordered list using 1., it should format.
* Try to create lists using the toolbar buttons, they should work.

## Diffs

**Changes** 🏗

* Added the tailwind 'prose' class to the editor, to ensure that lists display as lists
* Changed the onUpdate and onBlur handlers to not trim if you are creating a header or list as the first action in an empty field.

Resolves #2203

### DIFF
--- a/src/components/v5/shared/RichText/hooks.ts
+++ b/src/components/v5/shared/RichText/hooks.ts
@@ -79,15 +79,6 @@ export const useRichText = (
         },
       },
       content: field.value,
-      onUpdate: (props) => {
-        const trimmedText = props.editor.getText().trim();
-        const html =
-          props.editor.isEmpty || trimmedText.length === 0
-            ? ''
-            : props.editor.getHTML();
-
-        field.onChange(html);
-      },
     },
     [],
   );
@@ -106,11 +97,14 @@ export const useRichText = (
   useEffect(() => {
     const handleUpdate = ({ editor: textEditor }: { editor }) => {
       const trimmedText = textEditor.getText().trim();
+      const isCreatingHeading = textEditor.getAttributes('heading');
+      const isCreatingList = textEditor.getAttributes('list');
       const html =
-        textEditor.isEmpty || trimmedText.length === 0
+        (textEditor.isEmpty || trimmedText.length === 0) &&
+        !isCreatingHeading &&
+        !isCreatingList
           ? ''
           : textEditor.getHTML();
-
       field.onChange(html);
     };
 
--- a/src/components/v5/shared/RichText/hooks.ts
+++ b/src/components/v5/shared/RichText/hooks.ts
@@ -79,6 +79,18 @@ export const useRichText = (
         },
       },
       content: field.value,
+      onUpdate: (props) => {
+        const trimmedText = props.editor.getText().trim();
+        const isCreatingHeading = !!props.editor.getAttributes('heading');
+        const isCreatingList = !!props.editor.getAttributes('list');
+        const html =
+          (props.editor.isEmpty || trimmedText.length === 0) &&
+          !isCreatingHeading &&
+          !isCreatingList
+            ? ''
+            : props.editor.getHTML();
+        field.onChange(html);
+      },
     },
     [],
   );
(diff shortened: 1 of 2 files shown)


melyndav (changes requested):
Hey @davecreaser! Great start on this PR. A couple of pick-ups. Testing this, it looks like a lot more is broken than originally raised in the issue. Especially, styling has been removed that use to be present. Do you want this all in a separate issue? 

1. I noticed that there's additional padding that's now showing when the description box is expanded.

image
 
 
image


2. The numbered list font doesn't match Figma styling, particularly the font and spacing.

Currently:
image

Figma:
image
 
 
 3. Also same issue with the unordered list.
 
 Currently: 
image

 Figma: 
image

4. The quote block now adds "" around the text along with the UI change.

image

5. Styling gets stripped out upon completion state including spacing.
image

6. The block quote is showing as italic, although the italic is not selected in the WYSIWYG. I can not remove the italics. 
image

Italics are removed in the completed state. 
image

davecreaser:
Alright, styling issues should now be fixed. I removed the use of the prose class and instead added tailwind classes directly to the html output by tiptap. Seems like a better way of doing things!

While editing:

Screenshot 2024-04-15 at 14 01 05

-------

Completed state:

Screenshot 2024-04-15 at 14 08 22

iamsamgibbs (changes requested):
This is looking good! Just one tiny change: The "starter kit" extension includes a bunch of different extensions which means some of them are now getting loaded in twice where you've added the extra configuration options. You can just pass a configuration option to the starter kit extension to disable them to fix this issue.

Screenshot 2024-04-15 at 15 45 35

Otherwise, all looking good! No padding issue:

Screenshot 2024-04-15 at 15 52 55

And all looking as expected in both the editor and the completed action screen:

Screenshot 2024-04-15 at 15 51 44

Screenshot 2024-04-15 at 15 52 34

davecreaser:
@iamsamgibbs Nice catch, thanks. I've moved all the individual extension configurations into the starterkit configuration instead, I think it's intended more that way by on second look. Gets rid of the warning too. The only one that needs to stay separate from starterkit is the headers extension, since it has a call to the `extend` method which isn't there on the starterkit.

iamsamgibbs (approved):
Even better fix, tested and all working 👍

ArmandoGraterol (approved):
Looks good to me 👌 The issues raised on the initial issue and the others's comments are fixed. Nice job!

melyndav (changes requested):
Nice fixes on this one @davecreaser! It's looking more aligned with Figma, so appreciate the effort. 

image

Once feature we previously had that's been lost (not sure if it's part of these or previous refactors) is that all headings get stripped out, and the first paragraph block is shown when the description is collapsed. As you see now, the headings show without spacing, making it difficult to read the description snippet in its current state. 

image

Is this something you can resolve or would you like it in a separate issue?

iamsamgibbs:
Just a note that this isn't happening on master, so isn't something that has been affected by this PR:

Screenshot 2024-04-16 at 10 15 51

melyndav:
> Just a note that this isn't happening on master, so isn't something that has been affected by this PR:
> Screenshot 2024-04-16 at 10 15 51

Thanks @iamsamgibbs let's dismiss it for this PR and I'll raise an issue to have it fixed. Thanks!

melyndav (approved):
Approving @davecreaser. Will raise a separate issue to resolve the heading removal.

PR #2233, opened 2024-04-12
Fix: Ensure action fields are disabled when no permissions

### LINKED ISSUE #2221: Action sidebar fields should be disabled when you don't have permissions
## Steps to reproduce

1. On a fresh environment, connect your wallet to Dev wallet 2.
2. In GraphiQL, run this query to get the member invite code:
```
query MyQuery {
  getColony(id: "0x6dfa816d05d8a2FCc3f90F7bDB76650fAE16ee8b") {
    colonyMemberInviteCode
  }
}
```
3. Navigate to http://localhost:9091/invite/planex/%7BINVITE_CODE%7D
4. Join the colony and create a new action.
5. Select either the "Create agreement" or "Manage verified members" action types.

## Expected behaviour

* Every field on the form (except action type and title) should be disabled.
* On "Manage verified members" the "You don't have the correct permissions" error message should show.

## Actual behaviour

* The description field is not disabled on either.
* On "Manage verified members" the "Decision method" field is not disabled.
* On "Manage verified members" the "You don't have the correct permissions" error message does not show.

Screenshot 2024-04-11 at 11 36 30

Screenshot 2024-04-11 at 11 37 27

## Suggestions (optional)

See the "Simple payment" action form for an example of how this should behave.

Screenshot 2024-04-11 at 12 16 15

A similar fix for the "Payment builder" action form is in progress in #2123

Be aware that "Create agreement" is slightly different to all the other action types as it is the only one which requires the "Reputation weighting" extension to be installed, so the logic for disabling fields / showing error messages on this form may be slightly different.

iamsamgibbs:
## Description

All inputs (except title and action type) should be disabled when you lack the permissions for the action.

The description field was not disabled for "Manage verified members" or "Create agreement".
On "Manage verified members" the "Decision method" field is not disabled and the "You don't have the correct permissions" error message does not show.

This PR fixes these issues. It also fixes similar issues with "Manage reputation" which was not included in the original issue as it has only recently been merged into master.

For "Manage verified members" the form should only be interactive if the current user has the correct permissions in the root domain (having the administration role in a team only will not be enough).

The PR also handles a TODO comment to ensure "removing reputation" requires the "Arbitration" role and "awarding reputation" requires the "root" role.

And also disables the "Save draft" button on the "Create agreement" form when the "Reputation weighting" extension is not installed.

## Testing

This is best tested with three different users:
1. No permissions at all
2. Arbitration and Administration permissions in the "General" team
3. Arbitration and Administration permissions in any other team

**Create agreement**
All fields (including the "save draft" button) should be disabled unless the reputation weighting extension is installed.

**Manage verified members**
All fields should be disabled, unless you are user 2.

**Manage reputation"
When "awarding reputation", and team is "General" then all fields should be disabled, user you are user 2.
When "removing reputation" fields should not be disabled for user 2 ever, or for user 3 if the team matches the team they have permissions in.

## Diffs

**New stuff** ✨

* "Manage reputation" permissions are checked depending on whether awarding or removing reputation
* "Manage verified members" permissions are set to "Administration" and only valid if in the root domain
* "Save draft" button now disabled when reputation weighting is not installed.

Resolves #2221

Create agreement:
Screenshot 2024-04-12 at 16 03 27

Manage verified members:
Screenshot 2024-04-12 at 16 03 06

Manage reputation:
 
Screenshot 2024-04-12 at 16 01 50

### DIFF
--- a/src/components/v5/common/ActionSidebar/hooks/permissions/useHasNoDecisionMethods.ts
+++ b/src/components/v5/common/ActionSidebar/hooks/permissions/useHasNoDecisionMethods.ts
@@ -1,9 +1,10 @@
 import { useFormContext } from 'react-hook-form';
 
+import { Action } from '~constants/actions.ts';
 import { useAppContext } from '~context/AppContext/AppContext.ts';
 import { useColonyContext } from '~context/ColonyContext/ColonyContext.ts';
 import useEnabledExtensions from '~hooks/useEnabledExtensions.tsx';
-import { getAllUserRoles } from '~transformers';
+import { getAllUserRoles, getUserRolesForDomain } from '~transformers';
 
 import { ACTION_TYPE_FIELD_NAME } from '../../consts.ts';
 
@@ -24,6 +25,10 @@ const useHasNoDecisionMethods = () => {
     return true;
   }
 
+  if (!isVotingReputationEnabled && actionType === Action.CreateDecision) {
+    return true;
+  }
+
   if (isVotingReputationEnabled) {
     return false;
   }
@@ -31,6 +36,15 @@ const useHasNoDecisionMethods = () => {
   const requiredPermissions = getPermissionsNeededForAction(actionType, {});
   if (!requiredPermissions) {
     return false;
+  }
+
+  // Check if the user has the required permissions in root domain
+  // for action types which can only be actioned in root domain
+  const userRootRoles = getUserRolesForDomain(colony, user.walletAddress, 1);
+  if (!requiredPermissions.every((role) => userRootRoles.includes(role))) {
+    if (actionType === Action.ManageVerifiedMembers) {
+      return true;
+    }
   }
 
   // Check if the user has the required permissions in any domain
--- a/src/components/v5/common/ActionSidebar/hooks/permissions/useHasNoDecisionMethods.ts
+++ b/src/components/v5/common/ActionSidebar/hooks/permissions/useHasNoDecisionMethods.ts
@@ -44,26 +44,24 @@ const useHasNoDecisionMethods = () => {
 
   const requiredRolesDomain = getPermissionsDomainIdForAction(actionType, {});
 
-  // If the requiredRolesDomain is root, check the user has the required permissions in root
-  if (requiredRolesDomain === Id.RootDomain) {
-    const userRootRoles = getUserRolesForDomain(
-      colony,
-      user.walletAddress,
-      Id.RootDomain,
-    );
+  const userRootRoles = getUserRolesForDomain(
+    colony,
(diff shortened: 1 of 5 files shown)


src/components/v5/common/ActionSidebar/hooks/permissions/useHasNoDecisionMethods.ts
```diff
@@ -33,6 +38,15 @@ const useHasNoDecisionMethods = () => {
...
 
+  // Check if the user has the required permissions in root domain
+  // for action types which can only be actioned in root domain
+  const userRootRoles = getUserRolesForDomain(colony, user.walletAddress, 1);
+  if (!requiredPermissions.every((role) => userRootRoles.includes(role))) {
+    if (actionType === Action.ManageVerifiedMembers) {
+      return true;
+    }
+  }
+
```
jakubcolony, line 49: Would it be possible to somehow reuse `getPermissionsDomainIdForAction` helper here to determine which domain you need to check permissions in? It would be great to have those rules defined in one place

jakubcolony (dismissed):
Nice work fixing this @iamsamgibbs, all the cases described seem to work. All the validation is a bit convoluted, so well done finding your way around! 

image

bassgeta (approved):
Tried out different cases, it works, great job!
Maybe we can try out Jakub's suggestion too, that seems doable and sensible

bassgeta (dismissed):
Still good, nice one :tada:

src/components/v5/common/ActionSidebar/hooks/permissions/useHasNoDecisionMethods.ts
```diff
@@ -33,6 +42,25 @@ const useHasNoDecisionMethods = () => {
...
+    if (
+      !requiredPermissions.every((requiredPermission) =>
+        userRootRoles.includes(requiredPermission),
+      )
+    ) {
+      return true;
+    }
+  }
+
```
jakubcolony, line 63: Could those 2 checks (this and the `getAllUserRoles` below) be combined? `requiredRolesDomain` will be either `Id.RootDomain` or `undefined` (as it might depend on the form values which we don't know at this point) so it should hopefully be pretty straighforward

jakubcolony (dismissed):
Looks good, left you one extra suggestion - if it's not feasible to implement however, feel free to merge as is. Great work @iamsamgibbs!

rdig (approved):
Tested and working from my side as well. Nicely done!

Screenshot from 2024-04-16 23-17-45

bassgeta (approved):
Reapproving, let's get this badboi merged :tada: !

Issue #2080, opened 2024-03-21
Agreements: Status displaying invalid status


melyndav:
Teamwork task: [Agreements: Status displaying invalid status](https://tw.auroracreation.com/#/tasks/15854542)

## Description

In production, an agreement is showing with the status as 'invalid' with a 0 rendered in the card.

**_Important note: motions shouldn't be tested on production in the BetaColony_**

<img width="671" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/ff5a57fa-1c7e-44ea-87f0-fd55cc34d9a4">

The motion also didn't complete correctly and nothing is showing under each motion pill.
<img width="839" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/9ccb2ea6-5eba-4ca9-97ff-1e5d8ca08c64">

CzarekDryl:
I'm not able to replicate this on local env

rdig:
I think I know of a way to test this.

- Install Voting Reputation
- Create an agreement 
- Stake it, so that it shows up in the list
- Deprecate + Uninstall Voting Reputation

That should break the state of the agreement (motions, under the hood), and trigger this invalid state

CzarekDryl:
@rdig Right after uninstalling extension motion state change to `0` when using `useGetMotionStateQuery`. Could you please check if it's something with DB?
<img width="365" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/76940796/d1f8cf03-0b2f-4e52-9587-1a5d4c85dea2">

arrenv:
This likely requires us handling deprecated and uninstalled extensions in the database.

As it should still show the correct outcome, even if it was created with a previous version of the voting reputation contract.

rdig:
@arrenv is correct. We cannot currently handle motions from voting rep extensions that have been uninstalled. So for now, the UI needs to account for this case and handle it appropriately

arrenv:
@AdrianDudko For now, when this situation is detected, can we instead have the UI say "Uninstalled", instead of "Invalid".

joanna-pagepro:
@arrenv what about displaying this action in the action sidebar? Right now it looks like this: 
![image](https://github.com/JoinColony/colonyCDapp/assets/44369011/d771badf-c77d-4012-8d1b-5a9af4e911da)

Motion steps are empty

melyndav:
[...]

PR #2235, opened 2024-04-15
fix: uninstalled motion state

joanna-pagepro:
## Description

After uninstalling the `Reputation Weighted` extension, motion statuses are showing wrong status - `Invalid`. In such situation it should be `Uninstalled` and a box with the information about the uninstalled extension should be visible instead of the `Stepper` component inside the action preview (sidebar) component.

## Testing

* Step 1. Install `Reputation Weighted` extension if it's not installed yet
* Step 2. Create a `Create agreement` action and stake it, so that it shows up in the list
* Step 3. Deprecate and uninstall the `Reputation Weighted` extension
* Step 4. Go to the agreements page and open the agreement card

## Diffs

**New stuff** ✨

* new motion state `Uninstalled` shown when motion failed and voting reputation extension is uninstalled
* `ExtensionStatusBox` component to show that extension is uninstalled

**Changes** 🏗

* `getMotionState` function is also getting `isVotingReputationEnabled` arg right now, and based on that and networkMotionState we can return the `Uninstalled` status.

**Deletions** ⚰️

* 

Resolves https://github.com/JoinColony/colonyCDapp/issues/2080


melyndav (changes requested):
Thanks for your work on this @joanna-pagepro it's working as expected when I deprecate and uninstall.

image

However, if I install and enable the extension again, this state should remain. However, it goes back to the invalid state and shows the empty motion pills.

image

joanna-pagepro:
@melyndav this is what the API returns. @arrenv @rdig can you help here?

arrenv:
> @melyndav this is what the API returns. @arrenv @rdig can you help here?

Is there a way to interpret the response from the API? Such as, comparing the active extension ID with the ID in the motion id?

rdig:
So the the way this works is that once you uninstall the extension you have no more state for that motion, past or present.

Installed:
Screenshot from 2024-04-15 17-28-47
Screenshot from 2024-04-15 17-28-54

Uninstalled: (not voting rep extension active)
Screenshot from 2024-04-15 17-32-29
Screenshot from 2024-04-15 17-32-34

Reinstalled:
Screenshot from 2024-04-15 17-33-20
Screenshot from 2024-04-15 17-33-27

This is pretty easy to solve, at least from the UI perspective, in the context of this issue.

As long as your status is `Invalid` / network state is `0`, you can safely display the "uninstalled" banner

There shouldn't be any other instances where you get a `0` network state, so you can blanket assume that the extension was unistalled at some point

joanna-pagepro:
Ok, so right now if the status is `Invalid`, I'm showing `Uninstalled` pill

iamsamgibbs (changes requested):
Good start to this ticket!

The status is correctly showing "Uninstalled" for old motions which were done on previous installations of the extension.

Screenshot 2024-04-16 at 14 20 07

The status is correctly showing in the sidebar too.

Screenshot 2024-04-16 at 14 18 51

However, when the extension is installed again this status disappears:

Screenshot 2024-04-16 at 14 19 23

The motions also go to the "Uninstalled" state when the extension is deprecated (before it is uninstalled). From my understanding, when in the deprecated state it should still be possible to proceed through the remaining motion steps and finalize the motion.

Screenshot 2024-04-16 at 14 20 32

rdig:
@iamsamgibbs @joanna-pagepro 

> The motions also go to the "Uninstalled" state when the extension is deprecated (before it is uninstalled). From my understanding, when in the deprecated state it should still be possible to proceed through the remaining motion steps and finalize the motion.

Sam is correct in this assumption. This needs to be addressed as while the extension is deprecated, all currently active motions can still be interacted with, even more so we get the correct state from them from the action/motion object

arrenv:
> @rdig @melyndav how should the last step look like? Should there be a separate status for this state like `Extension deprecated` or something like this?

@joanna-pagepro When the extension is in a deprecated state the existing and past motions will still work as if it was enabled. You are just not able to create new ones. So, there should not need to be a different state for it.

Issue #2165, opened 2024-04-08
Finalize button appears with 0 stake and failed outcome


arrenv:
Teamwork task: [Finalize button appears with 0 stake and failed outcome](https://tw.auroracreation.com/#/tasks/15873332)

## Steps to reproduce

1. Ensure you have the Reputation extension installed with the testing configuration.
2. Create an action using the Reputation decision method.
3. Don't stake at all on the action.
4. Allow the time to pass and the action to fail.
5. Note the state of the card in the last stage of the process, which has a disabled Finalize button.

## Expected behaviour

* There should be no finalize button in this case disabled or otherwise.
* The copy in the header should align with the design.
* (Additionally) If the action goes to vote and fails, the "Finalize" button should instead be "Return stakes". As the action cannot be finalized and so the button is only needed to return user's stakes.

[Figma link](https://www.figma.com/file/l1dOM5qiQYwF0ElvKDqqjg/Design-System---Colony-v3?type=design&node-id=1782-94998&mode=design&t=739qVr2UwdcFScPU-4)

![image](https://github.com/JoinColony/colonyCDapp/assets/33682027/5c441201-568f-4399-a8ec-9ce878f4447c)

## Actual behaviour

* There is a finalize button that is visible but disabled. This can be confusing, as it implies it can be finalized but the button is just disabled.
* The info description in the card is incorrect.

![image](https://github.com/JoinColony/colonyCDapp/assets/33682027/8fc51ebd-fbf2-4a2a-bc9f-e109bceacb58)

Issue #2172, opened 2024-04-08
Remove option to finalize failed motions


arrenv:
Teamwork task: [Remove option to finalize failed motions](https://tw.auroracreation.com/#/tasks/15873333)

## Steps to reproduce

1. Create an action using the Reputation decision method.
2. Fully support and fully oppose the action.
3. Don't vote on the action and allow it to fail.
4. Wait until it fails and you are on the Finalize stage.
5. You will get a "Finalize" button that does not work, and you are not able to return stakes to uses.

## Expected behaviour

* You should not be able to finalize a failed action.
* You should be able to return stakes to users.

## Actual behaviour

* Users incorrectly have a finalize button visible in the UI that results in an error and a failed transaction.
* You are not able to return stakes to users.

![image](https://github.com/JoinColony/colonyCDapp/assets/33682027/8d321b65-46a9-4660-bffe-9d4895ff12d8)

joanna-pagepro:
@arrenv do we have a design for this?

AdrianDudko:
@arrenv I've found this but please confirm

https://www.figma.com/file/l1dOM5qiQYwF0ElvKDqqjg/Design-System---Colony-v3?type=design&node-id=1782-83711&mode=design&t=iHyuUhADbcwS2hz2-4
<img width="276" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/112850454/fa84a69a-c4ed-46a3-9bed-2cbaa2bcdf3e">

arrenv:
> @arrenv I've found this but please confirm

Yes, this is the correct design.

CzarekDryl:
@arrenv @AdrianDudko @rdig  I think that I will need some help of colony team - here is Raul comment on this one `the gist here is that you can have multiple failure states, some in which you have rewards, some in which you don't so we'll have to find something that accounts for both cases, which might be more tricky`

joanna-pagepro:
@arrenv @rdig so we can't show the reward and total here: 
![image](https://github.com/JoinColony/colonyCDapp/assets/44369011/d8545bf2-1414-46e3-9e19-3e3ffb61c482)
?

@rdig can I use `revealedVotes` field to check if we have the case described by @arrenv? Or is there some other way? Or should we show `Finalize` button only when `networkMotionState` is `NetworkMotionState.Finalizable`?

joanna-pagepro:
> > @arrenv @rdig so we can't show the reward and total here?
> 
[...]

PR #2236, opened 2024-04-15
fix: changed failed finalize step content

CzarekDryl:
You should not be able to finalize a failed action.
You should be able to return stakes to users.

## Testing

- Ensure you have the Reputation extension installed with the testing configuration.
- Create an action using the Reputation decision method.
- Don't stake at all on the action.
- Allow the time to pass and the action to fail.

Resolves [https://github.com/JoinColony/colonyCDapp/issues/2165](https://github.com/JoinColony/colonyCDapp/issues/2165)


iamsamgibbs (changes requested):
Good work on this ticket so far! It does resolve the issue you've described in the PR:

Screenshot 2024-04-15 at 16 29 27

But following the steps to reproduce in the original issue (#2172) (which includes staking both sides) I'm still experiencing the same issue:

Screenshot 2024-04-15 at 16 27 39

ArmandoGraterol (commented):
Looking good! In my env, the case in the initial issue is not happening. I followed the steps however I was actually able to finalize and claim the stakes. The case in the PR's description is fixed for me too.

CzarekDryl:
@iamsamgibbs @ArmandoGraterol My bad, I've linked wrong issue

iamsamgibbs (approved):
Ah okay! All good then in that case!

ArmandoGraterol (approved):
Looks good to me 👍 
Screenshot 2024-04-16 at 14 18 01

PR #2251, opened 2024-04-17
Add streaming payment testing UI and add amount conversion in saga

### LINKED ISSUE #2211: Streaming Payments API: Creating streams using permissions
Parent issue: https://github.com/JoinColony/colonyCDapp/issues/2207

- Saga needs to be modified to accept and set token and amount
- It should be possible to provide custom title 
- It should be possible to annotate the create transaction
- There should be a colony action created with a field linking to the stream

davecreaser:
## Description

* Add a button to the testing UI to test streaming payment create.
* Add a conversion of the amount in the create streaming payment saga to ensure the correct amount is sent.

## Testing

Test that streaming payments can be created in the db successfully.

* Step 1. Click on the create streaming payment button in the temporary testing UI on the dashboard.
* Step 2. Wait for it to complete.
* Step 3. Check that the payment was created in the db with the correct payload.

## Diffs

**New stuff** ✨

* New temporary button on the dashboard for testing.

**Changes** 🏗

* Changed the createStreamingPayment saga to ensure the correct amount is sent.

Resolves #2211

### DIFF
--- a/src/components/v5/payments/TmpAdvancedPayments.tsx
+++ b/src/components/v5/payments/TmpAdvancedPayments.tsx
@@ -3,7 +3,7 @@ import React, { useState } from 'react';
 
 import { useAppContext } from '~context/AppContext/AppContext.ts';
 import { useColonyContext } from '~context/ColonyContext/ColonyContext.ts';
-import { useGetExpenditureQuery } from '~gql';
+import { StreamingPaymentEndCondition, useGetExpenditureQuery } from '~gql';
 import useAsyncFunction from '~hooks/useAsyncFunction.ts';
 import useCurrentBlockTime from '~hooks/useCurrentBlockTime.ts';
 import useEnabledExtensions from '~hooks/useEnabledExtensions.ts';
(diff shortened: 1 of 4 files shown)


iamsamgibbs (approved):
This all looks good to me! Although it seems the `CreatePaymentMetadata` mutation is failing, I wonder if you could take a look at that here too as it is part of the same saga?

Screenshot 2024-04-18 at 09 34 42

Otherwise I'm able to access the extension page and install it.

Screenshot 2024-04-18 at 09 42 24

And I'm able to create a streaming payment and have it show up in the database.

Screenshot 2024-04-18 at 09 34 55

jakubcolony (approved):
All good here, nice one @davecreaser!

davecreaser:
> This all looks good to me! Although it seems the `CreatePaymentMetadata` mutation is failing, I wonder if you could take a look at that here too as it is part of the same saga?

Thanks @iamsamgibbs, I think @jakubcolony mentioned that he'd fix the `CreatePaymentMetadata` mutation, as it's a simple case of adding it to the auth. If you'd like me to do it instead though Jakub then let me know and I'll look into it!

PR #2253, opened 2024-04-17
Add saga to set domain skill threshold

### LINKED ISSUE #2140: [Multi-sig] feat: implement configuring of multi-sig extension per domain
### Story

As a *user moderating a colony*, I want to be able to set custom domain multi-sig thresholds. 

### Description

This enables that motions in each domain can have different thresholds (e.g. subteams need a lower threshold than the general one).
We need to enable users to be able to set custom thresholds per domain.
This means that each domain will have its own custom threshold, by default it inherits the global one.
We need to create a saga which will call [this contract](https://github.com/JoinColony/colonyNetwork/blob/029bd755d24458a12e6dd9167d7a915992845362/contracts/extensions/MultisigPermissions.sol#L166), but multiple times.
The settings get applied after the user clicks the `Save settings` button in the top right, so we need to do this for each domain.

The default value is `Inherit from colony`, which I suppose is the value when the return value of [this contract](https://github.com/JoinColony/colonyNetwork/blob/029bd755d24458a12e6dd9167d7a915992845362/contracts/extensions/MultisigPermissions.sol#L408) is `0` and the return value of `globalThreshold` is not `0`.

If `globalThreshold` is `0`, then that means we are using majority approval.

If it's anything else than that, it's `Fixed`.
To unset a fixed threshold call the `setDomainSkillThreshold` contract with a value of 0.

### Design

[Figma link](https://www.figma.com/file/SaO0Zj6e7yGqLzjEsxqVhy/Permissions?type=design&node-id=20450-30433&mode=design&t=wUsN5E7ncX19aOOU-4)

image

## Implementation

TODO: There's an edge case here, if we set the global threshold to 4, and set a domain's threshold to 4, we technically are using `Inherit from colony` but it was manually changed. Probably a non-issue, why would a user even do that :shrug: 

TODO: Can users set a domain to have majority approval WHILE the colony itself uses a fixed threshold?
image
at first glance it doesn't seem so

- [ ] User can set a domain to have a fixed threshold approval
- [ ] User can set a domain to inherit threshold from colony
- [ ] User can set a domain to use majority approval.
- [ ] User can alternate between all of these states

ArmandoGraterol:
Edited by @iamsamgibbs 

## Description

[block-ingestor counterpart](https://github.com/JoinColony/block-ingestor/pull/210)

As a user moderating a colony, I want to be able to set custom domain multi-sig thresholds.

This PR replaces the existing `setGlobalThreshold` saga with a new `setThresholds` saga which handles both the global threshold and domain threshold settings. The extensions setting UI has been expanded with a temporary UI to set the domain threshold settings. The block-ingestor has a new handler for the `DomainSkillThresholdSet` event.

## Testing

* Step 1 - Install the multi-sig extension
* Step 2 - Navigate to the extensions settings tab
* Step 3 - Make adjustments to both the colony settings and the individual domain threshold settings. (Note: If the colony threshold setting is set to 'majority' then this setting won't be available as a choice for the individual domains as it would be the same as 'inherit from colony wide'.)
* Step 4 - Run the following query and ensure the values reflect what you have entered in the form. Any domains set to 'inherit from colony wide' should have the same value as the colony threshold.

```
query MyQuery {
  listColonyExtensions {
    items {
      params {
        multiSig {
          colonyThreshold
          domainThresholds {
            domainId
            domainThreshold
          }
        }
      }
    }
  }
}
```

## Diffs

**New stuff** ✨

* `setThreshold` saga replaces `setGlobalThreshold` with multicall saga to handle both global threshold settings and all domain threshold settings
* Temporary extensions settings UI now includes domain threshold inputs

Screenshot 2024-05-21 at 12 36 18

Screenshot 2024-05-21 at 12 38 28

Resolves #2140


src/components/frame/Extensions/pages/MultiSigPage/MultiSigPageSetup.tsx
```diff
@@ -115,17 +197,32 @@ const MultiSigPageSetup: FC<MultiSigPageSetupProps> = ({ extensionData }) => {
...
   }, [extensionData, pathname, navigate]);
 
-  const getSetThresholdPayload = () => {
-    if (thresholdType === MultiSigThresholdType.MAJORITY_APPROVAL) {
-      return {
-        colonyAddress,
-        threshold: 0,
-      };
-    }
+  const getSetThresholdsPayload = () => {
```
bassgeta, line 200: just thinking if it still makes sense for this to be an object, since we are using an array on the block-ingestor and also `gql` returns an array too.

bassgeta (approved):
Seems legit, just dropped a comment if we want to get rid of a reduce and just use an array since we do that on the backend anyways

rdig (approved):
Can't spot anything that stands out. Good to go

bassgeta (approved):
Still looking, good, working too!

PR #2254, opened 2024-04-17
Store changes made using `editExpenditure` saga

### LINKED ISSUE #2245: Advanced Payments API: Store changes made using `editExpenditure` saga
The UI needs to display the initiator and summary of changes made when editing an expenditure. Unlike edit expenditure motions, edit action will need to store both the updated and previous state of properties it affects.

- schema will need to be modified, new action type introduced and relevant fields added to the action model
- block-ingestor handlers will need to be modified to create a new colony action when expenditure is edited

jakubcolony:
## Description

This PR contains schema updates that will allow for creating an action whenever expenditure is edited. All changes made to the expenditure will be stored on action's `expenditureSlotChanges` field.

It determines whether an edit action took place based on the presence of a multicall inside transaction trace and counting the relevant events within it.

block-ingestor PR: https://github.com/JoinColony/block-ingestor/pull/205

## Testing

Use the testing UI to create and edit expenditures. Note edit action only relates to locked expenditures, so you will need to do that before. You can edit the payload of edit expenditure saga to set different values in line `313` of `TmpAdvancedPayments.tsx`.

Every edit action should be stored in the DB and contain the `oldSlots` and `newSlots` under `expenditureSlotChanges` with correct values. It should not matter if edit transaction contains a single `setExpenditureState`/`setExpenditurePayout` call or more.

Since these methods are also called in other circumstances (e.g. setting expenditure payout when creating an expenditure), confirm no unwanted actions were created in the DB.

One of the queries I used for testing (allows you to confirm the relevant actions are created but also no unwanted actions get added):
```gql
query ListExpsAndActions {
    items {
      id
      slots {
        id
      }
      actions {
        items {
          id
          type
          expenditureSlotChanges {
            oldSlots {
              claimDelay
              id
              payoutModifier
              payouts {
                amount
                isClaimed
                tokenAddress
                networkFee
              }
              recipientAddress
            }
            newSlots {
              claimDelay
              payoutModifier
              id
              payouts {
                amount
                isClaimed
                networkFee
                tokenAddress
              }
              recipientAddress
            }
          }
        }
      }
    }
  }
  listColonyActions(filter: {type: {eq: EDIT_EXPENDITURE}}) {
    items {
      id
    }
  }
}
```

Resolves #2245


iamsamgibbs (changes requested):
Tried testing this, but was unable to edit the expenditure once locked.

Screenshot 2024-05-03 at 16 57 06

Screenshot 2024-05-03 at 16 56 55

When editing before locking, it does appear to work as intended though:

Screenshot 2024-05-03 at 16 56 51

jakubcolony:
@iamsamgibbs it looks like the UI was still thinking the expenditure is draft while it was locked - it seems to call `setExpenditureValues` which only applies for draft expenditures. Could you try waiting a few seconds and then refetching before editing?

iamsamgibbs (approved):
🤦‍♂️ Yep, just needed to be patient! All working as expected.

Screenshot 2024-05-03 at 18 08 02

Tested with changed values too:

Screenshot 2024-05-03 at 18 10 36

ArmandoGraterol (approved):
LGTM 👌 
Screenshot 2024-05-06 at 15 15 56

PR #2263, opened 2024-04-19
Fix: Created in field should only show current team or general

### LINKED ISSUE #2223: Created in should only show current and parent teams
## Steps to reproduce

1. Ensure you have voting reputation installed.
2. Ensure you have teams.
3. Create an action with the Reputation Decision method in a team.
4. Click on the `Created in` field and note the team options.

## Expected behaviour

* You should only be able to select to create the motion in the selected team, or a parent team.
* You should not have the option to select any other adjacent of child team.
* Changing the team above it should update the `Created in` team to match it.
* If the team has not been selected and the `Created in` field is clicked, then any team can be chosen, but it should also change the team field above it to match.

## Actual behaviour

* You are able to select from any teams in a Colony to create the motion in.

image

davecreaser:
## Description

* The created in field (in action forms) should be filtered based on the currently selected team.
* You should only be able to select to create the motion in the selected team, or general.
* You should not have the option to select any other team.
* Changing the team above it should update the created in field to match it.
* If the team above has not been selected and the created in field is changed, then any team can be chosen, but it should also change the team field above it to match, which in turn then filters the created in options.

https://github.com/JoinColony/colonyCDapp/assets/38098203/8ee06a8b-d1df-490f-995b-ab7e1cd2785d

## Testing

* Ensure you have voting reputation installed.
* Create an action with the Reputation Decision method in a team.
* Click on the created in field and check that the team options available to you is correct, based on the description above.

Test that the created in field is correctly filtered in the following places:

* Manage permissions
* Transfer funds
* Simple payment
* Edit a team

## Diffs

**New stuff** ✨

* New hook in the ActionSidebar folder to apply the same filtering logic in multiple places.

**Changes** 🏗

* Filter the action forms created in field options in the correct way.

Resolves #2223

### DIFF
--- a/src/components/v5/common/ActionSidebar/hooks/useFilterCreatedInField.ts
+++ b/src/components/v5/common/ActionSidebar/hooks/useFilterCreatedInField.ts
@@ -13,19 +13,23 @@ import {
 /**
  * Hook to filter the created in field in the action form, based on the current value of the team field passed in
  */
-const useFilterCreatedInField = (nameOfFieldToFilterOn: string) => {
+const useFilterCreatedInField = (
+  nameOfFieldToFilterOn: string,
+  onlyAllowRoot = false,
+) => {
   const { setValue, watch } = useFormContext();
   const selectedTeam = watch(nameOfFieldToFilterOn);
   const createdIn = watch(CREATED_IN_FIELD_NAME);
   const decisionMethod = watch(DECISION_METHOD_FIELD_NAME);
 
   useEffect(() => {
-    if (decisionMethod !== DecisionMethod.Reputation) return;
+    if (onlyAllowRoot || decisionMethod !== DecisionMethod.Reputation) return;
 
     if (!selectedTeam && !!createdIn && createdIn !== Id.RootDomain) {
       setValue(nameOfFieldToFilterOn, createdIn);
     }
   }, [
+    onlyAllowRoot,
     createdIn,
     decisionMethod,
     nameOfFieldToFilterOn,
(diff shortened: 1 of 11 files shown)


ArmandoGraterol (commented):
I'm not 100% sure about this but it I think it shouldn't be possible to create a upgrade colony version motion in any other domain other than Root/General, as after all, you are updating a whole-ass colony. It's not something domain specific:

Screenshot 2024-04-19 at 15 53 03

iamsamgibbs (changes requested):
Nice work on this so far! All the code looks great, but I've got a couple of very minor nitpicky issues that are definitely more quality of life improvements.

This isn't an issue introduced by this PR, but can quickly resolve it here - as Armando noticed, the 'Created in' field for 'Upgrade Colony version' should be read only.

Screenshot 2024-04-22 at 09 05 07

The 'Created in' field behaves a bit differently on 'Simple payment' compared to 'Transfer funds'. I think this is due to 'Transfer funds' having 'General' as the default team, whereas 'Simple payment' doesn't have a default. This means the 'Created in' field only shows 'General' as an option. It would be nice if these behaved consistently.

 Simple payment
Screenshot 2024-04-22 at 09 01 41

Transfer funds
Screenshot 2024-04-22 at 09 01 49

The 'Created in' field on 'Manage reputation' is also behaving slightly differently than the other actions. On the other actions, if you have selected a team and 'Created in' is set to 'General', when you change the team, the 'Created in' field will change from 'General' to the new team. On 'Manage reputation' it is staying on 'General'. Would be nice to either make all the other actions behave the same way as 'Manage reputation', or make 'Manage reputation' behave the same way as the others.


https://github.com/JoinColony/colonyCDapp/assets/34915414/72759d5b-8469-4991-a1b7-34675125935d

davecreaser:
Cheers guys, should be all consistent now! And I've made the Upgrade Colony Version createdIn field readonly too. 📈

melyndav (approved):
I've checked all the actions types and everything is working perfectly. Nice home run on this one @davecreaser! 🥇

Issue #2228, opened 2024-04-12
Team pill colour audit


melyndav:
## Description

** This is not a priority issue and is just part of general UI/UX QA and SOT maintenance ** 

The team pills/badges are used throughout all action types and activity feeds, allowing the user to perform actions within selected teams or sort by team information in filtering actions, etc. 

Currently, there is a very minor difference between the Figma component colour tokens and the ones in production. Such as three green teams showing with yellow borders. So a very quick check of the design system components are needed as part of this scope. 

The scope of this PR is to check all the team pills in our Design System and ensure they match in production correctly. Any issues such as incorrect border colors should be resolved. 

It's important that the main team colour which corresponds to the 'create new team' and 'edit team' action types are not impacted.

## Figma design system components
<img width="520" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/2b176183-dc96-49ef-a446-5404ab0df3e8">

https://www.figma.com/file/l1dOM5qiQYwF0ElvKDqqjg/Design-System---Colony-v3?type=design&node-id=1809-1165&mode=design&t=Q8lpfWcaTt0sGPJZ-4

Team colour tokens:
<img width="667" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/8fe6f098-c264-47be-a181-b92f0313a655">

Link: https://www.figma.com/file/l1dOM5qiQYwF0ElvKDqqjg/Design-System---Colony-v3?type=design&node-id=1776-34243&mode=design&t=Q8lpfWcaTt0sGPJZ-4

### Production issues identified:
Any green team type is showing with a yellow border:
<img width="196" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/5b6f3f81-cb43-4407-87d3-a2752325726d">

<img width="207" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/75efcf6b-8d63-4203-aa6a-7ad4892f500a">

## Root colour

The second part of this PR to update the default root colour. It's currently set as purple, but should be set to the default text-blue-400 colour. This blue token is reserved for the general team colour and can not be selected by any other team. Colonies are free to change this colour, but by default we've assigned the blue to help it stand out from the other options available. The team dropdown and other sears should show this blue colour by default. 

Currently:
[...]

PR #2269, opened 2024-04-22
Feat/#2228 update default root domain colour

davecreaser:
## Description

- Add a colour to the DomainColor enum called 'Root' which will be used as the default colour for the root domain (general) when a new colony is created.
- This colour should not be selectable when editing or creating a colony.
- If you edit the general domain to have a different colour, you should not be able to go back to the default colour.

Note: This PR is related to changes in the Block Ingestor here: https://github.com/JoinColony/block-ingestor/pull/208

## Testing

For this PR specifically, just check that the root colour (blue) is not selectable when editing or creating a team. There should only be 16 colours available to choose from, not 17.

If you want to test the full flow, you'll need to run block ingestor locally using this branch: https://github.com/JoinColony/block-ingestor/tree/feat/%232228-update-default-root-domain-colour

## Diffs

**New stuff** ✨

* Add a colour to the DomainColor enum called 'Root' which will be used as the default colour for the root domain (general) when a new colony is created.

Contributes to #2228

### DIFF
--- a/src/graphql/generated.ts
+++ b/src/graphql/generated.ts
@@ -1505,20 +1505,6 @@ export type CreateTransactionInput = {
   titleValues?: InputMaybe<Scalars['String']>;
 };
 
-/** Input data for creating a unique Colony within the Colony Network. Use this instead of the automatically generated `CreateColonyInput` input type */
-export type CreateUniqueColonyInput = {
-  /** Unique identifier for the Colony. This is the Colony's contract address */
-  colonyAddress: Scalars['ID'];
-  /** User id of creator to associate with further invite codes */
-  initiatorAddress: Scalars['ID'];
-  /** Unique identifier for the Colony's native token (this is its address) */
-  tokenAddress: Scalars['ID'];
-  /** The transaction hash of colony creation transaction */
-  transactionHash: Scalars['String'];
-  /** Type of the Colony (regular or MetaColony) */
-  type?: InputMaybe<ColonyType>;
-};
-
 /** Input data for creating a unique user within the Colony Network Use this instead of the automatically generated `CreateUserInput` input type */
 export type CreateUniqueUserInput = {
   /** Unique identifier for the user. This is the user's wallet address */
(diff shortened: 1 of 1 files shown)


src/graphql/generated.ts
```diff
@@ -1505,6 +1505,20 @@ export type CreateTransactionInput = {
...
+  /** User id of creator to associate with further invite codes */
+  initiatorAddress: Scalars['ID'];
+  /** Unique identifier for the Colony's native token (this is its address) */
+  tokenAddress: Scalars['ID'];
+  /** The transaction hash of colony creation transaction */
+  transactionHash: Scalars['String'];
+  /** Type of the Colony (regular or MetaColony) */
+  type?: InputMaybe<ColonyType>;
+};
+
```
jakubcolony, line 1521: Hmm, I think this is a leftover from before #2225, you might have to restart your dev and run codegen again.

jakubcolony (changes requested):
Changes look great, nice work @davecreaser, good to go as soon as the generated file is updated 👍

iamsamgibbs (changes requested):
Nice work on this so far @davecreaser 🎉

I can only see 16 colours as expected in the team colour picker and can confirm other teams cannot select the root colour.

Screenshot 2024-04-23 at 09 23 42

I've found if you change the 'General' team colour to any other colour, then it makes it impossible to change back to the default colour - not sure if this is the intended behaviour or not.

Screenshot 2024-04-23 at 09 24 22

And as Jakub mentioned, looks like codegen needs a re-run.

You could also add the most recent commit hash of your block-ingestor PR to the `colony-cdapp-dev-env-block-ingestor` docker file to get docker to build with that commit.

melyndav:
> I've found if you change the 'General' team colour to any other colour, then it makes it impossible to change back to the default colour - not sure if this is the intended behaviour or not.

FYI @iamsamgibbs it's intended. Default is the primary blue and then if they customize it's just the standard team option colours.

davecreaser:
Codegen has been rerun, looks cleaner now ;)

iamsamgibbs (approved):
Looks good to me now!

melyndav (approved):
Appreciate your work on this fix @davecreaser! 💯

PR #2292, opened 2024-04-25
Polygon and Arbitrum Network Details

rdig:
## Intro Note

## Current PR Details

This is a just a simple PR that adds network details into the `constants` file, there isn't much testing to be done.

If you really want to be thorough, you can test that the various networks render correctly by updating your `.env.local`'s `NETWORK_ID` value and connecting the correct wallet _(you'll need to use Metamask for that)_


src/types/network.ts
```diff
@@ -14,4 +18,6 @@ export const ColonyJSNetworkMapping = {
   [Network.Gnosis]: ColonyJSNetwork.Xdai,
   [Network.GnosisFork]: ColonyJSNetwork.XdaiQa,
   [Network.Ganache]: ColonyJSNetwork.Custom,
+  [Network.ArbitrumOne]: ColonyJSNetwork.ArbitrumOne,
+  [Network.ArbitrumSepolia]: ColonyJSNetwork.ArbitrumSepolia,
```
rdig, line 22: Note that only Arbitrum is listed here due to `colonyJS` not having support for Polygon.

This is ok for now, but will need updating if that ever becomes a requirement.

jakubcolony (approved):
All good 👍 

image
image

image
image

davecreaser (approved):
LGTM 👌

PR #2293, opened 2024-04-26
Refactor Chain ID and Native Skill ID to be stored as a `String`

rdig:
## Intro Note

## Current PR Details

Most of the work in this PR (as well as [on the Block Ingestor side](https://github.com/JoinColony/block-ingestor/pull/211)) has been done by @jakubcolony 

Changes here are to convert chain ids (colony model) and skill ids (domain model) to strings, as the new deployments now generate longer number than a 32-bit integer can store

Current:
Screenshot from 2024-04-26 11-49-50
New deployments:
Screenshot from 2024-04-26 11-50-13

Testing should be done by starting the dev environment, running the data creation script (or creating a colony manually) and checking, via the Graphql playground, if the chain ids and skill ids are indeed strings

Screenshot from 2024-04-26 11-54-51
Screenshot from 2024-04-26 11-55-01


rdig:
> `npm ci` is not working for me. It seems like there's something funky going on with the new package and MacOS:
> 
> Screenshot 2024-04-29 at 13 10 13

Yep, me and @jakubcolony spotted this as well and looking into fixing it

Thanks for confirming though :+1:

jakubcolony (approved):
`npm i` now works on my end, and the `create-data` script was able to create domains that I can see in the UI. LGTM 👍

rdig:
@ArmandoGraterol please re-test as it should be all good now

ArmandoGraterol (commented):
Latest commit allowed me to use `npm i` but now when I do `npm run dev` this is what I get:
Screenshot 2024-04-30 at 14 46 08

I've removed `node_modules` and started from scratch already.

ArmandoGraterol (approved):
Working as expected for me 👌 

Screenshot 2024-04-30 at 16 48 35

PR #2294, opened 2024-04-26
[Multi-sig] Update manage permissions saga to handle MultiSigRoleSet

### LINKED ISSUE #2146: [Multi-sig] feat: implement assigning of multi-sig roles
### Story

As a *user moderating a colony*, I want to be able to assign multi-sig roles to users.

### Description

Each colony contributor will have a new property called `multiSigRoles`. We need to add those in our graphql schema.
Essentially modify the `managePermissions` saga to also support the `Take actions via multi-sig` option.
Additionally, the role options displayed are different for multi-sig than Permissions, check the designs, essentially we don't have a `Mod` role here.

If that's selected, call [this contract](https://github.com/JoinColony/colonyNetwork/blob/029bd755d24458a12e6dd9167d7a915992845362/contracts/extensions/MultisigPermissions.sol#L433) and then modify the block-ingestor to handle the [`MultiSigRoleSet`](https://github.com/JoinColony/colonyNetwork/blob/029bd755d24458a12e6dd9167d7a915992845362/contracts/extensions/MultisigPermissions.sol#L480C14-L480C29) event.
It's probably handled in the same way as our current `managePermissions` handler works, just update a different property.

You will also need to write actions for this.

### Design

[Figma link](https://www.figma.com/file/SaO0Zj6e7yGqLzjEsxqVhy/Permissions?type=design&node-id=20450-34558&mode=design&t=iqzIcnoziy1lwDp6-4)

image

- [ ] Users can be assigned multi-sig specific roles
- [ ] The changes are reflected on the backend

iamsamgibbs:
## Description

[Block-ingestor counterpart](https://github.com/JoinColony/block-ingestor/pull/212)

This PR adds support to the 'Manage permissions' saga and block-ingestor handler to assign multi sig roles to users.

## Testing

1. Install the 'Multi-sig permissions' extension.
2. Run this query, any user who already has root permissions should be assigned root permissions with multi-sig. (Thanks to Bassgeta for implementing this on CDapp side in the previous PR, this PR adds the block-ingestor side to handle this and store it in the database).

```
query MyQuery {
  listColonyRoles(filter: {isMultiSig: {eq: true}}) {
    items {
      id
      isMultiSig
      role_0
      role_1
      role_2
      role_3
      role_5
      role_6
    }
  }
}
```

3. Create a new 'Manage permissions' action. The 'Take actions via Multi-sig' authority option will only show when the extension is installed.
4. Select a permissions type or select custom permissions.
5. Submit the action and run the query again to check the database for updates. Try adding / removing permissions from the same user to confirm the changes in the database are correct. ~~Please note the UI will need updating before it accurately reflects the selected permissions.~~ now implemented ~~Also note, role_6 (administration) is not used for multi-sig and will always be null.~~ not sure if this is actually the case, have put role_6 back in for now and will confirm.

## Diffs

**New stuff** ✨

* ColonyRole now has `isMultiSig` field and new id structure for multi-sig roles

**Changes** 🏗

* Manage permissions saga now handles multi sig roles
* Manage permissions block-ingestor handler now handles multi sig roles
* Manage permissions action sidebar UI changes and completed action UI changes

Resolves #2146 

Screenshot 2024-04-26 at 10 05 17

Screenshot 2024-04-26 at 10 52 17


ArmandoGraterol (commented):
Looking good! Dunno if this is worth updating in this PR, but the completed action state of manage permission should also reflect the new authority:
Screenshot 2024-04-29 at 14 13 06

It is currently assuming every action/motion is using permissions instead of multi-sig

iamsamgibbs:
> Looking good! Dunno if this is worth updating in this PR, but the completed action state of manage permission should also reflect the new authority:
> 
> It is currently assuming every action/motion is using permissions instead of multi-sig

I've made some changes today to ensure the correct permission tiers are shown on the action, and that the correct authority and permissions are shown on the completed action. The 'permission type' field on the action sidebar is also pre-selected if the user already has a multi-sig permission in the selected team and the multisig authority is selected. 

Screenshot 2024-04-30 at 16 42 29

Screenshot 2024-04-30 at 16 14 31

ArmandoGraterol (changes requested):
The CDapp is now crashing when I create an action/motion using these parameters:

Screenshot 2024-04-30 at 15 30 24
Screenshot 2024-04-30 at 15 31 00

It is also happening with the other authority type.

ArmandoGraterol (commented):
Thank you for the changes Sam! Only thing I've noticed is this:

Screenshot 2024-05-01 at 15 12 08

I've created this motion with multi-sig authority but the panel is showing the other option as if that was the one I selected.

iamsamgibbs:
🤦‍♂️ Had completely forgotten to handle motions, this is implemented now!

ArmandoGraterol (approved):
Looks good to me Sam! Thank you for the changes.

Screenshot 2024-05-02 at 15 09 57
Screenshot 2024-05-02 at 15 10 33

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -3060,7 +3068,7 @@ Snapshot of the user's full roles/permissions at a specific block
 type ColonyHistoricRole @model {
   """
   Unique identifier for the role snapshot
-  Format: `colonyAddress_domainNativeId_userAddress_blockNumber_roles`
+  Format: `<colonyAddress>_<domainNativeId>_<userAddress>_<blockNumber>_roles`  or `<colonyAddress>_<domainNativeId>_<userAddress>_<blockNumber>_multisig_roles`
```
rdig, line 3071: At some point in the past we used to "decompose" id's to extract various data from it _(ie: the user address)_

I don't think we still do it, but it's something to be aware, since that might break with the inclusion of two different "formats" of ids

rdig (approved):
Nothing stands out. Nicely done

PR #2304, opened 2024-04-30
Refactor Arbitrum Specific Urls

rdig:
## Intro Note

## Current PR Details

This PR updates the initial URLs we were using for Arbitrum to point to the various block explorer endpoints, as well as refactor the `getBlockExplorerLink` removing most of it's legacy logic and simplifying it's code.

Testing this is easier done in production, by going to any place where we link to an external token, address or transaction. One of the fastest place where you can access this is via the token popup:

Screenshot from 2024-04-30 17-04-32

Alternatively, you can **test this locally** as well, however, if you still use the local dev chain, you won't get a proper URL generated for the block explorer


jakubcolony (approved):
Changes look sensible, in production all the URLs work but I noticed there's a double slash in the URL:
image

rdig:
Nice catch, I'll look into that

rdig:
@jakubcolony fixed

jakubcolony (dismissed):

ArmandoGraterol (dismissed):
Nice changes! Looking good to me 👍

PR #2305, opened 2024-04-30
Introduce Arbitrum Specific Gas Estimation

rdig:
## Intro Note

## Current PR Details

This PR refactors the Redux Transaction system to allow us to provide transaction overrides, so that we can provide the `maxFeePerGas` and `maxPriorityFeePerGas` values to all direct transactions sent over Arbitrum _(not metatransactions though)_ 

After the above, we've updated the `estimateGasCost` saga, as well as the `getGasPrices` saga, to provide transaction gas estimates for the Arbitrum networks _(both Arbitrum One and Arbitrum Sepolia)_

The gas estimation part is novel, and different from anything we've used before, and it's done by making use of `ethers`'s `getFeeData` which provides the most real-time network specific prices. 

Testing this can only be done on a Arbitrum network, by disabling the metatransactions and sending a transaction

Screenshot from 2024-04-30 17-18-52

Then, at the same time _(delaying will not provide correct results)_, with your metamask wallet connected to the same Arbitrum network _(either One or Sepolia)_ instantiate a `ethers` provider _(`BrowserProvider` is the easiest)_ and call `getFeeData()`. The values you receive should be in line with what we're setting when sending the transaction.

Screenshot from 2024-04-30 17-23-00


jakubcolony (approved):
LGTM, tested a couple of transactions in QA (including creating a colony), all of which worked fine 👍

davecreaser:
LGTM 👍

davecreaser (approved):

PR #2306, opened 2024-04-30
Fix Motion Vote Reveal

### LINKED ISSUE #1790: Revealing side signature does not match the voted side signature
To replicate this, just go through a normal motion flow that goes through to the voting process:
- create a motion
- stake both sides
- vote for whatever side
- try to reveal your vote

You'll get a similar error to this: `voting-rep-secret-no-match`:
Screenshot from 2024-01-25 16-06-35

For reference I experienced this in a production environment, while trying to reveal my vote on this motion:

Screenshot from 2024-01-25 16-06-56

https://app.colony.io/privaterc5?tx=0x7ee04de3833295cfdd682323ef555f2fc38606a71b92adea6cf1014ec402863d

rdig:
## Intro Note

## Current PR Details

For a long time _(ever since we've introduced the `RetryProvider`)_ revealing your motion vote was broken, but since no one really tested the Gnosis production deployment, people didn't really notice _(this only affects you when you're not in dev mode, and using Metamask)_

This is combination problem. It's due to the way we determine voting sides, and the user of the `RetryProvider`. We cannot know which side the user actually voted for, so in order to know which side to reveal, we basically simulate _(by estimating gas)_ on both sides, and the one that doesn't fail _(using a `try`/`catch`)_, is the side the user voted for. Now the `RetryProvider` comes into play, as it's designed to retry in the event a call to the chain actually fails _(which includes gas estimates as well)_, meaning, our try/catch logic stopped working, as the catch block was never called.

The fix for this required us bypassing the `RetryProvider`, for this saga only, and use the "standard" one as to be able to actually make it fail.

Testing is fairly strait forward, just create a motion, get it to the voting stage, vote, then reveal. All while using Metamask as you wallet. 

Screenshot from 2024-04-30 17-52-38
Screenshot from 2024-04-30 17-52-47
Screenshot from 2024-04-30 17-53-15
Screenshot from 2024-04-30 17-53-02

Fixes #1790

### DIFF
--- a/src/redux/sagas/motions/revealVoteMotion.ts
+++ b/src/redux/sagas/motions/revealVoteMotion.ts
@@ -1,6 +1,12 @@
-import { ClientType, type AnyVotingReputationClient } from '@colony/colony-js';
-import { utils } from 'ethers';
+import {
+  ClientType,
+  type AnyVotingReputationClient,
+  Extension,
+} from '@colony/colony-js';
+import { utils, providers } from 'ethers';
 import { call, put, takeEvery } from 'redux-saga/effects';
+
+import { GANACHE_LOCAL_RPC_URL, isDev } from '~constants/index.ts';
 
 import { ActionTypes } from '../../actionTypes.ts';
 import { type AllActions, type Action } from '../../types/actions/index.ts';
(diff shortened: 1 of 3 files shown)


iamsamgibbs (approved):
Tested and all working as expected 👍 

https://github.com/JoinColony/colonyCDapp/assets/34915414/981e88f4-f9bf-48ad-908b-011fd91b2fe8

ArmandoGraterol (approved):
This is it! 👌 

Screenshot 2024-05-02 at 15 42 05

PR #2307, opened 2024-05-01
Make sure user is authenticated before sending a Transaction

rdig:
> ## Disclamer
>
>
>

## Current PR Details

This is a simple PR that adds the `authenticateWallet` call inside the `ActionForm` to ensure the user is authenticated before sending a transaction _(and potentially saving metadata)_

NOTE: _Due to the way the `createTransactionChannel` saga logic is set up I was unable to add this on the redux side, so had to resort to adding it on the component side. It's not really ideal, but it's good enough for now_

This is need for cases where the user gets de-authenticated in the middle of their browsing session. The two most common reasons for these are: the session token expires while browsing, or the session token is removed entirely _(auth proxy restarts -- either the service crashes, or the pod auto-scales)_. To account for this, we just make a call to `authenticateWallet`, which internally has logic to check if the user is actually authenticated _(and not attempt to re-authenticate again)_, or they aren't, at which point they'll send out the authentication API call _(and present the user with a message to sign)_

## Testing

To test this, the best way is to start the auth proxy separately, so you can trigger a session token reset. Connect with Metamask, restart auth proxy, then try to send a transaction.

Screenshot from 2024-05-01 12-58-24
Screenshot from 2024-05-01 12-58-43

Before you'll be asked to sign the new transaction, you should be asked to actually sign the authentication message.

### DIFF
--- a/src/auth/index.ts
+++ b/src/auth/index.ts
@@ -70,9 +70,11 @@ export const authenticateWallet = async (): Promise<void> => {
 
     const signature = await signer.signMessage(message);
 
-    await authProxyRequest('auth', {
+    return authProxyRequest('auth', {
       method: 'POST',
       body: JSON.stringify({ message, signature }),
     });
   }
+
+  return authCheck;
 };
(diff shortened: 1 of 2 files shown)


iamsamgibbs (approved):
Tested and working as expected! Nice change.

https://github.com/JoinColony/colonyCDapp/assets/34915414/3d168972-0ad8-41ae-91f5-b79144719dcc

ArmandoGraterol (approved):
Working as expected for me too 👍 Nice work!

PR #2308, opened 2024-05-01
Fix Colony Balance being subtracted twice for each Payment

### LINKED ISSUE #2298: Balances is calculating incorrectly in the UI
## Steps to reproduce

1. Take note of the balance in a colony. (A round number will make it easier)
2. Make a payment action and execute it.
3. Check the balance again, it should show the correct balance removed.
4. Reload the page, the balance will be less 2 x payment amount.

The issue is not within domains.

## Expected behaviour
* The balance should show correctly as of what the correct balance is on chain.

## Actual behaviour

* You can see the flashing of the value change where double deducting of values is happening
value-incorrect

* Domains are showing the correct value being deducted
image

This looks to have been Introduced in this PR: https://github.com/JoinColony/colonyCDapp/pull/1922/files#diff-d51f4db8269ce4854a3d43363cd407e5d4906f9a9fbb4554e2222a66405cb86d
This is the relevant logic here - https://github.com/JoinColony/colonyCDapp/blob/1a93e6e265ebe91332b143acacd42dccc95f418f/src/components/frame/v5/pages/BalancePage/partials/BalanceTable/hooks.tsx#L46

rdig:
> ## Disclamer
>
>
>

## Current PR Details

This fixes a issue where an amount being payed out of the colony would decrease the colony's balance for twice that amount.

This was due to a issue where the simple payment's expenditure _(one tx payment's expenditure)_ was not being updated properly, and showing a negative value. 

Combine that with the balance being fetched by lambdas with the payment value already subtracted, and the expenditure balance "added" to the (already subtracted) balance, and presto change-o you have double the amount removed

Note that the relevant, and actual code fix was done on the ingestor side via https://github.com/JoinColony/block-ingestor/pull/215 this PR just updates the docker commit hash to the relevant one, and it's purpouse is to aid in testing

## Testing

Create a fresh colony, mint a memorable amount _(`100` or `1000`)_. Create a payment for yet another memorable amount _(`10` or something)_.

Look up the expenditure in the database via the graphql explorer, it should have the balance `0`. Any other value in there means this is still broken. _(for one OneTX Payments, other expenditure types, can have non-zero values)_

Screenshot from 2024-05-01 13-32-14

Screenshot from 2024-05-01 13-33-58

Look up the colony's balance table, which should should your initial minted amount, minus the payment amount _(if it shows double the payment amount, then this is still broken)_

Screenshot from 2024-05-01 13-35-23

Fixes #2298

### DIFF
--- a/docker/colony-cdapp-dev-env-block-ingestor
+++ b/docker/colony-cdapp-dev-env-block-ingestor
@@ -1,6 +1,6 @@
 FROM colony-cdapp-dev-env/base:latest
 
-ENV BLOCK_INGESTOR_HASH=824b795fa01c03510908d44061a539fffe0f5e7f
+ENV BLOCK_INGESTOR_HASH=23132accc3c3b792042c5f1efc04a4a8ea849d4c
 
 # Declare volumes to set up metadata
 VOLUME [ "/colonyCDapp/amplify/mock-data" ]


iamsamgibbs (dismissed):
Tested and working! 👍 

Minted 100 tokens, paid 10, balance shows 0 and balance table shows correct amount removed.

Screenshot 2024-05-01 at 12 07 09

Screenshot 2024-05-01 at 12 07 19

ArmandoGraterol (approved):
Same as Sam! Both the expenditure and the table are showing the correct amounts 💯

PR #2274, opened 2024-04-23
Update UI of the wrong chain connected tooltip

davecreaser:
## Description

- Update the UI of the wrong chain connected warning & tooltip to match the Figma designs.

## Testing

* Step 1. Connect to a different network in metamask (not Ganache and not Arbitrum)
* Step 2. Connect the metamask wallet to Colony
* Step 3. Check and hover over the network pill in the top right. It should display the correct warnings and the correct tooltip as per designs.

## Diffs

**Changes** 🏗

* Update the UI of the wrong chain connected warning & tooltip

Resolves #2218

PR #2318, opened 2024-05-02
Visual indicator if user's wallet is connected to the wrong network

rdig:
> ## Disclamer
>
>
>

## Current PR Details

This PR implements a visual indication _(network pill lights up)_, if the user's wallet is connected to a network other than the current app is deployed to. This was done in a effort to prevent users reporting "transaction issues" while not sending the transaction on the correct chain to begin with.

This contains both the initial component logic, and the styles done by @davecreaser as part of #2274, but which were merged into the `master-arbitrum` branch, since this piece of logic only existed on that branch.

## Testing

Run the dev environment locally, on "ganache" _(hardhat under the hood)_ and connect your metamask wallet, while being connected to anything other than the local network

Screenshot from 2024-05-02 19-36-28

### DIFF
--- a/src/components/common/Extensions/UserNavigation/UserNavigation.tsx
+++ b/src/components/common/Extensions/UserNavigation/UserNavigation.tsx
@@ -1,7 +1,9 @@
 import { Cardholder, GearSix, List, X } from '@phosphor-icons/react';
 import React, { type FC } from 'react';
+import { defineMessages } from 'react-intl';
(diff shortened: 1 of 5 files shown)


ArmandoGraterol (approved):
I've tested this before, still working as expected 👌 

Screenshot 2024-05-02 at 15 54 06

jakubcolony (approved):
LGTM 👍 

image

PR #2320, opened 2024-05-03
Fix Colony Creation with Non Colony Tokens

### LINKED ISSUE #2264: Cannot create a colony with an existing token
## Steps to reproduce

1. Create a new colony on production with this existing token - https://arbiscan.io/token/0xf929de51d91c77e42f5090069e0ad7a09e513c73
2. Creation will fail on the first transaction of the create colony step.

## Expected behaviour

* You should be able to create a colony with an existing token, including the provided one.
* You should get feedback in the UI if there is an error.

## Actual behaviour

* You are not able to create a new colony with the provided token.
* There are no errors or feedback in the UI, only errors in the console.

image

rdig:
## Intro Note

## Current PR Details

Code for this fix was implemented while pair programming with @area 

This PR introduces a fix to allow colonies to be created with non colony network created tokens _(basically any random token out there, as long as it's ERC-20 standard compliant, however lax that standard might be)_

The problem here is that we were using another "trick" to determine what kind of token we were dealing with, by using a try/catch block, as similarly to the issue we experience with #2306, the `RetryProvider` would kick in and wouldn't actually allow the call to fail.

The `locked()` method on a token is implemented "only" by us, on tokens created via our contracts. And we assume that all tokens who have this method available are `Colony` tokens, while the ones that don't are `ERC-20` tokens.

Since the call to `locked()` failed for `ERC-20` tokens, the `RetryProvider` would kick in trying to retry that call, meaning that piece of code would never fail, and the colony creation saga would time out.

The fix here, involved adding exceptions in the `RetryProvider` for calls we expect to fail, and which we wish the `RetryProvider` would bypass when attempting to retry. The list currently includes only three methods but will most likely grow with time.

## Testing

This is not really strait forward to test, as there's no easy way to deploy a non colony token to the local dev environment.

We actually developed / fixed this while working against QA.

The easiest way to test this for you is to actually deploy this branch to QA and find a random ERC-20 token on Arbitrum Sepolia, and create a colony with that.

Fixes #2264

### DIFF
--- a/src/components/common/Onboarding/wizardSteps/CreateColony/TokenSelector.tsx
+++ b/src/components/common/Onboarding/wizardSteps/CreateColony/TokenSelector.tsx
@@ -52,6 +52,15 @@ const TokenSelector = ({
 
   useEffect(() => {
     if (isFetchingAddress) {
+      setIsLoading(true);
+    }
+    if (data) {
+      setIsLoading(false);
+    }
+  }, [data, isFetchingAddress, setIsLoading]);
+
+  useEffect(() => {
+    if (isFetchingAddress) {
       return;
     }
 
--- a/src/components/common/Onboarding/wizardSteps/CreateColony/TokenSelectorInput.tsx
+++ b/src/components/common/Onboarding/wizardSteps/CreateColony/TokenSelectorInput.tsx
@@ -4,7 +4,9 @@ import { useFormContext } from 'react-hook-form';
 import { defineMessages, useIntl } from 'react-intl';
 
 import { DEFAULT_NETWORK_INFO } from '~constants';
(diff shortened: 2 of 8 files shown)


jakubcolony (approved):
Tested the whole flow with USDT (https://sepolia.arbiscan.io/token/0x9aa40cc99973d8407a2ae7b2237d26e615ecafd2), all looking good 👍 Nice work

https://github.com/JoinColony/colonyCDapp/assets/112586815/c657947b-0629-4af7-8676-aff647a442f1

davecreaser (approved):
Tested with LINK: https://sepolia.arbiscan.io/token/0xb1d4538b4571d411f07960ef2838ce337fe1e80e

And all working fine 📈

PR #2322, opened 2024-05-03
Introduce Searchable Contributors Query

rdig:
## Intro Note

## Current PR Details

This introduces an `OpenSearch` for the `ColonyContributors` query, since, making it searchable, allows us to get the total items count, at the initial fetch time, even if we have pagination for that query.

Doing so allows us to display, in this case members count, after the initial load, in places like dashboard widgets, all at once, while the actual user data loads in the background.

## Testing

Testing this is quite strait forward, you'll just have to find a way to create a lot more users for one of the colonies, more than `100` since that's the pagination value _(realistically, more than 300, so you can actually see the effect)_

After refreshing you should see all the users you have in the database on your dashboard members widget count.

Screenshot from 2024-05-03 18-16-35
Screenshot from 2024-05-03 18-16-54
Screenshot from 2024-05-03 18-17-19


davecreaser (approved):

iamsamgibbs (approved):
All looks good to me!

Screenshot 2024-05-08 at 12 01 29
Screenshot 2024-05-08 at 11 59 24
Screenshot 2024-05-08 at 11 59 19
Screenshot 2024-05-08 at 11 59 14

PR #2324, opened 2024-05-06
fix: bindings not found

joanna-pagepro:
## Description

There's an error when I tried to run `npm run frontend`
image

The issue and suggested solutions can be found here: https://github.com/vitejs/vite-plugin-react-swc/issues/74

## Testing

I'm using Mac Sonoma 14.4.1

## Diffs

**New stuff** ✨

* `@swc/core` package

**Changes** 🏗

* 

**Deletions** ⚰️

*

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -117,6 +117,7 @@
         "@storybook/blocks": "^8.0.0",
         "@storybook/react-vite": "^8.0.0",
         "@storybook/test": "^8.0.0",
+        "@swc/core": "^1.4.17",
         "@tailwindcss/container-queries": "^0.1.1",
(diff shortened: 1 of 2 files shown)


ArmandoGraterol (approved):
Working as expected as far as I can notice. I also came across this error (Ventura OS) but my solution was to remove the `node_modules` + `package-lock`

joanna-pagepro:
@ArmandoGraterol The same solution works for me as well, but there were many changes in `package-lock` that could affect other packages, not related to this issue, and potentially break something else. I may be wrong, but if this solution is good let me know.

chmanie (approved):
This fixes it for me, too. Thanks @joanna-pagepro!

PR #2326, opened 2024-05-06
Disable Token Actions if Not Available

### LINKED ISSUE #2163: Hide "Mint tokens" action when using existing token
## Steps to reproduce

1. Create a new colony with an existing token.
2. Create a new action.
3. Select the `Mint tokens` action.
4. Confirm to mint, the transaction won't be triggered but the UI will be stuck in pending.

## Expected behaviour

* The `Mint tokens` action should be hidden if the colony does not have permissions to mint the token.
* This includes any where else where the Mint token action can be triggered, i.e. in the side menu and finances meatball menu.
* `Unlock tokens` should also be hidden.

image

image

## Actual behaviour

* You are able to Mint an existing token.
* When minting an existing token you don't have permissions for, the transaction will fail ungracefully.

image

rdig:
## Intro Note

## Current PR Details

This PR adds some checks to disable the _"Mint Token"_ and _"Unlock Token"_ actions if the colony doesn't have permissions to do so. This usually happens when you are using an external token, not one created by you, when creating the colony.

## Testing

Testing this is quite easy, just create a new colony, but instead of creating a new token along side it, just bring one from an existing colony _(`planex` for example)_. 

- The actions to unlock and mind should be grayed out in your action selector
- The "mint" option in the Balance page's kebab menu should not be visible
- The "mint tokens" link should not be visible in the "third menu" 

Screenshot from 2024-04-08 13-18-13

Resolves #2163

### DIFF
--- a/src/components/frame/Extensions/layouts/hooks.tsx
+++ b/src/components/frame/Extensions/layouts/hooks.tsx
@@ -106,7 +106,7 @@ export const useCalamityBannerInfo = (): UseCalamityBannerInfoReturnType => {
 
 export const useMainMenuItems = (hasTransactionId: boolean) => {
   const {
-    colony: { metadata },
+    colony: { metadata, status },
   } = useColonyContext();
 
   const {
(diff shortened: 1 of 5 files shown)


ArmandoGraterol (approved):
Working as expected 👍 

Screenshot 2024-05-06 at 16 29 57

Not related to this PR I think but I found this bug clicking the `Transfer Funds` item on the finances page:

Screenshot 2024-05-06 at 16 29 40

Reproducible through this menu:

Screenshot 2024-05-06 at 16 33 01

iamsamgibbs (approved):
All looks good to me! Tested and working as expected.

Screenshot 2024-05-07 at 10 14 07

Screenshot 2024-05-07 at 10 15 45

Screenshot 2024-05-07 at 10 22 23

PR #2327, opened 2024-05-06
Allow Domain Creation With Empty Descriptions

rdig:
## Intro Note

## Current PR Details

While the action sidebar validator allows you to set an empty domain (team) description when creating a new domain action, the schema did not actually support this, and would actually fail silently and not create to domain, even if the transaction went through.

This was live in production for a long time and nobody actually caught it.

This PR fixes that, but actually supporting this on the schema / db side as well.

## Testing

Testing is easy, just create a new domain and leave the description empty. You should not see any errors in the console (regarding this), as well as you should be able to query for it using the graphql playground where it should show up without a value as it's description


ArmandoGraterol (approved):
LGTM 👍 

Screenshot 2024-05-06 at 16 48 45

iamsamgibbs (dismissed):
All seems good to me!

Screenshot 2024-05-07 at 10 26 37

The empty gap on the completed action screen looks a bit odd to me, but that is probably just not being used to seeing it there. I don't believe any other actions have optional fields (except the description which does not show on the completed screen if empty) so there is not another example to compare against.

Screenshot 2024-05-07 at 10 26 00

PR #2330, opened 2024-05-07
Fix: Fund expenditure motion does not get created in the DB

### LINKED ISSUE #2302: Fund expenditure motion does not get created in the DB
## Steps to reproduce
Only possible on `feat/advanced-payments-2` branch, attempt to call `fundExpenditureMotion` saga with a correct payload.

## Expected behaviour
Fund expenditure motion is created in the DB.

## Actual behaviour
block-ingestor picks up the motion but does not create it in the DB.

jakubcolony:
## Description

This PR fixes a bug outlined in the issue where multicall motions funding an expenditure would not get picked up by block-ingestor. This was due to an overly simplified implementation of handling multicall motions. While my fix is far from getting it where it should be, it fixes the issue for the existing and upcoming UI.

## Testing

TBC

Resolves #2302

### DIFF
--- a/amplify/backend/api/colonycdapp/schema.graphql
+++ b/amplify/backend/api/colonycdapp/schema.graphql
@@ -2484,6 +2484,23 @@ type ColonyMotion @model {
   Edited expenditure slots associated with the motion, if any
   """
   editedExpenditureSlots: [ExpenditureSlot!]
+
+  """
+  In case of multicall motion funding an expenditure, array containing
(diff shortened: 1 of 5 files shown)


iamsamgibbs (approved):
Looks good to me! Code looks good and tested and working as expected.

Screenshot 2024-05-08 at 13 55 47

rdig (approved):
Seems all good

Issue #2310, opened 2024-05-01
Add date of transaction to permission card


arrenv:
### Story

As a USER, I want to be able to see the date and time when a transaction was executed.

### Description

In order to provide the required information for transactions when the permissions decision method is used, we need to include the date the transaction and action was executed, to do this, we will add an additional line to the permission card.

The date the action was created, in this date format:
    - For today: Today at XX:XXam/pm
    - For yesterday: Yesterday at XX:XXam/pm
    - Older: 12 Mar 2022 at XX:XXam/pm
    - The date and time should be best guess for the users timezone, and fallback to UTC+0

### Design

[Figma link](https://www.figma.com/file/l1dOM5qiQYwF0ElvKDqqjg/Design-System---Colony-v3?type=design&node-id=3701-19285&mode=design&t=v0nITJZ06JWwVgMc-4)

![image](https://github.com/JoinColony/colonyCDapp/assets/33682027/7fd9d4a9-297d-40a1-9e98-ab6d84cffcf1)

### This issue is complete when...

- [ ] You are able to see the date of the transaction in each action using permissions.

PR #2336, opened 2024-05-08
feat: added createdAt information

CzarekDryl:
## Testing

* Create any action using permissions

**New stuff** ✨

* Changed box description
* Added information when action was created
image

Resolves [2310](https://github.com/JoinColony/colonyCDapp/issues/2310)


melyndav (dismissed):
Looks great @CzarekDryl thanks for the update to this component. Nice work!

arrenv (dismissed):
Thank you @CzarekDryl, looks good to me :)

src/components/v5/common/ActionSidebar/partials/ActionSidebarContent/partials/utils.ts
```diff
@@ -0,0 +1,31 @@
+import { format, isToday, isYesterday } from 'date-fns';
+
+export const formatDate = (createdAt: string | undefined) => {
```
rdig, line 3: There's no need for this util as `react-intl` already exports a `FormattedDate` component which you can use directly, and already contains all the formatting options that you might need.

It also exports `FormattedDateParts` which gives you as much granular control as you'd want.

We already use it here: https://github.com/JoinColony/colonyCDapp/blob/feat/15899525-created/src/components/common/Extensions/UserHub/partials/StakesTab/partials/StakeItem.tsx#L28

rdig (changes requested):
The change itself is good, but please don't add redundant logic, when we already have ways of dealing with date and time formatting.

src/components/v5/common/ActionSidebar/partials/ActionSidebarContent/partials/utils.ts
CzarekDryl: @rdig I wasn't sure how FormattedDateParts works. Take a look to see if this is possible.

melyndav (approved):
Re-approving this one @CzarekDryl

rdig (approved):
All good now. 

Would have personally not had created the extra `formatData` fn, and just rendered everything in component directly, but It can go in as it is.

PR #2338, opened 2024-05-08
Feat: Edit streams using permissions

### LINKED ISSUE #2213: Streaming Payments API: Editing streams using permissions
Parent issue: https://github.com/JoinColony/colonyCDapp/issues/2208

- New saga needs to be implemented
- Relevant block-ingestor handlers need to be implemented 
- Editing via permissions doesn't require a corresponding colony action to be created
- Changes made to the stream need to be stored, as pairs of old and new values

iamsamgibbs:
[block-ingestor counterpart](https://github.com/JoinColony/block-ingestor/pull/220)

[colonyCDappAuthProxy counterpart](https://github.com/JoinColony/colonyCDappAuthProxy/pull/19)

## Description

* Create a saga to handle editing streaming payments
* Create relevant handlers on the block-ingestor

This PR adds a editStreamingPayment saga with a multicall for setTokenAmount, setStartTime and setEndTime.

The EndTimeSet event is already handled on the block-ingestor, this PR adds a StartTimeSet handler and tweaks the PaymentTokenUpdated handler to work with the new contract changes.

A handler has been added to auth proxy for the UpdateStreamingPaymentMetadata mutation which is used to update the endCondition and limit - it checks the initiator has permissions in either the domain the streaming payment was created in or the root domain.

## Testing

* Step 1 - Install the streaming payments extension
* Step 2 - Create a new streaming payment. Enter a transaction amount greater than 0, select 'Start one week from now' as the start time and 'End three weeks from now' as the end time. (Note it is not possible to edit the start time of a streaming payment which has already started.)

Screenshot 2024-06-24 at 12 20 40

* Step 3 - Wait a few seconds and click the `Use latest streaming payment native ID` button

Screenshot 2024-06-24 at 12 26 19

(You can also run the following query if you want to confirm the values:)

```
query MyQuery {
  listStreamingPayments {
    items {
      nativeId
      startTime
      endTime
      amount
      interval
      metadata {
        limitAmount
        endCondition
      }
    }
  }
}

```

* Step 4 - Change the transaction amount, set the start time to `Start two weeks from now`, the end time to `End four weeks from now` and change the interval. Ensure all the related toggles at the bottom are selected.

Screenshot 2024-06-24 at 12 27 36

* Step 5 - Click edit, then click refresh and confirm the values have changed:

Screenshot 2024-06-24 at 12 29 21

(Further testing, try updating different combinations of amount / interval / start time / end time)

* Step 6 - Select `When cancelled` as the end condition, and select the `Update end condition` toggle. Click edit, then refresh and note the updated values.

Screenshot 2024-06-24 at 12 30 35

Screenshot 2024-06-24 at 12 31 01

(Further testing, try updating the amount / interval / start time with the 'When cancelled' end condition)

* Step 7 - Select the 'Limit reached' end condition, set a limit. Ensure both related toggles are selected. Click edit, then refresh and confirm the values have updated.

Screenshot 2024-06-24 at 12 33 52

Screenshot 2024-06-24 at 12 34 53

* Step 8 - Change the amount and select the 'Update amount' toggle. We expect that changing the amount will also change the end time. The limit should remain unchanged.

Screenshot 2024-06-24 at 12 35 58

Screenshot 2024-06-24 at 12 36 16

(Further testing, try updating different combinations of amount / interval / start time / end time)

## Diffs

**New stuff** ✨

* Added editStreamingPayment saga
* Added StartTimeSet handler to block-ingestor
* Added UpdateStreamingPaymentMetadata mutation handler to auth proxy

**Changes** 🏗

* Tweaked PaymentTokenUpdated handler

Resolves #2213


rdig (changes requested):
I reviewed the code and it looks as expected, however I can't validate your testing scenario, as I'm getting the exact same start/end time + condition even after edit

Initial:
Screenshot from 2024-06-13 19-18-43

Change amount and edit:
Screenshot from 2024-06-13 19-22-25

Screenshot from 2024-06-13 19-22-38

I've rejected this until I can figure out what's going on, but I have a feeling I'm doing something wrong

iamsamgibbs:
@rdig My bad, I wrote enter the ID into the `Expenditure ID` field rather than `Streaming Payment ID` field. I've added a screenshot to the testing instructions to make it clearer.

Screenshot 2024-06-13 at 17 44 50

rdig:
Tested it like that as well, and getting a gas estimation error.

Created new streaming payment of 500
Screenshot from 2024-06-13 19-52-32

Tried to edit
Screenshot from 2024-06-13 19-53-19

Transaction fails to send
Screenshot from 2024-06-13 19-53-04

And console shows an estimation error
Screenshot from 2024-06-13 19-53-30

jakubcolony (changes requested):
Great work on this PR @iamsamgibbs!

Apologies for only putting the detailed scenarios in testing as opposed to the specs but hopefully it will make it easier for you to make the necessary changes.

Changing start time:
- Fixed time/when cancelled ✅

https://github.com/JoinColony/colonyCDapp/assets/112586815/95cbcaef-e44d-4c37-8544-bdac9cd381e5

- Limit reached ❌

https://github.com/JoinColony/colonyCDapp/assets/112586815/a9bf26ce-9378-45b8-a07e-69007f522684

**The end time should update accordingly (this is how limit works on the contract, by making the stream last exactly the time it takes to stream the limit amount)**

Changing end time:

- Fixed time/when cancelled ✅

https://github.com/JoinColony/colonyCDapp/assets/112586815/e59d2144-df1a-44e0-aeb2-290268137d54

- Limit reached ❌
**Saga should prevent modifying the end time manually when Limit reached condition is chosen**

Changing limit ❌

https://github.com/JoinColony/colonyCDapp/assets/112586815/0485ab5e-be20-4b5d-80a4-189f51fea496

**The end time should update in line with the new limit**

Changing amount: 

- Fixed time/when cancelled ✅

https://github.com/JoinColony/colonyCDapp/assets/112586815/1ec186d7-6089-4d45-a088-6fd9228a3991

- Limit reached ❌
**The end time should update accordingly**

Changing interval:

- Fixed time/when cancelled/limit reached ❌

https://github.com/JoinColony/colonyCDapp/assets/112586815/fb263ab4-3a34-4c56-8ac0-b8004c5761dc

**The interval did not update in the DB**

Changing token address:

- Fixed time/when cancelled/limit reached ❌

https://github.com/JoinColony/colonyCDapp/assets/112586815/7c967386-b451-4069-a436-d6caf0b47c7e

**The token address did not update in the DB**

I think it will be much easier not to allow changing the end condition, as that will have various consequences depending on which conditions you switch between. There's also no product requirement to facilitate that or a way to do it on the contracts.

I also noticed the saga requires to pass in the start time, amount, decimals and interval, even if we are not editing those properties. Could we make them optional?

While testing, I realised someone could edit the start/end time for a stream with "Limit reached" end condition but there is no code in block-ingestor that would handle updating the limit on the metadata (so it will be incorrect). We should probably add some code to do that, and maybe that way we won't need to update it directly in the saga.

iamsamgibbs:
> I think it will be much easier not to allow changing the end condition, as that will have various consequences depending on which conditions you switch between. There's also no product requirement to facilitate that or a way to do it on the contracts.

Can you confirm this is definitely the case @jakubcolony / @arrenv ?

The specs on notion list "ends" as an editable field, and the figma design shows the end condition as part of the "ends" field so it isn't super clear.

Screenshot 2024-06-24 at 11 38 49

Screenshot 2024-06-24 at 11 39 26

Removing this would simplify things a little, although I'd still need to calculate the end time dependant on the end condition anyway so wouldn't massively change things.

iamsamgibbs:
> I also noticed the saga requires to pass in the start time, amount, decimals and interval, even if we are not editing those properties. Could we make them optional?

Done! (I figured the form will have all the values available to it anyway to pass in, but doesn't hurt to make them optional).

> While testing, I realised someone could edit the start/end time for a stream with "Limit reached" end condition but there is no code in block-ingestor that would handle updating the limit on the metadata (so it will be incorrect). We should probably add some code to do that, and maybe that way we won't need to update it directly in the saga.

I've added this to the block-ingestor, I've had to handle it on `setStartTime`, `setEndTime` and `paymentTokenUpdated` as in theory any one of these could signify a change in limit without the others being triggered so in rare cases the metadata will be updated 3 times back to back but the final update should always be the correct one.

iamsamgibbs:
@jakubcolony This is ready for testing again (once you're back!) but please feel free to address the above questions first and I can make changes if required before you review - can jump on a call to discuss in more detail if that would help

jakubcolony:
@iamsamgibbs 

> The contracts do not allow for editing the token address: 

My bad, that makes sense 👍 

> Question - Is it feasible to drop limitAmount from the metadata model and just calculate it in the UI on the frontend? Seeing as it is purely a UI abstraction anyway and can be calculated from the startTime, endTime, interval and amount. Then we don't need to worry about updating it anywhere and we just calculate it from the other values when we display it.

I think that makes sense, let's go for it! When you remove it, could you open an issue to add a helper/hook to calculate the limit amount in the UI as it would be something needed by Pagepro?

iamsamgibbs:
@jakubcolony This is ready to re-review with limit amount removed from the streamingPaymentMetadata.

I've cherry-picked one of your schema changes as it is on the block-ingestor branch as was causing me issues getting block-ingestor running.

Let me know how you'd like to approach colonyCDappAuthProxy#19 too - happy to move it to a lambda, or leave it in the auth proxy.

src/redux/sagas/expenditures/editStreamingPayment.ts
```diff
@@ -0,0 +1,291 @@
...
+
+  const batchKey = 'editStreamingPayment';
+
+  const txChannel = yield call(getTxChannel, meta.id);
+
+  try {
+    const { decimals: tokenDecimals } =
+      tokens?.items.find(
+        (token) => token?.token.tokenAddress === streamingPayment.tokenAddress,
+      )?.token || {};
```
jakubcolony, line 74: There's a util `getSelectedToken` you could use here

src/redux/sagas/expenditures/editStreamingPayment.ts
```diff
@@ -0,0 +1,291 @@
...
+          adminChildSkillIndex,
+          streamingPayment.nativeId,
+          time,
+        ]),
+      );
+    };
+
+    // @NOTE: The multicall will fail if we set the new start time to be before the current end time
+    // Or the new end time to be before the current start time
+    // If both are true, then first set the endTime to the maximum possible end time
```
jakubcolony, line 193: Great catch!

src/utils/streamingPayments.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  const originalAmount = BigNumber.from(amount).div(
+    BigNumber.from(10).pow(tokenDecimals),
+  );
+
+  const limit = BigNumber.from(endTime)
+    .sub(startTime)
+    .mul(originalAmount)
+    .div(interval);
+  return limit.toString();
```
jakubcolony, line 31: I feel like this could return a wei value, and leave the conversion down to UI, as this creates a nice separation of concerns and simplifies the calculations

jakubcolony (approved):
This looks great @iamsamgibbs, a lot of effort went into this task.

The new temp UI made testing so much simpler.

All the previously failing cases work, sorry for not double checking but you were absolutely right with token address not being editable.

The limit calculation works fine as well, nice idea with removing it from the metadata!

I have also confirmed with @arrenv we want to allow changing end condition, and it works flawlessly in this PR 🙂 

image


I left a few minor comments, but overall very nice job 💪

iamsamgibbs:
@jakubcolony I've made the suggested changes :)

jakubcolony (approved):
LGTM 👍

rdig (approved):

src/redux/sagas/utils/getEndTimeByEndCondition.ts
```diff
@@ -0,0 +1,51 @@
+import { BigNumber } from 'ethers';
+
+import { StreamingPaymentEndCondition } from '~gql';
+
+// Maximum uint256
+export const TIMESTAMP_IN_FUTURE = BigNumber.from(2).pow(256).sub(1);
```
davecreaser, line 6: So, although it was me who originally made this number, I've since realised we can just use `constants.MaxUint256` from ethers 😂 

Not really a "change request" but just a tiny detail!

davecreaser (approved):
Looking really good @iamsamgibbs, super nice work! 🎉 

Tested your steps, and everything worked as expected:
Screenshot 2024-07-12 at 14 23 01
Screenshot 2024-07-12 at 14 24 10
Screenshot 2024-07-12 at 14 24 19
Screenshot 2024-07-12 at 14 25 08
Screenshot 2024-07-12 at 14 26 09
Screenshot 2024-07-12 at 14 28 17

mmioana (approved):
Amazing work on this one @iamsamgibbs 🎉 

I managed to go through all testing steps flawlessly
Screenshot 2024-07-12 at 15 28 13
Screenshot 2024-07-12 at 15 30 49
Screenshot 2024-07-12 at 15 31 07

Screenshot 2024-07-12 at 15 32 20
Screenshot 2024-07-12 at 15 33 07
Screenshot 2024-07-12 at 15 33 43

However, only the GraphQL query from the testing steps should be updated as there is no `limitAmount` in the metadata section
Screenshot 2024-07-12 at 15 29 38
Screenshot 2024-07-12 at 15 40 22

PR #2127, opened 2024-04-03
Get user transactions from database instead of redux state

chmanie:
## Description

Sorry for another big one :(

This PR moves the transaction state to Apollo, from redux. It _does not_ remove sagas but it enables a path forward to refactor this gradually. All sagas _should_ still work, except that the transaction state is read from the Apollo cache / the database as opposed to redux.

Redux as a whole also wasn't removed yet, as some of the sagas depend on other parts of the state, still (like gas prices estimation). This can be done in a follow-up PR.

Another thing that isn't in scope is the fix of the transactions list paging (didn't work before), for that I'm going to open yet another PR.

I used optimistic UI updates for the transaction updates so that changes should be reflected as fast as before. 

## Testing

Basically everything should work as before. Please create some transactions and/or bigger transaction groups to test this.

## Diffs

**Changes** 🏗

* All redux transaction state now lives in Apollo

**Deletions** ⚰️

* Lots of transaction related redux state management boilerplate

Resolves #2171

PR #2341, opened 2024-05-09
Fix transaction paging

### LINKED ISSUE #2332: Paging of transactions is broken
## Steps to reproduce

1. Have more than 20 transactions in the transactions tab
2. Scroll down

## Expected behaviour

More transactions should load (paging)

## Actual behaviour

Nothing really happens

chmanie:
## Description

This PR fixes the transaction list paging. It's based on #2127 and should probably only be reviewed after that one is merged.

Here we fix the loading by actually passing in the correct element into Apollo's pagination loader. As the old solution was quite cumbersome and brittle, I completely rewrote the caching to facilitate Apollo functions instead of custom written ones. This included adding a cache query-field merge function.
Also, I am now using a more react-y approach for setting the root element for the intersection observer. It's a little bit better as the code is in the same place as the actual element, but for a proper solution something like [`react-intersection-observer`](https://www.npmjs.com/package/react-intersection-observer) should be used.

To actually test this properly (also for the reviewers), I added the creation of transaction entries in the database for the `create-data` script.

## Testing

Re-create your dev environment after checking out this branch.

Go to the UserHub, click Transactions and scroll all the way down.

## Diffs

**New stuff** ✨

- Transactions are now created in the `create-data` script.

**Changes** 🏗

* Page loading was fixed and the state/caching strategy for the pagination was completely rewritten. Now there's a loading spinner showing at the bottom of the transactions, showing when there's more to load.

**Deletions** ⚰️

* Weird insersection observer code

Closes #2332


scripts/create-data.js
```diff
@@ -369,15 +424,42 @@ const mintTokens = async (
 
   // mint
   const mintTokens = await colonyClient.mintTokens(amount);
-  delay();
+  await delay();
```
chmanie, line 427: `delay`s without `await` don't do anything

chmanie:
I found an issue with this and it seems it has to be done properly. Putting back in draft state

chmanie:
This is considered done now and can be reviewed, once #2127 is merged.

iamsamgibbs (commented):
This is working great!

https://github.com/user-attachments/assets/11cf5ea3-c03b-496a-b1fd-df3ae4076737

I made some changes to `create-data.js` on master recently so looks like it needs updating again here, but otherwise is working nicely!

mmioana (approved):
Really cool stuff, @chmanie 🎉 

https://github.com/user-attachments/assets/e040a16f-6bbe-42d5-90bb-81f5b0cfdc3f

iamsamgibbs (approved):
🎉 Still working!

https://github.com/user-attachments/assets/fbc9f348-2992-45fb-a479-843b27bb69de

PR #2353, opened 2024-05-10
Fix: useHasNoDecisionMethods hook causes edit colony details completed action to crash

### LINKED ISSUE #2278: Error with completed transaction after editing colony details
## Steps to reproduce

This was experienced in production, this is the relevant transaction: https://app.colony.io/shapeshift?tx=0x6ed4ce9f7aec89a690b3b554553cac1fcccb67822887c9a6313f5ee3d6095bdc

This was the process taken:
1. Edit the details of the colony.
2. Use the permissions decision method.
3. Execute the action.

Let me know if you need an invite code to the colony.

## Expected behaviour

* You should be able to visit the action in the side panel.

## Actual behaviour

* Opening the action crashes the app with a white screen with the error as of the screenshot below.

image

iamsamgibbs:
## Description

The CDapp was crashing when loading the completed action for edit colony details.

This PR adds a check to the `useHasNoDecisionMethods` hook so that if `watch` is undefined it will return false. `watch` will only ever be undefined if the `useHasNoDecisionMethods` hook is called outside of a form (as is the case on the edit colony details completed action screen).

## Testing

Create a new edit colony details action. The completed action screen should show without issue.

## Diffs

**Changes** 🏗

* The `useHasNoDecisionMethods` hook returns false if `watch` is undefined.

Resolves #2278

### DIFF
--- a/src/components/v5/common/ActionSidebar/hooks/permissions/useHasNoDecisionMethods.ts
+++ b/src/components/v5/common/ActionSidebar/hooks/permissions/useHasNoDecisionMethods.ts
@@ -22,7 +22,12 @@ const useHasNoDecisionMethods = () => {
   const { user } = useAppContext();
   const { isVotingReputationEnabled } = useEnabledExtensions();
 
-  const { watch } = useFormContext();
+  const { watch } = useFormContext() || {};
+
+  if (!watch) {
+    return false;
+  }
+
   const actionType = watch(ACTION_TYPE_FIELD_NAME);
 
   if (!user) {


jakubcolony (approved):
Nice fix, the completed action view is not crashing anymore:

https://github.com/JoinColony/colonyCDapp/assets/112586815/972b1bdc-9201-4751-94cf-5749682c5d02

melyndav (approved):
Back to working as expected, thanks for the quick fix and fast pick up @iamsamgibbs!

PR #2360, opened 2024-05-10
Remove reputation weighted step validation

### LINKED ISSUE #2284: Remove incorrect validation on Reputation Weighted extension set up
## Steps to reproduce

1. Install the Reputation Weighted extension.
2. On the settings screen, set the value for "Required Stake" to a fraction of a percent.
3. Click the Enable button.
4. The form will validate incorrectly.

## Expected behaviour

* The correct validation design should appear.
* The only validation should be that it is a positive number.
* This should apply for other fields with the same issue.
* This is actually being caused by the step value in the field, this can either be removed or lowered as much as possible.

image

## Actual behaviour

* Incorrect validation is appearing when trying to enable the extension.

image

davecreaser:
## Description

- Remove reputation weighted step validation

## Testing

- Install the Reputation Weighted extension.
- On the settings screen, set the value for "Required Stake" to a fraction of a percent.
- Click the Enable button.
- The form should allow you to do it.

Resolves #2284

### DIFF
--- a/src/components/frame/Extensions/pages/LazyConsensusPage/consts.tsx
+++ b/src/components/frame/Extensions/pages/LazyConsensusPage/consts.tsx
@@ -94,7 +94,7 @@ export const initialExtensionContent = [
           name: 'params.totalStakeFraction',
           min: 1,
           max: 50,
-          step: 1,
+          step: 'any',
         },
         accordionItem: [
           {


marcocolony (approved):
@davecreaser Nice one, all looks good. 
Tested the following scenarios and all works:
- Decimal percentage are accepted (0.05%, 5.6%, 0.0005%, .5%)
- Negative values: Not accepted
- Any value above 50%: Not accepted
image

rdig (approved):
Good to go!

iamsamgibbs (approved):
All good! Tested the same scenarios as Marco without issue.

Screenshot 2024-05-15 at 08 49 52

Issue #2287, opened 2024-04-25
Side panel close feature for desktop & table modes 


melyndav:
Teamwork task: [Side panel close feature for desktop & table modes](https://tw.auroracreation.com/#/tasks/15899876)

## Description

Adding close feature within the side panel on desktop and tablet will enable uses to close the side panel via the header. This adds an extra close mechanism to the side panel giving greater usability control for users within the CDapp. 

The close icon should be available to the user in all states of the side panel including:

1. During the active action stage where a user is creating a new action;
2. When the transaction is completed; and
3. And when the transaction is in a motion state. 
4. When the side panel is in the full width state.

If the user selects the close icon during the active action stage and has changed content within the action type chosen, the closing modal should be triggered (which matches the current modal trigger).

For all other states, the side panel should close when selected by the user without the modal triggering. 

<img width="456" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/8d1a4146-dcc5-4398-becd-8a0307d11878">

## Figma:

Design: https://www.figma.com/file/ZQ8JrzynhZDz7yXLXCqhEC/Project-22?type=design&node-id=2517-46291&mode=design&t=DqHj2pkiZy7zPIO9-4

Component link: https://www.figma.com/file/l1dOM5qiQYwF0ElvKDqqjg/Design-System---Colony-v3?type=design&node-id=1782-12590&mode=design&t=Px8zAdUedZZyscvZ-4

melyndav:
Hi @AdrianDudko - could you please estimate on this feature and send the estimation to Arren? Thank you :)

PR #2365, opened 2024-05-13
feat: action sidebar close button

joanna-pagepro:
## Description

Adding close feature within the side panel on desktop and tablet will enable uses to close the side panel via the header. This adds an extra close mechanism to the side panel giving greater usability control for users within the CDapp.

The close icon should be available to the user in all states of the side panel including:

During the active action stage where a user is creating a new action;
When the transaction is completed; and
And when the transaction is in a motion state.
When the side panel is in the full width state.
If the user selects the close icon during the active action stage and has changed content within the action type chosen, the closing modal should be triggered (which matches the current modal trigger).

For all other states, the side panel should close when selected by the user without the modal triggering.

## Testing

Scenario 1:
* Step 1. Open sidepanel
* Step 2. Close without updating any form field

Scenario 2:
* Step 1. Open again and edit any field
* Step 2. Click on close button - modal with confirmation message should appear

Scenario 3:
* Step 1. Click on the completed action (from activity table on dashboard page)
* Step 2. Close the sidebar

## Diffs

**New stuff** ✨

* 

**Changes** 🏗

* close button now visible on desktop

**Deletions** ⚰️

* 

## TODO

Resolves https://github.com/JoinColony/colonyCDapp/issues/2287

### DIFF
--- a/src/components/v5/common/ActionSidebar/ActionSidebar.tsx
+++ b/src/components/v5/common/ActionSidebar/ActionSidebar.tsx
@@ -127,17 +127,16 @@ const ActionSidebar: FC<PropsWithChildren<ActionSidebarProps>> = ({
       ref={registerContainerRef}
     >
       <div className="flex w-full items-center justify-between border-b border-gray-200 px-6 py-4">
-        {isMobile ? (
+        <div className="flex items-center gap-2">
           <button
             type="button"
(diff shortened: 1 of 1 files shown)


melyndav (approved):
Hey @joanna-pagepro thanks for adding this feature in! Everything is working as expected. Thanks for your work on this.

rdig (approved):
All good

PR #2382, opened 2024-05-15
Fetch Tokens Lambda Handle "Irregular" Token Contracts

rdig:
## ERC-20 Rant

As always, when dealing with an ERC-20 token, I will rant about it's lax standard implementation: https://ethereum.org/en/developers/docs/standards/tokens/erc-20/

Due to the way it's set up and "valid" and "standard" token is a contract that is not required to have a "name", a "symbol" or "decimal" places. These are basic values everyone in the crypto world got to accept and expect.

It's mind boggling to me that these are not enforced.

## Description

This is exactly what happened in this case, we've had to deal with a token that didn't have decimal places, and the `decimal()` method on the contract did not exist.

Due to that, our lambda failed to fetch, and fail to save the token in the dabase, breaking some colonies in the process.

This fix puts in fallback values for all three: name, symbol and decimal by assuming tokens don't have them at all.

For the name we fall back to the token's address, for the symbol, we use the first `6` characters from the address (including the `0x` hex prefix) while for the decimals, we fall back to `0`

Also note, that the console log statements were purposefully left in, as they only show up in our back end logs, and they are quite useful when debugging issues. 

Screenshot from 2024-05-15 19-17-30
_PS: Check out the value on the symbol_

## Testing

As for testing, I can theoretically tell you how to test, but have no practical way to offer you.

**Local Dev**

In  theory, if you deploy, using the hardhat console, a ERC-20 token contract that has either the name, symbol or decimals values and methods removed (get the basic contract, remove them, then compile it), you could test this locally, but I have no easy way in which I could tell you how to do that

**QA**

Find a token that doesn't have the name, symbol or decimals methods on it's contract and add that to an existing colony 

_(Don't personally know of any)_

**Production**

Use token https://arbiscan.io/token/0x43C4717e367e35Bd65C04C9C8E646B7a51679670 and add it to one of your colonies

### DIFF
--- a/amplify/backend/function/fetchTokenFromChain/src/index.js
+++ b/amplify/backend/function/fetchTokenFromChain/src/index.js
@@ -104,9 +104,24 @@ exports.handler = async (event) => {
         basicTokenAbi,
         provider,
       );
-      const name = await tokenFromChain.name();
-      const symbol = await tokenFromChain.symbol();
-      const decimals = await tokenFromChain.decimals();
+      let name = checksummedAddress;
+      let symbol = checksummedAddress.slice(0, 6);
+      let decimals = 0;
+      try {
+        name = await tokenFromChain.name();
+      } catch (error) {
+        console.log(`TOKEN NAME NOT AVAILABLE, FALLING BACK TO: "${name}"`, error);
+      }
+      try {
+        symbol = await tokenFromChain.symbol();
+      } catch (error) {
+        console.log(`TOKEN SYMBOL NOT AVAILABLE, FALLING BACK TO: "${symbol}"`, error);
+      }
+      try {
+        decimals = await tokenFromChain.decimals();
+      } catch (error) {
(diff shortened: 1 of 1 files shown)


chmanie:
I wonder if we shouldn't fall back to 18 decimals as this is pretty much 99.9% of the tokens out there

rdig:
> I wonder if we shouldn't fall back to 18 decimals as this is pretty much 99.9% of the tokens out there

We decided that we'll treat it the same way arbiscan does, by defaulting to zero (`0`). This is the token in question that actually prompted this fix: https://arbiscan.io/address/0x43C4717e367e35Bd65C04C9C8E646B7a51679670

area:
I also approve of defaulting to 0 and not 18. :+1:

iamsamgibbs:
Also approving on the same grounds as Dave! Code looks good, will attempt to test locally unless it gets merged in the meantime.

iamsamgibbs (approved):

PR #2396, opened 2024-05-20
Feat/15900757 split payment

joanna-pagepro:
## Description

Split payment action

## Scope 

**The scope provided by @arrenv** 
https://www.notion.so/colony/Split-Payments-Action-60f0989d5b834278b68448f459884798

**Design:**
https://www.figma.com/design/5V8pr7iMwXsT9L3VAZsmUt/Colony-v5?node-id=3804-317494&t=1XeV28kMWBtONodB-4

_note from @AdrianDudko_ 

_For reviewers - please note that there might be some discrepancies between the implementation included in this PR and the scope described in the notion doc. 
Some of the work was made a few months ago, based on the provided design and might not align with what's in the notion doc._

_"dropdown component with checkboxes for easily adding multiple recipients at once" - agreed and confirmed that will be done later_

## Testing

* Step 1. Open action sidebar
* Step 2. Select `Split payment` action type
* Step 3. Select one of the available distribution methods
  * Equal - amount and percentage for each payment should be calculated automatically based on the value added to `amount` field
  * Unequal - you can edit amount and percentage for each payment, but they should sum up to the value of the total amount or to 100% if percent is edited
  * Reputation percentage - amount and percentage for each payment should be calculated automatically based on the value added to `amount` field and recipient's reputation
* Step 4. Provide all other needed data and submit the form
* Step 5. Check completed state of the action through all of the steps

## Diffs

**New stuff** ✨

* SplitPayment completed action component

**Changes** 🏗

* changes in `SplitPaymentForm` component and related components
* completed action for Payment builder and Split payment share the same components, so the structure of the folders was updated accordingly. There's `AdvancedPayments` directory now that includes both actions. 
* `PaymentBuilderWidget` renamed to `AdvancedPaymentsWidget`

**Deletions** ⚰️

* 

## TODO

- [ ] Add TODOs

Resolves - there's no github task


iamsamgibbs (changes requested):
Nice start to this ticket so far, the completed action screen all looks good.

Screenshot 2024-05-24 at 09 04 25

I've noticed a few issues with the form. If you add a new row before selecting a split type, there'll be some error validation messages that only disappear when you try submitting.

https://github.com/JoinColony/colonyCDapp/assets/34915414/c36366f6-49a2-4df8-b845-ab0f7513190f

The form won't let you submit percentage values that take it over 100%, but it doesn't show any error validation message and instantly changes the value. Feels like the better user experience would be to let it go greater than 100% but show the validation message. In this example I'm planning to set the first payment to 70% then the second to 30%, but as I change the first one it is instantly changed.

https://github.com/JoinColony/colonyCDapp/assets/34915414/75338a26-6dfd-47d1-a418-10466893406b

There are some weird tabbing issues in the table.

https://github.com/JoinColony/colonyCDapp/assets/34915414/a191f786-62ea-4e24-8588-92354fcc821d

And the value of both the amount field and the percentage field seem to be linked as they both change at the same time.

https://github.com/JoinColony/colonyCDapp/assets/34915414/433df09c-9a4e-4a83-946e-2591b7cb2043

joanna-pagepro:
@iamsamgibbs 

> And the value of both the amount field and the percentage field seem to be linked as they both change at the same time.

I'm not sure what is wrong here. Both values are connected and they have to update when one of them is changed.

I've made some changes. Can you review it again? :)

iamsamgibbs:
> @iamsamgibbs
> 
> > And the value of both the amount field and the percentage field seem to be linked as they both change at the same time.
> 
> I'm not sure what is wrong here. Both values are connected and they have to update when one of them is changed.
> 
> I've made some changes. Can you review it again? :)

Oh of course yeah 🤦‍♂️ You are correct. I'll spin up the branch again later today if I have a chance to review the changes.

marcocolony (changes requested):
Hey @joanna-pagepro, I've found a few issues:

For the equal split, whenever the percentage doesn't add up to a round figure, we get 15 decimal places. It should round up to only 4 decimal places.

Also, the percentage input field should only allow for 4 decimal places to be entered when selecting **Unequal** distribution 
 
Export-1716807544285

Export-1716808044390


There should be no pagination UI here, as per the [spec](https://www.notion.so/colony/Split-Payments-Action-60f0989d5b834278b68448f459884798?pvs=4#20704de7ed8b44e19174dc8a28edc073)  
Export-1716812516690

Also when adding a new row, the **Add recipient ** button is hidden and requires the user to manually scroll down in the action panel. Ideally, the table should scroll down automatically after a new row is added.

Export-1716812810129
 
The UI breaks when over 18 decimal places are entered on the amount input field when **unequal**  distribution is selected. This field should not accept over 18 decimal places to be entered 

Export-1716813454870

I've also gone through the issues raised by @iamsamgibbs and they all seem to have been addressed 🙌

AdrianDudko:
@joanna-pagepro @marcocolony @arrenv 

Ass agreed, due to the global nature of the change request, it has been moved to a separate issue https://github.com/JoinColony/colonyCDapp/issues/2449

> Also when adding a new row, the **Add recipient ** button is hidden and requires the user to manually scroll down in the action panel. Ideally, the table should scroll down automatically after a new row is added.
> 
> Export-1716812810129 [ Export-1716812810129 ](https://private-user-images.githubusercontent.com/165801260/334101301-a0e1bf22-c310-40a0-b9ba-9f71a28990e7.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTY4ODg1MDgsIm5iZiI6MTcxNjg4ODIwOCwicGF0aCI6Ii8xNjU4MDEyNjAvMzM0MTAxMzAxLWEwZTFiZjIyLWMzMTAtNDBhMC1iOWJhLTlmNzFhMjg5OTBlNy5naWY_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjQwNTI4JTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI0MDUyOFQwOTIzMjhaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT1jYTg5NGRiMzJlZmY2NTc2NWJiMTdlZDdjZjNlMDQ3ODc2Y2FhZDQ1OTczNjkwMjQzNWI5Mzg5ZDkzNjg1MGNlJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZhY3Rvcl9pZD0wJmtleV9pZD0wJnJlcG9faWQ9MCJ9.6SyjjRECgdT7IPl0mn-Y00mmTBGEaNDD9SdgwIsdqNo) [ ](https://private-user-images.githubusercontent.com/165801260/334101301-a0e1bf22-c310-40a0-b9ba-9f71a28990e7.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTY4ODg1MDgsIm5iZiI6MTcxNjg4ODIwOCwicGF0aCI6Ii8xNjU4MDEyNjAvMzM0MTAxMzAxLWEwZTFiZjIyLWMzMTAtNDBhMC1iOWJhLTlmNzFhMjg5OTBlNy5naWY_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjQwNTI4JTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI0MDUyOFQwOTIzMjhaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT1jYTg5NGRiMzJlZmY2NTc2NWJiMTdlZDdjZjNlMDQ3ODc2Y2FhZDQ1OTczNjkwMjQzNWI5Mzg5ZDkzNjg1MGNlJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZhY3Rvcl9pZD0wJmtleV9pZD0wJnJlcG9faWQ9MCJ9.6SyjjRECgdT7IPl0mn-Y00mmTBGEaNDD9SdgwIsdqNo)

joanna-pagepro:
@marcocolony @iamsamgibbs can you review it again?

arrenv (changes requested):
@joanna-pagepro Thank you for working on this one.

This seems like it could do with a rebase, I am seeing a lot of changes that don't seem relevant.

1. There are gray boxes on the disabled value inputs, these should not be disabled, as of the spec these inputs should be editable. The distribution type should just update to Unequal if manually changed. It is too confusing to have the state of the input change between options.
image

2. We should use our validation for these errors, not the default validation.
image

3. This tooltip should not be here, it interferes with the select member popup.
image
image

4. I am getting a weird issue if typing at the start of a number, where on the third digit it jumps to the end.
input-weird-placement

5. I get a flashing of values when the table updates. It updates correctly, then changes to 0, then back to the correct value. I have slowed down the gif for dramatic effect. This also happens when I enter in any value, it seems.
flashing-zero

6. I have auto calculated values for Reputation, but it fails validation.
image

7. Validation fails in this case, where it does not equal 100%. It should still allow you to do this if the option is set to equal.
image

8. Making the payment resulted in a white screen and the following error:
image

joanna-pagepro:
@arrenv please review again

iamsamgibbs (changes requested):
Nice work on this! All the previous issues appear to be resolved. I've noticed a few new things however.

The error validation messages need fixing:

Screenshot 2024-06-10 at 09 04 19

It is impossible to edit the amount or percentage when  "Reputation percentage" is selected. It should be possible to edit the amount and for it to switch to "Unequal". (Same as it does when editing an amount with "Equal" selected).

https://github.com/JoinColony/colonyCDapp/assets/34915414/4048dec7-3c44-4ebb-a9a2-cbe4fe9b7ea0

It is possible to submit amounts which do not equal 100%.

https://github.com/JoinColony/colonyCDapp/assets/34915414/ebe26365-808a-4e4b-a535-537a5ede06d7

Screenshot 2024-06-10 at 09 13 09

The percentages do not update when the total amount is changed, resulting in some weird behaviours (and can make it impossible to submit the form even when amounts are valid):

https://github.com/JoinColony/colonyCDapp/assets/34915414/a9283328-292d-4382-9447-5daa92be7806

Screenshot 2024-06-10 at 09 40 31

arrenv (changes requested):
@joanna-pagepro Thank you for the updates.

1. I still seem to be seeing the grey background on the disabled state, but also, now seeing the width being larger for the amount input in the table. Can this just be no background and the standard on load width.
image

2. Can we update this validation message to "Percent must be less than or equal to 100".
image

3. The reputation percentage option seems to have stopped working even though reputation is loading.
image

4. Status is showing as "Passed" in the header, when it should just be in the "Review" stage.
image

joanna-pagepro:
@arrenv 
1. fields inside the table are disabled until you add an amount in one of the fields above. Otherwise, all of them should be editable. Let me know if that's ok or if you expect it to work differently.
2. updated
3. it started working this morning, I haven't changed anything. Can you check it again?
4. it should show correct status now

arrenv (changes requested):
@joanna-pagepro thank you for updating.

1. I would prefer that they were not disabled, I tested locally, and all it does when you enter a value is change the distribution method. So, I think this is fine to not disable the inputs. Especially as it adds the grey background, which we don't have anywhere else.

2. Can we please change the placeholder values for these inputs from 0 for both to "Enter amount" for the Amount column and "Enter value" for the Percent column. This is something we are doing elsewhere as well to reduce user confusion and improve the user experience.

3. There is a horizontal alignment issue with the amount appearing higher then the other columns.
image

4. I changed the token type in the amount up the top and it did not update the token in the table.
image

5. Validation for higher then 100% did not appear this time and the percent did not increase when I manually entered in a value in the amount field.
image

6. Whenever I create the payment, it crashes the app with the following errors:
image

arrenv:
@joanna-pagepro To add additional clarification regarding disabled fields in the table if no amount is entered.

By remove the disable attribute, you get the following gif, which would be the desired experience, with the only additional requirement being that when the amount is entered up the top, the percent value would update.

split-difference

joanna-pagepro:
@arrenv @iamsamgibbs @jakubcolony can you check again?

arrenv (changes requested):
@joanna-pagepro Thank you for your continuing hard work.

1. The user does not have any available decision methods, despite Staking Advanced Payments being available. It should else show that the user does not have permissions using the same card as in other actions if there are no available decision methods.
image

2. The horizontal alignment issue is still present across columns.
image

Everything else looks good.

joanna-pagepro:

iamsamgibbs (changes requested):
The horizontal alignment issue appears to be resolved:
 
Screenshot 2024-06-27 at 10 33 56

However, the permissions issue is still occurring:

Screenshot 2024-06-27 at 10 32 34

I was logged in as a user with no permissions. Reputation weighted extension was uninstalled. Staking advanced payments extension was deprecated. Also experienced the same when Staking advanced payments extension was uninstalled.

The amount column and percentage column should have `Enter amount` and `Enter value` as placeholder text.

Screenshot 2024-06-27 at 10 36 54

joanna-pagepro:
@iamsamgibbs where did you find the information about the placeholders? On a design it's `0` for both fields.

iamsamgibbs:
> @iamsamgibbs where did you find the information about the placeholders? On a design it's `0` for both fields.

Arren's second point in this comment: https://github.com/JoinColony/colonyCDapp/pull/2396#pullrequestreview-2116180027

joanna-pagepro:
Thanks! I missed this one

joanna-pagepro:
@arrenv @iamsamgibbs code updated, can you check if it's ok now?

iamsamgibbs (changes requested):
This is looking better! Placeholders are correct now:

Screenshot 2024-06-28 at 11 11 46

Permissions are being checked correctly with the error notice displayed. But the disabled amount fields have a grey background, this should be removed.
 
Screenshot 2024-06-28 at 11 13 15

joanna-pagepro:
@iamsamgibbs can you explain why they shouldn't be grey? All of the fields are disabled in this case and I think that we should show it.

iamsamgibbs:
> @iamsamgibbs can you explain why they shouldn't be grey? All of the fields are disabled in this case and I think that we should show it.

 Sorry my screenshot might not have been clear enough. The grey background on the amount fields here:
 
Screenshot 2024-06-28 at 13 07 13

Should be removed so it looks like this:

Screenshot 2024-06-28 at 13 10 12

I've also just noticed 'enter value' has the wrong text colour when disabled, it should match the other fields.

Screenshot 2024-06-28 at 13 10 20

iamsamgibbs (approved):
Thanks for the update! This looks good now, let's get it merged!

Screenshot 2024-06-28 at 16 16 47

arrenv (changes requested):
@joanna-pagepro Thank you for all the updates. It is looking really good now. I seem to be having a couple of issues at the moment.

1. I am not able to use the Reputation distribution type due to validation, with the default reputation of the create data script and any value for amount.
image

2. Creating a Split payment with Staking does not seem to work. Clicking Create payment in the modal does not do anything.
    image

    I tested with an Advanced payment, and it worked as expected.
image

    I did see this error, but was not able to replicate it and does not seem related.
image

joanna-pagepro:
@arrenv staking decision method should work right now. I couldn't reproduce the bug with the reputation percentage - it was always 100%. Can you test it again?

arrenv (changes requested):
@joanna-pagepro Thank you for the update. The staking method works now! 

It looks like a rebase has introduced a couple of issues, though, and I still get the issue with Reputation distribution.

1. The input widths not working correctly when you enter in a value.
image

2. The amount placeholder is flashing in when entering in values.
amount-placeholder

3. I am still get the validation issue with the Reputation distribution method. I have tested again a few times with fresh dev environments using the create-data script and I still get the issue with different combinations of members, such as:
    - leela, fry, amy
    - leela, fry, amy, amy, fry
    - amy, amy, fry

   image

   image

   image

src/components/v5/common/ActionSidebar/partials/forms/SplitPaymentForm/partials/SplitPaymentPayoutsTotal/types.ts
```diff
@@ -0,0 +1,10 @@
+import { type Token } from '~types/graphql.ts';
+
+import { type SplitPaymentRecipientsFieldModel } from '../SplitPaymentRecipientsField/types.ts';
+
+export interface SplitPaymentPayoutsTotalProps {
+  data: SplitPaymentRecipientsFieldModel[];
+  token: Token;
+  moveDecimals?: boolean;
```
jakubcolony, line 8: It's a bit unclear what this does (could be converting either way), could you rename it to make it clear it's converting to WEI or from WEI?

src/components/v5/common/CompletedAction/partials/SplitPayment/partials/SplitPaymentTable/hooks.tsx
```diff
@@ -0,0 +1,105 @@
...
+import RecipientField from '../../../PaymentBuilder/partials/PaymentBuilderTable/partials/RecipientField/RecipientField.tsx';
+
+import { type SplitPaymentTableModel } from './types.ts';
+
+export const useGetSplitPaymentColumns = (
+  data: SplitPaymentTableModel[],
+  status: ExpenditureStatus,
+  isLoading?: boolean,
+) => {
+  const dataRef = useWrapWithRef(data);
```
jakubcolony, line 24: Can we go away from using this hook? It's not actually doing anything apart from 'hiding' the value from React eslint rules so that it doesn't complain about it missing from `useMemo` deps

A much better option that makes the intentions clear would be to just disable the eslint rule, and leave a comment about the reasons.

jakubcolony (changes requested):
Nice work @joanna-pagepro, apologies for a delayed review but I was off following the all-hands. 

The feature is functioning well, only picked up some minor issues.

Equal split:
 - 1 recipient:
image
- 2 recipients:
image

- 3 recipients:
image

Unsure if we want to handle those repeating decimals differently (e.g. truncate after 4 digits) but I don't see anything relevant in specs.

Unequal split works fine, allowed me to set my own amounts/percentages with validation.

 I got to a state when there's no reputation in the colony:

image

Should that trigger a "no reputation in colony" validation error or similar? @arrenv probably more of a question to you.

When `Equal` distribution is selected, clicking in and out of the percent field causes the entire form to flicker:

https://github.com/JoinColony/colonyCDapp/assets/112586815/7cfb7d69-8714-4f12-b723-4d2b1c1cf742

arrenv:
@joanna-pagepro @jakubcolony 

We need to handle both cases here:
- If reputation is not able to be loaded due to an API issue.
    - We have the same error when staking to create a payment.
- If there is no reputation in the colony.
    - We have the same error when selecting the reputation decision method.

joanna-pagepro:
@arrenv @jakubcolony can you check if everything is working as it should?

arrenv (changes requested):
@joanna-pagepro Thank you for your continued work. Looks like all issues are resolved.

Although, sorry, I did notice one thing.

1. I noticed that the number in the Split payment metadata does not have commas.
image

    #### In simple payments
    image

joanna-pagepro:
@arrenv it should be ok now

arrenv (changes requested):
@joanna-pagepro Tested a bunch and everything seemed to work well. Unfortunately, I ended up running into this issue, it happened about 1 in every 10 attempts.

I'm not sure what led to it, but the only thing that pops up in my mind is that the payment info was not available for when loading the transaction view.

error-split-pay

joanna-pagepro:
@arrenv can you check it again?

arrenv (approved):
@joanna-pagepro Great work, everything tested is working great. Thank you for your hard work!

joanna-pagepro:
@jakubcolony can you check as well? You requested changes in this PR

jakubcolony (approved):
Code-wise all looks good, thank you for making the changes.

I have noticed we're showing decimals even when they're not needed, I let @arrenv know and let him decide if it should hold this PR or not:

image

arrenv:
@jakubcolony Thank you, and good pickup. I will create a separate issue for it. As it would be good to close this one out as is.

PR #2420, opened 2024-05-21
Fix/2358 Truncate token symbol length

### LINKED ISSUE #2358: Limit the length of token symbol string in all cases
## Steps to reproduce

1. Add a token to a Colony that has a Symbol longer then 5 characters. Easiest way would like to do so in the DB.
2. Check all instances where token symbols are used.
3. Note there is no limit on how many characters will show in the UI.

## Expected behaviour

* Token symbol strings should be limited to 5 characters.
* Symbols longer than 5 characters should have an ellipsis.

## Actual behaviour

* The UI does not limit tokens that have a symbol longer then 5 characters, breaking the UI.

image

image

image

rumzledz:
## Description

Token symbol lengths can potentially break the app's UI so this limits the characters being shown for a token's symbol.

## Testing

Please check all areas where token symbols are rendered. Apologies in advance if I have missed some areas. I will update those areas asap if you spot them.

## Diffs

**New stuff** ✨

A new string util function named `truncateTokenSymbol` which is to be used specifically for truncating token symbols to a maximum of 5 characters.

Resolves #2358

### DIFF
--- a/src/components/common/Extensions/UserHub/partials/ReputationTab/partials/Balance.tsx
+++ b/src/components/common/Extensions/UserHub/partials/ReputationTab/partials/Balance.tsx
@@ -85,7 +85,7 @@ const Balance: FC<BalanceProps> = ({ nativeToken, wallet, onTabChange }) => {
             className={reputationTabClasses.numeral}
             value={tokenBalanceData?.balance ?? 0}
             decimals={tokenDecimals}
-            suffix={` ${multiLineTextEllipsis(nativeToken?.symbol ?? '', 5) || 'CLNY'}`}
+            suffix={` ${multiLineTextEllipsis(nativeToken?.symbol ?? 'CLNY', 5)}`}
             appearance={{ size: 'small' }}
           />
         </div>
@@ -127,7 +127,7 @@ const Balance: FC<BalanceProps> = ({ nativeToken, wallet, onTabChange }) => {
               className={reputationTabClasses.numeral}
               value={tokenBalanceData?.activeBalance ?? 0}
               decimals={tokenDecimals}
-              suffix={` ${multiLineTextEllipsis(nativeToken?.symbol ?? '', 5) || 'CLNY'}`}
+              suffix={` ${multiLineTextEllipsis(nativeToken?.symbol ?? 'CLNY', 5)}`}
               appearance={{ size: 'small' }}
             />
           </div>
@@ -178,7 +178,7 @@ const Balance: FC<BalanceProps> = ({ nativeToken, wallet, onTabChange }) => {
               className={reputationTabClasses.numeral}
               value={tokenBalanceData?.lockedBalance ?? 0}
               decimals={tokenDecimals}
-              suffix={` ${multiLineTextEllipsis(nativeToken?.symbol ?? '', 5) || 'CLNY'}`}
+              suffix={` ${multiLineTextEllipsis(nativeToken?.symbol ?? 'CLNY', 5)}`}
               appearance={{ size: 'small' }}
             />
           </div>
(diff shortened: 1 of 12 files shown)


src/components/common/Extensions/UserHub/partials/ReputationTab/partials/Balance.tsx
```diff
@@ -84,7 +85,7 @@ const Balance: FC<BalanceProps> = ({ nativeToken, wallet, onTabChange }) => {
             className={reputationTabClasses.numeral}
             value={tokenBalanceData?.balance ?? 0}
             decimals={tokenDecimals}
-            suffix={` ${nativeToken?.symbol || 'CLNY'}`}
+            suffix={` ${truncateTokenSymbol(nativeToken?.symbol) || 'CLNY'}`}
```
rdig, line 88: Why does this have CLNY as a fall back, it really shouldn't be here. @jakubcolony any ideas

src/utils/strings.ts
```diff
@@ -190,3 +190,26 @@ export const stripAndremoveHeadingsFromHTML = (content: string) => {
...
+ */
+export const truncateTokenSymbol = (tokenSymbol?: string) => {
+  if (tokenSymbol) {
+    return tokenSymbol.length > 5
+      ? `${tokenSymbol.slice(0, 5)}...`
+      : tokenSymbol;
+  }
+
+  return tokenSymbol;
+};
```
rdig, line 215: Why not make this generic, and configurable via a argument. We really don't need this to be so specific

rdig (changes requested):
Rejecting just because the util can be made more generic.

src/utils/strings.ts
rumzledz: @rdig it was only because I couldn't come up with another use case for it in the codebase and wanted the function to be very "oddly specific" about token length 👀 but yes you're right! There's nothing stopping me from making this more generic 😄 Will push some changes!

rumzledz: Hey @rdig I've removed that function and I'm now using the existing string truncation function. All pushed 🚀

rdig (changes requested):
Almost there. Picked up on something after your changes which I also want you to address

src/components/common/Extensions/UserHub/partials/ReputationTab/partials/Balance.tsx
```diff
@@ -84,7 +85,7 @@ const Balance: FC<BalanceProps> = ({ nativeToken, wallet, onTabChange }) => {
             className={reputationTabClasses.numeral}
             value={tokenBalanceData?.balance ?? 0}
             decimals={tokenDecimals}
-            suffix={` ${nativeToken?.symbol || 'CLNY'}`}
+            suffix={` ${multiLineTextEllipsis(nativeToken?.symbol ?? '', 5) || 'CLNY'}`}
```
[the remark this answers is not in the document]
rumzledz: Doh 🤦‍♂️  Yes sorry silly mistake

rumzledz:
> Almost there. Picked up on something after your changes which I also want you to address

Thanks for this @rdig , I've just pushed the changes!

rdig (dismissed):
All good now. Nicely done!

melyndav (approved):
Hey @rumzledz solid work on this PR! Everything looks as intended with the issue fix in the core areas.

image

image

image



Just one request, can we also update the native pill on the dashboard next to the Colony title? These break when a longer string is used as per my screenshots.

Pill in question:
image

Breaking with large string
image

 
 And also on the Colony details page:
 
image

melyndav (changes requested):
Updating to requested changes and not approved.

rumzledz:
> Updating to requested changes and not approved.

Thanks for showing me those places @melyndav ! Will apply the fix to those as well 👌

rumzledz:
Hey @melyndav I've just truncated the token symbol in the Pill component as well. Can you please re-check when you're free? Thank you!

melyndav (approved):
@rumzledz nice fix on those pills! Thanks for updating the symbol here as well. 🦖

rdig (approved):
Re-approving so you can merge

PR #2421, opened 2024-05-21
fix: Disable token selection when mint form is active

### LINKED ISSUE #2383: Mint tokens: Non native tokens are selectable
## Steps to reproduce

Tested on Production, QA and local dev 

1. Create a new Action
2. Select the **Mint tokens** action type
3. Open token selector
4. Select a non-native token
5. Select decision method 
6. Create payment 

## Expected behaviour

Token selector should be disabled when this action type is selected since the only mintable token should be the colony's native token. 

## Actual behaviour

1. You can select a non-native token
2. You can clock on **Mint tokens** and sign the transaction
3. Once the action is complete,  the action data displays the colony's native token instead of the non-native token that was initially selected. 
4. The amount of tokens entered is successfully minted. 

Export-1715814484027

rumzledz:
## Description

This will disable the token selection when the mint form is presented to the user.

## Testing

1. Visit a Colony you've joined
2. Bring up the Action form
3. Set the Action type field to "Mint tokens"
4.  Verify that:
  a.) Hovering on the token symbol doesn't do anything
  b.) Clicking the token symbol doesn't do anything

## Diffs

**New stuff** ✨

I noticed that the `ACTION_TYPE_FIELD_NAME` action is being tracked in multiple places so I made a new hook called `useActionActionType` to keep things dry.

**Changes** 🏗

Token selection should now be disabled when the mint form is active.

Resolves #2383

### DIFF
--- a/src/components/v5/common/ActionSidebar/ActionTypeSelect.tsx
+++ b/src/components/v5/common/ActionSidebar/ActionTypeSelect.tsx
@@ -1,7 +1,7 @@
 import { FilePlus, WarningCircle } from '@phosphor-icons/react';
 import clsx from 'clsx';
 import React, { type FC, useState } from 'react';
-import { useController, useFormContext, useWatch } from 'react-hook-form';
+import { useController, useFormContext } from 'react-hook-form';
 
 import { useAdditionalFormOptionsContext } from '~context/AdditionalFormOptionsContext/AdditionalFormOptionsContext.ts';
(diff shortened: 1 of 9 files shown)


marcocolony (approved):
@rumzledz Thanks for picking this one up! Looks great, well done.

Export-1716373995616

melyndav (approved):
Nice work again @rumzledz! Thanks for sorting this one out. 🦖

bassgeta (approved):
Stellar job with this, works, and I like the new hook!

PR #2431, opened 2024-05-22
fix: Update the total balances currency conversion arithmetic

### LINKED ISSUE #2334: Currency conversion balances are not showing correctly on dashboard
## Description

Currency conversion balances are not showing correctly on dashboard.

This is noticeable in this colony - https://app.colony.io/boston

This was resolved in this PR previously, but for the balances page - https://github.com/JoinColony/colonyCDapp/issues/1936

## Expected behaviour

* You should see accurate total funds balance in the user denominated currency.

## Actual behaviour

* Total funds balance is not showing accurately in the users denominated currency.

image

rumzledz:
## Description

Previously, we were converting the wei version of the balance by its currency equivalent which was producing a "wei-like" format of the conversion price.

In this PR, it is now first converting the wei version of the balance to its unit version before applying the final currency conversion.
 
## Testing

1. Visit your Colony's Balances page
2. Take note of the "Total funds" value
3. Click on the the Total Funds cell
4. Verify that it matches the "Balance" value for ETH

## Diffs

**Changes** 🏗

Added an extra step to first derive the non-wei value of the token's balance prior to finalising its currency conversion.

| before | after |
| ------ | ------ |
| Screenshot 2024-05-23 at 00 15 39 | image |

Resolves #2334

### DIFF
--- a/src/components/v5/frame/ColonyHome/partials/TokenBalance/hooks.ts
+++ b/src/components/v5/frame/ColonyHome/partials/TokenBalance/hooks.ts
@@ -17,7 +17,7 @@ const calculateTotalFunds = async (
     ?.filter(notNull)
     .filter(({ domain }) => !!domain?.isRoot)
     .reduce(
-      async (total, { balance, token: { tokenAddress } }) => {
+      async (total, { balance, token: { tokenAddress, decimals } }) => {
         const currentPrice = await fetchCurrentPrice({
           contractAddress: tokenAddress,
           conversionDenomination: currency,
@@ -27,7 +27,11 @@ const calculateTotalFunds = async (
           isError = true;
         }
 
-        return (await total).add(new Decimal(balance).mul(currentPrice ?? 0));
+        const balanceInWeiToToken = new Decimal(balance).div(10 ** decimals);
+
+        return (await total).add(
+          new Decimal(balanceInWeiToToken).mul(currentPrice ?? 0),
+        );
       },
       Promise.resolve(new Decimal(0)),
     );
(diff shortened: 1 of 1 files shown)


src/components/v5/frame/ColonyHome/partials/TokenBalance/hooks.ts
```diff
@@ -27,7 +27,11 @@ const calculateTotalFunds = async (
           isError = true;
         }
 
-        return (await total).add(new Decimal(balance).mul(currentPrice ?? 0));
+        const balanceInWeiToToken = new Decimal(balance).div(10 ** decimals);
```
jakubcolony, line 30: It's gonna be a nit-pick, it all works correct but since you're dividing by 10^decimals, the result is ETH rather than WEI. Would you mind updating the naming to reflect that?

jakubcolony (approved):
The balance looks correct now, great work!
Screenshot 2024-05-23 at 11 25 32

I also checked the balance table to make sure there was no regression there:
Screenshot 2024-05-23 at 11 25 39

src/components/v5/frame/ColonyHome/partials/TokenBalance/hooks.ts
rumzledz: @jakubcolony oh I was under the impression that because this loop is going through your balances, I thought that you will also have non-ETH token balances in there with their own respective decimals as well which is why I wanted to call it something generic like "Token"

rumzledz: @jakubcolony thanks for explaining that ETH could also mean the actual token unit, and is not just talking specifically about the Ethereum token! Changes pushed! 🚀

jakubcolony (approved):

rdig (approved):
All good.

However, as a side note I want to mention that we really should stop using `Decimals` as a lib as it's really error prone.

We should change it instead to just be used via BigNumber (or BigInt, if we refactor everything)

Issue #2368, opened 2024-05-13
Decimals issue in claim delay input


arrenv:
Teamwork task: [Decimals issue in claim delay input](https://tw.auroracreation.com/#/tasks/15912172)

## Steps to reproduce

1. Create a Payment builder action.
2. Enter in 100th of a decimal in the claim delay field, e.g. 0.05
3. You won't be able to enter it in.
4. Enter in a 10th of a decimal in the claim delay fields e.g. 0.5
5. You will be able to enter it in.
6. Now add 0s before the 5 as much as you can, it will convert to a scientific notation at 7 decimal places.
7. On submission it will validate incorrectly.

## Expected behaviour

* You should be able to enter in multiple decimals places into the claim delay field.
* The limit should be 4 decimal places, e.g. 0.0001.
* You should be able to enter in decimals without the incorrect validation.

## Actual behaviour

* You are not able to enter in 100th of a decimals into the claim delay field.
* There is no limit to the number of decimals you can enter.
* You can incorrect validation when using decimals due to the step attribute.

![decimals-on-delay](https://github.com/JoinColony/colonyCDapp/assets/33682027/33daba71-f48d-4ee5-ba1f-f14652edb6ad)

![image](https://github.com/JoinColony/colonyCDapp/assets/33682027/e5dbc4ed-d089-4d25-a248-50415a51fb8b)

PR #2435, opened 2024-05-23
fix: claim delay field validation

joanna-pagepro:
## Description

Claim delay field validation updated so that a user has to add a number between 0.0001 and 99999.

## Testing

* Step 1. Open action sidebar and select `Payment builder` action type
* Step 2. Edit `Claim delay` field inside the table with payments and check if the validation works as it should.

## Diffs

**New stuff** ✨

* `ClaimDelayField` component - uses `cleave-zen` inside

**Changes** 🏗

* `delay` field validation updated
* `claim delay` column value formatting updated for completed action

**Deletions** ⚰️

* 

## TODO

Resolves https://github.com/JoinColony/colonyCDapp/issues/2368

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/forms/PaymentBuilderForm/partials/PaymentBuilderRecipientsField/hooks.tsx
+++ b/src/components/v5/common/ActionSidebar/partials/forms/PaymentBuilderForm/partials/PaymentBuilderRecipientsField/hooks.tsx
@@ -1,13 +1,15 @@
 import { createColumnHelper, type ColumnDef } from '@tanstack/react-table';
+import { BigNumber } from 'ethers';
 import React, { useMemo } from 'react';
 import { useFormContext } from 'react-hook-form';
 
 import { useColonyContext } from '~context/ColonyContext/ColonyContext.ts';
 import useWrapWithRef from '~hooks/useWrapWithRef.ts';
+import { convertPeriodToHours } from '~utils/extensions.ts';
 import { formatText } from '~utils/intl.ts';
 import AmountField from '~v5/common/ActionSidebar/partials/AmountField/index.ts';
-import FormInputBase from '~v5/common/Fields/InputBase/FormInputBase.tsx';
 
+import ClaimDelayField from '../ClaimDelayField/ClaimDelayField.tsx';
 import PaymentBuilderPayoutsTotal from '../PaymentBuilderPayoutsTotal/index.ts';
 
 import {
@@ -25,15 +27,44 @@ export const useRecipientsFieldTableColumns = (
 
   const dataRef = useWrapWithRef(data);
   const expendituresGlobalClaimDelayHours = useMemo(() => {
-    if (typeof expendituresGlobalClaimDelay !== 'number') {
+    if (typeof expendituresGlobalClaimDelay !== 'string') {
       return null;
     }
 
-    return expendituresGlobalClaimDelay / (60 * 60);
+    return convertPeriodToHours(expendituresGlobalClaimDelay);
   }, [expendituresGlobalClaimDelay]);
   const { watch } = useFormContext();
   const selectedTeam = watch('from');
   const hasMoreThanOneToken = data.length > 1;
+
+  const footerData = useMemo(
+    () => ({
+      amountFooter: hasMoreThanOneToken
+        ? () => (
+            <PaymentBuilderPayoutsTotal
+              data={dataRef.current}
+              moveDecimals
+              itemClassName="justify-end md:justify-start"
+              buttonClassName="justify-end md:justify-start"
+            />
+          )
+        : undefined,
+      recipientFooter: hasMoreThanOneToken
+        ? () => (
+            <span className="flex min-h-[1.875rem] items-center text-xs text-gray-400">
+              {dataRef.current.length <= 7
+                ? formatText({ id: 'table.footer.total' })
+                : formatText(
+                    { id: 'table.footer.totalPayments' },
+                    { payments: dataRef.current.length },
+                  )}
+            </span>
+          )
+        : undefined,
+    }),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [hasMoreThanOneToken],
+  );
 
   const columns: ColumnDef<PaymentBuilderRecipientsTableModel, string>[] =
     useMemo(() => {
@@ -45,20 +76,12 @@ export const useRecipientsFieldTableColumns = (
           id: 'recipient',
           header: () => formatText({ id: 'table.row.recipient' }),
           cell: ({ row }) => (
-            <UserSelectRow row={row} name={`${name}.${row.index}.recipient`} />
+            <UserSelectRow
+              key={row.id}
+              name={`${name}.${row.index}.recipient`}
+            />
           ),
-          footer: hasMoreThanOneToken
-            ? () => (
-                <span className="flex min-h-[1.875rem] items-center text-xs text-gray-400">
-                  {data.length <= 7
-                    ? formatText({ id: 'table.footer.total' })
-                    : formatText(
-                        { id: 'table.footer.totalPayments' },
-                        { payments: data.length },
-                      )}
-                </span>
-              )
-            : undefined,
+          footer: footerData.recipientFooter,
         }),
         columnHelper.display({
           id: 'amount',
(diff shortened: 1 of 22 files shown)


davecreaser (approved):
All looks good to me. Tested and the validation is working as expected for both min and max.

marcocolony (changes requested):
Hey @joanna-pagepro, thanks for picking this one up!

- Minimum and maximum values validation is working ✅ 
- Only up to 4 decimal points can be entered ✅ 
 - I'm able to proceed with the payment creation when a number bigger than one is entered ✅ 
- Values over 99999 can be entered ❌ 
- The error message that was previously displayed whenever a decimal number was entered is no longer there, but I'm unable to proceed and an error is thrown in the console ❌ 

Export-1716478734658

Export-1716479096742

image

arrenv:
@joanna-pagepro 
If possible, it would be good to solve the live validation lag issue when we have a lot of rows.

If not possible, then it would be good to limit the field length to 5 digits with 0.0001 and 99999.

If that is not possible due to the decimal place, then it would still be good to limit the length to 5 digits with 0.001 and 99999

joanna-pagepro:
@marcocolony can you review it again?

marcocolony (changes requested):
Hey @joanna-pagepro 

I don't think it's quite fixed just yet, please see the examples I tested below  

Export-1716569121298

melyndav (changes requested):
Hey @joanna-pagepro great start on this one.

When attempting to test this, I get default validation.

image


Can we remove `step="0.001"` to avoid the default validation?

joanna-pagepro:
@marcocolony @melyndav can you check again?

marcocolony (changes requested):
@joanna-pagepro Thanks for looking into this again

I'm afraid I'm still able to enter as many digits as I want here.
Export-1716817246817


But the decimal values look great, always limited to 4 🙌

joanna-pagepro:
@marcocolony as discussed during the last daily - you will see an error on submit, it's not validated live. You should be able to enter as many digits as you want, but you shouldn't be able to submit the form if the value is bigger than 99999 or smaller than 0.0001

marcocolony (approved):
Thank's @joanna-pagepro, everything looks good!

src/components/v5/common/ActionSidebar/partials/forms/PaymentBuilderForm/partials/PaymentBuilderRecipientsField/hooks.tsx
```diff
@@ -119,9 +113,11 @@ export const useRecipientsFieldTableColumns = (
...
                 header: () => formatText({ id: 'table.column.totalDelay' }),
                 cell: ({ row }) => {
-                  const totalHours =
-                    expendituresGlobalClaimDelayHours +
-                    (dataRef.current[row.index]?.delay || 0);
+                  const totalHours = BigNumber.from(
+                    expendituresGlobalClaimDelayHours,
+                  )
+                    .add(BigNumber.from(dataRef.current[row.index]?.delay || 0))
+                    .toNumber();
```
jakubcolony, line 120: This can't be easily tested since there's no saga to set the global claim delay, however I spotted some issues:

The problem with `toNumber` here is that it's possible for `expendituresGlobalClaimDelayHours` to be > 2^53 which will cause it to throw an error (and it wouldn't be caught). All claim delay calculations should use BigNumbers (great shout on already switching it here) and strings.

This made me realise the `expendituresGlobalClaimDelayHours` will not be set correctly because of line 29 where we check if global claim delay's type is `number`, which it will never be.

And, if `dataRef.current[row.index]?.delay` is a floating point number, `BigNumber` will throw an error. 

Feel free to ping me on Discord and I'll provide you with instructions on how to set the global claim delay for testing.

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/PaymentBuilderTable/PaymentBuilderTable.tsx
```diff
@@ -94,9 +93,7 @@ const useGetPaymentBuilderColumns = (
         staticSize:
           status === ExpenditureStatus.Finalized ? '6.875rem' : undefined,
         cell: ({ row }) => {
-          const formattedHours = Math.floor(
-            Number(row.original.claimDelay) / 3600,
-          );
+          const formattedHours = row.original.claimDelay / 3600;
```
jakubcolony, line 96: Should use `BigNumber`

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/PaymentBuilderTable/PaymentBuilderTable.tsx
```diff
@@ -171,9 +168,9 @@ const PaymentBuilderTable: FC<PaymentBuilderTableProps> = ({
         (item) =>
           item.payouts?.map((payout) => ({
             recipient: item.recipientAddress || '',
-            claimDelay: BigNumber.from(item.claimDelay || '0')
-              .add(expendituresGlobalClaimDelay ?? '0')
-              .toString(),
+            claimDelay:
+              Number(item.claimDelay) +
+              Number(expendituresGlobalClaimDelay ?? '0'),
```
jakubcolony, line 173: Should use `BigNumber`

jakubcolony (changes requested):
Nice start @joanna-pagepro, please be aware that incoming claim delay (fetched from the DB) can be up to 2^256 so we need to do all the maths operations using `BigNumber`.

On a side note, allowing setting the claim delay as fractional number exposes us to rounding issues such as this one. It's not a code issue, as it converted 0.1111 hr to 400 mins, then 400 mins / 3600 = 0.11111111111 seconds. @arrenv something to be aware of I suppose.


https://github.com/JoinColony/colonyCDapp/assets/112586815/79535867-2e08-4ba9-baf4-afa738fb6361

arrenv:
> On a side note, allowing setting the claim delay as fractional number exposes us to rounding issues such as this one. It's not a code issue, as it converted 0.1111 hr to 400 mins, then 400 mins / 3600 = 0.11111111111 seconds. @arrenv something to be aware of I suppose.

@jakubcolony I am not understanding the rounding to 400 mins here? Could we not just convert it to seconds in all cases? i.e. 0.1111 * 3600 then 0.1111 / 3600?

jakubcolony:
@arrenv Apologies, I meant seconds! 

0.1111 hours * 3600 = 399.96 ≈ 400 seconds (what we send and store on chain)

Then back in the UI:

400 seconds / 3600 = 0.11111111111 hours

The solution could simply be to round to 4 decimal places when we display the delay too.

melyndav (commented):
Hey @joanna-pagepro just confirming if you're still seeing the validation as I am?

image

joanna-pagepro:
@melyndav no, I'm not seeing this. I've disabled the default form validation, but maybe it doesn't work the same on every browser. I will check it on Monday.

arrenv:
> @arrenv Apologies, I meant seconds!
> 
> 0.1111 hours * 3600 = 399.96 ≈ 400 seconds (what we send and store on chain)
> 
> Then back in the UI:
> 
> 400 seconds / 3600 = 0.11111111111 hours
> 
> The solution could simply be to round to 4 decimal places when we display the delay too.

Oh, I made an error in mine as well, haha. Makes sense the rounding of the seconds.

@joanna-pagepro Can we please round to 4 decimal places on the completed view in the UI as @jakubcolony has suggested.

joanna-pagepro:
@arrenv @jakubcolony can you review again?

jakubcolony (changes requested):
Codewise looking good, however there's an issue with claim delays starting with multiple zeros:

https://github.com/JoinColony/colonyCDapp/assets/112586815/07da7eb7-9b43-488d-a5bf-e3a7519d58ae

jakubcolony (changes requested):
Not sure what happened since last review but I started to see the native browser validation:

https://github.com/JoinColony/colonyCDapp/assets/112586815/d7d62779-4703-4142-8fc3-44fc7394c5cf

joanna-pagepro:
@jakubcolony it should be ok now

bassgeta (approved):
Nice one, works for minimum and maximum checks :ok_hand: 
image
LGTM!

jakubcolony (changes requested):
I'm unable to set the claim delay to 0 since the validation requires it to be at least `0.0001`. I'm not sure this is correct, you should not be able to set it to something such as `0.00000001` but it should still be possible to create a payment with no claim delay.

image

joanna-pagepro:
@jakubcolony can you check if it's ok now?

jakubcolony (changes requested):
Thanks for the updates @joanna-pagepro, can confirm I'm able to set the claim delay to `0` now.

Sometimes I'm seeing different claim delay in the completed action view than what I entered (e.g. `0.0001` becomes `0.0000`, `0.0002` becomes `0.0003`). I suspect this could be due to rounding but could you please check if there's anything we can do in the UI to prevent this?

https://github.com/JoinColony/colonyCDapp/assets/112586815/1d2817c1-86ab-4eb8-955d-7d79feff096c

I also noticed the 4 decimal places truncation in the completed action view also adds trailing zeros:
image

image

image

Ideally we would only show the significant figures, without trailing zeros.

joanna-pagepro:
@jakubcolony you get a different value in the form and later in the completed action because of the rounding up the number before sending it to the API. If it's smaller than 0.5h then you will always get 0. When I tried to remove this rounding, I've got this error: 
image
Can you help me with this?

jakubcolony:
@joanna-pagepro thanks for sharing, I think we won't be able to do anything about the rounding of anything less than 0.0002 as this rounds to 0 seconds on chain. In that case, my only comment would be around the truncation of trailing zeros (e.g. showing `0.5` instead of `0.5000`)

joanna-pagepro:
@jakubcolony @melyndav please check again

jakubcolony (dismissed):
Looks great now, thanks @joanna-pagepro 👍 

image

joanna-pagepro:
@melyndav can you review it again?

arrenv (changes requested):
@joanna-pagepro This one certainly has been a long one. Thanks for being persistent with it.

I tested a few scenarios with decimals, I have some odd behavior were decimals were validate to have a max limit with one number but not another.

e.g.
`0.0000000000111111111111111111111111111111111` validated
`00000000000000.0000000000000000001` did not validate

Also, I understand validation is on click, but can we still have a max width so that we don't have the field overflow over the end of the table, like in the screenshot.

image

joanna-pagepro:

arrenv (changes requested):
@joanna-pagepro Thank you. Unfortunately, the issue is now happening with integers.

image

joanna-pagepro:
@arrenv and now? :)

arrenv (approved):
@joanna-pagepro All looks good now, thank you!

PR #2442, opened 2024-05-24
fix: Show correct roles in the Completed Action component

### LINKED ISSUE #2303: Updated permissions do not show correctly on completed action
## Steps to reproduce

1. Give custom permissions to a user (for example both funding and administration).
2. The custom permissions toggles will accurately reflect the assigned permissions.
3. Update the custom permissions for the user (for example, remove funding and add root).
4. The custom permission toggles will not accurately reflect the assigned permissions.

https://github.com/JoinColony/colonyCDapp/assets/34915414/50adf23d-58e6-42f2-97bf-bacd44332176

This can also be seen when changing permission tiers:

1. Give 'Payer' permissions to a user.
2. The completed action screen will show that the user has been assigned the payer role.
3. Update the permissions for the user to 'Owner'.
4. The completed action screen will show the custom permission toggles and they will not accurately reflect the assigned permissions.

https://github.com/JoinColony/colonyCDapp/assets/34915414/9f2a7c8d-efb7-44af-9844-e10b02844172

## Expected behaviour

* The custom permission toggles should accurately reflect the assigned permissions.

## Actual behaviour

* The custom permission toggles do not accurately reflect the assigned permissions.

## Suggestions (optional)

Whenever permissions are updated, the block-ingestor creates a ColonyAction with the `SetUserRoles` type.

It appears that the roles stored here are currently only recording the change in permissions rather than the current permissions state. This means, if 'funding' is set to true in the first action, then it is left unchanged in the second action, it won't have changed and thus appears as false in the recorded action.

Either block-ingestor should be changed to store the roles as they currently are, or CDapp should be changed to fetch the ColonyRole permissions or work through the history of changes to reach the correct state of permissions. If there is no reason for block-ingestor to store the change, it is likely simpler to make the change there - the only complication is that the contract events only emit the changes in permissions, so the previous set of permissions would need to be fetched first to accurately reflect the current state.

rumzledz:
## Description

The Completed Action component for Permission updates should now show the most up-to-date information for a user.

| Adding permissions | Removing permissions |
| ---- | ---- |
| Update Permissions | remove permissions |

## Testing

**Prerequisite**: Select a user who already has at least one custom permission enabled or enable at least 1 custom permission for a user, we'll call this user as (USER).

### Adding new custom permissions

1. Bring up the new Action form and choose Manage Permission
2. Select (USER)
3. Enable two more custom permissions
4. Submit the form
5. Verify that all the permissions for this user are correctly reflected in the Completed Action component

### Removing new custom permissions

1. Bring up the new Action form and choose Manage Permission
2. Select (USER)
3. Disable at least 1 custom permission
4. Submit the form
5. Verify that all the permissions for this user are correctly reflected in the Completed Action component

## Diffs

**New stuff** ✨

- A new GraphQL query for fetching a user's historic role data

**Changes** 🏗

- The Completed Action page for Role updates now take into account historic role information on top of the recently toggled role(s) to build a complete view of a user's roles.

Resolves #2303


src/components/v5/common/CompletedAction/partials/SetUserRoles/SetUserRoles.tsx
```diff
@@ -81,8 +97,22 @@ const SetUserRoles = ({ action }: Props) => {
...
+  const { data: historicRoles } = useGetColonyHistoricRoleRolesQuery({
+    variables: {
+      id: `${colonyAddress}_${fromDomain?.nativeId}_${recipientAddress}_${blockNumber}_roles`,
+    },
+    fetchPolicy: 'cache-and-network',
+  });
+
+  const userColonyRoles = transformActionRolesToColonyRoles(
+    roles,
+    historicRoles?.getColonyHistoricRole,
```
davecreaser, line 113: Since this function expects two arguments, and the second one here is conditional, it may be a good idea to add a fallback to an empty object or something?

davecreaser: Ah, ignore me, I just looked into it and the type includes null or undefined :)

davecreaser (dismissed):
LGTM! Nice job 👍

src/components/v5/common/CompletedAction/partials/SetUserRoles/SetUserRoles.tsx
davecreaser: You may just need to consider that inside the function you call `const combinedRoles = { ...historicRoles };`which could be unhappy if historicRoles is indeed `null`.

rumzledz: Thanks for approving @davecreaser ! Yep that should be fine because in that scenario, `combinedRoles` will just equal to an empty object `{ }` then it will be populated with the new roles that were toggled -- the function will still pretty much behave like it did before 👌

src/components/v5/common/CompletedAction/partials/SetUserRoles/SetUserRoles.tsx
```diff
@@ -45,11 +50,22 @@ interface Props {
...
 const transformActionRolesToColonyRoles = (
   roles: ColonyActionRoles | null | undefined,
+  historicRoles: GetColonyHistoricRoleRolesQuery['getColonyHistoricRole'],
 ): ColonyRole[] => {
   if (!roles) return [];
-  const roleKeys = Object.keys(roles);
+
+  const combinedRoles = { ...historicRoles };
+
+  for (const [key] of Object.entries(roles)) {
```
bassgeta, line 59: I think we can do `[key, value]` here since we're using `Object.entries`

bassgeta (dismissed):
It works, just dropped a small comment for a loop. Gj :muscle:

src/components/v5/common/CompletedAction/partials/SetUserRoles/SetUserRoles.tsx
rumzledz: @bassgeta yep but in this scenario, I only need the `key` 🔑

rumzledz: @bassgeta I think my brain rotted 😂 I am in fact referencing values from `roles` ! will push a change

rumzledz:
I've pushed another commit to clean up that `for` loop a little bit @bassgeta @davecreaser . Can I please get another review? Cheers!

bassgeta (approved):
Still legit! :muscle:

davecreaser (approved):
All good! 🚀

src/components/v5/common/CompletedAction/partials/SetUserRoles/SetUserRoles.tsx
```diff
@@ -45,11 +50,22 @@ interface Props {
...
+
+  const combinedRoles = { ...historicRoles };
+
+  for (const [key, value] of Object.entries(roles)) {
+    if (value !== null) {
+      combinedRoles[key] = value;
+    }
+  }
+
+  const roleKeys = Object.keys(combinedRoles);
```
jakubcolony, line 65: I believe (please double check though) historic roles entry is a snapshot of the entire permissions state so we should be able to use it without merging with with `roles` - which would make that first argument redundant.

src/components/v5/common/CompletedAction/partials/SetUserRoles/SetUserRoles.tsx
```diff
@@ -81,8 +97,22 @@ const SetUserRoles = ({ action }: Props) => {
...
     isMotion,
     annotation,
+    blockNumber,
+    colonyAddress,
   } = action;
-  const userColonyRoles = transformActionRolesToColonyRoles(roles);
+
+  const { data: historicRoles } = useGetColonyHistoricRoleRolesQuery({
+    variables: {
+      id: `${colonyAddress}_${fromDomain?.nativeId}_${recipientAddress}_${blockNumber}_roles`,
```
jakubcolony, line 106: Please add a util to construct this ID similar to others in `utils/databaseId`

Issue #2414, opened 2024-05-21
Fix Payment builder table loading state


arrenv:
Teamwork task: [Fix Payment builder table loading state](https://tw.auroracreation.com/#/tasks/15917761)

## Steps to reproduce

1. Create a Payment builder action.
2. Add a few rows to the payment.
3. Create the payment, keeping an eye on the payment builder table.

## Expected behaviour

* You should see the skeleton loader with 3 rows until all the table data has loaded.

## Actual behaviour

* You see default 0x0000...0000 recipient values on first load.
* Only one row loads after the skeleton loader disappears, until the rest of the rows load in.

![image](https://github.com/JoinColony/colonyCDapp/assets/33682027/814f83f7-bb42-462a-9322-a26630a8a95e)

![image](https://github.com/JoinColony/colonyCDapp/assets/33682027/b00f2f26-8d65-498a-8549-7f9d9c811165)

![image](https://github.com/JoinColony/colonyCDapp/assets/33682027/628e35ee-a29d-4115-a7e4-5c9717984e0c)

PR #2451, opened 2024-05-28
fix: payment builder table loading state

joanna-pagepro:
## Description

Payment builder table loading state

## Testing

* Step 1. Create a Payment builder action.
* Step 2. Add a few rows to the payment.
* Step 3. Create the payment, keeping an eye on the payment builder table.

## Diffs

**New stuff** ✨

* 

**Changes** 🏗

* loading state passed to `RecipientField` and `UserPopover` components

**Deletions** ⚰️

* 

## TODO

- 

Resolves https://github.com/JoinColony/colonyCDapp/issues/2414

### DIFF
--- a/src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/PaymentBuilderTable/PaymentBuilderTable.tsx
+++ b/src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/PaymentBuilderTable/PaymentBuilderTable.tsx
@@ -4,6 +4,7 @@ import { BigNumber } from 'ethers';
 import React, { type FC, useMemo } from 'react';
 
 import { useColonyContext } from '~context/ColonyContext/ColonyContext.ts';
+import { useMemberContext } from '~context/MemberContext/MemberContext.ts';
 import { type ExpenditureSlotFragment, ExpenditureStatus } from '~gql';
 import { useTablet } from '~hooks';
 import useCurrentBlockTime from '~hooks/useCurrentBlockTime.ts';
@@ -50,17 +51,23 @@ const useGetPaymentBuilderColumns = ({
     [blockTime, finalizedTimestamp, slots],
   );
 
+  const { loading: isColonyContributorDataLoading } = useMemberContext();
+
+  const isDataLoading = isLoading || isColonyContributorDataLoading;
+
   return useMemo(
     () => [
       paymentBuilderColumnHelper.accessor('recipient', {
         enableSorting: false,
         header: formatText({ id: 'table.row.recipient' }),
-        cell: ({ row }) => (
-          <RecipientField
-            isLoading={isLoading}
-            address={row.original.recipient}
-          />
-        ),
+        cell: ({ row }) =>
+          isLoading ? (
+            <div className="flex w-full items-center">
+              <div className="h-4 w-full overflow-hidden rounded skeleton" />
+            </div>
+          ) : (
+            <RecipientField address={row.original.recipient} />
+          ),
         footer: hasMoreThanOneToken
           ? () => (
               <span className="flex min-h-[1.875rem] items-center text-xs text-gray-400">
@@ -79,16 +86,24 @@ const useGetPaymentBuilderColumns = ({
         header: formatText({ id: 'table.row.amount' }),
         footer: hasMoreThanOneToken
           ? () => (
-              <PaymentBuilderPayoutsTotal
-                data={dataRef.current}
-                itemClassName="justify-end md:justify-start"
-                buttonClassName="justify-end md:justify-start"
-              />
+              <>
+                {isLoading ? (
+                  <div className="flex w-3/4 items-center">
+                    <div className="h-4 w-full overflow-hidden rounded skeleton" />
+                  </div>
+                ) : (
+                  <PaymentBuilderPayoutsTotal
+                    data={dataRef.current}
+                    itemClassName="justify-end md:justify-start"
+                    buttonClassName="justify-end md:justify-start"
+                  />
+                )}
+              </>
             )
           : undefined,
         cell: ({ row }) => (
           <AmountField
-            isLoading={isLoading}
+            isLoading={isDataLoading}
             amount={row.original.amount}
             tokenAddress={row.original.tokenAddress}
           />
@@ -104,7 +119,7 @@ const useGetPaymentBuilderColumns = ({
             Number(row.original.claimDelay) / 3600,
           );
 
-          return !isLoading ? (
+          return !isDataLoading ? (
             <span className="text-md text-gray-900">
               {formatText(
                 { id: 'table.column.claimDelayField' },
@@ -147,13 +162,14 @@ const useGetPaymentBuilderColumns = ({
           ]
         : []),
     ],
+    // Rule added so that we won't have to pass data or dataRef as dependencies to avoid unnecesary re-renders
+    // eslint-disable-next-line react-hooks/exhaustive-deps
     [
       claimablePayouts,
-      data.length,
-      dataRef,
       fetchCurrentBlockTime,
       finalizedTimestamp,
       hasMoreThanOneToken,
+      isDataLoading,
       isLoading,
       isTablet,
       status,
@@ -165,6 +181,7 @@ const PaymentBuilderTable: FC<PaymentBuilderTableProps> = ({
   items,
   status,
   finalizedTimestamp,
+  isLoading,
 }) => {
   const isTablet = useTablet();
   const {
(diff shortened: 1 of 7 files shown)


marcocolony (changes requested):
Hey @joanna-pagepro, thanks for looking into this!

The default address no longer shows up, but the loading state looks to be a bit inconsistent from what I have experienced. 

image

Export-1716900885744

Export-1716900885744

Here's the expected behaviour for the loading state:
- You should see the skeleton loader with 3 rows until all the table data has loaded.

Actual current  behaviour:
- Data seems to be displayed as it is loading 
- Skeleton loading appears only for the exact data that has yet to be loaded
- If there's a row with data yet to load, all the other rows with complete data will still be visible

davecreaser (changes requested):
Nice one! Code-wide all good apart from one minor comment.

The skeleton loading state also looks good to me, but we might need to check with @JoinColony/product about Marco's comments on the exact requirements here.

jakubcolony (changes requested):
Code-wise looks good (other than what Dave mentioned).

I'm also seeing the data appear incrementally and skeleton only showing on the user popover which is expected as the on-chain events are processed sequentially.

To show skeleton loader until all data is loaded you could store the expected count of payouts when expenditure is created and display the loader until the `expenditure` object from API has matching count of slots. You would need to check all of those slots have recipients set, and payouts with expected amount and claim delay. Feel free to ping me if you require more details.

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/PaymentBuilderTable/PaymentBuilderTable.tsx
```diff
@@ -50,14 +51,18 @@ const useGetPaymentBuilderColumns = ({
     [blockTime, finalizedTimestamp, slots],
   );
 
+  const { loading: isColonyContributorDataLoading } = useMemberContext();
```
bassgeta, line 54: I actually think the solution you've implemented in `UserInfoPopover` fetches data from the same source, so we don't need to pass `isLoading` to `UserPopover`.
Maybe we could try removing that property. We can also try and see what happens if we revert the loading flag changes here (maybe it won't work though)

bassgeta (changes requested):
Overall the changes are quite nice, we are slowly consolidating these popover components!
I dropped one comment, I think that we can stop passing `isLoading` to `UserPopover` and `UserInfoPopover`, since you made it that `UserInfoPopover` fetches data both from the member context and graphql

jakubcolony (dismissed):
> @jakubcolony The problem was that data related to the recipient was fetched inside the UserInfoPopover component, separately from the expenditure data. I used member context to get rid of this effect. Let me know if my solution is ok, or if I should revert the changes. I couldn't see any other issues, so if you see anything else, please also let me know.

I think that's fine, but please take a look at Kerry's comment as you might not need to pass the additional `isLoading` prop.

Nice work on this 👍 


https://github.com/JoinColony/colonyCDapp/assets/112586815/f2302b16-4e10-484f-b4e7-a7d0ddbd8b62

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/PaymentBuilderTable/PaymentBuilderTable.tsx
joanna-pagepro: @bassgeta it's used to pass the loading state of the data from the table, so I think this prop is also needed so that skeleton loader will be shown for all of the table items simultaneously. I also want to show the loading state on other items from the table when user data is being fetched.

joanna-pagepro:
@davecreaser @bassgeta can you review this one again?

jakubcolony (approved):

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/PaymentBuilderTable/PaymentBuilderTable.tsx
```diff
@@ -154,7 +159,7 @@ const useGetPaymentBuilderColumns = ({
       fetchCurrentBlockTime,
       finalizedTimestamp,
       hasMoreThanOneToken,
-      isLoading,
```
davecreaser, line 157: On line 65 (inside the useMemo) you still have `isLoading={isLoading}` which means there's now a missing dependency in the dependency array of this useMemo call.

You'll need to either keep isLoading as a dependency or change that line to `isLoading={isDataLoading}` to fix it. I assume it's just a missed thing and the second option is what you intended!

davecreaser (changes requested):
All good to me except one small thing!

melyndav (approved):
This looks great @joanna-pagepro, thanks for your work on making this fix. 💯

jakubcolony (dismissed):
re-approving

iamsamgibbs (dismissed):
Looks good to me! All working as intended.

https://github.com/JoinColony/colonyCDapp/assets/34915414/dc14d273-2971-443a-82c8-4baf8c0ae5fa

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/PaymentBuilderTable/PaymentBuilderTable.tsx
```diff
@@ -147,13 +152,14 @@ const useGetPaymentBuilderColumns = ({
           ]
         : []),
     ],
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
bassgeta, line 155: Why are we disabling exhaustive deps here? I think we should not do it when possible, or just write a comment why. Most of the times, we want to keep this rule on.

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/PaymentBuilderTable/partials/RecipientField/RecipientField.tsx
```diff
@@ -1,37 +1,16 @@
...
-      )}
-    </>
-  );
-};
+const RecipientField: FC<RecipientFieldProps> = ({ address, isLoading }) => (
+  <UserPopover
+    size={18}
+    walletAddress={address}
+    withVerifiedBadge
+    isLoading={isLoading}
```
bassgeta, line 12: This is just a major issue with how the popovers currently work, but this is further complicating their behaviour.
We are planning a refactor of them, but could we please for the moment just display a skeleton if `isLoading` is true, and stop propagating this flag to both `UserInfoPopover` and `UserPopover` components?

bassgeta (changes requested):
There's some tech debt that's impacting this PR, and I don't want to further increase it.
Would it be possible for the moment to just display the loader manually for the popovers? It can also be extracted into one, I would just not like see this prop propagation going on

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/PaymentBuilderTable/PaymentBuilderTable.tsx
joanna-pagepro: @bassgeta sorry, but I don't understand. The task is to add consistent styles to the table when any of its data is being loaded. There was some inconsistency here because `UserInfoPopover` had the loading state added inside and the skeleton was shown a bit longer than for other columns. Right now I'm getting information about two things - one about loading action data (via `isLoading` arg) and the second one about loading contributor data (`isColonyContributorDataLoading`). I know that `isColonyContributorLoading` is already checked in `UserInfoPopover` and that's why I'm passing there only the loading state of the action data.

bassgeta: I am sorry, I should have been more thorough in my comment.
Basically the changes done here introduce a short circuit in `UserPopover` and `UserInfoPopover` to display a skeleton when a passed `isLoading` prop is `true`. At the same time, the component will still fetch the contributor data, even if you just want to display that skeleton.
While we are now conveniently displaying the loading state, since we are looking at the same loading flags, it may be abused in the future and lead to faulty behaviour. 
And generally if you want a different component rendered from a child, you don't introduce a short circuit, that's not a good pattern.
If you want the parent to render a loader instead of the child component, it should render the loader itself, not pass an `isLoading` prop to an uncontrolled component that fetches its own data.

Again, this issue stems from the fact that our `UserInfoPopover` component does way too much, but this change adds to its tech debt. If you really feel like not doing this, just add a @TODO comment to remove the flag and where it is used, and we will sort it when refactoring this component :)

joanna-pagepro:
@bassgeta can you check if it's ok now?

iamsamgibbs (approved):
Still looks good to me!


https://github.com/JoinColony/colonyCDapp/assets/34915414/fcfba291-8a1b-46ad-8422-96500c7b4361

melyndav (approved):
This one still looks good to me @joanna-pagepro! Thanks again for your hard work on this 🥇

arrenv (approved):

Issue #2117, opened 2024-04-02
Verified members link not working in MemberPopover


arrenv:
Teamwork task: [Verified members link not working in MemberPopover](https://tw.auroracreation.com/#/tasks/15912194)

## Steps to reproduce
**Members page**
1. Go to the members page.
3. Click on a member who is not verified to open their Memberpopover.
4. Click on the `Add as verified member` link.
5. The button won't do anything.

**Action panel**
1. Create a new action.
2. Select the Simple payments action type.
4. Select a member who is not verified.
5. Click on the yellow warning icon.
6. You will have the same issue.

## Expected behaviour

* The `Add as verified member` link should open the `Manage verified members` action type with the user pre-selected with the Add member option selected.

## Actual behaviour

* The  `Add as verified member` link does nothing.

![image](https://github.com/JoinColony/colonyCDapp/assets/33682027/dd2c495b-3410-4985-b66c-49c301cf81bb)

![image](https://github.com/JoinColony/colonyCDapp/assets/33682027/ae1ab64b-e7a9-4b84-8ae2-ff6e75058426)

PR #2458, opened 2024-05-30
fix: add verified member from user popover

CzarekDryl:
## Testing

Members page

* Go to the members page.
* Click on a member who is not verified to open their Memberpopover.
* Click on the Add as verified member link.

Action panel

* Create a new action.
* Select the Simple payments action type.
* Select a member who is not verified.
* Click on the yellow warning icon.

## Diffs

**Changes** 🏗

* `Add as verified member` button will now open sidebar with selected action and member.

Resolves [#2117](https://github.com/JoinColony/colonyCDapp/issues/2117)

### DIFF
--- a/src/components/v5/shared/UserInfoPopover/partials/UserNotVerified.tsx
+++ b/src/components/v5/shared/UserInfoPopover/partials/UserNotVerified.tsx
@@ -22,7 +22,7 @@ const UserNotVerified: FC<UserNotVerifiedProps> = ({
 }) => {
   const {
     actionSidebarToggle: [
-      ,
+      isActionSidebarOpen,
       { toggleOn: toggleActionSidebarOn, toggleOff: toggleActionSidebarOff },
     ],
   } = useActionSidebarContext();
@@ -35,7 +35,11 @@ const UserNotVerified: FC<UserNotVerifiedProps> = ({
         <button
           type="button"
           onClick={() => {
-            toggleActionSidebarOff();
+            const timeout = isActionSidebarOpen ? 500 : 0;
+
+            if (isActionSidebarOpen) {
+              toggleActionSidebarOff();
+            }
 
             setTimeout(() => {
               onClick?.();
@@ -44,7 +48,7 @@ const UserNotVerified: FC<UserNotVerifiedProps> = ({
                 members: [{ value: walletAddress }],
                 manageMembers: ManageMembersType.Add,
               });
-            }, 500);
+            }, timeout);
           }}
         >
           {formatText({ id: 'add.verified.member' })}
--- a/src/components/v5/shared/UserInfoPopover/partials/UserNotVerified.tsx
+++ b/src/components/v5/shared/UserInfoPopover/partials/UserNotVerified.tsx
@@ -12,11 +12,13 @@ const displayName = 'v5.UserInfoPopover.partials.UserNotVerified';
 export interface UserNotVerifiedProps {
   description?: React.ReactNode;
   walletAddress: string;
+  onClick?: () => void;
 }
 
 const UserNotVerified: FC<UserNotVerifiedProps> = ({
   description,
   walletAddress,
+  onClick,
 }) => {
   const {
     actionSidebarToggle: [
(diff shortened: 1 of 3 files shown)


melyndav (dismissed):
Works as expected @CzarekDryl thank you!

iamsamgibbs (changes requested):
Nice start to this ticket, but I'm having a couple of issues. The user popover stays open when opening the sidebar.

Screenshot 2024-05-30 at 16 50 45

This is particularly bad on mobile.

Screenshot 2024-05-30 at 16 50 59

Clicking the exclamation mark in the recipient select opens the select menu rather than the user popover.

Screenshot 2024-05-30 at 16 53 53

src/components/v5/shared/UserInfoPopover/partials/UserNotVerified.tsx
```diff
@@ -1,22 +1,49 @@
...
+          onClick={() => {
+            toggleActionSidebarOff();
+
+            setTimeout(() => {
+              toggleActionSidebarOn({
+                [ACTION_TYPE_FIELD_NAME]: Action.ManageVerifiedMembers,
+                members: [{ value: walletAddress }],
+                manageMembers: ManageMembersType.Add,
+              });
+            }, 500);
```
rumzledz, line 44: Hey @CzarekDryl is the timeout necessary here? I'm new to the codebase so just trying to understand the reasoning behind it

bassgeta: +1 on this question, maybe there was an issue here which could be solved with `requestAnimationFrame`?

bassgeta (changes requested):
I have the same question as Romeo regarding the timeout
image
and I can confirm the popup doesn't close when the sidebar is opened.

I think you can safely pass a `closePopup` function to the component and call it after opening the sidebar.
Other than that, nice job, simple and clean solution!

iamsamgibbs (changes requested):
Thank you for fixing the issue with the popover staying open.

Clicking the yellow exclamation mark still does not open the popover though:

https://github.com/JoinColony/colonyCDapp/assets/34915414/ef455f41-1206-4018-af4f-fd637bbd7875

src/components/v5/shared/UserInfoPopover/partials/UserNotVerified.tsx
iamsamgibbs: I think this will need reworking to remove the timeout, the 500ms delay makes this feel very unresponsive:

https://github.com/JoinColony/colonyCDapp/assets/34915414/fb1f38de-5868-450a-99aa-a905b9cdf1e5

With the timeout removed for comparison:

https://github.com/JoinColony/colonyCDapp/assets/34915414/e7246009-f50b-4202-9365-570f84e2b9b7

I assume this is purely to handle the scenario where the sidebar is already open and it needs to switch to a different action type? I imagine it would be okay for the form to instantly switch to another action without the sidebar closing and reopening. If the closing and reopening is a requirement, then maybe you can add a check here to see where it is being called from. Or might just need to add something to `toggleActionSidebarOff` to detect when the animation has finished / needs to run at all.

iamsamgibbs (changes requested):
This is working now!

https://github.com/JoinColony/colonyCDapp/assets/34915414/4d6253a1-2f9b-4635-a731-f859c0806294

But I've left a comment regarding the timeout.

src/components/v5/shared/UserInfoPopover/partials/UserNotVerified.tsx
CzarekDryl: @iamsamgibbs This should work better now, I've removed timeout when sidebar is closed.

iamsamgibbs (approved):
Thanks for the change, this is working a lot better.


https://github.com/JoinColony/colonyCDapp/assets/34915414/a5c709b5-8146-4688-a20d-48588784e81e


https://github.com/JoinColony/colonyCDapp/assets/34915414/53be1972-5515-4cf3-b1b8-e3e2d8a4a99a

bassgeta (approved):
Nice one with this, clicking the popover from the members tab correctly opens the sidebar, and sweitching the action type when clicking on the popover in something like simple payments also works.
Good work on this, let's merge it :ship:

PR #2470, opened 2024-06-03
Maintenance: Add ESLint for max function parameters

### LINKED ISSUE #2466: Enable an ESlint rule for max-params
## Description

The main motivation is to avoid having functions which can appear ambiguous when used. This will also help decrease the chances of human error when passing arguments to a function.

I guess we can take a vote on what our threshold is but I want to suggest a maximum of 3 params per function before we decide to encapsulate all the params into an object param instead.

This does not mean that we cannot use an object param even if we only need a single value. Please do so it if helps with code readability.

We should enforce this with the [corresponding ESLint rule](https://eslint.org/docs/latest/rules/max-params).

## Without object parameters

```js
function calculateTotalPrice(price, taxRate, discount, shipping) {
    return (price + (price * taxRate) - discount + shipping);
}

// Usage
const totalPrice = calculateTotalPrice(100, 0.07, 10, 5);
```

## With object parameters

```js
function calculateTotalPrice({ price, taxRate, discount, shipping }) {
    return (price + (price * taxRate) - discount + shipping);
}

// Usage
const totalPrice = calculateTotalPrice({
    price: 100,
    taxRate: 0.07,
    discount: 10,
    shipping: 5
});
```

- [ ] We add an ESLint rule for max-params in the `.eslintrc.js` file: `'max-params': ['error', { max: 3 }],`
- [ ] Running `npm run lint` does not report any errors

rumzledz:
## Description

This PR introduces a new ESLint rule named `max-params`, which restricts functions to a maximum of 3 parameters. Moving forward, if a function requires more than 3 parameters, the new standard is to use an object parameter instead.

## Testing

Critical paths to test:
- Installing an extension
- Making a Simple Payment and using the Permissions and Reputation decision methods
- Removing and Awarding Reputation for a user. Reputation can be quickly mined by visiting this URL: [http://127.0.0.1:3001/reputation/monitor/toggle](http://127.0.0.1:3001/reputation/monitor/toggle)
- Transactions tab should still show correct information
- Creating a new Colony

## Diffs

**New stuff** ✨

- A new ESLint rule called [max-params](https://eslint.org/docs/latest/rules/max-params) which will throw an error for functions that have more than 3 parameters.

**Changes** 🏗

There is no need to fill in parameters in between, as we are passing values through object fields. This approach should reduce human error. An example of this change is:

| Before | After |
| -------- | -------- |
| image | image |

**Deletions** ⚰️

* Removed an unused `useActionDialogStatus` hook
* Removed an unused `setupEnablingGroupTransactions` function
* Removed an unused param `error` from the `useGroupedTransactionContent` hook

## Todo

- [ ] After this PR is merged, merge the [max-params frontend living standard PR](https://github.com/JoinColony/ColonyFrontEndLivingStandard/pull/24)

Resolves #2466


src/components/common/Extensions/UserHub/partials/TransactionsTab/partials/GroupedTransactionContent.tsx
```diff
@@ -56,16 +56,15 @@ const GroupedTransactionContent: FC<GroupedTransactionContentProps> = ({
     toggleCancelConfirmation,
     handleCancelTransaction,
     canBeSigned,
-  } = useGroupedTransactionContent(
+  } = useGroupedTransactionContent({
     id,
-    error,
```
rdig, line 61: why was `error` removed from here? not used ?

src/components/common/Extensions/UserHub/partials/TransactionsTab/partials/hooks.tsx
```diff
@@ -8,18 +9,27 @@ import {
...
   transactionRetry,
 } from '~redux/actionCreators/index.ts';
+import { type TransactionId } from '~redux/immutable/index.ts';
 import Toast from '~shared/Extensions/Toast/index.ts';
+import { type ExtendedClientType } from '~types/transactions.ts';
 
-export const useGroupedTransactionContent = (
+export const useGroupedTransactionContent = ({
   id,
-  error,
```
rdig, line 15: same as above

rdig (approved):
Looks good. Ran `lint` and `typecheck` and no errors thrown, so I'd say this one is ready

Screenshot from 2024-06-04 23-37-09
Screenshot from 2024-06-04 23-38-04

jakubcolony (approved):
Ran `create-data` and clicked through the pages, all seems to be working fine. Nice work!

iamsamgibbs (approved):
This looks good to me! Tested all the suggested paths and was able to lint and typecheck without issue.

Screenshot 2024-06-05 at 09 23 01

src/components/common/Extensions/UserHub/partials/TransactionsTab/partials/GroupedTransactionContent.tsx
rumzledz: Yep! It's unused here

src/components/common/Extensions/UserHub/partials/TransactionsTab/partials/hooks.tsx
rumzledz: Yep! Another function param whose day has come!

PR #2472, opened 2024-06-03
Feat: Handle network fee validation on payment

### LINKED ISSUE #2279: Handle network fee validation on payment
## Steps to reproduce

1. Create a payment action that is the exact amount of total funds in the colony e.g. total balance is 10, make a payment of 10.
2. The transaction will fail with no errors in the UI.
3. The action stays in the pending state.
4. If you try again with 9.9999 it will go through.

_Bonus: we also need to add the network fee to the form, please let product know if you want to tackle this as well in this PR._

## Expected behaviour

* You should not be able to make a payment if there is not enough to cover the gas fees and the payment.
* A not enough funds validation message should be shown with the message: "Not enough funds to cover the payment and network fees"

image

## Actual behaviour

* The transaction is reverted with the error: "execution reverted: arithmetic underflow or overflow"

image

jakubcolony:
## Description

Amended payment action validation to take network fee into consideration.

I noticed the same issue affected payment builder so that got fixed too.

## Testing

Create a payment (both simple and payment builder) with amount very close to available balance. Validation error should appear. 

## Diffs

**Changes** 🏗

* `hasEnoughFundsValidation` accepts optional `networkInverseFee` param, calculates and adds the fee to the amount if specified
* `hasEnoughFundsValidation` accepts params object instead of arguments

Resolves #2279


iamsamgibbs (changes requested):
Nice work on this! Working as expected for simple payments:

Screenshot 2024-06-04 at 09 24 31

But I'm having an issue when submitting with Payment Builder:

Screenshot 2024-06-04 at 09 25 09

jakubcolony:
@iamsamgibbs thanks, should be fixed

bassgeta (approved):
Nice one, like that we use params for that bigboi helper now
image
Payment builder also displays the error
image


Just a question @jakubcolony , is it fine that the payment builder verified this on submit and simple payment verifies it on change?

iamsamgibbs (approved):
Thanks for the fix Jakub, all working and looks good to me!

Screenshot 2024-06-05 at 08 54 39
Screenshot 2024-06-05 at 08 55 17

iamsamgibbs (approved):

bassgeta (approved):
LGTM :ship:

Issue #2434, opened 2024-05-23
Staking Advanced Payments Extension


arrenv:
Teamwork task: [Staking Advanced Payments Extension](https://tw.auroracreation.com/#/tasks/15924180)

## Related issues/PR's

- [ ] https://github.com/JoinColony/colonyCDapp/pull/2386

### Story

- As a USER, I would like to be able to create advanced payments by staking.
- As a USER, I would like to be able to install, deprecate and uninstall the Staking Advanced Payments extension.
- As a USER, I would like to see know what the Staking Advanced Payment is and what is it used for.
- As a USER, I would like to be able to manage the Staking Advanced Payments extension.

### Description

Users will be able to create advanced payments by staking as a new action type to be used within a colony. Enabling this extension will add the new decision method “Staking” to Advanced Payment types including Payment builder, Split payments, and Staged payments to allow anyone to start the payment process by staking their own tokens.

### Design

Extension details page - [Figma Link](https://www.figma.com/design/6qONTE1uD9K4yoIbfFzidg/Extensions?node-id=736-13870&t=vPkF3ZfomdCFcufq-4)

![image](https://github.com/JoinColony/colonyCDapp/assets/33682027/503d535b-9269-4a13-963c-c2f0850cb04e)

Settings page - [Figma link](https://www.figma.com/design/6qONTE1uD9K4yoIbfFzidg/Extensions?node-id=1188-11839&t=vPkF3ZfomdCFcufq-4)

![image](https://github.com/JoinColony/colonyCDapp/assets/33682027/7190b23a-ba96-46b6-a599-b5108d6cadd0)

Uninstall warning - [Figma link](https://www.figma.com/design/6qONTE1uD9K4yoIbfFzidg/Extensions?node-id=1188-11839&t=vPkF3ZfomdCFcufq-4)

![image](https://github.com/JoinColony/colonyCDapp/assets/33682027/9db26b48-4a0b-4252-89ce-eabb0a36ffe2)

## Acceptance Criteria

- **General requirements**
    - Should be in alignment with desktop/tablet/mobile design from Designs.
    - Should use existing components and styles where possible for consistency.

- **Extension tabs update**
    - Add to all extensions.
    - Move the tabs up the top.

- **Extension details page:**
    - The extension icon and name are correct as of the design.
    - Extension type:
        - Payments.
    - Overview:
        - Description
            
[...]

PR #2340, opened 2024-05-08
feat: create staked expenditure

joanna-pagepro:
## Description

Creating a Payment Builder action with the `Staking` decision method, when the Staked expenditure extension is enabled.

## Testing

* Step 1. Make sure that the Staked expenditure extension is installed and enabled and create a Payment builder action with the `Staking` decision method.
* Step 2. When a modal appears, click the `Create payment` button

## Diffs

**New stuff** ✨

* `CreateStakedExpenditureModal` component
* `FinalizeWithStakingInfo` component

**Changes** 🏗

* `ActionButtons` component - `onSubmitClick` prop added to pass the action to show a modal instead of submit
* `ActionSidebarContent`component - added `CreateStakedExpenditureModal` modal
* `DecisionMethodField` and `DecisionMethodRow` components - added `Staking` decision method
* `PaymentBuilderWidget` - if the expenditure is staked `FinalizeWithStakingInfo` is shown instead of `FinalizeWithPermissionsInfo` in `Create` and `Review` steps
* `Button` component - div wrapper added to the loader

**Deletions** ⚰️

* 

## TODO

Contributes to https://github.com/JoinColony/colonyCDapp/issues/2344

PR #2474, opened 2024-06-04
Feat/15924180 staking extension page

joanna-pagepro:
## Description

Details page for staked expenditure extension

## Testing

* Step 1. Go to `Extensions` page
* Step 2. Click on `Staked Expenditure`
* Step 3. Install the extension
* Step 4. Change `Required stake` percentage value and click `Save changes` - button should appear if the field was edited
* Step 5. Deprecate the extension
* Step 6. Reenable extension
* Step 7. Deprecate the extension again and the uninstall

## Diffs

**New stuff** ✨

* SaveChangesButton - button is shown when the stake fraction is edited
* StakedExpenditureSettings - form visible for the user with right permissions and when the extension is enabled
* getTextChunks - function with reusable chunks

**Changes** 🏗

* InstallButton - useEffect with handleEnableExtension function to enable staked expenditure extension right after installing it
* ExtensionMethods.SAVE_CHANGES added to waitForDbAfterExtensionAction function
* ExtensionDetailsPage components updated to show staked expenditure extension related items
* SpecialInput - updated so that `onChange` will be passed from `useController` hook and not set separately - now all of the formStates are updated correctly

**Deletions** ⚰️

* 

## TODO

- 

Resolves https://github.com/JoinColony/colonyCDapp/issues/2434


src/components/frame/Extensions/pages/ExtensionDetailsPage/ExtensionDetailsPage.tsx
```diff
@@ -82,50 +102,71 @@ const ExtensionDetailsPage: FC = () => {
...
   });
 
   const SetupComponent = SetupComponentMap[extensionData.extensionId];
 
+  const actionType =
+    isStakedExpenditureExtension &&
+    extensionData.isInitialized &&
+    extensionData.isEnabled
+      ? ActionTypes.SET_STAKE_FRACTION
+      : ActionTypes.EXTENSION_ENABLE;
```
iamsamgibbs, line 115: There will be other extensions with a settings tab like this, so can this be made more generic to support other action types depending on the extension rather than a ternary?

src/components/frame/Extensions/pages/ExtensionDetailsPage/ExtensionDetailsPage.tsx
```diff
@@ -69,11 +77,23 @@ const ExtensionDetailsPage: FC = () => {
...
+      isStakedExpenditureExtension &&
+      extensionData.isInitialized &&
+      'params' in extensionData &&
+      extensionData.params?.stakedExpenditure?.stakeFraction
+        ? {
+            stakeFraction: getStakeFraction(
+              extensionData.params.stakedExpenditure.stakeFraction,
+            ),
+          }
+        : initialValues,
```
iamsamgibbs, line 96: Same as my other comment on this file, can we make this more reusable?

iamsamgibbs (changes requested):
Nice start on this, a lot of big changes in here and lots of useful stuff. I've put in a couple of comments in one file of where this could be made more reusable for other extensions, looks like there are some spots in the other files that could be adapted in the same way.

I was able to deprecate, uninstall and reinstall the extension without issue.

A couple of issues. The 'save changes' button causes some elements to move about:

https://github.com/JoinColony/colonyCDapp/assets/34915414/88ea7419-2a5d-41d0-97e3-12d320e76c0f

This spinner stays for ages after saving, eventually there is an error message, but after refreshing the value is persisted, so it seems like it is saving correctly.

Screenshot 2024-06-06 at 10 04 17
Screenshot 2024-06-06 at 10 05 27

joanna-pagepro:
@iamsamgibbs can you check if everything is ok now?

src/components/frame/Extensions/pages/ExtensionDetailsPage/utils.tsx
```diff
@@ -134,50 +149,85 @@ export const waitForDbAfterExtensionAction = ({
...
+          refetchExtensionData,
+          latestStakeFraction: convertFractionToWei(
+            fieldValues.params?.stakeFraction,
+          ),
+        });
+      } else {
+        await waitForDbAfterExtensionAction({
+          method: ExtensionMethods.ENABLE,
+          refetchExtensionData,
+        });
```
iamsamgibbs, line 189: Would be nice to make this more reusable for other extensions too.

src/components/frame/Extensions/pages/partials/ActionButtons.tsx
```diff
@@ -45,6 +50,14 @@ const ActionButtons: FC<ActionButtonProps> = ({
     extensionData.uninstallable &&
     !extensionData.isDeprecated;
 
+  /* To save changes, a user must have the root permission. */
+  const isSaveChangesButtonVisible =
+    extensionData.extensionId === Extension.StakedExpenditure &&
+    userHasRoot &&
+    isInstalledExtensionData(extensionData) &&
+    !extensionData.isDeprecated &&
+    activeTab === ActiveTab.Settings;
```
iamsamgibbs, line 59: And here

src/components/frame/Extensions/pages/partials/EnableButton.tsx
```diff
@@ -61,7 +61,8 @@ const EnableButton = ({
     isInstalledExtensionData(extensionData) &&
     canExtensionBeInitialized(extensionData.extensionId) &&
     !extensionData.isDeprecated &&
-    !extensionData.isInitialized;
+    !extensionData.isInitialized &&
+    extensionData.extensionId !== Extension.StakedExpenditure;
```
iamsamgibbs, line 65: And here

iamsamgibbs (changes requested):
Thanks for the changes, this is looking a lot better.

UI no longer shifts when the 'Save changes' button appears:

Screenshot 2024-06-07 at 10 43 41

The loading spinner issue is resolved, but the 'Save changes' button stays visible after saving and doesn't respect the new stake value, meaning it will disappear if you enter the previous value:

https://github.com/JoinColony/colonyCDapp/assets/34915414/ae8f4b73-133c-4abf-87f6-115e08b04fbe

Also dropped some comments for where the code could be made more reusable for settings pages for other extensions.

joanna-pagepro:
@iamsamgibbs can you take a look again?

arrenv (changes requested):
Thank you @joanna-pagepro! Some things I picked up.

1. The One Transaction Payment extension details has gone missing.
image

2. Bold items and headings are not present as of the design.

    image

    #### Design
    image

3. The images are not correct as of the design. They should be:
image
image

4. The heading should be "Staking Advanced Payments".
image

5. The extension item should also be updated to match the design, ensuring the version is pulled is accurate and pulled from the API ([Figma link](https://www.figma.com/design/6qONTE1uD9K4yoIbfFzidg/Extensions?node-id=1564-15479&t=lrXzVK6Kzy0L2q92-4))
image
image

6. I uninstalled and then re-installed the extension, but then ran into this issue with missing permissions, I had to reload the page to get them to disappear. It should only show this banner if the permissions are actually missing.
image

7. I understand that this is outside the scope of this PR, but, just to double check, after installing this extension, I was not able to create a Payment builder action using Staking.
image

8. I modified the Required stake value from 1% to 3%, I uninstalled, then reinstalled and the value shows as 3% again. Reloading the page then shows it correctly at 1%.
image

joanna-pagepro:
@arrenv 
1 - 5 should be fixed now
6, 8 - I'm still working on it
7 - you can't see it because the changes related to this decision method are still in code review https://github.com/JoinColony/colonyCDapp/pull/2340

joanna-pagepro:
@arrenv 6 and 8 are now also ready

arrenv (changes requested):
@joanna-pagepro Great work, thank you for the updates. Just a couple of very quick ones.

1. A minor issue now is with the value going back to one after saving. Not a massive issue, but if you can think of a way to resolve it, it would certainly be a less confusing experience.
update-value-extension

2. The images are showing blurry, can you instead use the images I have sent you via direct message.

src/components/frame/Extensions/pages/ExtensionDetailsPage/utils.tsx
```diff
@@ -209,3 +294,74 @@ export const mapExtensionActionPayload = (
...
+};
+
+export const getActionData = (extensionData: AnyExtensionData) => {
+  const isExtensionInstalled = isInstalledExtensionData(extensionData);
+
+  switch (extensionData.extensionId) {
+    case Extension.StakedExpenditure: {
+      if (isExtensionInstalled && extensionData.isEnabled) {
+        return ActionTypes.SET_STAKE_FRACTION;
+      }
```
jakubcolony, line 359: I might be wrong but I think this is where the confusion around #2498 could have aroused from. Both contracts and the saga should support deprecating but it would be prevented by this if check since deprecated extension has its `isEnabled` set to false.
@arrenv

src/components/frame/Extensions/pages/partials/InstallButton.tsx
```diff
@@ -35,28 +58,136 @@ const InstallButton = ({ extensionData }: InstallButtonProps) => {
...
+
+  const isExtensionInstalled =
+    extensionData && isInstalledExtensionData(extensionData);
+
+  useEffect(() => {
+    const handleEnableExtension = async () => {
+      setWaitingForActionConfirmation(true);
+      try {
+        setIsPolling(true);
+        const extensionRoles = isExtensionInstalled
```
jakubcolony, line 78: Is it intentional to run this whole handler if `isExtensionInstalled` is false?

src/components/frame/Extensions/pages/partials/InstallButton.tsx
```diff
@@ -35,28 +58,136 @@ const InstallButton = ({ extensionData }: InstallButtonProps) => {
...
+          : [];
+        const missingPermissions = extensionData.neededColonyPermissions.filter(
+          (neededRole) => {
+            return !userHasRole(extensionRoles, neededRole);
+          },
+        );
+
+        const updatedExtensionData = {
+          ...extensionData,
+          missingColonyPermissions: missingPermissions,
```
jakubcolony, line 89: This should be already handled by `useExtensionData`. Is there some race condition that forces you to edit them again?

src/components/frame/Extensions/pages/ExtensionDetailsPage/ParamDetailsTab.tsx
```diff
@@ -1,29 +1,50 @@
...
 
 const displayName = 'pages.ExtensionDetailsPage.ParamDetailsTab';
 
-const ParamDetailsTab: FC<ParamDetailsTabProps> = ({ extension, params }) => {
+const ParamDetailsTab: FC<ParamDetailsTabProps> = ({ extensionData }) => {
+  const { user } = useAppContext();
+  const { colony } = useColonyContext();
+
+  const params: Record<string, string> | null =
+    extensionData.params?.[camelCase(extensionData.extensionId)];
```
jakubcolony, line 25: Should that also use the `getExtensionParams` helper?

src/components/frame/Extensions/pages/ExtensionDetailsPage/utils.tsx
jakubcolony: I'm not sure it's the right time to make this generic, no other extension offers updating parameters after enabling.

jakubcolony (dismissed):
Nice start on this PR @joanna-pagepro 

It looks like @rumzledz went on to implement that same feature in #2500. While we'll have to close that PR, it might be worth for you to see how he approached the problem, especially around introducing the new context to avoid prop drilling. I think this PR could use some of that

Functionally I haven't noticed any issues apart from those already raised.

https://github.com/JoinColony/colonyCDapp/assets/112586815/e5628bbf-edcd-4f9b-8af1-ae1f4bf0f93f
Screenshot 2024-06-10 at 19 30 58
Screenshot 2024-06-10 at 19 31 03

iamsamgibbs:
> I'm not sure it's the right time to make this generic, no other extension offers updating parameters after enabling.

@jakubcolony I requested making it more generic as multisig will offer updating parameters after enabling.

src/components/frame/Extensions/pages/partials/InstallButton.tsx
joanna-pagepro: I got an error that this field is missing

src/components/frame/Extensions/pages/ExtensionDetailsPage/utils.tsx
joanna-pagepro: @jakubcolony I don't understand. There's a separate saga that is used to reenable the extension. Please check `src/components/frame/Extensions/pages/ExtensionDetailsPage/partials/ExtensionDetails/hooks.tsx`. 
image

src/utils/extensions.ts
```diff
@@ -81,6 +84,15 @@ export const canExtensionBeInitialized = (extension: Extension) => {
...
   new Decimal(percentage).mul(new Decimal(10).pow(16)).toString();
 
+export const getDefaultStakeFraction = (nativeTokensDecimals: number) =>
+  BigNumber.from(10)
+    .pow(nativeTokensDecimals)
+    .mul(DEFAULT_STAKE_PERCENTAGE)
+    .div(100);
+
+export const getStakeFraction = (value: string) =>
+  new Decimal(value).div(new Decimal(10).pow(16)).toString();
```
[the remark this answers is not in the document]
joanna-pagepro: @jakubcolony should the native token decimals value be also updated for `convertFractionToWei` function? `getStakeFraction` is used to transform the percentage value that we get from the API to the one that is displayed in the stake fraction field.

rumzledz (changes requested):
Hey @joanna-pagepro I started quite recently so nice to meet you! 😄  Great start on this PR! I did notice a couple of areas in the UI that could use some improvements:

- The Image Carousel should now only be viewable inside the Overview tab
Screenshot 2024-06-11 at 14 51 56
- The processing of this value will need some investigation
Screenshot 2024-06-11 at 14 52 49
- On the Figma designs, there seems to be a new validation message introduced for this field. It should say "Percentage must be greater than 0" if the value is 0
Screenshot 2024-06-11 at 14 57 47
- When viewing this tab as a User with insufficient permissions, the field row should be center aligned on the x-axis and ideally should have a reasonable amount of gap between the label and the percentage amount
Screenshot 2024-06-11 at 15 00 42
As in it should preserve this x-axis alignment:
Screenshot 2024-06-11 at 15 04 27
- Then finally, this is an ultra nitpick do forgive me but I'm just basing it on the designs -- I think there should be a max-width for the field's subtitle so the text doesn't become too long, which makes it easier to read as a chunk. On Figma it's currently 480px
image

src/components/frame/Extensions/pages/ExtensionDetailsPage/utils.tsx
jakubcolony: @joanna-pagepro I'm referring to the issue #2498 which states:
> It should be possible to edit the "Required stake" percentage value when the extension is in a deprecated state. This is possible on the contracts, however, according to recent updates as of this PR https://github.com/JoinColony/colonyCDapp/pull/2474, it is currently not supported by the saga.

I don't think this is correct, the `setStakeFraction` saga supports editing stake fraction for deprecated extensions. It is the UI that needs to be modified to allow that.

The if check I commented on would be one of the things that might need to change, as currently it either sets the stake fraction (for non-deprecated extensions) or enables the deprecated ones. There is no path to set stake fraction for deprecated extension without enabling it.

Nonetheless, I just wanted to mention if for future reference, it's out of scope of this PR.

joanna-pagepro: @jakubcolony I guess that we have a small misunderstanding here. I asked @arrenv if a user should be able to change the stake fraction while reenabling the extension and not if he should be able to change the stake fraction while the extension is deprecated without reenabling it. As I understand from your comment, you want a user to be able to modify the stake fraction while in the deprecated state without reenabling it.

jakubcolony: @joanna-pagepro thanks for explaining, I have checked with Arren and being able to set stake fraction for a deprecated extension is one of the requirements and should be part of this PR. 

> I asked @arrenv if a user should be able to change the stake fraction while reenabling the extension and not if he should be able to change the stake fraction while the extension is deprecated without reenabling it. As I understand from your comment, you want a user to be able to modify the stake fraction while in the deprecated state without reenabling it.

I think the exact behaviour needs to be confirmed with Arren, the point I was trying to make is that it's ready to be implemented now, and doesn't require any saga modifications.

joanna-pagepro:
@rumzledz thanks for your feedback. Can you check if everything is ok now?
@jakubcolony @arrenv @iamsamgibbs can you also take a look? :)

jakubcolony:
> @jakubcolony I requested making it more generic as multisig will offer updating parameters after enabling.

Sure, when that gets implemented I'm all in for making this UI generic/reusable, however until updating Multisig params is implemented we should hold off with making things generic as that usually results in a badly applied abstraction (which makes sense as you don't know what you're abstracting at that point, it is much easier to refactor 2 concrete use cases). Even more so that they're on two separate feature branches

arrenv:
@joanna-pagepro To confirm, when the extension is deprecated, it should be possible to save the parameters still, but also be able to enable the extension again. So, we need to show these buttons next to each other.

The design for how this should look is here - [Figma link](https://www.figma.com/design/SaO0Zj6e7yGqLzjEsxqVhy/Permissions?node-id=21305-15997&t=aMLGPQXMXkipDbXm-4)

image

iamsamgibbs (changes requested):
This is looking better, save button is now behaving correctly when the value is updated.

https://github.com/JoinColony/colonyCDapp/assets/34915414/2bc68284-a0e6-4460-91c7-e8eb7655d522

Rejecting for now until it is possible to update stake whilst deprecated.

joanna-pagepro:
@arrenv @iamsamgibbs @rumzledz @jakubcolony please review again

iamsamgibbs (approved):
This looks good to me! Able to edit the amount whilst deprecated. Uninstalling and reinstalling working as expected. Nice job!

I'm guessing this is the intended functionality, but it feels a little unintuitive to me that if I change the amount whilst deprecated and click enable, then it will enable with the previous amount rather than also updating.

 

https://github.com/JoinColony/colonyCDapp/assets/34915414/1c71b04f-40ff-47ea-98c9-302f7b1eecb9

arrenv (approved):
All looks good to me, thank you @joanna-pagepro

rumzledz (approved):
You've finessed the issues I've raised @joanna-pagepro ✨

jakubcolony:
I think this might need a rebase @joanna-pagepro 

BTW. I finally figured out `git rebase --onto` which lets you skip those old commits from pre-rebase parent branch and only deal with yours (same as doing cherry-pick, except you don't need a new branch and all that hassle).

You can simply run:
```sh
git rebase --onto feat/advanced-payments-ui <last-commit-hash-before-your-changes>
```

In case of this PR the last commit before your changes is `3ebb4db1df4c7347f5584cb8f23cfd70394e8370`

PR #2478, opened 2024-06-05
Fix: Show the Expenditure's description on the Completed Action component

### LINKED ISSUE #2373: Descriptions don't show by default until hard refresh
When I add a description to the payment, it does not show when the transaction is complete until I refresh the page. 

Video: https://github.com/JoinColony/colonyCDapp/assets/82127904/1738e9ff-4579-4cc9-92c5-3e471296833a

## Steps to reproduce

1. Create a new payment builder action
2. Complete all the fields including the description
3. Complete the transaction
4. The description does not show
5. Hard refresh
5. The description shows

## Expected behavior

* The description should show without the user requiring a hard refresh to see the description.

rumzledz:
## Description

Currently, there's a bit of a race condition that occurs whereby polling for a Colony Action gets cancelled prematurely even though the intention is to only stop it when the Action Sidebar unmounts.

This PR ensures the original intent of stopping a poll only when the component unmounts is respected.

2373_asap

## Testing

**Prerequisite**: If you don't currently have a network monitoring tool for GraphQL queries, install the [GraphQL Network Inspector browser add-on](https://chromewebstore.google.com/detail/graphql-network-inspector/ndlbedplllcgconngcnfmkadhokfaaln?hl=en-GB)

1. Bring up the Payment Builder action form
2. Fill in all required fields
3. Make sure to enter a description
4. Submit the form
5. Open your GraphQL monitoring tool
6. When the Completed Action component mounts, verify that the description field comes up without the need for a browser reload and that this behaviour is consistent with how a Simple Payment action behaves on production
7. Close the Completed Action component
8. Verify that the `GetColonyAction` GraphQL requests have stopped

## Diffs

**Changes** 🏗

I isolated the concern of stopping the poll in a separate effect and gave that effect an empty array dependency to ensure that its return statement only ever gets called when the component unmounts, and not when there's a state change elsewhere.

Resolves #2373

### DIFF
--- a/src/components/v5/common/ActionSidebar/hooks/useGetColonyAction.ts
+++ b/src/components/v5/common/ActionSidebar/hooks/useGetColonyAction.ts
@@ -51,6 +51,14 @@ const useGetColonyAction = (transactionHash?: string) => {
 
   const action = actionData?.getColonyAction;
 
+  const clearPollingCancellationTimer = () => {
+    if (pollTimerRef.current) {
+      clearTimeout(pollTimerRef.current);
+
+      pollTimerRef.current = null;
+    }
+  };
+
   useEffect(() => {
     const shouldPool = !isInvalidTx && !action;
 
@@ -68,11 +76,7 @@ const useGetColonyAction = (transactionHash?: string) => {
       return;
     }
 
-    if (pollTimerRef.current) {
-      clearTimeout(pollTimerRef.current);
-
-      pollTimerRef.current = null;
-    }
+    clearPollingCancellationTimer();
 
     pollTimerRef.current = setTimeout(stopPollingForAction, POLLING_TIMEOUT);
 
@@ -95,9 +99,7 @@ const useGetColonyAction = (transactionHash?: string) => {
       // This effect should receive an empty array dependency to ensure that
       // it only ever calls the return statement when its node unmounts,
       // and not when any other state gets updated.
-      if (pollTimerRef.current) {
-        clearTimeout(pollTimerRef.current);
-      }
+      clearPollingCancellationTimer();
 
       stopPollingForAction();
     };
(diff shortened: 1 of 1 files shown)


rdig (approved):
So in the truest sense, this issue is fixed, the description does indeed show up after completion

[Screencast from 2024-06-07 00-31-50.webm](https://github.com/JoinColony/colonyCDapp/assets/1193222/d8301509-1243-473f-b9c4-c8f325bd6164)

However I do see a very long delay (basically the block time). I guess in production, this will be faster (at least on Arbitrum), but I still think this interaction is a tad bit wonky

rumzledz:
> So in the truest sense, this issue is fixed, the description does indeed show up after completion
> 
>  Screencast.from.2024-06-07.00-31-50.webm 
> However I do see a very long delay (basically the block time). I guess in production, this will be faster (at least on Arbitrum), but I still think this interaction is a tad bit wonky

Thanks for reviewing @rdig ! I agree about the wonky bit and unfortunately this is the current behaviour on `master` for all other actions like "Mint token", "Simple payment" etc. I found that it's because annotations takes longer to reach our DB for some reason. I don't understand how the IPFS upload works behind the scenes for now but I think it's the cause of the delay. This screenshot shows the poll timings for a "Mint token" action on `master`. Sometimes it takes even more `GetColonyAction` polls before the annotation becomes available 😕

image

iamsamgibbs (approved):
Seems to resolve the issue, description is now showing on the completed action.

https://github.com/JoinColony/colonyCDapp/assets/34915414/0d17c7d3-7a50-4478-a90a-5bdd7e50c879

jakubcolony (approved):
Takes a bit of time but it does appear:
image

Very nice fix! 👍

PR #2479, opened 2024-06-05
Fix: Handle transactions that don't exist

### LINKED ISSUE #2335: Handle transactions that don't exist
#### Case 1:
1. Navigate to the following URL: `http://localhost:9091/planex/activity?tx=0x`
2. The action panel will load with an incorrect state showing the process sidebar.

#### Case 1:
1. Navigate to the following URL: `http://localhost:9091/planex/activity?tx=0xfd44e904a267f7c1377486f03020effc50139df775007be4efb7f5bf102b8a7b`
2. The action panel will stay in a continuous loading state,

## Expected behaviour

* If a transaction is not able to be found, then it should show an error state.
* Report a bug button should work the same as the 404 page "Report a bug" button.
* The "Go to activity page" should navigate to the activity page `{colony_url}/activity`.
* The "Create a new action" button should open a blank action side panel.

[Figma Design - Transaction not found](https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-v5?type=design&node-id=23947-236394&mode=design&t=0ojMp5MGZ3osfbXo-4)

image

## Actual behaviour

* Transactions that don't exist are trying to load or showing an incorrect state showing the process sidebar without any details.

#### Case 1:
image

#### Case 2:
image

davecreaser:
## Description

- Properly handle the case in the action sidebar when opening an action that doesn't exist.
- Two cases are covered:
  - When the transaction hash is incorrect
  - When the transaction can't be found in the db

## Testing

- Navigate to the following URL: `http://localhost:9091/planex/activity?tx=0x`
- The action panel should load with a 404 message.

- Navigate to the following URL: `http://localhost:9091/planex/activity?tx=0xfd44e904a267f7c1377486f03020effc50139df775007be4efb7f5bf102b8a7b`
- The action panel should try to load the action for 20 seconds, before throwing a 404 (with a retry button).

## Diffs

**New stuff** ✨

* New reusable 404 message component. The 404 page now uses this too.

**Changes** 🏗

* Update the action sidebar to handle not found actions.
* Update the 404 page to use a reusable component.

## TODO

- [x] Check with @arrenv and @melyndav about the retry button for the second case, since it's different functionality from the original issue / designs. It came about from conversation with some dev team members as a nice feature to include, since an action might take some time to be created in the db, so it being missing might just be a time issue. This change can easily be removed if it's not wanted however.

Resolves #2335


iamsamgibbs (changes requested):
Nice work! Code changes look good to me, both testing examples working as expected and 404 page is unaffected.

Screenshot 2024-06-06 at 08 57 22
Screenshot 2024-06-06 at 09 05 32
Screenshot 2024-06-06 at 09 10 03

However it seems to have affected the staking stage of motions:

https://github.com/JoinColony/colonyCDapp/assets/34915414/3647cf48-3eee-4389-b4e5-60628adde0de

src/components/v5/common/FourOFourMessage/FourOFourMessage.tsx
```diff
@@ -0,0 +1,104 @@
...
+  linksTitle: {
+    id: `${displayName}.linksTitle`,
+    defaultMessage: 'Helpful links',
+  },
+});
+
+const FourOFourMessage: FC<FourOFourMessageProps> = ({
+  description,
+  links,
+  primaryLinkButton,
```
bassgeta, line 40: could we maybe just compose these in? so we just render the `primaryLinkButton` as passed next to the report button.
The links are a bit trickier, but I think it makes sense to just compose them in and have the parent component control if the link is external or internal. It's a bit tricky to get the styles in (maybe we could wrap each of them in a div and apply the styles there, or try CSS selectors from this component).

bassgeta (changes requested):
Nice work on this, a nice and sane solution! I am getting the same issue Sam spotted and dropped a comment for composing stuff in instead of passing in objects.

davecreaser:
Cheers @iamsamgibbs and @bassgeta - issues are now fixed and I've switched the props to use composition 🚀

iamsamgibbs (dismissed):
All good now! Motion flow is restored.

bassgeta (approved):
Calm and composed :sunglasses: 
Nice one, the bug has been fixed so it's time to :ship:

arrenv:
I think the enhancement is an good.

I would suggest updating the copy to be more relevant in this case:

"The transaction you're looking for could not be loaded or does not exist. Please check the link is correct, retry, or look for the transaction on the activity page."

davecreaser:
Thanks @arrenv, copy is now updated! We are now merging after three approvals, so feel free to add yours here if you would like it to be merged 😆 

@iamsamgibbs and @bassgeta - a reapproval is required if you please 😄

iamsamgibbs (approved):
Still working and new copy is in. 👍 

Screenshot 2024-06-10 at 10 13 04

arrenv (approved):
Thanks @davecreaser, all working as expected for me.

Regarding the description, I should have not added "in this case", as I had meant for the updated description to be used for all cases :)

However, I think the conditional copy is fine.

davecreaser:
@arrenv @bassgeta @iamsamgibbs I had to rebase and it wiped all the approvals I'm afraid 🙃 Could I steal them back please?

bassgeta (approved):
re-base, re-approve, let's get this one shipped :ship:

iamsamgibbs (approved):

arrenv (commented):
All looks good still!

arrenv (approved):

PR #2482, opened 2024-06-05
Fix: Handle image upload error due to fingerprint blocking

### LINKED ISSUE #2262: Users have reported not being able to upload a token logo
## Steps to reproduce

This was run into when trying to add a token image to a create a colony with a new token process.

I have not been able to replicate this, however, Etienne has been able to. This was reported by a user and Etienne, who both have the following setup:
- Mac OSx
- Brave Version 1.65.114 Chromium: 124.0.6367.60 (Offizieller Build) (arm64)

## Expected behaviour

* All users should be able to upload a token logo when within the validation limitations.

## Actual behaviour

* Image uploads is giving a corrupted image error

image

rumzledz:
## Description

There's currently an error being thrown by the `pica` lib whenever a file is being selected in our image uploader due to Brave's fingerprint blocking mechanism which is enabled when the Brave Shields are up.

image

I'm waiting for Arren's thoughts on this but last time I discussed this with Raul & Chris, it was decided that the most optimal solution **for now** is to describe to the user what the error is. Because the preferred fix is much more complex and will take up a lot of time. Here's a proposal for the error message:

<div align="center">

image

</div>

## Testing

I will need @rdig's or @chmanie's thoughts with this one. I am unable to find a way to replicate it locally. From what I understand, the on-demand qa environment isn't set up yet so I'm guessing we should just take a gamble and test this once it's deployed to the QA environment 🎲 

But for the actual steps:

1. Install the Brave browser if you don't have it already
2. Make sure Brave Shields are up and "Block fingerprinting" is enabled
Screenshot 2024-06-07 at 10 47 44
3. Go to Manage Profile page
4. Select a file to upload as your avatar
5. Verify that you can see the following error message: `"Your browser blocked the upload. Check if fingerprint blocking is enabled and temporarily disable it before trying again."`
image

## Diffs

**New stuff** ✨

* A new error message to display when we capture a fingerprint blocking related error whenever a user selects an file to upload.

Resolves #2262

### DIFF
--- a/src/components/v5/common/AvatarUploader/hooks.ts
+++ b/src/components/v5/common/AvatarUploader/hooks.ts
@@ -47,6 +47,12 @@ export const useAvatarUploader = ({ updateFn }: UseAvatarUploaderProps) => {
     } catch (e) {
       if (e.message.includes('exceeded the maximum')) {
         setUploadAvatarError(DropzoneErrors.TOO_LARGE);
+      } else if (
+        e.message.includes(
+          "Pica: cannot use getImageData on canvas, make sure fingerprinting protection isn't enabled",
+        )
+      ) {
+        setUploadAvatarError(DropzoneErrors.FINGERPRINT_ENABLED);
       } else {
         setUploadAvatarError(DropzoneErrors.DEFAULT);
       }
--- a/src/components/v5/common/AvatarUploader/partials/ErrorContent.tsx
+++ b/src/components/v5/common/AvatarUploader/partials/ErrorContent.tsx
@@ -29,7 +29,7 @@ const ErrorContent: FC<ErrorContentProps> = ({
         </div>
       </div>
       <div className="flex w-full flex-col gap-1">
-        <div className="items-top flex justify-between">
+        <div className="items-top flex justify-between gap-2">
           <span className=" text-left text-negative-400 text-1">
             {formatMessage(errorMessage)}
           </span>
--- a/src/components/v5/common/AvatarUploader/utils.ts
+++ b/src/components/v5/common/AvatarUploader/utils.ts
@@ -16,7 +16,7 @@ const MSG = defineMessages({
   fingerprintError: {
     id: `${displayName}.fingerprintError`,
     defaultMessage:
-      'You currently have fingerprint blocking enabled. Please disable it before selecting a file.',
+      'Your browser currently have fingerprint blocking enabled. Please disable it before selecting a file.',
   },
   fileSizeError: {
     id: `${displayName}.fileSizeError`,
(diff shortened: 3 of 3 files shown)


arrenv:
@rumzledz I am okay with this for the time being, although I did see after a quick search that the issue with fingerprint blocking has false positive issues, which could explain others experience of this on other operating systems and browsers, so, may need to be revisited.

As you mentioned, not being able to test locally means I cannot verify this works as expected.

Lastly, I would suggest to update the copy slightly. Suggested copy:

"Your browser blocked the upload. Check if fingerprint blocking is enabled and temporarily disable it before trying again."

rdig (dismissed):
I'm good with this, however please do update the copy to the one Arren suggested before merging it

rumzledz:
> @rumzledz I am okay with this for the time being, although I did see after a quick search that the issue with fingerprint blocking has false positive issues, which could explain others experience of this on other operating systems and browsers, so, may need to be revisited.
> 
> As you mentioned, not being able to test locally means I cannot verify this works as expected.
> 
> Lastly, I would suggest to update the copy slightly. Suggested copy:
> 
> "Your browser blocked the upload. Check if fingerprint blocking is enabled and temporarily disable it before trying again."

All right I shall update the copy! 👌

rumzledz:
@rdig @arrenv pushed!
Screenshot 2024-06-11 at 15 28 32

src/components/v5/common/AvatarUploader/hooks.ts
```diff
@@ -47,6 +47,12 @@ export const useAvatarUploader = ({ updateFn }: UseAvatarUploaderProps) => {
     } catch (e) {
       if (e.message.includes('exceeded the maximum')) {
         setUploadAvatarError(DropzoneErrors.TOO_LARGE);
+      } else if (
+        e.message.includes(
+          "Pica: cannot use getImageData on canvas, make sure fingerprinting protection isn't enabled",
```
rdig, line 52: I'd normally store this as a `enum` or just plain `type`, but since it's a single error message, it feel kinda "weird" to do that.

rdig (approved):
All good from my end. Nicely done!

bassgeta (approved):
Nice find and nice solution, shows up in my brave browser :sunglasses:

arrenv (approved):
Thank you @rumzledz, looking forward to testing it out :)

src/components/v5/common/AvatarUploader/hooks.ts
rumzledz: I was debating on whether or not to store it as an enum at first, but I thought I'll just add it as a plain string for now since it's only ever used here and this will likely go away once the proper solution gets applied. BUT if we do end up having another error message to handle, I shall use an enum to make things a bit more tidy! 😄 @rdig

Issue #2158, opened 2024-04-05
[Multi-sig] feat: cancel a motion as a signee


bassgeta:
### Story

As a *user rejecting a multi-sig motion*, I want to be able to cancel it if the threshold has passed.

### Description

If the motion is fully signed to be rejected, we can click the `Finalize` button, but it should reject the motion by calling [this contract](https://github.com/JoinColony/colonyNetwork/blob/029bd755d24458a12e6dd9167d7a915992845362/contracts/extensions/MultisigPermissions.sol#L300) instead of the `execute` one.

The block-ingestor needs to catch the [`MotionCancelled`](https://github.com/JoinColony/colonyNetwork/blob/029bd755d24458a12e6dd9167d7a915992845362/contracts/extensions/MultisigPermissions.sol#L42C9-L42C24) event and update the motion state to `CANCELLED`/isFinalized to true *TBD: we need to determine if we're adding new motion types in the database, or reusing existing ones*

The widget needs to poll the motion query and change it's state to `Rejected`
![image](https://github.com/JoinColony/colonyCDapp/assets/23449297/58846136-dbb7-4e74-87c9-a3ec8214a356)

No changes should be applied.

### Design

[Figma link](https://www.figma.com/file/SaO0Zj6e7yGqLzjEsxqVhy/Permissions?type=design&node-id=20450-35152&mode=design&t=iqzIcnoziy1lwDp6-4)

### This issue is complete when...

- [ ] A motion has enough reject votes the `Cancel` button appears
- [ ] The user clicks the button the motion should get cancelled
- [ ] Block ingestor receives the cancel event
- [ ] The changes should be reflected in the database, no changes applied

PR #2489, opened 2024-06-06
[Multi-sig] Feat: Reject multisig and cancel motion

### LINKED ISSUE #2154: [Multi-sig] feat: rejecting a multi-sig motion
### Story

As a *user eligible to sign a multi-sig motion*, I want to be able to reject it.

### Description

Do note that you are actually signing with a user, but we also need to keep track of how many votes we have per role.

We need to first determine the required role for the motion. The potential signatories are all the members with relevant roles in the motion's `createdIn` domain and **ALL** the role holders in the `Root` domain.. The list is paginated by clicking `Load more` (so we can create a new gql query for this, we can check if it's possible to ramp up the caching as a lot of these members will be fetched already).

For displaying the threshold data, we need to get the current signees from the motion data (we need to determine what this property will be called) and fetching the domain's multi-sig threshold by calling [this contract](https://github.com/JoinColony/colonyNetwork/blob/029bd755d24458a12e6dd9167d7a915992845362/contracts/extensions/MultisigPermissions.sol#L408). *TBD: we need to check if contracts will return it directly or if we need to do [these shennanigans](https://www.figma.com/file/SaO0Zj6e7yGqLzjEsxqVhy/Permissions?type=design&node-id=20450-35051&mode=design&t=iqzIcnoziy1lwDp6-4)*

This issue's scope is only for rejecting a motion. After clicking the meatball menu and clicking `Reject action` inside it, we should immediately call [this contract](https://github.com/JoinColony/colonyNetwork/blob/029bd755d24458a12e6dd9167d7a915992845362/contracts/extensions/MultisigPermissions.sol#L224).

The block-ingestor will listen to [this event](https://github.com/JoinColony/colonyNetwork/blob/029bd755d24458a12e6dd9167d7a915992845362/contracts/extensions/MultisigPermissions.sol#L45). Upon receiving it should change the rejection voters property on the motion (actual structure TBD). Block-ingestor also need to keep track of which user voted for which role which we can get from this event.

upon clicking the `Reject` button, we need to call [this contract](https://github.com/JoinColony/colonyNetwork/blob/029bd755d24458a12e6dd9167d7a915992845362/contracts/extensions/MultisigPermissions.sol#L224) with the vote set to `Vote.Reject`. We should be constantly polling/subscribe to the motion query. The signees are displayed on the top
image.
As soon as we get an entry for signers that have rejected, we display the `Reject` step in the stepper component.

Make sure not to display users which have approved it as potential signees.

### Design

[Figma link](https://www.figma.com/file/SaO0Zj6e7yGqLzjEsxqVhy/Permissions?type=design&node-id=20450-34979&mode=design&t=iqzIcnoziy1lwDp6-4)

image
image

- [ ] The correct number is shown as the required threshold (so we need to check domains that have a custom one, majority approval or inheriting from colony)
- [ ] Potential signees are shown in the widget, the list being paginated
- [ ] The signees are shown on top, followed by all potential signees 
- [ ] Upon clicking `Reject`, the vote gets added to the motion and the user is now a signee for the reject section

iamsamgibbs:
## Description

[Related block-ingestor counterpart](https://github.com/JoinColony/block-ingestor/pull/237)

This PR implements the option to reject a multisig motion.

## Testing

You will need 2 browsers, so invite `amy` or whoever you wish to `planex`.
For checking multi sig motion state, use this handy query
```
query MyQuery {
  getColonyAction(
    id: "THE_TX_HASH"
  ) {
    multiSigId
    id
    multiSigData {
      isExecuted
      isSuccess
      isRejected
      nativeMultiSigId
      signatures {
        items {
          multiSigId
          role
          userAddress
          vote
        }
      }
    }
    type
    isMultiSig
  }
}
```

1. Install the multi sig extension and set the global threshold to 2 (easiest to test). (you may need to refresh the app so multi sig pops up as decision method)
2. Assign Amy `Owner` roles with the authority of `Take actions via Multi-Sig`
3. Create a mint tokens action with the decision method set to `Multi-sig`
image
4. Confirm the reject stage of the multisig motion is not visible in the sidebar.

Screenshot 2024-06-10 at 14 48 03

4. Copy over the url to the second browser, where you need to be logged in with `amy` (You'll need to have accepted an invite to the colony first).
6. Select `Reject action` from the meatball menu:

Screenshot 2024-06-10 at 14 47 47

7. This should vote for reject and open the reject stage of the multisig motion in the sidebar (it will take a few moments for your vote to show - this will be handled more cleanly in the proper UI).

Screenshot 2024-06-10 at 14 50 12

8. Back in the first browser, you should now see the reject stage and Amy's vote. Choose to vote for reject instead of approve.
9. After the threshold for reject has passed, the `Finalize rejection` button should appear below the voting button. You can verify the votes and the state by running the gql query.

Screenshot 2024-06-10 at 14 52 01

10.  Click the `Finalize rejection` button, after some moments, the transaction should be cancelled, the sidebar should change to reflect that the multi sig has been cancelled. You can run the query to make sure :)

Screenshot 2024-06-07 at 08 35 31

Further testing:
Create another mint multisig action and try rejecting via the meatball menu as the user who created the action. This should instantly cancel the multisig motion, skipping the entire reject stage (it is also possible to cancel a multisig motion this way even if approve / reject has surpassed the voting threshold but not yet been finalized).

## Diffs

**New stuff** ✨

* Added "reject step" to the multisig widget
* Added new `RejectionChanged` handler to block-ingestor

**Changes** 🏗

* Meatball menu now cancels multisig motion if owner, or sets "reject step" to show if not owner.
* Vote button now accepts a vote type button for approve or reject.
* Refactored some helper functions on block-ingestor to reuse them for both approve and reject

Resolves #2154 and #2158


src/components/v5/common/CompletedAction/partials/MultiSigMeatballMenu/MultiSigMeatballMenu.tsx
```diff
@@ -23,24 +29,50 @@ const displayName = 'v5.common.CompletedAction.partials.MultiSigMeatballMenu';
...
+
+  const handleCancelMultiSigAsOwner = () => {
     cancelMultiSigMotion({
       colonyAddress: colony.colonyAddress,
       motionId: multiSigData?.nativeMultiSigId,
     });
   };
 
+  const handleRejectMultiSig = async () => {
+    const voteForPayload: VoteOnMultiSigActionPayload = {
```
bassgeta, line 62: `rejectVotePayload` would make more sense

bassgeta (changes requested):
Stellar work on this, it works, juuuust a small variable name we can change spotted :D 
Rejecting as owner works:
image
Rejection shows up and also the vote for approve disappears:
image
And it can be rejected:
image

bassgeta (approved):
Naming changes make sense, stuff still works, let's go :sunglasses:

rumzledz (approved):
Stellar work @iamsamgibbs ✨ 

**Rejected the action as another user**
Screenshot 2024-06-13 at 17 40 13

**Finalised the rejection and getting the GraphQL result**
Screenshot 2024-06-13 at 17 41 32

**Rejecting the action as the user who created the action**
cancel mint as minter

PR #2515, opened 2024-06-11
Feat: Handle the cancelAndWaive method call for Streaming Payments

### LINKED ISSUE #2217: Streaming Payments API: handle `cancelAndWaive` contract call
When cancelling a stream, caller can specify a list of token addresses to waive existing claims for. This will reduce any amount claimable to 0 and needs to be taken into account when calculating available payments.

The existing cancelStreamingPayment saga should be modified to accept `shouldWaiveClaim` parameter or similar.

~~Alex will be adding a new event to this method so this will require a new colony-js release.~~ There is a `ClaimWaived` event that block-ingestor needs to handle.

There should be a property of a payout in the DB indicating whether it had its claims waived when stream was cancelled.

rumzledz:
## Description

Handle the `cancelAndWaive` method call for Streaming Payments.

Here are the required block-ingestor updates: [block-ingestor#239: handle the StreamingPayments contract's ClaimWaived event emission](https://github.com/JoinColony/block-ingestor/pull/239).

## Testing

1. Enable the Streaming Payments extension
2. Go to your Colony's dashboard and using the debug UI, create a Streaming Payment
create_streaming_payment
3. Open the GraphQL Playground http://localhost:20002 and run the following query:
> This query will show all streaming payments
```
query MyQuery {
  listStreamingPayments {
    items {
      isWaived
      isCancelled
      amount
      nativeId
    }
  }
}
``` 
4. Verify that your streaming payment's `isCancelled` and `isWaived` fields are both set to `null`:
```
{
  "isWaived": null,
  "isCancelled": null,
  "amount": "1000000000000000000",
  "nativeId": 5 // your nativeId will vary
},
```
5. Press the "Cancel and Waive" button
cancel_and_waive_button
6. Run the query in Step 3 and verify that your streaming payment's `isCancelled` and `isWaived` fields are both set to `true`
```
{
  "isWaived": true,
  "isCancelled": true,
  "amount": "1000000000000000000",
  "nativeId": 5
},
```

## Diffs

**New stuff** ✨

* ~A new saga called `cancelAndWaiveStreamingPaymentAction` to handle the `cancelAndWaive` action for a streaming payment. I could have repurposed the existing `cancelStreamingPayment` saga but then I'd also have to litter it with a lot of `if that, else that` statements because we're not accepting annotations for the `cancelAndWaive` actio, at least for now, which means I won't need a Colony Manager Client and an extra Channel for the annotation upload. It also makes it easier to read.~

P.S. ~I could already sense @rdig asking and yes, this one will likely need a migration 👀~ Jakub just explained to me that streaming payments is not actually a live feature yet so no migration required, IGNORE ME!

Resolves #2217


rdig (approved):
Codewise all looks good. Nice work! :1st_place_medal: 

Tested this according to your script and it behaves as expecte:
Screenshot from 2024-06-11 19-17-01
Screenshot from 2024-06-11 19-17-10
Screenshot from 2024-06-11 19-18-16
Screenshot from 2024-06-11 19-18-30

A note however that this branch, at least on my machine is being "plagued" by the infamous node modules / swc / rollup issue

Screenshot from 2024-06-11 19-11-05

Screenshot from 2024-06-11 19-11-56

I had to install both `@rollup/rollup-linux-x64-gnu` and `@swc/core` in order to get the frontend to work

davecreaser (approved):
Good stuff, tested locally and all working as expected! 🚀

src/components/v5/payments/TmpAdvancedPayments.tsx
```diff
@@ -32,6 +32,7 @@ import { type ReleaseExpenditureStageMotionPayload } from '~redux/sagas/motions/
 import {
   type CancelStreamingPaymentPayload,
   type CancelStakedExpenditurePayload,
+  type CancelAndWaitveStreamingPaymentPayload,
```
jakubcolony, line 35: little typo here

src/redux/sagas/expenditures/cancelAndWaiveStreamingPayment.ts
```diff
@@ -0,0 +1,70 @@
...
+
+import {
+  type ChannelDefinition,
+  createTransaction,
+  createTransactionChannels,
+  waitForTxResult,
+} from '../transactions/index.ts';
+import { initiateTransaction, putError, takeFrom } from '../utils/index.ts';
+
+function* cancelAndWaiveStreamingPaymentAction({
```
jakubcolony, line 14: It would be great to validate the user is the stream's recipient (contract requirements) before sending any transactions

jakubcolony (changes requested):
No issues functionally, very nice work.

One of the requirements was to modify the existing `cancelStreamingPayment` saga as the process is very similar, could you have a look if that's something straightforward?

Also left a comment about validation in the saga, but otherwise looks great!

rumzledz:
> No issues functionally, very nice work.
> 
> One of the requirements was to modify the existing `cancelStreamingPayment` saga as the process is very similar, could you have a look if that's something straightforward?
> 
> Also left a comment about validation in the saga, but otherwise looks great!

Hey @jakubcolony I actually have a little section on my PR description regarding this but basically the only petty reason was I just didn't want to write a bunch of if else statements to differentiate the `cancel` and `cancelAndWaive` contract calls, a conditional for the contract function params because they're different & doing a `annotationMessage && `isWaived` to make sure we're not creating annotations for the `cancelAndWaive` flow. It's totally doable though. And sure I'll have a look at the validation stuff!

src/redux/sagas/expenditures/cancelStreamingPayment.ts
```diff
@@ -20,25 +25,48 @@ import {
...
   meta,
-  payload: { colonyAddress, streamingPayment, annotationMessage },
+  payload: { colonyAddress, streamingPayment, annotationMessage, shouldWaive },
 }: Action<ActionTypes.STREAMING_PAYMENT_CANCEL>) {
-  const batchKey = 'cancelStreamingPayment';
+  // To make the flow more readable
+  const isCancelMethod = !shouldWaive;
+  const isCancelAndWaiveMethod = shouldWaive;
+
+  const batchKey = `cancel${isCancelMethod ? '' : 'AndWaive'}StreamingPayment`;
```
jakubcolony, line 34: I think it's fine to use the same batch key, saves us an extra i18n message entry

jakubcolony (approved):
Looks good to me, great work @rumzledz 👍 

image

image

src/redux/sagas/expenditures/cancelStreamingPayment.ts
rumzledz: Oh, I thought we wanted to differentiate the ones that were just cancelled and ones that were cancelled + waived in places like the User Hub Transactions tab 🤔 @jakubcolony

rumzledz: Now using the same batch key for both claimed & claimed+waived method calls @jakubcolony 

Screenshot 2024-06-13 at 11 11 08
Screenshot 2024-06-13 at 11 11 00

PR #2525, opened 2024-06-12
Fix: Use correct asset platform when calling CoinGecko

### LINKED ISSUE #2418: Currency conversion is not working and is wrong in production
## Steps to reproduce

1. Add funds to a colony with value, examples included in screenshots USDC, ETH, and ARB.
2. Check the balance of funds in the users denominated currency.
3. The values are either incorrect or missing.

## Expected behaviour

* The currency conversions should show a value if it has a value.
* The currency conversion value should show correctly.

## Actual behaviour

* Token values are not being shown, when a value exists.
* Token values are showing wrong, if a value does exist.

image

image

image

image

davecreaser:
## Description

- Use the network_id environment variable to determine which asset platform should be used while calling the CoinGecko api.

## Testing

A little trickier to test properly, but there's some things you can do:

- Go to the balances table and refresh the page with the network tab open in dev tools.
- If you check the network tab you should find some calls to the CoinGecko api:

Screenshot 2024-06-12 at 17 30 58

- Change`NETWORK_ID="ganache"` in your .env.local file to be `NETWORK_ID="gnosis"`
- Repeat the first two steps
- You should now see the call being made with `xdai` instead of `arbitrum-one` as the asset platform

Screenshot 2024-06-12 at 17 38 30

The main idea here is that on production, it is currently incorrectly calling with `xdai` as the asset platform, whereas it should be using `arbitrum-one`. The network_id environment variable on production is `arbitrumOne`.

## Diffs

**Changes** 🏗

* Update the CoinGecko config to use the network_id environment variable to determine which asset platform should be used when calling the api.

Resolves #2418

### DIFF
--- a/.github/workflows/build-deploy.yaml
+++ b/.github/workflows/build-deploy.yaml
@@ -36,7 +36,7 @@ jobs:
           echo "$GITHUB_CONTEXT"
 
       - name: Configure AWS credentials
-        uses: aws-actions/configure-aws-credentials@v2
+        uses: aws-actions/configure-aws-credentials@v3
         with:
           aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID_QA }}
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY_QA }}
@@ -92,7 +92,7 @@ jobs:
 
     steps:
       - name: Configure AWS credentials for ECR
-        uses: aws-actions/configure-aws-credentials@v2
+        uses: aws-actions/configure-aws-credentials@v3
         with:
           aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID_QA }}
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY_QA }}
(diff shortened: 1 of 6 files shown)


rdig:
@bogdan-1337 why did you make this commit 290da64c8952840e6667f293fd8daf9b0610bf03 on this branch ?

Is this for arbitrum or gnosis ?

iamsamgibbs (approved):
This looks good to me! Changes make sense and tested using the steps provided.

Screenshot 2024-06-13 at 10 47 34
Screenshot 2024-06-13 at 10 48 24

rumzledz (approved):
Nice one @davecreaser , I learnt something new while reviewing this PR 😎 

- `NETWORK_ID="ganache"
Screenshot 2024-06-14 at 12 24 27

- `NETWORK_ID="gnosis"
Screenshot 2024-06-14 at 12 25 11

rdig (approved):
Changes look good. Initial knee jerk reaction would be to tell you to default to Gnosis, not Arbitrum since this will be merged into `master` and that is still a gnosis default branch.

However soon that won't be the case and I won't fight too much against this.

Screenshot from 2024-06-14 21-50-26
Screenshot from 2024-06-14 21-50-53
Screenshot from 2024-06-14 21-51-47

However I'm still curious about the change that @bogdan-1337 pushed to this branch, and especially if it needs a similar change on the `master-arbitrum` side

bogdan-1337:
> @bogdan-1337 why did you make this commit [290da64](https://github.com/JoinColony/colonyCDapp/commit/290da64c8952840e6667f293fd8daf9b0610bf03) on this branch ?
> 
> Is this for arbitrum or gnosis ?

Updated the version for the action, it was causing issues for the build on this branch, but not for other branches so not sure what that's about, but anyway good to have it updated to the new version so can be merged to any branch.

Issue #2189, opened 2024-04-10
[Multi-sig] feat: implement move funds multi-sig motion


bassgeta:
### Story

As a *user moderating a colony*, I want to be able to create and finalize a multi-sig move funds motion.

### Description

After the groundwork has been laid with #2150 , implement the move funds motion.

Additional task at this issue is to double check if #2197 resolved the failing execution properly.
Essentially try to move funds from one domain to another as a multi-sig motion, fast forward the time, empty the first domain of funds, and then see if the motion gets correctly rejected.

### This issue is complete when...

- [ ] The move funds multi-sig motion can be created
- [ ] The motion can be finalized
- [ ] Only users with [adequate permissions](https://www.notion.so/colony/Multi-sig-Permissions-6aaf0b95309f4754a243d54e8d9c4189?pvs=4#fec3699a5fd241a2925bd2809494b157) can create them

PR #2532, opened 2024-06-13
Feat/2197 multi sig max time

### LINKED ISSUE #2197: [Multi-sig] feat: handle multi-sig execution after max time has passed
### Story

Motions which are not finalized or closed within a 1 week timeframe, are executed, but their actions are not applied.

### Description
If the `execute` contract is called after a week and it fails, it will emit the `MotionExecuted` event with `success` set to `false`. 
We need to handle that and "close" the motion, so it won't display as pending.
We also need to handle the following case in block-ingestor:

> After one week, anyone with the multi-sig permissions (not just the creator) can cancel the motion without requiring approval.

- [ ] Active multi-sig motions where execution fails after a week, are marked as closed (action is not applied)
- [ ] Active multi-sig motions can be cancelled by any permission holders after a week

bassgeta:
## Description

This PR handles the case for any user to be able to reject a multi sig motion after a week and adds block-ingestor support for when a multi sig motion executes, but the action fails, it rejects the motion.
This can't be tested yet, it will be rechecked as part of #2189 
[Block ingestor PR](https://github.com/JoinColony/block-ingestor/pull/241)

## Testing

We're gonna be hacking dates hard. You will again need 2 users and 2 browsers.

1. Run the dev environment, install the multi sig extension
2. As `leela` give user `amy` `Owner` permissions for `Take actions via Multi-Sig`
3. As `leela` create a mint tokens motion, go to the other browser's colony homepage and check for the latest action in the activity widget and don't open it yet
image

4. Open up your dev console and paste in:
```JS
var oneWeekInMilliseconds = 7 * 24 * 60 * 60 * 1000 + 100;

Date.now = function() {
    return new Date().getTime() + oneWeekInMilliseconds;
};
```
5. Run `curl -X POST --data '{"jsonrpc":"2.0","method":"evm_increaseTime","params":[604900000],"id":1}' localhost:8545` in a terminal
6. This should've fast forwarded your motion for a week, and the console will return a week later for `Date.now()` until refresh.
7. Click `Reject action` from the meatball menu as `amy`
image
8. The motion should get rejected, the sidebar changing

## Diffs

**New stuff** ✨

* Another check if user can reject a motion directly if a week has passed
* In the block ingestor, if a motion executed btu the action failed, we just mark the multi sig motion as rejected (can't reproduce yet)

**Changes** 🏗

* Removed the `isSuccess` flag from a multiSig object

Resolves  #2197


src/components/v5/common/CompletedAction/partials/MultiSigMeatballMenu/MultiSigMeatballMenu.tsx
```diff
@@ -73,6 +75,15 @@ const MultiSigMeatballMenu: FC<MultiSigMeatballMenuProps> = ({
...
   };
 
+  const canCancelMultiSig =
+    !isRejected &&
+    (isOwner ||
+      Date.now() - new Date(multiSigData.createdAt).getTime() >
+        MULTISIG_EXPIRY_PERIOD);
+
+  const canRejectMultiSig =
+    !canCancelMultiSig && !isRejected && !isOwner && !showRejectMultiSigStep;
```
iamsamgibbs, line 85: I wonder if this variable should be renamed to `canRejectImmediately` / `canCancelWithoutVote` or something to avoid confusion between cancel and reject?

iamsamgibbs (approved):
Nice work on this! Code all looks good - tested following the instructions and working as expected.

Screenshot 2024-06-14 at 18 31 20

davecreaser (approved):
All looking good to me, nice one. Tested and working too 🥇 

Screenshot 2024-06-17 at 12 06 42

Screenshot 2024-06-17 at 12 06 50

rdig (approved):
Tested according to your script and works as expected _(although I had visual trouble finding the meatball menu to reject the motion)_
Screenshot from 2024-06-18 01-02-22
Screenshot from 2024-06-18 01-09-17
Screenshot from 2024-06-18 01-09-22

PR #2542, opened 2024-06-14
Feat: Set correct end time for streaming payments based on end condition

### LINKED ISSUE #2300: Streaming Payments API: Set correct `endTime` depending on end condition
Parent issue: https://github.com/JoinColony/colonyCDapp/issues/2207

Depending on the selected end condition, the `createStreamingPayment` saga should set the `endTime` parameter accordingly: 
- `WHEN_CANCELLED` should set the end time to max possible `uint256` value
- `LIMIT_REACHED` should calculate the timestamp at which limit will be reached and set it as the end time
- `FIXED_TIME` should accept end time as saga payload

davecreaser:
## Description

- Clean up the tmpAdvancedPayments ui as the file was growing too big and hard to read imo.
- Change the type of the timestamps stored for end time and start time for streaming payments to account for potentially long values. 
- Edit the createStreamingPayment saga so that it sets the endTime parameter according to the end condition.
  - For `WHEN_CANCELLED`, it sets the end time to the maximum possible uint256 value (2^256 - 1).
  - For `LIMIT_REACHED`, it calculates the timestamp at which the set limit would be reached and sets it as the end time.
  - For `FIXED_TIME`, it simply takes the end time passed in the saga payload.

See Block Ingestor PR here: https://github.com/JoinColony/block-ingestor/pull/242

## Testing

You may need to restart your dev environment after checking out this branch. Also be aware that there is a new block ingestor commit hash being used.

Ensure you also have streaming payments extension installed.

* Use the temporary streaming payments UI to create three streaming payments:
  * One with fixed time end condition. You should immediately be able to claim the amount multiplied by seven. (By default, the temporary UI creates streaming payments with start times one week in the past, an interval of one day, and an end time of one week in the future). 
  * One with a limit reached end condition (be sure to set a limit in the input box). It's easy to test this is working if you set the limit to _less_ than the amount multiplied by seven. You should only be able to claim the limit.
  *  One with a when cancelled end condition. To check this one, you'll need to go into `http://localhost:20002/` and use the following query to look for it. Check that the endTime of this streaming payment is set to "115792089237316195423570985008687907853269984665640564039457584007913129639935". While here, you could also check that the endTimes for the other two are different, and that the limited one has a lower endTime (assuming you put a low limit of less than two weeks worth of claims).
  
```
query MyQuery {
  listStreamingPayments {
    items {
      amount
      startTime
      nativeId
      interval
      id
      endTime
      createdAt
      metadata {
        endCondition
      }
    }
  }
}
```   

## Diffs

**New stuff** ✨

* New file `src/components/v5/payments/TmpStreamingPayments.tsx` which takes the streaming payments stuff and separates it out from the other advanced payments testing UI. It's 95% copy / pasted stuff.

**Changes** 🏗

* Updated createStreamingPayment saga to handle end times properly
* Updated types and ran graphql codegen 
* New block ingestor commit

Resolves #2300


iamsamgibbs (approved):
Nice work on this, all looks good and working as expected. Nice UI changes too.

Screenshot 2024-06-14 at 18 04 58
Screenshot 2024-06-14 at 18 05 24
Screenshot 2024-06-14 at 18 05 44
Screenshot 2024-06-14 at 18 06 08

src/redux/sagas/expenditures/createStreamingPayment.ts
```diff
@@ -103,6 +104,43 @@ function* createStreamingPaymentAction({
...
+
+      case StreamingPaymentEndCondition.WhenCancelled:
+        realEndTimestamp = TIMESTAMP_IN_FUTURE;
+        break;
+
+      default:
+        realEndTimestamp = TIMESTAMP_IN_FUTURE;
+        break;
+    }
+
```
jakubcolony, line 143: Perhaps this could live in a helper so it can also be reused by the motion version of the saga?

jakubcolony: Actually, will be easier if I move it in #2212

jakubcolony (approved):
Nice work @davecreaser, the numbers are looking correct 👍

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -3393,8 +3393,8 @@ type StreamingPayment @model {
   createdAt: AWSDateTime!
   recipientAddress: String!
   nativeDomainId: Int!
-  startTime: AWSTimestamp!
-  endTime: AWSTimestamp!
+  startTime: String!
+  endTime: String!
```
rdig, line 3397: What's the reason we're using plain strings to store the timestamps as opposed to the native AWS type ?

If it's because we're storing timestamps rather than actual time, maybe we should rename them to reflect that fact.

src/redux/sagas/expenditures/createStreamingPayment.ts
rdig: Agree, this should be extracted

rdig (approved):
Tested and everything seems to be functioning as expected. Nicely done.

Screenshot from 2024-06-18 23-14-03
Screenshot from 2024-06-18 23-15-04
Screenshot from 2024-06-18 23-15-37
Screenshot from 2024-06-18 23-16-07

amplify/backend/api/colonycdapp/schema.graphql
davecreaser: So the AWSTimestamp scalar wouldn't allow anything bigger than a 32 bit number, and since we're now using potentially 256 bits I had to change to a string.

PR #2544, opened 2024-06-16
Streaming Payments API: Creating streams using motions

### LINKED ISSUE #2212: Streaming Payments API: Creating streams using motions

jakubcolony:
## Description

This PR adds a saga to create a streaming payment using motions.

## Testing

...

## TODO

- [ ] Correctly set end time (similar to #2542, but the `Limit` variant is somewhat tricky as we don't know the start timestamp until the motion is finalized)

Resolves #2212


jakubcolony:
Marking as blocked, it requires updates to the StreamingPayment's `create` method to treat `endTime` as duration if `startTime == 0` to allow setting stream end condition to "Limit reached"

PR #2545, opened 2024-06-16
Include expenditure stakes amount in User Hub staked balance

### LINKED ISSUE #2537: Staked funds missing in UserHub for Staked Expenditures
## Steps to reproduce

1. Ensure you have the Staking advanced payment extension enabled.
2. Create an advanced payment using the Staking decision method.
3. Check the Userhub to see if you have any tokens staked.

## Expected behaviour

* You should show an accurate value of staked tokens to the user.

## Actual behaviour

* The value of staked tokens is not showing at all.

image

jakubcolony:
## Description

This PR ensures amounts staked on expenditures show up in the User Hub Staked section. 

The main change was switching `getUserTokenBalance` lambda to rely on `UserStake` which represents any stake the user does in the platform. It also allowed for a removal of the motions-specific `ColonyStake` model.

`ColonyStake` references also needed to be removed from block-ingestor - I decided not to bore you with another PR but here's [the commit](https://github.com/JoinColony/block-ingestor/pull/225/commits/ff6f5b651be38a9a8719b66b70a6eedcb2c3c6ff).

## Testing

- Validate staked amount behaviour when staking and finalizing motion
- Validate staked amount behaviour when staking for and reclaiming stake after releasing staked expenditure (use testing UI for that one)

Resolves #2537


amplify/backend/function/getUserTokenBalance/src/utils.js
```diff
@@ -47,15 +44,27 @@ const getStakedTokens = async (
...
+    data.getUserStakesInColony.items.forEach((item) => {
+      totalAmount = totalAmount.add(item.amount);
+    });
+  } else {
+    console.error(
+      `Could not get user ${walletAddress} stakes in colony ${colonyAddress}`,
+    );
+  }
+
+  return totalAmount.toString();
```
mmioana, line 67: Given we use a limit in the getUserStakesInColony results, shouldn't we make use of the pagination feature to ensure the amount is properly computed?

mmioana (commented):
Nice job, @jakubcolony!

amplify/backend/function/getUserTokenBalance/src/utils.js
jakubcolony: That's a fair point, I thought about it but then came to a conclusion it's pretty unlikely for any user to have more than 10,000 stakes, which is when pagination would be required. 

It's great that you're keeping an eye out on this, if there wasn't @index on userAddress, amplify would get 10,000 records (for any user in any colony), and only then apply filtering, and we would hit pagination issues much sooner.

mmioana: Fair, was thinking the limit is high enough ftm, just to keep in mind to implement pagination at some point to be sure we will not run into issues in the future

jakubcolony: @mmioana would you mind approving? Thanks!

src/components/common/Extensions/UserHub/partials/BalanceTab/partials/PendingReputation/PendingReputation.tsx
```diff
@@ -6,7 +6,7 @@ import useUserReputation from '~hooks/useUserReputation.ts';
 import TimeRelative from '~shared/TimeRelative/index.ts';
 import TitleLabel from '~v5/shared/TitleLabel/index.ts';
 
-import reputationTabClasses from '../../ReputationTab.styles.ts';
+import reputationTabClasses from '../../BalanceTab.styles.ts';
```
davecreaser, line 9: Nitpick, but should this object being imported be renamed to balanceTabClasses or similar? I see it's used in both Balance and Reputation things, so the file name change is a little confusing atm.

davecreaser:
All looks good, the total balance updates correctly when I stake in a reputation-weighted motion. Just one question came up from my testing there. I see that the Staked overview shows `0` in the userhub balance tab, but when I go into the Stakes tab via the `view` button I can see that I've staked almost 5 CREDS. Is that expected behaviour, since it's not claimable or something? I'm a little unsure on what the behaviour here should be.

Screenshot 2024-06-19 at 16 40 29
Screenshot 2024-06-19 at 16 55 12

Another thing which I might just not be understanding properly (it may be a dev time based thing) is when I use the temporary UI to create a staked expenditure, and then click `fund` the total balance does NOT update, but I see the `staked` amount in the balance tab go up. Maybe you could shed some light on what I should be expecting there!

mmioana (approved):

rumzledz (approved):
I can confirm that the User Hub gets an updated Staked tokens value after performing a Staked Expenditure and that once a user reclaims it, the value is updated as well.

**Side note: ** It still shows > 4 CREDS staked on this recording because I have another staked expenditure that I haven't reclaimed yet.

src/components/common/Extensions/UserHub/partials/BalanceTab/partials/PendingReputation/PendingReputation.tsx
jakubcolony: of course it should!

jakubcolony:
@davecreaser I think you accidentally pasted the same screenshot twice so I'm not sure what the issue could be. Could you try reproducing again?

davecreaser:
> @davecreaser Both your screenshots shows the Balance tab so I'm not sure what status the stake in question was in. Could you try reproducing again? It is expected for the stake to be shown in the Stakes list even after it's been claimed (and not be included in the total Staked balance).

Right, of course that would be helpful to show the stakes tab 😆 🤦 Here it is again:

Screenshot 2024-06-21 at 12 25 15
Screenshot 2024-06-21 at 12 25 24

As I said, this might be expected, I just wanted to check. Basically the `Staked` amount in the balance tab is 0, even though there is tokens being staked, as shown in the stakes tab. Is it 0 because I can't access those tokens yet?

jakubcolony:
@davecreaser should be fixed and ready for another look

iamsamgibbs (approved):
Changes look good and seems like the numbers are adding up now:

Screenshot 2024-06-27 at 15 49 00

Screenshot 2024-06-27 at 15 49 05

After claiming:

Screenshot 2024-06-27 at 15 50 26

Screenshot 2024-06-27 at 15 50 32

I don't think this has been caused by any of the changes here (so still approving it) but I don't seem to be able to create a staked payment here:

https://github.com/JoinColony/colonyCDapp/assets/34915414/f9b272bd-b1cd-4d3b-8db8-e0af48c37bbd

PR #2551, opened 2024-06-17
feat: manage tokens action ui

joanna-pagepro:
## Description

Manage tokens action UI

## Testing

* Step 1. Open action sidebar
* Step 2. Add a new token or remove the existing one

## Diffs

**New stuff** ✨

* `TokensTable` in `CompletedActions`
* `some` validator

**Changes** 🏗

* `TokenStatus` moved to `src/types/actions.ts` to use inside a form and in the completed action view
* moved action-related components into one directory

**Deletions** ⚰️

* removed the components that are not needed anymore, like `TokenRow`

## TODO

Resolves ?? (@AdrianDudko please link the issue)


iamsamgibbs (changes requested):
Thanks for picking up this is @joanna-pagepro ! Great to see this all coming together and a good start to the issue!

The completed action screen is looking great:

Screenshot 2024-06-17 at 14 06 52

And so is the motion screen:

Screenshot 2024-06-17 at 14 29 59

There are a few issues I've found though.

1. Required field validation error

Screenshot 2024-06-17 at 14 03 30

2. Adding new empty rows shows the 'Duplicate tokens are not allowed' error.

Screenshot 2024-06-17 at 14 03 37

3. It shouldn't be possible to click 'Ether' here and open the search box.

https://github.com/JoinColony/colonyCDapp/assets/34915414/2513382c-df05-457b-858f-06b05b53713c

This also means it is possible to replace the token and get the table into a weird state.

https://github.com/JoinColony/colonyCDapp/assets/34915414/374f8ce8-c220-47d3-8bf7-da098b03fb4c

4. On mobile, the table headings should be smaller:

Screenshot 2024-06-17 at 14 08 36

Figma:

Screenshot 2024-06-17 at 14 08 41

5. Token information including blockie / logo and symbol should be displayed in the edit form:

Screenshot 2024-06-17 at 14 11 20

Screenshot 2024-06-17 at 14 11 29

6. It is possible to add the native token twice and then will be impossible to remove.

https://github.com/JoinColony/colonyCDapp/assets/34915414/801bda7b-3c4f-4f4a-922b-b61d8b5ebf87

7. Invalid tokens can be added.

https://github.com/JoinColony/colonyCDapp/assets/34915414/a4576252-ed87-459a-8d55-bd8546178936

8. Replacing an existing token in the table leads to some weird behaviours.

https://github.com/JoinColony/colonyCDapp/assets/34915414/6a49a300-4b17-4750-a3e6-9d99edb18b14

Another example:

https://github.com/JoinColony/colonyCDapp/assets/34915414/8342f1dd-0dac-4f03-b27a-157c84252784

9. No results are shown in the token popover.

Screenshot 2024-06-17 at 14 23 57

Let me know if you have any questions regarding any of this issues.

iamsamgibbs:
(Regarding point 9 - Raul is working on a query for this)

src/types/actions.ts
```diff
@@ -168,3 +168,10 @@ export enum DecisionMethod {
   Permissions = 'Permissions',
   Reputation = 'Reputation',
 }
+
+export enum TokenStatus {
+  Added = 'Added',
+  Removed = 'Removed',
+  Unaffected = 'Unaffected',
+  NotEditable = 'NotEditable',
+}
```
jakubcolony, line 177: Considering this status only exists in the context of ActionSidebar form, should it be moved closer to that component?

jakubcolony (changes requested):
Nice start @joanna-pagepro, overall the feature is pretty functional 👍 

**I noticed the following issues:**
- Weird behaviour of the token input: attempting to enter address manually closes the dropdown instantly (I entered `0` in the field). It seems to work fine on the second attempt.

https://github.com/JoinColony/colonyCDapp/assets/112586815/f242d725-ef7e-4b2d-a679-9c464196e4ae

- Entering a well-formed address but one that's not a token causes vertical scrollbar to appear even though nothing overflows yet:
image

- There is no validation for well-formed addresses that aren't token addresses:

https://github.com/JoinColony/colonyCDapp/assets/112586815/0bda6650-7908-4c2a-b2b6-c9a2534c8da7

@iamsamgibbs @arrenv this might need your input as to where the validation is performed (presumably both frontend and block-ingestor?)

- Cancelling token removal causes the token to be removed from the array and validation error to come up (instead of just the "Removed" pill disappearing):

https://github.com/JoinColony/colonyCDapp/assets/112586815/98bce288-0b30-46c3-945a-eb7fb596ae55

---

**Other flows I tested that are working fine:**

- Adding a new token:

https://github.com/JoinColony/colonyCDapp/assets/112586815/8b71d94c-0cf2-46b4-91cb-c40f11772418

- Removing a token:


https://github.com/JoinColony/colonyCDapp/assets/112586815/e6fe76ba-6142-4636-98b6-b4815001684d

- Adding and removing tokens at the same time:

https://github.com/JoinColony/colonyCDapp/assets/112586815/232dc969-9cc8-497b-a453-a6336d41c108

- Validation when leaving empty rows works fine:
image

- Attempting to submit without making any changes:

image

- Duplicate token validation:

image

- Removing one of the added tokens:

https://github.com/JoinColony/colonyCDapp/assets/112586815/3640bbec-6e1e-4281-bb40-9615cd05653a

iamsamgibbs (changes requested):
Thanks for the extra work on this it is looking a lot better. I've experienced the same issues as Jakub. There is also this translation string error when the token is not found.

https://github.com/JoinColony/colonyCDapp/assets/34915414/55102ad9-3940-48d1-a784-4ca263d51567

@jakubcolony looks like you may have uncovered an issue with `useGetTokenFromEverywhereQuery` - it is being used on the token select component here (and on block-ingestor) so it should be validating tokens. I tried it here with the colony address (`0x88E8a4d2E64cA82ecd6bD5d8b11B2Fe9253939CC`) and again altering just the last character which was not allowed.

https://github.com/JoinColony/colonyCDapp/assets/34915414/91676560-c1cc-41f6-9feb-734410297b46

arrenv:
> There is no validation for well-formed addresses that aren't token addresses:
> 
> @iamsamgibbs @arrenv this might need your input as to where the validation is performed (presumably both frontend and block-ingestor?)

Yes, client side and in the block-ingestor make sense. I assume it would create all sorts of issues to have a non-token contract address in there.

joanna-pagepro:
Everything should be ok now, except one thing. Right now when you create a new action and remove the token that is added to the colony and then cancel this removing action, you will see an error with no changes in the table. I'm not able to add it to submit action only. Maybe you have some ideas on how to handle this case? If not, then maybe we can leave it like this and create a separate issue to come back to this later?

iamsamgibbs (changes requested):
Thanks for your work on this @joanna-pagepro - it is looking really close now!

I think the error validation you have mentioned in the comment above is completely fine.

I've only noticed two minor issues.

When adding a duplicate of an existing token, if you attempt to remove the original it will mark is as "removed" rather than removing the duplicate leading to some weird behaviour.

https://github.com/JoinColony/colonyCDapp/assets/34915414/479b1672-e461-4817-aae3-5a09c72a8bb0

Perhaps you could include all the colony tokens as part of the check for this error message (it seems as though it only checks the native token currently), then it won't be possible to have a duplicate of an existing token in the form.
 
Screenshot 2024-06-24 at 13 59 42

 On mobile, the token names are truncated much earlier than they should be. The kebab menu also seems to be taking up more space than is necessary on those rows too.
 
Screenshot 2024-06-24 at 14 03 08

Screenshot 2024-06-24 at 14 04 19

(The completed view is fine)

Screenshot 2024-06-24 at 14 09 46

---

Just so that it is noted, properly formatted token addresses are being accepted here, even if they are not valid tokens.
 
Screenshot 2024-06-24 at 14 05 34

But this is also the case in the 'create new colony' flow and is a result of issues with `useGetTokenFromEverywhereQuery`

Screenshot 2024-06-24 at 14 05 52

So I think this should be addressed separately to avoid holding this up.

joanna-pagepro:
Thanks @iamsamgibbs! Can you check if it's better now? :)

iamsamgibbs (changes requested):
Thanks for the changes!

This issue is resolved:

Screenshot 2024-06-25 at 12 57 15

And truncation is looking better on mobile (although it has made me think about the replacement of the added / removed pill - it might be better for it to sit inline with the symbol)

Screenshot 2024-06-25 at 12 59 35

However, these changes have affected tablet and desktop:

Screenshot 2024-06-25 at 12 55 33
Screenshot 2024-06-25 at 12 55 46

Also affecting tables elsewhere in the app.

Screenshot 2024-06-25 at 12 54 00
Screenshot 2024-06-25 at 12 54 52

joanna-pagepro:
@iamsamgibbs tables should be ok now. I think that we should ask @melyndav about the pill placement on mobile :)

joanna-pagepro:
@jakubcolony can you also take a look?

iamsamgibbs (approved):
Thanks for the changes! This all looks good to me now, tables elsewhere are restored and mobile is looking good now. 🎉

Screenshot 2024-06-27 at 09 53 48

Screenshot 2024-06-27 at 09 53 54

src/components/v5/common/ActionSidebar/partials/TokenSelect/partials/TokenSearchSelect/TokenSearchSelect.tsx
```diff
@@ -0,0 +1,152 @@
...
+  ) => {
+    const [searchValue, setSearchValue] = useState('');
+    const [debouncedSearchValue, setDebouncedSearchValue] = useState('');
+    const isMobile = useMobile();
+    const filteredList = useSearchSelect(items, debouncedSearchValue);
+
+    useEffect(() => {
+      const debounceTimeout = setTimeout(() => {
+        setDebouncedSearchValue(searchValue);
+      }, 500);
```
rumzledz, line 55: Not sure why we need to set up another 500 millisecond timeout here when we can just fully rely on the `debounce ()` function from `lodash` to handle it for us, or maybe there some edge-case I'm not aware of 🤔 But was thinking we can have a more concise and leaner approach to this:

```js
const [searchValue, setSearchValue] = useState('');
const [debouncedSearchValue, setDebouncedSearchValue] = useState('');
const isMobile = useMobile();
const filteredList = useSearchSelect(items, debouncedSearchValue);

// The only handler for delaying a state update for `debouncedSearchValue`
const debouncedSearchValueSetter = useCallback(
  debounce((value) => setDebouncedSearchValue(value), 500),
  [],
);

useEffect(() => {
  debouncedSetSearchValue(searchValue);

  return () => {
    // Maybe this was the behaviour you intended instead of a setTimeout?
    debouncedSetSearchValue.cancel();
  };
}, [searchValue, debouncedSetSearchValue]);

const onChange = useCallback(
  (value) => {
    onSearch?.(value);
    setSearchValue(value);
    // Removed the if statement here because in my example, we don't need `handleSearch` anymore
    // since the debounce logic is now handled by the useEffect when a state update is performed for `searchValue`
    // The sequence is:
    // 1. `setSearchValue(value)` <-- state update is performed for `searchValue`
    // 2. `useEffect` picks this up and invokes `debouncedSetSearchValue(searchValue)`
    // 3. debounce is performed within `debouncedSetSearchValue()`
  },
  [onSearch],
);
```

In my example, I've removed the following which to me felt redundant:
- `useMemo`
- `setTimeout`

rumzledz (approved):
This seems to have been a sizeable piece of work @joanna-pagepro, awesome stuff ✨ 

I can confirm that I can add and remove tokens in individual actions and the completed action component is showing correct information


I can also simultaneously add and remove tokens in one action and the completed action component is showing correct information
2551_b

arrenv (changes requested):
@joanna-pagepro Thank you for your work on this. All looks good, my only request would be to make sure we are using skeleton loader, rather then the spinners.

image

src/components/v5/common/ActionSidebar/partials/forms/ManageTokensForm/partials/TokensTable/TokensTable.tsx
```diff
@@ -0,0 +1,133 @@
...
+        if (!token) {
+          return;
+        }
+
+        trigger(`${name}.${index}.token`);
+      });
+    });
+
+    return () => subscription.unsubscribe();
+  }, [name, trigger, watch]);
```
jakubcolony, line 87: This component has both `useWatch` and the callback version of watch API, is there a reason for it?

The [docs](https://react-hook-form.com/docs/usewatch) suggest `useWatch` can offer better performance, and I'd lean towards it for better readability/consistency.

src/components/v5/common/ActionSidebar/partials/forms/ManageTokensForm/partials/TokensTable/TokensTable.tsx
```diff
@@ -0,0 +1,133 @@
...
+              !isTablet,
+            '[&_table]:table-auto': isTablet,
+          })}
+          getRowId={({ key }) => key}
+          columns={columns}
+          data={data}
+          getMenuProps={getMenuProps}
+          isDisabled={isDisabled}
+          verticalLayout={isTablet}
+          renderCellWrapper={(_, content) => content}
```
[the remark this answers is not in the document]
joanna-pagepro: This is needed to remove a wrapper from each cell

Nortsova:
Thanks again for your PR! Nice work ✨! 

I can confirm that tokens are added and removed successfully

video1

I can confirm that motion works: 
image

I can confirm that there is an error if no changes in the table:
image

I see that I can choose the same token that I already have:
image

It will show a duplication error, but still I can select this duplicate:
image

I can add multiple empty fields: 
image

Please, ignore me if some of my comments are expected behaviour.

Thanks!

src/components/v5/common/CompletedAction/partials/ManageTokens/partials/TokenStatusPill/TokenStatusPill.tsx
```diff
@@ -0,0 +1,31 @@
...
+  if (!showNewLabel && !showRemovedLabel) {
+    return null;
+  }
+
+  return (
+    <PillsBase
+      className={clsx('flex-shrink-0', {
+        'bg-negative-100 text-negative-400': showRemovedLabel,
+        'bg-success-100 text-success-400': showNewLabel,
+      })}
```
Nortsova, line 23: Just a small note: I propose using a `variant` prop instead of passing a class to this component. However, this change would require refactoring the entire `PillsBase` component and its usage. Therefore, it should be handled in a separate maintenance ticket. :)

joanna-pagepro:
@jakubcolony @arrenv @Nortsova can you check again?

src/components/v5/common/CompletedAction/partials/ManageTokens/partials/TokenStatusPill/TokenStatusPill.tsx
Nortsova: I created issue for that, no changes needed right now 😊 : https://github.com/JoinColony/colonyCDapp/issues/2666

Nortsova:
Thank you, @joanna-pagepro, for your quick fixes 🎉. I am approving this PR. Let's wait for feedback from @arrenv  and @jakubcolony  to ensure we haven't missed anything. 😊

Nortsova (approved):

arrenv (changes requested):
@joanna-pagepro Thank you. Apologies for not including a spec originally with my feedback. Can we change how the skeleton appears.

It would be good to have it more consistent with other skeleton loaders, looking like the following:
- width: 140px;
- height: 24px;

image


As opposed to this:

image

joanna-pagepro:
@arrenv loaders updated
@jakubcolony can you review it again?

iamsamgibbs (changes requested):
Thanks for the changes, the loading state looks good to me now.

Screenshot 2024-07-08 at 10 49 59
 
And is still functioning as expected.

Screenshot 2024-07-08 at 10 53 27

However, I've found this issue:

https://github.com/JoinColony/colonyCDapp/assets/34915414/b2c1247c-a33f-4c8f-9868-9031be3d5a9b

arrenv (changes requested):
@joanna-pagepro The skeleton is much better now, thank you.

I have just replicated the same issue as @iamsamgibbs though, essentially, the token should not be clickable and the dropdown should not appear if the token is existing and is being removed.

joanna-pagepro:
@arrenv @iamsamgibbs it should be ok now
@jakubcolony can you review it again?

iamsamgibbs (approved):
This looks good to me now! Previous issue is resolved. I'm able to create with permissions and via motion no problem.

Screenshot 2024-07-08 at 16 00 26

Screenshot 2024-07-08 at 15 59 02

Screenshot 2024-07-08 at 15 59 14

arrenv (approved):
Nice work @joanna-pagepro, thank you!

jakubcolony (approved):
This works well, nice work @joanna-pagepro


https://github.com/JoinColony/colonyCDapp/assets/112586815/ab307ff7-fc49-4f2e-bc92-a9f83a9fb69f

melyndav (approved):
Really nice work on this PR @joanna-pagepro, appreciate your effort on Manage tokens! 👍

Issue #2548, opened 2024-06-17
Unable to open settings menu on mobile when action sidebar is open


iamsamgibbs:
## Steps to reproduce

1. Set screen size to mobile
2. Create a new action to open the action sidebar
3. Attempt to open the settings menu

## Expected behaviour

* The settings menu opens on top of the action sidebar

## Actual behaviour

* The settings menu appears to open, but is obstructed by the action sidebar

## Suggestions

Probably just a z-index issue

https://github.com/JoinColony/colonyCDapp/assets/34915414/6ed6414f-ee8d-4dbd-a622-73660b3eabb4

PR #2567, opened 2024-06-18
fix: Unable to open settings menu on mobile when action sidebar is open

Nortsova:
## Description
Fixed opening settings in action sidebar. The problem was that the settings had top: `100% !important`, which helped position it under the header. However, in the case of the action sidebar, the "relative"-fixed value was higher on the wrapper with the header plus action form, so the settings screen was positioned down, under the action form.

Position: relative on the correct div fixed that.

## Testing

* Step 1 Set screen size to mobile
* Step 2 Create a new action to open the action sidebar
* Step 3 Attempt to open the settings menu

image

(Resolves | Contributes to) #2548

### DIFF
--- a/src/components/v5/common/ActionSidebar/ActionSidebar.tsx
+++ b/src/components/v5/common/ActionSidebar/ActionSidebar.tsx
@@ -235,49 +235,51 @@ const ActionSidebar: FC<PropsWithChildren<ActionSidebarProps>> = ({
       )}
       ref={registerContainerRef}
     >
-      <div className="flex w-full items-center justify-between border-b border-gray-200 px-6 py-4">
-        <div className="flex items-center gap-2">
-          <button
-            type="button"
-            className="flex items-center justify-center py-2.5 text-gray-400 transition sm:hover:text-blue-400"
-            onClick={closeSidebarClick}
-            aria-label={formatText({ id: 'ariaLabel.closeModal' })}
-          >
-            <X size={18} />
-          </button>
-          {!isMobile && (
-            <div className="flex items-center gap-4">
-              <button
-                type="button"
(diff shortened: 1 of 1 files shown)


src/components/v5/common/ActionSidebar/ActionSidebar.tsx
```diff
@@ -235,49 +235,51 @@ const ActionSidebar: FC<PropsWithChildren<ActionSidebarProps>> = ({
...
-                  <ArrowLineRight size={18} />
-                ) : (
-                  <ArrowsOutSimple size={18} />
-                )}
-              </button>
-              {action && !isMotion && !expenditure && (
-                <PillsBase
-                  className="bg-success-100 text-success-400"
-                  isCapitalized={false}
+      <div className="relative">
```
Nortsova, line 238: here is not that much changes, only added this div with relative class. I desided create a new div because when I added it to line below header border was hidden under opened settings menu

iamsamgibbs (approved):
Great job! This all looks good to me, and the settings menu appears unaffected elsewhere - nice fix 👍 

https://github.com/JoinColony/colonyCDapp/assets/34915414/a3e7f6ee-83e1-442b-b071-772a4e9a5856

rdig (approved):
Tested and can confirm it now works, I can open the settings menu while having an action open as well.

Nice solution. :100: 

[Screencast from 2024-06-19 00-00-18.webm](https://github.com/JoinColony/colonyCDapp/assets/1193222/136e02fd-5e52-4778-b412-4cafb164d1f9)

melyndav (approved):
Nice one @Nortsova, everything working as expected here! Appreciate your work 🌵

PR #2513, opened 2024-06-10
Staged Payments API: Releasing staged payments 

jakubcolony:
## Description

Please see issue for description.

Additionally, this PR fixes a regression introduced by myself some time ago where staged payments did not have their claim delays set to max possible value (and were therefore "released" straight after creation).

## Testing

Make sure you have Staged Expenditure and Voting Reputation extensions enabled.

**Setting correct claim delay on payment creation**
There was a regression that caused the stages to be immediately claimable.
- Create a staged payment
- Confirm the claim delay for all slots is set to max uint256 value

**Releasing staged payments (as owner)**
- Create a staged payment - you can adjust the parameters in `TmpAdvancedPayments`, by default there'll be 2 slots
- Lock, fund and finalize
- Test releasing single and multiple stages, confirm the funds have been paid out and run the query below to verify the data

**Releasing staged payment via arbitration**
- Create a staged payment as someone else
- Lock, fund and finalize
- As a user with `Arbitration` permissions, test releasing single and multiple stages

**Release staged payment via motion**
- Create a staged payment, lock, fund and finalize it
- Create a motion to release single and multiple stages, have it finalize and confirm the stages have been released

```gql
query ListStagedPayments {
  listExpenditures(filter: {type: {eq: STAGED}}) {
    items {
      nativeId
      slots {
        id
        payouts {
          amount
          isClaimed # should be `true` after releasing
        }
        claimDelay # max uint256 before releasing, 0 after
      }
      firstEditTransactionHash
      actions {
        items {
          id
          type
          expenditureSlotIds
          motionData {
            expenditureSlotIds
          }
        }
      }
    }
  }
}

```

There's a bunch of additional changes/refactoring in [block-ingestor](https://github.com/JoinColony/block-ingestor/pull/243).

TODO:
- [x] Clean up `ExpenditureStage` GraphQL type
- [x] Fix issue with claim delay not being set to max value for staged payments

Resolves #2486

PR #2573, opened 2024-06-18
Feat: Add `forward-time` script advancing EVM time

jakubcolony:
## Description

This is a cherry-pick of a commit from #2513 as I want to make it available before it gets merged. 

The days of manually copying curls from https://www.notion.so/colony/Time-in-development-f1283a43c09e46bb97f15accd438069b and adjusting the params (pity if you pasted into console before doing so!) are finally gone, as now you can simply run:

```
npm run forward-time -- <number_of_hours>
```

Which will do all the "heavy" lifting for you. 

On a side note, I really don't like the `--` bit but it's necessary when you run a script using npm run.

**Note:** as @rdig mentioned, advancing time when reputation monitor is on will not work properly, and similarly any time-bound features such as staking won't function, so please make sure to disable it before adjusting time manually.

## Testing

Navigate to some time-sensitive feature, e.g. motion staking. Run the script and it should advance the time as expected.

### DIFF
--- a/scripts/forward-time.js
+++ b/scripts/forward-time.js
@@ -1,0 +1,59 @@
+/**
+ * Script forwarding the local blockchain time by a given number of hours
+ */
+const fetch = require('node-fetch');
+
+const args = process.argv.slice(2);
+if (args.length !== 1) {
+  console.error('Usage: npm run forward-time -- <hours>');
+  process.exit(1);
+}
+
+const hours = parseFloat(args[0]);
+if (isNaN(hours) || hours <= 0) {
+  console.error('Please provide a valid number of hours.');
+  process.exit(1);
+}
+
+const seconds = hours * 60 * 60;
+
+const forwardTime = async () => {
+  try {
+    let response = await fetch('http://localhost:8545', {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify({
+        jsonrpc: '2.0',
+        method: 'evm_increaseTime',
+        params: [seconds],
+        id: 1,
+      }),
+    });
+
+    if (!response.ok) {
+      throw new Error('Request failed');
+    }
+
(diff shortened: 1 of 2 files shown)


rdig (approved):
It was about time :)

Do note _(and maybe warn people about)_ that if the reputation monitor is toggled on, which itself forwards time, running this script will actually forward time twice _(at least)_

rumzledz (approved):
It's now time for me to get rid of my alias 😂 

```bash
# Colony Advance Block Time
advanceTime() {
  curl -X POST --data '{"jsonrpc":"2.0","method":"evm_increaseTime","params":[3600],"id":1}' localhost:8545
}
```

scripts/forward-time.js
```diff
@@ -0,0 +1,59 @@
...
+    response = await fetch('http://localhost:8545', {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify({
+        jsonrpc: '2.0',
+        method: 'evm_mine',
+        params: [],
+        id: 1,
+      }),
+    });
```
mmioana, line 46: Would it make sense to extract the API call into a separate method to which we pass the method and params?

```
const postRequest = ({ method, seconds }) => 
  fetch('http://localhost:8545', {
      method: 'POST',
      headers: { 'Content-Type': 'application/json' },
      body: JSON.stringify({
        jsonrpc: '2.0',
        method,
        params: seconds ? [seconds] : [],
        id: 1,
      }),
    });
```

and then call it with
```
response = await postRequest({method: 'evm_increaseTime', seconds})
```
or
```
response = await postRequest({method: 'evm_mine'})
```

mmioana (approved):
Great work @jakubcolony 👏

scripts/forward-time.js
jakubcolony: Potentially, it's a simple script that I do not envisage changing though

Issue #1768, opened 2024-01-21
Use enums instead of `as const` objects


jakubcolony:
This issue is to replace all occurrences of `as const` objects with string enums.

Example:
```ts
export const TOKENS_MODAL_TYPES = {
  activate: 'activate',
  deactivate: 'deactivate',
} as const;
```

should become:
```ts
export enum TokensModalType {
  Activate = 'activate',
  Deactivate = 'deactivate'
}
```

rdig:
You're now assigned to it, @malcyb123

PR #2576, opened 2024-06-19
Fix: update permissions upon member field change

### LINKED ISSUE #2547: Permissions field not properly reset when changing members in action sidebar
## Steps to reproduce

1. Create a new `Manage permissions` action.
2. Fill out the entire form ensuring you select a member and a permissions type. (Note the table at the bottom providing details on the selected permission).
3. Change the `Member` field to select a different member.

## Expected behaviour

* The new member is selected and the selected permission remains selected.

## Actual behaviour

* The new member is selected and the selected permission visually appears to still be selected in the field, but the value has actually been cleared - the details table at the bottom of the sidebar will disappear and you will get a validation message when attempting to submit the form.

https://github.com/JoinColony/colonyCDapp/assets/34915414/c0aec897-ad82-48e7-bb90-3e1bfd46382e

mmioana:
## Description

- Update permissions upon member field change in Manage Permissions action

## Testing

TODO: Check permissions are updated when changing the member in the Manage Permissions action

* Step 1. Create a Manage Permissions action
* Step 2. Select a member
* Step 3. Select permissions
* Step 4. Select another member
* Step 5.1. Check permissions are reset if user has no permissions already set
* Step 5.2. Check permissions are set based on existing user permissions
* Step 6. Check you can continue to next step without getting an error message for the permissions field if it's already filled

Resolves #2547

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/forms/ManagePermissionsForm/consts.ts
+++ b/src/components/v5/common/ActionSidebar/partials/forms/ManagePermissionsForm/consts.ts
@@ -3,7 +3,10 @@ import { type InferType, mixed, object, string, number } from 'yup';
 
 import { CUSTOM_USER_ROLE } from '~constants/permissions.ts';
 import { formatText } from '~utils/intl.ts';
-import { ACTION_BASE_VALIDATION_SCHEMA } from '~v5/common/ActionSidebar/consts.ts';
+import {
+  ACTION_BASE_VALIDATION_SCHEMA,
+  DECISION_METHOD_FIELD_NAME,
+} from '~v5/common/ActionSidebar/consts.ts';
 import { type CardSelectOption } from '~v5/common/Fields/CardSelect/types.ts';
 
 export const FIELD_NAME = {
@@ -13,17 +16,21 @@ export const FIELD_NAME = {
   ROLE: 'role',
   AUTHORITY: 'authority',
   PERMISSIONS: 'permissions',
-  DECISION_METHOD: 'decisionMethod',
+  DECISION_METHOD: DECISION_METHOD_FIELD_NAME,
 } as const;
+
+type PermissionRole = `role_${(typeof AVAILABLE_ROLES)[number]}`;
 
 export const validationSchema = object()
   .shape({
-    member: string().required(),
-    team: number().required(),
-    createdIn: number().required(),
-    role: string().required(),
-    authority: string().required(),
-    permissions: mixed<Partial<Record<string, boolean>>>().test(
+    [FIELD_NAME.MEMBER]: string().required(),
+    [FIELD_NAME.TEAM]: number().required(),
+    [FIELD_NAME.CREATED_IN]: number().required(),
+    [FIELD_NAME.ROLE]: string().required(),
+    [FIELD_NAME.AUTHORITY]: string().required(),
+    [FIELD_NAME.PERMISSIONS]: mixed<
+      Partial<Record<PermissionRole, boolean>>
+    >().test(
       'permissions',
       'You have to select at least one permission.',
       (value, { parent }) => {
@@ -34,7 +41,7 @@ export const validationSchema = object()
         return Object.values(value || {}).some(Boolean);
       },
     ),
-    decisionMethod: string().defined(),
+    [FIELD_NAME.DECISION_METHOD]: string().defined(),
   })
   .defined()
   .concat(ACTION_BASE_VALIDATION_SCHEMA);
--- a/src/components/v5/common/ActionSidebar/partials/forms/ManagePermissionsForm/consts.ts
+++ b/src/components/v5/common/ActionSidebar/partials/forms/ManagePermissionsForm/consts.ts
@@ -6,6 +6,16 @@ import { formatText } from '~utils/intl.ts';
 import { ACTION_BASE_VALIDATION_SCHEMA } from '~v5/common/ActionSidebar/consts.ts';
 import { type CardSelectOption } from '~v5/common/Fields/CardSelect/types.ts';
 
+export const FIELD_NAME = {
+  MEMBER: 'member',
+  TEAM: 'team',
+  CREATED_IN: 'createdIn',
+  ROLE: 'role',
+  AUTHORITY: 'authority',
+  PERMISSIONS: 'permissions',
+  DECISION_METHOD: 'decisionMethod',
+} as const;
+
 export const validationSchema = object()
   .shape({
     member: string().required(),
--- a/src/components/v5/common/ActionSidebar/partials/forms/ManagePermissionsForm/hooks.ts
+++ b/src/components/v5/common/ActionSidebar/partials/forms/ManagePermissionsForm/hooks.ts
@@ -72,6 +72,7 @@ export const useManagePermissions = (
 
       setValue(FIELD_NAME.ROLE, roleValue, {
         shouldValidate: !!roleValue || (isSubmitted && !isValid),
+        shouldDirty: true,
       });
 
       if (userRole.role !== UserRole.Custom) {
--- a/src/components/v5/common/ActionSidebar/partials/forms/ManagePermissionsForm/hooks.ts
+++ b/src/components/v5/common/ActionSidebar/partials/forms/ManagePermissionsForm/hooks.ts
@@ -32,8 +32,11 @@ export const useManagePermissions = (
   const decisionMethod: DecisionMethod | undefined = useWatch({
     name: FIELD_NAME.DECISION_METHOD,
   });
-  const { setValue, watch } =
-    useFormContext<Partial<ManagePermissionsFormValues>>();
+  const {
+    setValue,
+    watch,
+    formState: { isValid, isSubmitted },
+  } = useFormContext<Partial<ManagePermissionsFormValues>>();
   const { colony } = useColonyContext();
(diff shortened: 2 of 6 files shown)


iamsamgibbs (approved):
Thanks for picking up my issue, this all looks good to me!

https://github.com/JoinColony/colonyCDapp/assets/34915414/d9253be7-83e8-4dc5-aab7-0f1e211ec3c8

I like the new `FIELD_NAME` const, although it might be better as an enum just for the sake of consistency as they are more widely used throughout the app.

src/components/v5/common/ActionSidebar/partials/forms/ManagePermissionsForm/consts.ts
```diff
@@ -6,6 +6,16 @@ import { formatText } from '~utils/intl.ts';
...
 import { type CardSelectOption } from '~v5/common/Fields/CardSelect/types.ts';
 
+export const FIELD_NAME = {
+  MEMBER: 'member',
+  TEAM: 'team',
+  CREATED_IN: 'createdIn',
+  ROLE: 'role',
+  AUTHORITY: 'authority',
+  PERMISSIONS: 'permissions',
+  DECISION_METHOD: 'decisionMethod',
```
jakubcolony, line 16: Looks like we ended up with 2 sources for the decision method field name, this will only work as long as `DECISION_METHOD_FIELD_NAME` has exactly the same value, which is not obvious.

Please either set the value of this property to the existing `DECISION_METHOD_FIELD_NAME`, or pass `FIELD_NAME.DECISION_METHOD` as a prop to `DecisionMethodField`, so that we don't rely on 2 separate variables holding the same value.

Is there a reason for using `as const` objects as opposed to string-based enums? I understand they solve some problems with numeric enums but otherwise I think enums would be the preference as per #1768.

jakubcolony (changes requested):
Nice work on this, I like the removal of hardcoded field names!

Interestingly I was not able to reproduce the original issue on `master` (see recording below). @iamsamgibbs was I doing something wrong?
 
I left a comment regarding having 2 sources of field name for the decision method, but I also spotted a potential regression:

On `master`, selecting a user with existing permissions automatically sets the value for that field:

https://github.com/JoinColony/colonyCDapp/assets/112586815/faa858a8-1a57-4237-a0e1-57a899edbeca

Whereas on your branch it doesn't happen:

https://github.com/JoinColony/colonyCDapp/assets/112586815/d4ca1842-4ab7-463b-88b6-2f022c5e8918

iamsamgibbs:
> Interestingly I was not able to reproduce the original issue on `master` (see recording below). @iamsamgibbs was I doing something wrong?

It seems to only occur when redoing a completed action:

https://github.com/JoinColony/colonyCDapp/assets/34915414/d9aabb0d-bbfc-4241-9239-c2352850238b

(Although I'm sure I also experienced this when creating a new action, but I can't seem to recreate it now)

src/components/v5/common/ActionSidebar/partials/forms/ManagePermissionsForm/hooks.ts
```diff
@@ -66,7 +66,10 @@ export const useManagePermissions = (
       });
       const userRole = getRole(userPermissions);
 
-      setValue('role', userRole.permissions.length ? userRole.role : undefined);
+      setValue(
+        FIELD_NAME.ROLE,
+        userRole.permissions.length ? userRole.role : undefined,
+      );
```
rumzledz, line 72: I could see the intention for using an enum to achieve a strictly typed field name approach. My concern here is that we're declaring the field name in a separate object which means there's a chance that it can be out-of-sync with the validation schema. Currently, we're inferring the type of the schema to identify the field names.

So in the case of:

```js
// Before
const decisionMethod: DecisionMethod | undefined = useWatch({
  name: DECISION_METHOD_FIELD_NAME,
});
  
 // After
const decisionMethod: DecisionMethod | undefined = useWatch({
  name: FIELD_NAME.DECISION_METHOD,
});
```

We've now essentially created 3 different sources for the field names:
1. The validation schema's inferred type
2. The current constant variables we're using on `master` i.e. `DECISION_METHOD_FIELD_NAME`
3. Then the inferred type from the validation schema itself

A better approach would be to pass the type to `useWatch`:

```js
const decisionMethod: DecisionMethod | undefined = useWatch<ManagePermissionsFormValues>({
  name: 'this will be a keyof ManagePermissionsFormValues',
})
```

Then to address the loosely typed `ActionFormRow` component, we can amend that component a little bit by adding a generic type to its interface so instead of

```js
<FormCardSelect
          name={FIELD_NAME.ROLE}
          cardClassName="max-w-[calc(100vw-2.5rem)] md:max-w-sm md:px-4 md:[&_.section-title]:px-2"
          renderSelectedValue={(option, placeholder) =>
            getRoleLabel(option?.value) || placeholder
```

We can do:

```js
<FormCardSelect<ManagePermissionsFormValues>
          name="keyof ManagePermissionsFormValues"
          cardClassName="max-w-[calc(100vw-2.5rem)] md:max-w-sm md:px-4 md:[&_.section-title]:px-2"
          renderSelectedValue={(option, placeholder) =>
            getRoleLabel(option?.value) || placeholder
```

But perhaps this will increase the overall complexity and hours of the ticket 😅

iamsamgibbs (dismissed):
Nice work on this! Thanks for making the changes we discussed with Romeo. (Apologies for not getting this all clarified when originally writing the ticket!)

https://github.com/JoinColony/colonyCDapp/assets/34915414/4432f766-4d24-473c-a426-b2a3bb49e44d

I've found one tiny issue if you are able to fix it (but I've still approved as it such an edge case)

https://github.com/JoinColony/colonyCDapp/assets/34915414/a3bebd44-d5ec-4f94-81d1-b168d32a7a92

When redoing an action and changing members, the permissions field will reset as intended, but the text will remain `text-gray-900` rather than returning to the placeholder `text-gray-400`

Screenshot 2024-06-20 at 16 09 46

Screenshot 2024-06-20 at 16 09 57

src/components/v5/common/ActionSidebar/partials/forms/ManagePermissionsForm/consts.ts
mmioana: I have changed the `FIELD_NAME.DECISION_METHOD` to reference `DECISION_METHOD_FIELD_NAME`, so we don't end up with multiple sources of truth. Didn't do it from the beginning as I saw many fields being spread across multiple files and was thinking a bigger refactor task might be needed.

As for the `as const` objects, I have no strong opinion for one or the other. 
However, when trying to refactor to an enum, the `setValue( `${FIELD_NAME.PERMISSIONS}.role_${colonyRole}`, userRole.permissions.includes(colonyRole))` triggered an error that boolean can't be assigned to type never, which I'm not sure yet why it appeared.

iamsamgibbs (approved):
Thanks for fixing up this small edge case too! Everything else is still working as intended.

https://github.com/JoinColony/colonyCDapp/assets/34915414/c19eebe6-380f-47d1-a56f-b59ad5802f3c

src/components/v5/common/ActionSidebar/partials/forms/ManagePermissionsForm/hooks.ts
```diff
@@ -23,36 +22,40 @@ import {
...
+      setValue(FIELD_NAME.AUTHORITY, Authority.Own);
     }
   }, [isModeRoleSelected, setValue]);
 
   useEffect(() => {
     const { unsubscribe } = watch(({ member, team }, { name }) => {
       if (
         !name ||
-        !['team', 'member'].includes(name) ||
+        !([FIELD_NAME.MEMBER, FIELD_NAME.TEAM] as string[]).includes(name) ||
```
bassgeta, line 58: is the casting necessary here? does the `as const` break type inferring?

bassgeta (approved):
1. 
image
2.
image
3.
image

All works as intended, love the changes with field names as constants, we need to do that for all the forms down the line!

rumzledz (dismissed):
I can confirm that the Permissions are being populated correctly based on a member's permission with a selected team. Awesome work @mmioana ✨ 



A quick side note: The testing steps were missing the bit about selecting a Team which is needed for the entire formula 😂 But yeah, all good!

src/components/v5/common/ActionSidebar/partials/forms/ManagePermissionsForm/consts.ts
jakubcolony: I think that makes it a good exception to the "prefer enums over const object" guideline.

Also came across this https://github.com/react-hook-form/react-hook-form/issues/8898

jakubcolony (changes requested):
Setting existing permissions works fine now:

https://github.com/JoinColony/colonyCDapp/assets/112586815/5a14a5d8-80fe-4cca-b99f-116327dbd700

It also works fine for existing custom permissions, however when I start switching the domain, they seem to go rogue:


https://github.com/JoinColony/colonyCDapp/assets/112586815/13e173ea-6354-4063-941e-00efa008a443

mmioana:
@jakubcolony you're right! 
It was a memoization problem with the table cells, preventing the FormSwitch to properly update

However, after discussing with @rumzledz this PR conflicts with the changes he's doing for the forms, so I'll only leave my PR open in case some of the changes I've made are of any help

PR #2583, opened 2024-06-20
fix: Select member incorrect tooltip on actions

### LINKED ISSUE #2514: Select member incorrect tooltip on actions
## Steps to reproduce

1. Open any action type that has a select member input field
2. Hover over the placeholder
3. The incorrect tooltip will show as per the following screenshot

image

## Expected behaviour

This tooltip should not show on any select member placeholder across all action types and action tables.

Nortsova:
## Description
Fix: Select member incorrect tooltip on actions

## Testing
1. Open any action type that has a select member input field
2. Hover over the placeholder
3. The tooltip will NOT show

image

4. The tooltip will show only for error state

image

Resolves #2514

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/UserSelect/UserSelect.tsx
+++ b/src/components/v5/common/ActionSidebar/partials/UserSelect/UserSelect.tsx
@@ -89,8 +89,6 @@ const UserSelect: FC<UserSelectProps> = ({
 
   const userName = getUserName();
   const isUserAddressValid = isAddress(field.value);
-  const isUserAddressInvalid =
-    !isUserAddressValid && (selectedUser || field.value);
 
   const toggler = (
     <button
@@ -159,13 +157,9 @@ const UserSelect: FC<UserSelectProps> = ({
 
   const selectedUserContent = (
     <>
-      {tooltipContent || isUserAddressInvalid ? (
+      {tooltipContent ? (
         <Tooltip
-          tooltipContent={
-            isUserAddressInvalid
-              ? formatText({ id: 'actionSidebar.addressErrorTooltip' })
-              : tooltipContent
-          }
+          tooltipContent={tooltipContent}
           placement="top"
           // selectTriggerRef={(triggerRef) => {
           //   if (!triggerRef) {
--- a/src/components/v5/common/ActionSidebar/partials/UserSelect/UserSelect.tsx
+++ b/src/components/v5/common/ActionSidebar/partials/UserSelect/UserSelect.tsx
@@ -89,6 +89,8 @@ const UserSelect: FC<UserSelectProps> = ({
 
   const userName = getUserName();
   const isUserAddressValid = isAddress(field.value);
+  const isUserAddressInvalid =
+    !isUserAddressValid && (selectedUser || field.value);
 
   const toggler = (
     <button
@@ -157,12 +159,12 @@ const UserSelect: FC<UserSelectProps> = ({
 
   const selectedUserContent = (
     <>
-      {tooltipContent || !isUserAddressValid ? (
+      {tooltipContent || isUserAddressInvalid ? (
         <Tooltip
           tooltipContent={
-            isUserAddressValid
-              ? tooltipContent
-              : formatText({ id: 'actionSidebar.addressErrorTooltip' })
+            isUserAddressInvalid
+              ? formatText({ id: 'actionSidebar.addressErrorTooltip' })
+              : tooltipContent
           }
           // selectTriggerRef={(triggerRef) => {
           //   if (!triggerRef) {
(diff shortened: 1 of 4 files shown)


src/components/v5/common/ActionSidebar/partials/UserSelect/UserSelect.tsx
```diff
@@ -171,8 +173,8 @@ const UserSelect: FC<UserSelectProps> = ({
 
           //   return triggerRef.querySelector(`.${LABEL_CLASSNAME}`);
           // }}
-          trigger={!isUserAddressValid ? 'hover' : undefined}
-          placement={isUserAddressValid ? 'top' : 'bottom'}
+          trigger={isUserAddressInvalid ? 'hover' : undefined}
```
iamsamgibbs, line 176: I don't think this line actually does anything as the default trigger type for the tooltip component is hover anyway - so this entire line can be removed.

src/components/v5/common/ActionSidebar/partials/UserSelect/UserSelect.tsx
```diff
@@ -171,8 +173,8 @@ const UserSelect: FC<UserSelectProps> = ({
...
+          placement={isUserAddressInvalid ? 'bottom' : 'top'}
```
iamsamgibbs, line 177: Feels a bit weird to me to have this one specific tooltip placed below, but I've asked Mel to clarify whether this is necessary or not. Will add a comment here.

iamsamgibbs (changes requested):
This looks good to me! The tooltip is now displaying correctly:

Screenshot 2024-06-20 at 15 30 40

I've left two small comments, they're both on tweaks to the existing code rather than additions, but I'm not sure either is actually necessary.

src/components/v5/common/ActionSidebar/partials/UserSelect/UserSelect.tsx
iamsamgibbs: Just for reference, the only other place I could find this tooltip being used:

Screenshot 2024-06-20 at 15 30 20

iamsamgibbs: > Feels a bit weird to me to have this one specific tooltip placed below, but I've asked Mel to clarify whether this is necessary or not. Will add a comment here.

Confirmation from Mel that all tooltips can also be positioned above, so this ternary can be removed. :)

Nortsova: thanks! updated code according to this comments :) 

image

iamsamgibbs (dismissed):
Thanks for the changes, looks good to me!

Screenshot 2024-06-21 at 11 34 35
Screenshot 2024-06-21 at 11 34 44

melyndav (dismissed):
Great work on this PR @Nortsova! Tooltips are removed where requested. 🦘 

image

rumzledz (changes requested):
I've follows the specified testing steps:

> 1. Open any action type that has a select member input field
> 2. Hover over the placeholder
> 3. The tooltip will NOT show
> 4. The tooltip will show only for error state

But it seems the tooltip does not show for an error state. @melyndav @Nortsova can you please confirm if these changes only cover scenarios where the address itself is incorrect? And not when the field has an error in general? Thank you!

2583 confirm if bug

mmioana (commented):
When adding an inexistent address the tooltip shows as expected

Screenshot 2024-07-01 at 19 15 30

However, if you enter a wrong value (anything but an address), there is no tooltip displayed. Is this behaviour intended?
Screenshot 2024-07-01 at 19 14 50

melyndav:
@Nortsova @rumzledz this 'address error' state has been incorrectly picked up from the CSV upload feature for advanced payments table. It should not be inherited into any member field outside this. 

When an error occurs, the normal error validation should happen where the field highlights red and the standard error notification box is shown below the action fields.

image

No custom error tool-tip should be used in the top action fields.

Nortsova:
According to @melyndav comments I removed tooltip from any place except Advanced payment - table.

Also additional fix for table overflow:
| Before | After |
| ---- | ---- |
| 
image
 | 
image
 |


## Examples with no tooltip:
### Manage reputation:
image
---

### Manage permissions:
image
---

### Simple payment
image
---

rumzledz (approved):
@Nortsova I saw that you fixed the Advance Payment's "Payments" table overflow as well, awesome stuff ✨ 

I can confirm that the "Address error" error state only shows a tooltip on hover when the Action form is set to Advanced Payments:

melyndav (approved):
Nice work on this PR @Nortsova and all the updates included. Thanks again 🐰

mmioana (approved):
Awesome work @Nortsova 👏 

Now the tooltip is shown only in the Advanced payments action
Screenshot 2024-07-04 at 16 49 24
Screenshot 2024-07-04 at 16 49 09
Screenshot 2024-07-04 at 16 49 45
Screenshot 2024-07-04 at 16 50 09

Issue #2456, opened 2024-05-29
Fix issue with overlapping transaction pending pill


arrenv:
Teamwork task: [Fix issue with overlapping transaction pending pill](https://tw.auroracreation.com/#/tasks/15935062)

## Steps to reproduce

1. Create an action, so that the side panel is open.
2. Click on the UserHub.
3. Click on the Activate button.
4. Activate some tokens.
5. Notice the transaction pending button overlapping.

## Expected behaviour

* You should not see a double up of the transaction pending pill/button

## Actual behaviour

* There is currently an overlapping of the transaction pending button when in a modal.

![image](https://github.com/JoinColony/colonyCDapp/assets/33682027/837d7d44-89cd-483a-916f-c7f70b79d2d9)

PR #2637, opened 2024-07-02
fix: user hub placement

joanna-pagepro:
## Description

Fix for transaction pending button overlapping issue

## Testing

* Step 1. Create an action, so that the side panel is open.
* Step 2. Click on the UserHub.
* Step 3. Click on the Activate button.
* Step 4. Activate some tokens.
* Step 5. Check if the transaction pending button is still overlapping.

## Diffs

**New stuff** ✨

* 

**Changes** 🏗

* blur-related styles are now added to elements with `modal-blur` or `modal-blur-navigation` classes
* there's no duplication of the user navigation component on the desktop view - the same component is shown when a modal is open and closed
* The user navigation component is moved to the top when the `shouldShowHeader` prop is passed to the `Modal` component

**Deletions** ⚰️

* 

## TODO

- [ ] Add TODOs

Resolves https://github.com/JoinColony/colonyCDapp/issues/2456

### DIFF
--- a/src/components/frame/Extensions/layouts/ColonyLayout.tsx
+++ b/src/components/frame/Extensions/layouts/ColonyLayout.tsx
@@ -1,3 +1,4 @@
+import clsx from 'clsx';
 import { AnimatePresence } from 'framer-motion';
 import React, {
   type FC,
@@ -19,7 +20,7 @@ import { useColonyContext } from '~context/ColonyContext/ColonyContext.ts';
 import { useColonyCreatedModalContext } from '~context/ColonyCreateModalContext/ColonyCreatedModalContext.ts';
 import { useMemberModalContext } from '~context/MemberModalContext/MemberModalContext.ts';
 import { usePageHeadingContext } from '~context/PageHeadingContext/PageHeadingContext.ts';
-import { useTokensModalContext } from '~context/TokensModalContext/TokensModalContext.ts';
+import { useTablet } from '~hooks';
 import { TX_SEARCH_PARAM } from '~routes/index.ts';
 import ActionSidebar from '~v5/common/ActionSidebar/index.ts';
 import ColonyCreatedModal from '~v5/common/Modals/ColonyCreatedModal/index.ts';
@@ -56,6 +57,7 @@ const ColonyLayout: FC<PropsWithChildren> = ({ children }) => {
   const [isActionSidebarOpen, { toggleOn: toggleActionSidebarOn }] =
     actionSidebarToggle;
   const txButtons = useGetTxButtons();
+  const isTablet = useTablet();
 
   const {
     isMemberModalOpen,
@@ -67,7 +69,6 @@ const ColonyLayout: FC<PropsWithChildren> = ({ children }) => {
     useColonyCreatedModalContext();
   // const [isInviteMembersModalOpen, setIsInviteMembersModalOpen] =
   //   useState(false);
-  const { isTokensModalOpen } = useTokensModalContext();
 
   const { calamityBannerItems, canUpgrade } = useCalamityBannerInfo();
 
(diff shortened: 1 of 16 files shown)


melyndav (changes requested):
Hey @joanna-pagepro 🦖 

When I attempt to test this, the userhub closes the side panel and then automatically closes straight after as shown in the recording below:

Export-1720010849272

melyndav (dismissed):
Thanks for that fix @joanna-pagepro, I've been able to test it and everything is looking resolved now. nice work! 👍 

Export-1720688781133

src/context/ActionSidebarContext/ActionSidebarContextProvider.tsx
```diff
@@ -51,7 +53,12 @@ const ActionSidebarContextProvider: FC<PropsWithChildren> = ({ children }) => {
 
   actionSidebarUseRegisterOnBeforeCloseCallback((element) => {
     const isClickedInside = isElementInsideModalOrPortal(element);
-    if (!isClickedInside) {
+    const navigationWrapper = document.querySelector('.modal-blur-navigation');
```
mmioana, line 56: Can we please refactor `document.querySelector` in a utils function?
```
const getElementWithSelector = (selector) => document.querySelector(selector);
```
and use it here `const navigationWrapper = getElementWithSelector('.modal-blur-navigation')`

src/components/v5/shared/Modal/Modal.tsx
```diff
@@ -30,15 +24,29 @@ const Modal: FC<PropsWithChildren<ModalProps>> = ({
...
+      shouldShowHeader &&
+      !document.body.classList.contains('show-header-in-modal')
+    ) {
+      document.body.classList.add('show-header-in-modal');
+    }
+
+    return () => {
+      document.body.classList.remove('show-header-in-modal');
+    };
+  }, [isOpen, shouldShowHeader]);
```
mmioana, line 44: Can we extract this in a hook?

```
const useAddClassToElement = ({ shouldAddClass, element, className }) => {

 useEffect(() => {
    if (
      shouldAddClass &&
      !element.classList.contains(className)
    ) {
      element.classList.add(className);
    }

    return () => {
      element.classList.remove(className);
    };
  }, [shouldAddClass, element, className]);

}
```

And use it here as 
```
useAddClassToElement({ 
  shouldAddClass: isOpen && shouldShowHeader,
  className: 'show-header-in-modal',
  element: document.body
})
```

src/context/ActionSidebarContext/ActionSidebarContextProvider.tsx
```diff
@@ -51,7 +53,12 @@ const ActionSidebarContextProvider: FC<PropsWithChildren> = ({ children }) => {
...
+
+    if (
+      !isClickedInside ||
+      (navigationWrapper && navigationWrapper.contains(element) && !isTablet)
```
mmioana, line 60: Can we please refactor this part `navigationWrapper && navigationWrapper.contains(element)` into an utils function?
```
const isChildOf = (parent, child) => parent && parent.contains(child);
```
and use it here `(isChildOf(navigationWrapper, element) && !isTablet)`

src/components/frame/Extensions/layouts/ColonyLayout.tsx
```diff
@@ -91,17 +92,22 @@ const ColonyLayout: FC<PropsWithChildren> = ({ children }) => {
...
-              {/* Hide Initially */}
-              {/* {!isActionSidebarOpen ? (
+    (isHidden?: boolean) => (
+      <UserNavigationWrapper
+        txButtons={txButtons}
+        userHub={userHub}
+        className={clsx(
+          'modal-blur-navigation [.show-header-in-modal_&]:z-[1001]',
+          {
+            'relative z-[11]': !isTablet,
```
mmioana, line 102: Is anything stopping us from using `z-20`?

mmioana (changes requested):
Cool work @joanna-pagepro ✨ 
I didn't manage to reproduce the issue anymore 👍  however I left some code refactoring comments

src/context/ActionSidebarContext/ActionSidebarContextProvider.tsx
joanna-pagepro: What is the advantage of doing so?

src/components/frame/Extensions/layouts/ColonyLayout.tsx
joanna-pagepro: I just think that 11 is enough here. I haven't tested it with 20

joanna-pagepro:
@mmioana can you check the changes and my answers?

src/context/ActionSidebarContext/ActionSidebarContextProvider.tsx
mmioana: The advantage would be to abstract the way elements are selected. 
In case additional logic will be needed in the future, eg. log an error if the element is not found for a specific selector, only the utility function will need to be updated.

src/components/frame/Extensions/layouts/ColonyLayout.tsx
mmioana: In this case, we could add these values (11 and 1001) in `tailwind.config.js` if the already defined values are not usable in this scenario

mmioana (changes requested):
Thank you @joanna-pagepro for your prompt changes! 💮 
Could please tackle also the other two comments?

joanna-pagepro:
@mmioana done

mmioana (approved):
Thank you @joanna-pagepro for your work! 🚀 💟

melyndav (approved):
Re-approving this one @joanna-pagepro! Thanks again for your work. 🤩

PR #2654, opened 2024-07-04
Feat: Edit colony description and objective for multi-sig

mmioana:
## Description

Create and finalize a multi-sig edit colony details and manage colony objective motions.

- [X] The edit colony details multi-sig motion can be created
- [X] The manage colony objective multi-sig motion can be created
- [X] The edit colony details motion can be finalized
- [X] The manage colony objective motion can be finalized
- [X] Only users with [adequate permissions](https://www.notion.so/colony/Multi-sig-Permissions-6aaf0b95309f4754a243d54e8d9c4189?pvs=4#fec3699a5fd241a2925bd2809494b157) can create them

## Testing

Check Edit Colony details motion and Manage objective motion can be created with both Multi-Sig and Permissions decision method.

### Edit Colony details action
* Step 1. Install Multi-Sig extension
* Step 2. Create new action - Edit Colony details 
* Step 3. Select Multi-Sig decision method
* Step 4. Update Colony description field
* Step 5. Create action
* Step 6. Finalise action - only for Multi-Sig
* Step 7. Check colony description was updated on dashboard
* Step 8. 🔁 Repeat above steps for the Permissions decision method 🔁

### Manage objective action
* Step 1. Install Multi-Sig extension
* Step 2. Create new action - Manage objective
* Step 3. Select Multi-Sig decision method
* Step 4. Update Objective description field
* Step 5. Create action
* Step 6. Finalise action - only for Multi-Sig
* Step 7. Check colony objective was updated on dashboard
* Step 8. 🔁 Repeat above steps for the Permissions decision method 🔁

[Block ingestor PR](https://github.com/JoinColony/block-ingestor/pull/250)
Resolves [#2191](https://github.com/JoinColony/colonyCDapp/issues/2191)

### DIFF
--- a/src/components/v5/common/CompletedAction/CompletedAction.tsx
+++ b/src/components/v5/common/CompletedAction/CompletedAction.tsx
@@ -120,7 +120,6 @@ const CompletedAction = ({ action }: CompletedActionProps) => {
       case ColonyActionType.ColonyEditMotion:
       case ColonyActionType.EditExpenditureMotion:
       case ColonyActionType.FundExpenditureMotion:
-      case ExtendedColonyActionType.UpdateColonyObjective:
         // @NOTE: Enabling those 2 above temporarily
         return <Motions transactionId={action.transactionHash} />;
       case ColonyActionType.CreateExpenditure:
--- a/src/components/v5/common/CompletedAction/CompletedAction.tsx
+++ b/src/components/v5/common/CompletedAction/CompletedAction.tsx
@@ -85,6 +85,7 @@ const CompletedAction = ({ action }: CompletedActionProps) => {
         return <ManageReputation action={action} />;
       case ColonyActionType.ColonyEdit:
       case ColonyActionType.ColonyEditMotion:
+      case ColonyActionType.ColonyEditMultisig:
         return <EditColonyDetails action={action} />;
       case ExtendedColonyActionType.UpdateColonyObjective:
       case ExtendedColonyActionType.UpdateColonyObjectiveMotion:
--- a/src/components/v5/common/CompletedAction/CompletedAction.tsx
+++ b/src/components/v5/common/CompletedAction/CompletedAction.tsx
@@ -89,6 +89,7 @@ const CompletedAction = ({ action }: CompletedActionProps) => {
         return <EditColonyDetails action={action} />;
       case ExtendedColonyActionType.UpdateColonyObjective:
       case ExtendedColonyActionType.UpdateColonyObjectiveMotion:
+      case ExtendedColonyActionType.UpdateColonyObjectiveMultisig:
         return <UpgradeColonyObjective action={action} />;
       // @TODO: Connect this to the reputation actions
       /* case ColonyActionType.EmitDomainReputationReward:
--- a/src/components/v5/common/CompletedAction/CompletedAction.tsx
+++ b/src/components/v5/common/CompletedAction/CompletedAction.tsx
@@ -120,6 +120,7 @@ const CompletedAction = ({ action }: CompletedActionProps) => {
       case ColonyActionType.ColonyEditMotion:
       case ColonyActionType.EditExpenditureMotion:
       case ColonyActionType.FundExpenditureMotion:
+      case ExtendedColonyActionType.UpdateColonyObjective:
         // @NOTE: Enabling those 2 above temporarily
         return <Motions transactionId={action.transactionHash} />;
       case ColonyActionType.CreateExpenditure:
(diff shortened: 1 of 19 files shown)


Nortsova (changes requested):
Thanks for your PR, @mmioana!

I can confirm that after the "Edit colony" action is created with a "Multi-sig" motion, changes are not applied by default.

video-1-before-approve

After finalizing this action, the changes apply to the colony description:

video-2-after-approve
 
I can also confirm that the "Edit colony" action created with a "Permissions" motion has changes applied instantly:

video-3-edit-permissions

The "Manage objective" action created with "Multi-sig" motion is not applied instantly (before finalization):

objective-1-not-finalize

It is applied after the action is finalized:

objective-2-finalized

However, for the "Manage objective" action with "Permissions," the app crashes with this error after submission:

image

Also I found an error with the description field where the max length is 200, but a colony has already been created with more than 3k symbols. I'm not sure where this mistake originates, but is it possible to solve it easily? 😊

image

Thank you! ✨

mmioana:
Hi @Nortsova thank you for taking the time to review this PR. 💟 

You are right! With muti-sig changes should be visible only after the action is finalised - I will update the testing steps.

For the 'Manage objective' with permissions seems there was a bug, but now it should be solved ✨ 

Regarding the colony description/objective description length I have already updated the create-data script on master to truncate the saved values. If it helps, I can cherry-pick it into this branch.

mmioana:
@Nortsova you were right! Actually the colony description character limit was relying on the objective description char limit, so in the create-data script I completely omitted it.
Now there should be a MAX_COLONY_DESCRIPTION which ensures the colony description is not larger than 200 chars also for the test data

rumzledz (changes requested):

rumzledz:
@mmioana I only found out about this, you should also update the `en-actions.ts` file to include a translation entry for your multi-sig action so it doesn't end up as "Generic action..." in the `Latest activity` table and "Generic" in the action type field in the Completed Action component:

image

image

rumzledz (approved):

Nortsova (changes requested):
Hey @mmioana , thank you for the updates :) 

The "Manage Objectives with Permissions" feature is working now, thanks! 💫

However, I found a small issue: after using "Manage Objectives with Multi-Sig," the company description gets removed (social links as well, I didn't show it on video):

manage-objective-bug

And another thing: I'm not sure if this was mentioned to be done, but right now I can submit "Edit Colony Details" and "Manage Objectives" even if the form hasn't been changed.

So for now, I am rejecting only due to the bug where the description and links are removed after the "Manage Objective" motion with multi-sig is called. ✨

bassgeta (changes requested):
Nice work on this :ok_hand: 
I was able to edit details via Multi-sig
image
and I can confirm that other stuff doesn't get wiped anymore
image
Managing an objective also works!
image

Maybe we can just rework how voting reputation gets the child index, so we use the local helpers and after that it's good to go in my book!

src/redux/sagas/motions/editColonyMotion.ts
```diff
@@ -129,33 +118,96 @@ function* editColonyMotion({
...
+
+      const rootDomain = colonyDomains.find((domain) =>
+        BigNumber.from(domain.nativeId).eq(Id.RootDomain),
+      );
+
+      if (!rootDomain) {
+        throw new Error('Cannot find rootDomain in colony domains');
+      }
+
+      const childSkillIndex = yield call(
```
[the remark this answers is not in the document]
mmioana: "If we don't need the permissions domain ID (we don't in this case, edit colony always happens in root)"
"As long as the parentDomainId and domainId match, there's no network call. Child skill index, as far as I know, is just a number that relates two domains together so if 2 domains are the same, it's always that MaxUint256"

src/components/v5/common/CompletedAction/CompletedAction.tsx
```diff
@@ -117,6 +120,7 @@ const CompletedAction = ({ action }: CompletedActionProps) => {
       case ColonyActionType.ColonyEditMotion:
       case ColonyActionType.EditExpenditureMotion:
       case ColonyActionType.FundExpenditureMotion:
+      case ExtendedColonyActionType.UpdateColonyObjectiveMotion:
         // @NOTE: Enabling those 2 above temporarily
```
iamsamgibbs, line 124: Tiny tiny note, but it would be better to have this two lines higher as it now makes the comment incorrect.

src/i18n/en-actions.ts
iamsamgibbs, line 1: Nice extra pickups here :)

iamsamgibbs (approved):
Great work on this! It is all working great, and all the code looks great. Nice job! 🎉 

Screenshot 2024-07-12 at 10 17 35

Screenshot 2024-07-12 at 10 17 41

Screenshot 2024-07-12 at 10 03 15

Screenshot 2024-07-12 at 10 04 09

Screenshot 2024-07-12 at 10 05 25

Screenshot 2024-07-12 at 10 05 59

Screenshot 2024-07-12 at 10 07 40

davecreaser (approved):
Nice job! 🎉 

Code looks all good, and works as expected when testing:

Screenshot 2024-07-12 at 12 03 16

Issue #2192, opened 2024-04-10
[Multi-sig] feat: implement manage roles multi-sig motion


bassgeta:
### Story

As a *user moderating a colony*, I want to be able to create and finalize a multi-sig manage roles motion.

### Description

After the groundwork has been laid with #2150 , implement the manage roles motion.

### Permissions Issue

Whilst you're here, it'd also be great to fix an existing issue with manage permissions:

How the manage permissions saga is currently setup: In the root domain, only users with the root role can create a manage permissions action. In any subdomain, only users with the architecture role can create a manage permissions action.

`domainId === Id.RootDomain ? ColonyRole.Root : ColonyRole.Architecture`

How it should be setup:  In the root domain, only users with the root role can create a manage permissions action. In any subdomain, only users with the **root role OR architecture role** can create a manage permissions action.

It should be fairly straightforward to adapt the saga to check for either permission when in a sub domain. However, this change also affects `PERMISSIONS_NEEDED_FOR_ACTION`, `getPermissionsNeededForAction` and `addressHasRoles` which are primarily used to show an error validation message when attempting to create an action you do not have permissions for. These currently have an array of required permissions, they will need to be adjusted to accommodate an array of valid permission sets where any set is acceptable.

As for multi-sig specifically, changing `PERMISSIONS_NEEDED_FOR_ACTION` will have knock on effects to `useEligibleSignees` and `useDomainThresholds` which will need adjusting. If a user has either permission it should be possible for them to vote towards the motion (you will need to check the exact implementation on the contract to check if the permission in the multisig motion is created with holds any relevance).

This will likely also impact how approvals / rejections / thresholds are calculated.

### This issue is complete when...

- [ ] The manage roles multi-sig motion can be created
- [ ] The motion can be finalized
- [ ] Only users with [adequate permissions](https://www.notion.so/colony/Multi-sig-Permissions-6aaf0b95309f4754a243d54e8d9c4189?pvs=4#fec3699a5fd241a2925bd2809494b157) can create them
- [ ] The standard `Manage Permissions` saga has been updated to allow either root or architecture in sub domains
- [ ] The necessary knock on issues this will cause are addressed

iamsamgibbs:
Just noting here it is no longer a requirement to be able to create (or vote on) a multi-sig manage permissions motion with the `root` permission in a sub-domain.

It is still required when using the permissions decision method.

I will see if anything can be simplified thanks to this change.

PR #2674, opened 2024-07-08
Feat: Require any permission refactor

iamsamgibbs:
## Description

When creating a `Manage Permissions` action in a sub-domain via permissions it should be possible to use either `root` or `architecture` permissions. This PR addresses an existing issue where it was only possible to use `architecture`.

This change has had major knock on implications to how permissions are checked elsewhere as they now need to accommodate for this EITHER permissions check. `PERMISSIONS_NEEDED_FOR_ACTION` has been refactored into arrays of valid role groups to achieve this.

Luckily this has meant some quality of life improvements have been implemented. For example, the `Manage reputation` action requires `root` if awarding reputation or `arbitration` if removing reputation. Previously the permission checks required award / remove to be selected, now we can check if the user has either permission and show the permissions notice earlier if they don't have either.

This PR also introduces `getAnyPermissionProofsLocal` to get permission proofs if a user has any of the valid roles. This is now used in the action saga for `Manage Permissions`, but will also be helpful for multi-sig voting on actions which allow voting on multiple roles (Simple Payment).

Also fixed up a bunch of permissions bugs as I came across them, such as checking permissions in the `createdIn` field rather than the `from` / `team` fields (this wasn't necessary before multi-sig as permissions don't get checked for reputation voting, but now this check is necessary). `Edit domain` wasn't checking for permissions in a sub-domain. `Simple payment` wasn't checking for all permissions in the same sub-domain.

### Testing a user can create a `Manage Permissions` action in a sub-domain with the `root` permission.

1. Create a `Manage permissions` action.
2. Select Amy as the user, General as the team and Custom permissions. Select `Take actions on their own` as the authority, and use the permissions decision method. Select the `root` custom permission.
3. Amy should now have `root` permissions in general. (Note it is not possible to have root permissions in any sub-domain.
4. Switch to Amy in an incognito browser.
5. Create a `Manage permissions` action.
6. Select any sub-domain as the team, any user, any set of permissions.
7. It should be possible to submit the action, and the relevant changes should be applied. (Note: You will need to refresh in order to see the new permissions in the user popover)

Screenshot 2024-07-15 at 10 31 13

Also feel free to test the same using the `architecture` role in either a sub-domain or the root domain as either role should be valid.

### Testing other actions have not been affected by the changes without the reputation voting or multi-sig extension installed

This is a bit of a pain to test thoroughly as there are so many different scenarios. I have done my best to account for every eventuality, so hopefully some general testing is sufficient here.

The permissions notification banner should show as expected when a user does not have the relevant permissions in any domain. Testing this is a bit of a pain, but essentially give a user only enough permissions to perform an action in a relevant domain and check the permissions banner does not appear (or test the opposite, deliberately make sure a user does not have sufficient permissions and check the banner does appear). Check `PERMISSIONS_NEEDED_FOR_ACTION` to confirm which role(s) are required for each action.

Screenshot 2024-07-15 at 10 34 48

Particular actions of interest are: `Simple payment` as it requires three different roles all in the same domain. `Manage reputation` as it uses different roles dependant on whether you are awarding or removing reputation. `Manage permissions` as it allows different roles dependant on which domain the action is created in.

Test this without the reputation voting or multi-sig extension installed as the logic works slightly differently when they are not installed (you should later test again with either and / or both installed to check this still functions as expected).

### Testing a user can create a `Manage permissions` action in a sub-domain with the `architecture` **multi-sig** permission

Unlike core permissions, when creating a `Manage permissions` actions in a sub-domain using **multi-sig** it is only possible to use the `architecture` role and not the `root` role.

1. Install the multi-sig extension, set a fixed threshold of 2.
2. Using the manage permissions action, give Amy custom permissions for just the `architecture` role in general. Set authority to multi-sig and create the action using the permissions decision method.
3. Switch to Amy in an incognito browser and create a `Manage permissions` multi-sig action.
4. Select the 'multi-sig' decision method. Then select 'general' as the team. The permissions banner should show as it would require the `root` multi-sig permission.
5. Switch to any other team and the permissions banner should disappear as sub-domains require the `architecture` role.

Feel free to go through the rest of the multi-sig flow to confirm it functions as expected.

Could also test this further by assigning `root` multi-sig permissions to a user. They should be able to create a `Manage permissions` multi-sig action in General, but not in any other team.

### Testing other actions have not been affected by the changes without extensions installed

As above but testing again now that the multi-sig extension is installed. Ideally also worth testing with both extensions installed, then with just the voting reputation extension installed - but that is a lot of testing.

## Diffs

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

* Changes go here

**Deletions** ⚰️

* Deletions go here

## TODO

One TODO left in the code, but it applies to such a tiny edge case that I don't feel it is worth addressing here. (The `useHasNoDecisionMethods` function could show the permissions banner sooner if we refactored it to return rolesPerDomain, then check that all the required roles are in a single domain - currently it returns all the permissions a user has in any domain. Once the `from` domain is selected, this check does occur and the banner shows, so it would only be a very minor improvement and would require a fairly large refactor)

Contributes to #2192


src/redux/sagas/utils/proofs.ts
```diff
@@ -196,52 +206,124 @@ const getMultiPermissionProofsLocal = async ({
...
       requiredDomainId,
-      requiredColonyRoles: requiredColonyRole,
+      requiredColonyRoles: roleGroup,
       permissionAddress,
       isMultiSig,
     });
   }
-  return getSinglePermissionProofsLocal({
+  const roleGroupSet = requiredColonyRoles as RoleGroupSet;
+  // [[ColonyRole.Root]] // [[ColonyRole.Root, ColonyRole.Architecture]] // [[ColonyRole.Root], [ColonyRole.Architecture]]
```
rdig, line 320: Leftover ?

rdig (approved):
Holy moly this was a lot a work. Thanks for putting in the time Sam and refactoring this. I know it wasn't particularly easy _(or pleasant)_

Below I've tested most cases you listed out or that I could think of, and everything seems to be working as expected.

Once again, stellar work! :100: 

Using Root Permissions:
Screenshot from 2024-07-16 00-19-11
Screenshot from 2024-07-16 00-20-21

Using Architecture Permissions:
Screenshot from 2024-07-16 00-21-37
Screenshot from 2024-07-16 00-22-24

Simple payment using incorrect permissions:
Screenshot from 2024-07-16 00-26-45

Simple payment using correct permissions:
Screenshot from 2024-07-16 00-27-40
Screenshot from 2024-07-16 00-28-17

Manage permissions with incorrect permissions:
Screenshot from 2024-07-16 00-31-50

Manage permissions with correct permissions:
Screenshot from 2024-07-16 00-32-30
Screenshot from 2024-07-16 00-32-54

Manage permissions with multisig in root:
Screenshot from 2024-07-16 00-35-24
Screenshot from 2024-07-16 00-36-44

Manage permissions with multisig in subdomain:
Screenshot from 2024-07-16 00-36-51

Simple payment with incorrect permissions, while multisig is installed:
Screenshot from 2024-07-16 00-38-28

Simple payment with correct permissions, while multisig is installed:
Screenshot from 2024-07-16 00-39-22
Screenshot from 2024-07-16 00-39-43

bassgeta (approved):
Really great work on this! It's an annoying as hell issue, but a nice solution with 2d arrays we should be able to cover all edge cases :)
Amy can manage permissions as root anywhere:
image
Amy with just funding Multi-sig can't create an action:
image
She can create it
image
But not via permissions:
image
Amy also can't modify permissions in root via multisig without root
image
But can do so in subteams:
image

I've tested editing a subteam and I was able to sign it as `amy` with architecture and finalize the motion:
image

Let's merge this badboi and onwards to victory!

src/redux/sagas/utils/proofs.ts
iamsamgibbs: Intentional to show the different permission sets that we are expecting to get through to this part of the function. I should probably have made the comment clearer though!

davecreaser (approved):
Awesome work Sam! This looks like it was a huge pain, so nice job 😆

Code all looks excellent (just one question/comment from something that stood out).

Testing all went well too, I tested a few scenarios without going totally insane, here's some examples:

Managing permissions as root:
Screenshot 2024-07-19 at 16 37 49
Screenshot 2024-07-19 at 16 39 25
Screenshot 2024-07-19 at 16 39 51

Trying to do architecture / root stuff as Root:
Screenshot 2024-07-19 at 16 41 53
Screenshot 2024-07-19 at 16 42 14

Testing with Payer permissions in a sub team:
Screenshot 2024-07-19 at 16 42 38
Screenshot 2024-07-19 at 16 42 49
Screenshot 2024-07-19 at 16 42 56

Multi-Sig testing:
Screenshot 2024-07-19 at 16 46 03
Screenshot 2024-07-19 at 16 47 36
Screenshot 2024-07-19 at 16 47 46

src/components/v5/common/ActionSidebar/hooks/permissions/helpers.ts
```diff
@@ -35,9 +36,50 @@ export const getPermissionsNeededForAction = (
...
     case Action.ManagePermissions: {
-      return formValues.createdIn === Id.RootDomain
+      const { decisionMethod, createdIn, team } = formValues;
+
+      const isMotion = decisionMethod !== DecisionMethod.Permissions;
+
+      let createdInDomain;
+      if (!isMotion) {
+        createdInDomain = team;
+      } else if (team) {
```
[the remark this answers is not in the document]
iamsamgibbs: This is intentional, I should probably add a comment here as Kerry asked about the same thing!

I want to use the `createdIn` field as the value, but only once the `team` field has a value.

When you switch to a motion decision method, the `createdIn` field gets set to `General` straight away, whilst the `Team` field remains empty. So without this check, if you don't have permissions in `General` the `no permissions banner` would show even if you have permissions in a sub-domain, which felt misleading to me - so better to wait until a team has been selected.

PR #2675, opened 2024-07-08
Action Sidebar cleanup

chmanie:
## Description

**Don't be afraid! Most of the changed files were just moved around**

Apparently there was a lot of code that hung out unused in the `ActionSidebar` component. It seems most of it was copied into the `CompletedAction` component without being deleted in the `ActionSidebar` itself. I tried to identify all of what was left unused and removed it. In this process I moved a lot of files to the `CompletedAction` component as it is now only used there.

This PR serves as a base for some greater `ActionSidebar` refactoring action.

## Testing

Try to use the `ActionSidebar` as you're used to. If you can find an edge case that was used and I couldn't find, please report it :) Thank you!

## Diffs

**Deletions** ⚰️

* Loads of unused code in the `ActionSidebar`


src/components/v5/common/ActionSidebar/ActionSidebar.tsx
```diff
@@ -286,31 +114,14 @@ const ActionSidebar: FC<PropsWithChildren<ActionSidebarProps>> = ({
...
-                <MotionOutcomeBadge motionState={motionState} />
               </div>
             )}
-            {isMobile && getShareButton()}
           </div>
           <div>{children}</div>
         </div>
       </div>
-      {getSidebarContent()}
+      <ActionSidebarContent defaultValues={initialValues} />
```
davecreaser, line 124: ☺️ So much nicer!

src/components/v5/common/ActionSidebar/partials/ActionSidebarContent/ActionSidebarContent.tsx
```diff
@@ -30,36 +31,33 @@ import useHasActionPermissions from '../../hooks/permissions/useHasActionPermiss
 import useHasNoDecisionMethods from '../../hooks/permissions/useHasNoDecisionMethods.ts';
 import useActionFormProps from '../../hooks/useActionFormProps.ts';
 import useSidebarActionForm from '../../hooks/useSidebarActionForm.ts';
+import { type ActionFormBaseProps } from '../../types.ts';
```
davecreaser, line 34: I think the above 4 imports would be better off being absolute imports. Generally I think it's good to have relative imports only for files that are in the same folder, or, at a stretch the parent folder. Any more than that and I start to get itchy 🤣 Happy to hear opinions if you disagree though!

(This also kinda matches what is written [here](https://github.com/JoinColony/ColonyFrontEndLivingStandard/blob/f09a6d489c6b496a067b34838dece3107da1a15e/repo/index.md?plain=1#L43))

src/components/v5/common/ActionSidebar/partials/ActionSidebarContent/ActionSidebarContent.tsx
```diff
@@ -30,36 +31,33 @@ import useHasActionPermissions from '../../hooks/permissions/useHasActionPermiss
...
-import {
-  type ActionSidebarContentProps,
-  type ActionSidebarFormContentProps,
-} from './types.ts';
+import { type ActionSidebarContentProps } from './types.ts';
 
 const displayName = 'v5.common.ActionsContent.partials.ActionSidebarContent';
 
-const ActionSidebarFormContent: FC<ActionSidebarFormContentProps> = ({
+const ActionSidebarFormContent: FC<ActionFormBaseProps> = ({
```
davecreaser, line 47: I'm not really a big fan of having two React components made inside one file, especially ones this size, so I'd prefer that they were split into two files. I do see the reasoning behind doing it like this since they're so tightly coupled, but I do think it'd be cleaner to separate them. Again, happy to hear opposing opinions though!

davecreaser (changes requested):
Really nice to have the journey started on this cleanup! 🧹 

Just left a couple of comments code-wise. They are a little nitpicky perhaps, but since this is a code cleanup PR, where better to be nitpicky eh?

Also found a few issues when testing:

- I can't open completed actions, it's just opening an empty Form.

https://github.com/user-attachments/assets/423e5e18-e314-446e-8b7d-f561ab2f0daa

- Closing the sidepanel doesn't remove the transaction hash query parameter, so when I refresh the page the sidepanel is opening again when it should be closed.

https://github.com/user-attachments/assets/89255543-7ae9-4744-97ef-4dee99d47a8c

- After closing the sidepanel, I can't reopen the same action again (possibly related to the above issue with the query param?)

https://github.com/user-attachments/assets/fc43403b-917c-49bc-b6c7-d6a9548dd555

- After completing an action, it does not progress to the completed action state, but instead just allows me to keep creating the action

https://github.com/user-attachments/assets/0c1acf66-56e3-4338-9c09-5bc7c1064348

src/components/v5/common/ActionSidebar/partials/ActionSidebarContent/ActionSidebarContent.tsx
chmanie: I agree. But it's not something that I introduced in this PR :P

chmanie:
Put back to draft as this is not nearly as done as I thought it was.

chmanie:
Closed in favor of #2820

PR #2682, opened 2024-07-09
Feat: Add extra checks to fetchTokenFromChain to better validate address

### LINKED ISSUE #2671: Any properly formatted address is accepted as a valid token address
## Steps to reproduce

1. Run `node scripts/create-colony.url.js` in your terminal and open the URL in your browser.
2. Follow the create colony flow and select "Use an existing token".
3. Paste in any properly formatted address `0x0000000000000000000000000000000000000001`, Colony addresses will also work.

Screenshot 2024-07-08 at 15 10 43

## Expected behaviour

* An error notice should display the entered address is not a valid token address.

Screenshot 2024-07-08 at 14 57 50

## Actual behaviour

* No error notice will display and the address is accepted as a valid token address.

Screenshot 2024-07-08 at 15 00 57

## Suggestions

This is caused by the `fetchTokenFromChain` lambda not actually validating if the address is a valid token address. It will be an issue when the manage tokens refactor is merged as it also uses this lambda.

I've tried looking into this a little bit and these might be helpful:

`provider.getCode` (https://docs.ethers.org/v5/api/providers/provider/#Provider-getCode) will return `0x` if no contract is deployed to the address, which partially solves the problem, but does not confirm if it is a token.

`provider.getBalance` will return 0 even if the address is not a token address, but was also returning 0 when testing with the CREDS wallet address and ƓƓƓ wallet address.

iamsamgibbs:
## Description

This PR adds some validation to the `fetchTokenFromChain` lambda to better check is the provided address is a "valid" token. (It is not really possible to check if it is a "valid" token, but it implements the following: )

* The function already returns the token if it exists in our database
* If it does not exist as a token in our database, it will now check if it exists as either a colony address or a user address and will throw an error if it is already in use for one of these as it therefore means it cannot be a token address.
* If it still does not exist in our database, it will call the required ERC20 functions - if none of these throw an error, it is likely an ERC20 token contract

This is probably still not foolproof, but it does at the very least prevent users from putting in colony addresses instead of token addresses and wondering why it is not working.

## Testing

* Run the `create-data` script
* Run the `node scripts/create-colony-url.js`
* Follow the URL and follow the colony creation flow, select to use an existing token
* Open the amplify console in docker if you want to check the error messages
* Test out some different addresses. Copy the colony address and paste it in. Copy a user address and paste it in. Copy a valid token address (CREDS / ƓƓƓ) and confirm that is allowed. Paste in any other properly formatted address: `0x0000000000000000000000000000000000000001`

Screenshot 2024-07-09 at 16 35 49

## Diffs

**New stuff** ✨

* Check address is not already in use by a colony or user
* Check none of the required ERC20 functions throw an error

Resolves #2671

### DIFF
--- a/amplify/backend/function/fetchTokenFromChain/src/index.js
+++ b/amplify/backend/function/fetchTokenFromChain/src/index.js
@@ -13,7 +13,12 @@ const {
  * So that we can always ensure it follows the latest schema
  * (currently it's just saved statically)
  */
-const { createToken, getTokenByAddress } = require('./graphql');
+const {
+  createToken,
+  getTokenByAddress,
+  getColonyByAddress,
+  getUserByAddress,
+} = require('./graphql');
 
 let apiKey = 'da2-fakeApiId123456';
 let graphqlURL = 'http://localhost:20002/graphql';
@@ -94,6 +99,30 @@ exports.handler = async (event) => {
 
   if (tokenQuery && !token) {
     try {
+      // Check this address is not currently in use by a colony
+      const colonyQuery = await graphqlRequest(
+        getColonyByAddress,
+        { address: tokenAddress },
+        graphqlURL,
+        apiKey,
+      );
+
+      if (colonyQuery?.data?.getColonyByAddress?.items.length > 0) {
+        throw new Error(`${tokenAddress} is already in use by a colony.`);
+      }
+
+      // Check this address is not currently in use by a user
+      const userQuery = await graphqlRequest(
+        getUserByAddress,
+        { address: tokenAddress },
+        graphqlURL,
+        apiKey,
+      );
+
+      if (userQuery?.data?.getUserByAddress?.items.length > 0) {
+        throw new Error(`${tokenAddress} is already in use by a user.`);
+      }
+
       /*
        * Attempt to fetch it from the chain
        */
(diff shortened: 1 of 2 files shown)


davecreaser (approved):
Very nice @iamsamgibbs ! I tested the flow as suggested and got the correct errors in the front end and in Amplify:

Screenshot 2024-07-09 at 17 44 50

Screenshot 2024-07-09 at 17 37 42

Also just double checked creating a colony with a new token and it worked fine. 📈

amplify/backend/function/fetchTokenFromChain/src/index.js
```diff
@@ -104,23 +133,56 @@ exports.handler = async (event) => {
...
+
+      // Call each of the following required ERC20 functions using their function selectors
+      // If these contract calls succeed without throwing an error, it is likely an ERC20 token contract
+      const erc20Functions = [
+        '0x18160ddd', // totalSupply
+        '0x70a08231', // balanceOf(address)
+        '0xa9059cbb', // transfer(address,uint256)
+        '0x095ea7b3', // approve(address,uint256)
+        '0xdd62ed3e', // allowance(address,address)
+        '0x23b872dd', // transferFrom(address,address,uint256)
```
rdig, line 153: I assume these came from @area :P

rdig (approved):
I like this attempt at tightening the token check, but as you said, it will catch most cases, but not all.

This is why I always complain about tokens, as their standard is lacking making them a pain to work with.

I do like the approach of checking for basically all addresses that we know of (via the db) to at least ensure we've done most we can to check them.

Nicely done Sam!

PR #2684, opened 2024-07-09
Fix: CreatedIn Field should be read only for Move Funds motions

### LINKED ISSUE #2683: User should not be able to set createdIn when making a Move Funds motion
## Steps to reproduce

1. Install the reputation weighted extension (then refresh the page).
2. Create a new action, and select Transfer Funds as the action type, and Reputation as the decision method.
3. Change the created in field. 

## Expected behaviour

* You should not be able to change the created in field when making this type of action with this type of decision method.
* Right now, only General (the root domain) should be allowed to be used, since no other domain will work. The created in domain must be a parent domain of both the To and the From domains.

## Actual behaviour

* You can change it, and select to create the action in the same domain as the `from` domain. When the action is created however, it reverts back to General. Very confusing!

davecreaser:
## Description

- Set the createdIn field to read only, and fix it as root, when making a motion of type Move Funds.
- Right now, only General (the root domain) should be allowed to be used when making this type of motion, since no other domain will work. The created in domain must be a parent domain of both the To and the From domains.

## Testing

- Install the reputation weighted extension (then refresh the page).
- Create a new action, and select Transfer Funds as the action type, and Reputation as the decision method.
- Try to change the created in field. You should not be able to, and it should be `General`.

Screenshot 2024-07-09 at 17 16 03

## Diffs

**Changes** 🏗

* Set the createdIn field to read only, and fix it as root, when making a motion of type Move Funds.

Resolves #2683

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/forms/TransferFundsForm/TransferFundsForm.tsx
+++ b/src/components/v5/common/ActionSidebar/partials/forms/TransferFundsForm/TransferFundsForm.tsx
@@ -25,7 +25,7 @@ const TransferFundsForm: FC<ActionFormBaseProps> = ({ getFormOptions }) => {
   const selectedTeam = watch('from');
 
   const hasNoDecisionMethods = useHasNoDecisionMethods();
-  const createdInFilterFn = useFilterCreatedInField('from');
+  const createdInFilterFn = useFilterCreatedInField('from', true);
 
   return (
     <>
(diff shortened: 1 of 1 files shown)


mmioana (approved):
Awesome work @davecreaser 

When creating a Transfer funds action and selection the Reputation decision method, the Created In field shows and is not editable
[screen-capture.webm](https://github.com/JoinColony/colonyCDapp/assets/32430018/22a3a95d-41de-4503-9700-422ddb00ad3a)

iamsamgibbs (approved):
Nice work with this Dave, can confirm `Transfer Funds` motion can only be created in general.

Screenshot 2024-07-10 at 13 44 40

bassgeta (approved):
image
Nice one, can't click on the little bugger anymore :ship:

melyndav (approved):
Looking good, johhny dapp @davecreaser! Nice update and works as expected. 🚶

Issue #2658, opened 2024-07-05
[Multi-sig] feat: quality of life improvements on the multi-sig branch


bassgeta:
### Story

This is a catch-all issue for all minor improvements on the multi-sig branch.

### Description

List of improvements needed:
1. After multisig is installed, invalidate the query for decision methods, so you can immediately create an action - @mmioana 
2. Install multisig and immediately give it permissions so we don't need the `Enable permissions banner` @mmioana 
3. After users are given multi-sig permissions (including when the extension is installed and as a result of a manage permissions action / motion finalization), invalidate the permission holder query so we can immediately create multi-sig actions with them as potential signers @mmioana 
4. Unify types and how our `getMotionParams` helpers in the sagas work like
5. Go over and remove any useless TODO comments
6. Add rejected at date to the `MultiSig` widget @mmioana 
7. The MultiSig SVG isn't according to the designs @mmioana 
8. Rework `VoteButton` and `RemoveVoteButton` to not use `ActionForm` but use `ActionButton` instead (loader is an issue) @bassgeta 
9. Instead of `isMultiSig?: boolean;` , maybe we can define a literal type here so like: `motionType: 'multiSig' | 'reputation' | 'someOtherMotionType'`
10. Check the Permissions page is showing users correctly with all filter options @iamsamgibbs - I investigated this, there are issues, but there are also issues on master so have opened a new issue to resolve it there first (#2784, #2785) Will need fixing up here after these are resolved
11. Refactor `createEditDomainReputationMotion` saga to use updated `getChildIndex` and `getPermissionsProof` functions & fix "Latest activity" bug for domain action related entries @rumzledz 
12. When the extension is enabled, users should be assigned multi-sig roles which reflect their current permission roles. For example, if Amy has custom permissions in Serenity to have the funding and architecture roles, after multi-sig has installed, she should have funding and architecture multi-sig roles in Serenity. (Update this in the `handleMultiSigInstall` function.) - this can be a bigboi, opened its own issue for it [here](https://github.com/JoinColony/colonyCDapp/issues/2765) 
13. Once #2565 is merged into main, rebase feat/multi-sig and check it is still working as expected, but that it also works in a sensible way with the new multi-sig authority. (There are a couple of other PRs kicking around at the moment which touch this form too)
14. ~~Rejecting as the motion owner should automatically cancel the motion (I thought PagePro were handling this as part of the UI issue, so double check this with Kerry first)~~ @iamsamgibbs 
[...]

PR #2686, opened 2024-07-10
Fix: Refresh multi-sig extension after enabled permissions

mmioana:
Refresh multi-sig extension after enabled permissions 
Fix error toast translation keys

## Description

- After multisig is installed, invalidate the query for decision methods, so you can immediately create an action

## Testing

After installing multisig extension check action using Multi-Sig decision method can immediately be created without page refresh. 

* Step 1. Go to Colony > Admin > Extensions > Multi-sig Permissions
* Step 2. Install extension
Screenshot 2024-07-10 at 10 52 39
* Step 3. Enable permissions
Screenshot 2024-07-10 at 10 52 49
* Step 4. Open "Create new action" sidebar
* Step 5. Check Multi-Sig is among decision methods
Screenshot 2024-07-10 at 10 53 17

Contributes to #2658


src/components/frame/Extensions/pages/ExtensionDetailsPage/hooks.ts
```diff
@@ -0,0 +1,59 @@
...
+
+import { useColonyContext } from '~context/ColonyContext/ColonyContext.ts';
+import useExtensionData, { ExtensionMethods } from '~hooks/useExtensionData.ts';
+
+import {
+  waitForColonyPermissions,
+  waitForDbAfterExtensionAction,
+} from './utils.tsx';
+
+export const useGetExtensionsViews = async () => {
```
rdig, line 11: This will need to be refactored at some point to pull data from our DB rather than from that subgraph endpoint

rdig (approved):
Works as expected. Nice work!

Screenshot from 2024-07-12 00-58-23
Screenshot from 2024-07-12 00-58-41
Screenshot from 2024-07-12 00-58-52
Screenshot from 2024-07-12 00-59-08
Screenshot from 2024-07-12 00-59-18

iamsamgibbs (approved):
Nice work on this! Tested and working as expected - will make the remaining multi-sig issues much nicer to test!

https://github.com/user-attachments/assets/c6f12094-aaa2-48f4-af8f-0ca5084e31cf

There is the issue that the contributors do not get re-fetched after the extension is installed so it can't fetch threshold data correctly - but this is already a point on the QoL issue.

https://github.com/user-attachments/assets/35d883b6-ab8a-4129-98b7-aad769177669

PR #2688, opened 2024-07-10
Feat: "Validate" tokens on token select and in manage tokens sagas

### LINKED ISSUE #2687: [Manage Tokens] Tokens should be validated as best as possible
Currently the user can enter a properly formatted address into the form and submit the saga which will fail.

This issue should use the `getTokenFromEverywhere` query to validate the token value in the UI.

The Manage Token action saga and motion saga should also validate the token as an extra precaution.

iamsamgibbs:
## Description

Currently token addresses are not validated when added them to the manage tokens form, meaning any properly formatted address (including colony addresses and user addresses) are accepted.

This PR adds extra validation to the form to prevent these values being added using recent changes to `fetchTokenFromChain`. It also adds the same checks to the manage tokens sagas as an extra precaution.

## Testing

* Enable the reputation voting extension
* Create a 'Manage tokens' action.
* Add a new token to the table. Try using an existing colony address, a user address, a properly formatted address (`0x0000000000000000000000000000000000000001`) these should all be rejected. You can also try a valid token from another colony - if doing this in Planex you will either have to switch to Wayne and add CREDS there, or first remove GGG and re-add it to test this.
* Create the motion and complete the flow to check the expected changes are applied.

## Diffs

**New stuff** ✨

* Token selector component checks if token is "valid"
* Manage tokens sagas "validate" tokens

Resolves #2687


iamsamgibbs:
🎉 @rdig Filtering out the tokens is already being worked on here: #2692 :)

rdig:
Oh. Cool. Ignore me then :)

davecreaser (approved):
Nice work @iamsamgibbs, code looks great and everything is working well.

Colony address:
Screenshot 2024-07-15 at 09 58 25

User address:
Screenshot 2024-07-15 at 09 58 38

Correctly formatted but non existent address:
Screenshot 2024-07-15 at 09 58 57

Adding Creds to Wayne:
Screenshot 2024-07-15 at 10 02 31
Screenshot 2024-07-15 at 10 03 17

PR #2689, opened 2024-07-10
Fix/hackathon bank details form

Nortsova:
Please do **not review**, this is a hackathon fixes. Did basic fixes to make bank details form work for US, but still need UI work and refactoring work on that component


src/components/frame/v5/pages/UserCryptoToFiatPage/partials/BankDetailsModal/index.tsx
mmioana, line 1: Would it make sense to extract an onSubmit handler for the `BankDetailsForm` component?

mmioana (commented):
Amazing work on this one @Nortsova!
Code-wise looks good to me, only had a small refactoring comment if you feel it makes sense at this moment ✨

PR #2691, opened 2024-07-11
Add hardhat:trace npm command

kronosapiens:
## Description

Add support for transaction tracing through `hardhat-tracer`. Transactions from the containerized network running at `https://localhost:8545` can be traced as follows:

```
npm run hardhat:cmd -- trace --rpc http://localhost:8545  --hash 0x80917b158addd7b750495792a0ee7632e3fe7d29adc031c5d0ce14ffc3368ab9
```

In addition, calldata can be traced as follows:

```
npm run hardhat:cmd -- tracecall --rpc http://localhost:8545  --to 0x78876Ba649623A5cC3a370509fe4AD5aCa26dfB4 --data 0x04f3bcec
```

Note the `--` used to preface the command arguments.

Requires network hash `31c11b8fa3bb0521c51304ecd0b4814af40293ee` or later.

Companion to https://github.com/JoinColony/colonyNetwork/pull/1274

## Testing

* Step 1. Spin up a dockerized network instance
* Step 2. Make a few transactions
* Step 3. Trace one by calling `npm run hardhat:cmd -- trace`


area:
Can we add an equivalent for `tracecall` as well? If a transaction generated on the frontend is sufficiently wrong that `estimateGas` fails, it won't be broadcast as a transaction (and so won't have a hash), but a trace would still be extremely useful in that scenario.

area:
This is fun to see! Was hoping to show something more complex like a motion finalizing that I'd set up through the front-end, but unfortunately the network commit is based on `develop`, which has colony v16 which is not yet supported.

area:
Ran in to a possible issue:
```
 npx hardhat trace --hash 0x881b5d7776caa80bb1ded29ba981a93466297ef0de39fb36aa37bc7217d75ebe --rpc http://127.0.0.1:8545
Nothing to compile
Activating mainnet fork at block 1759
thread 'tokio-runtime-worker' panicked at /build/crates/edr_provider/src/data.rs:2464:18:
current time must be after fork block: SystemTimeError(11403967.830346821s)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
I _think_ this is because the create-data script used in the frontend forwards time, so we're going to need to figure something out here, otherwise practically this won't be able to be used :(

area:
In the short term, [a hack along these lines](https://stackoverflow.com/questions/29556879/is-it-possible-change-date-in-docker-container) might work. Being in a docker being a real bonus here (despite needing the hack) - I would not want to change the actual system time to solve this. It is going to be a problem for us in the network repository when we want to use this feature though (which I really do!)

In the medium term, I've reported this [upstream](https://github.com/zemse/hardhat-tracer/issues/74) and will hopefully be fixed.

kronosapiens:
Would it be possible to initialize the node with an older timestamp, so even after accounting for all the `forward-time` operations we still end up behind the system clock?

area:
Good thinking! The [initialDate](https://hardhat.org/hardhat-network/docs/reference#initialdate) config option should get us out of this pickle.

area:
Yep, that solves the issue nicely. Had to hack around in various places to support the new colony versions, so I don't think this can be merged yet, but both commands work nicely even after the `create-data` script is used.

screenshot-2024-07-17_11-39-07

kronosapiens:
Hey @jakubcolony there are some build issues on the network side; for some reason the tracing plugin is causing connection issues with our reputation tests. We're investigating it and will hopefully sort it out soon.

kronosapiens:
@jakubcolony this is ready for review now

jakubcolony (approved):
Looks good to me, I've tried a few actions in the UI and got traces that look right (as far as I can tell):

- Mint tokens:
image

- Payment (had to cut the screenshot due to its length):
image

- Create domain:
image

Nice one @kronosapiens 👍

davecreaser (approved):
Nice one @kronosapiens - everything seems to be working, and though I admit I don't know exactly what is expected in the output, everything I see does make sense!

After installing an extension:
Screenshot 2024-08-20 at 12 52 36

After managing permissions:
Screenshot 2024-08-20 at 12 51 24

After managing reputation:
Screenshot 2024-08-20 at 13 06 19

PR #2694, opened 2024-07-11
Feat/2658 remove multi sig permissions banner

mmioana:
Enable multi-sig extension without showing the permissions banner

## Description

- Install multisig and immediately give it permissions so we don't need the Enable permissions banner
- As a quick fix, I added the logic to trigger the extension enabling in the PermissionsNeededBanner component

## Testing

After installing multisig extension check if permission banner still shows up and extension is enabled without page refresh.

* Step 1. Go to Colony > Admin > Extensions > Multi-sig Permissions
* Step 2. Install extension
Screenshot 2024-07-10 at 10 52 39
* Step 3. This banner should **NOT show up**
Screenshot 2024-07-10 at 10 52 49
* Step 4. Check extension is enabled by creating a new action
* Step 5. Check Multi-Sig is among decision methods

Contributes to [#2658](https://github.com/JoinColony/colonyCDapp/issues/2658)

### DIFF
--- a/src/components/frame/Extensions/pages/ExtensionDetailsPage/PermissionsNeededBanner.tsx
+++ b/src/components/frame/Extensions/pages/ExtensionDetailsPage/PermissionsNeededBanner.tsx
@@ -37,13 +37,13 @@ interface Props {
 }
 
 const PermissionsNeededBanner = ({ extensionData }: Props) => {
+  const shouldDisplay =
+    extensionData.extensionId !== Extension.MultisigPermissions;
   const { colony } = useColonyContext();
   const { user } = useAppContext();
   const { checkExtensionEnabled } = useCheckExtensionEnabled(
     extensionData.extensionId ?? '',
   );
-
-  const [shouldDisplay, setShouldDisplay] = useState(false);
   const [isPermissionEnabled, setIsPermissionEnabled] = useState(false);
   const userHasRoles = addressHasRoles({
     requiredRolesDomains: [Id.RootDomain],
@@ -81,9 +81,8 @@ const PermissionsNeededBanner = ({ extensionData }: Props) => {
         // Enable Extension.MultisigPermissions by default
         enableAndCheckStatus();
       }
-    } else {
-      setShouldDisplay(true);
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [extensionData.extensionId, colony.colonyAddress]);
 
   const getBanner = () => {
--- a/src/components/frame/Extensions/pages/ExtensionDetailsPage/PermissionsNeededBanner.tsx
+++ b/src/components/frame/Extensions/pages/ExtensionDetailsPage/PermissionsNeededBanner.tsx
@@ -1,6 +1,6 @@
-import { ColonyRole, Id } from '@colony/colony-js';
+import { ColonyRole, Extension, Id } from '@colony/colony-js';
 import { CheckCircle, WarningCircle } from '@phosphor-icons/react';
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import { FormattedMessage, defineMessages } from 'react-intl';
 
 import { useAppContext } from '~context/AppContext/AppContext.ts';
@@ -43,6 +43,7 @@ const PermissionsNeededBanner = ({ extensionData }: Props) => {
     extensionData.extensionId ?? '',
   );
 
+  const [shouldDisplay, setShouldDisplay] = useState(false);
   const [isPermissionEnabled, setIsPermissionEnabled] = useState(false);
   const userHasRoles = addressHasRoles({
     requiredRolesDomains: [Id.RootDomain],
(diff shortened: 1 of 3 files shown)


iamsamgibbs (changes requested):
Nice attempt at resolving this issue, but I think there are some issues with this approach.

If you test installing this with a metamask wallet and reject the transaction to enable to extension permissions, then you get into a state where the banner will never show despite it not having the correct permissions.

https://github.com/user-attachments/assets/d12b554a-5d6a-4061-b047-35fc73acb135

The same will happen if you close or refresh the page before the third transaction.

https://github.com/user-attachments/assets/79b2c320-13e0-4c34-ae15-4bb5a92f9bec

I think this solution would be good enough for now if the banner still showed when the permissions haven't been assigned.

I think a better long term solution would be to use a multicall to install and enable the extension in the same transaction. (Though admittedly I've not looked at the extension installation flow to know if this is actually sensible.)

iamsamgibbs:
> Thank you @iamsamgibbs for your review! I will look into the issue, however do we use anywhere multicall for the transactions?

No problem! I think it is worth sanity checking with the other devs that multicall is the right approach here. There is a multicall example in the `colonyCreate` saga, it even looks like the commented out sections are to enable the staked expenditure extension after it has installed. At a quick glance it looks like there are 4 separate transactions each using multicall to call the same contract function multiple times. src/redux/sagas/colony/colonyCreate.ts

But multicall can also be used to call different contract functions in the same transaction: src/redux/sagas/utils/setExpenditureStateHelpers.ts

I would think for this we could call the install method, then the enable method in the same multicall. This is assuming we don't need any data generated by the install method in order to call the enable method (I'm not sure if the enable method needs the extension hash or anything)

mmioana:
@iamsamgibbs thank you for your throughout explanation!
Unfortunately for the extension enabling we need the extension address within the colony it was installed into
```
// Extension address after it successfully installed and present in the DB
  const extensionAddress = yield call(
    waitForExtensionDataFromDB,
    colonyAddress,
    extensionHash,
  );
```

In this case it seems the extension enabling within the same transaction is not something we can do, so I would propose to stick then to the first solution (aka the one done in the first commit on this PR https://github.com/JoinColony/colonyCDapp/pull/2694/commits/2c5b79771a62a5a31cd1d42b326db6070e1ac614 )

rdig:
Should this be put back in Draft mode until you figure out a solution ?

iamsamgibbs:
> @iamsamgibbs thank you for your throughout explanation!
> 
> Unfortunately for the extension enabling we need the extension address within the colony it was installed into
> 
> ```
> 
> // Extension address after it successfully installed and present in the DB
> 
>   const extensionAddress = yield call(
> 
>     waitForExtensionDataFromDB,
> 
>     colonyAddress,
> 
>     extensionHash,
> 
>   );
> 
> ```
> 
> 
> 
> In this case it seems the extension enabling within the same transaction is not something we can do, so I would propose to stick then to the first solution (aka the one done in the first commit on this PR https://github.com/JoinColony/colonyCDapp/pull/2694/commits/2c5b79771a62a5a31cd1d42b326db6070e1ac614 )

Ahh in that case let's stick with this, is it ready for more testing?

mmioana:
I will move it back to draft until I revert to the previous commit

mmioana:
@iamsamgibbs if you want to re-test this, please give it a go 🙏

src/components/frame/Extensions/pages/ExtensionDetailsPage/PermissionsNeededBanner.tsx
```diff
@@ -43,6 +43,7 @@ const PermissionsNeededBanner = ({ extensionData }: Props) => {
     extensionData.extensionId ?? '',
   );
 
+  const [shouldDisplay, setShouldDisplay] = useState(false);
```
bassgeta, line 46: I think this can just be a variable like
`const shouldDisplay = extensionData.extensionId === Extension.MultisigPermissions` and then we don't need to set state

bassgeta (changes requested):
This is actually such a simple solution, love it!
It works, it gets autoenabled and the flow is simple, maybe we can just do without 1 `useState` variable and after that it's a :ship: from me!

bassgeta (approved):
Nice and simple solution, still works like a charm, let's merge this quality of lime improvement :sunglasses:

davecreaser:
Hey @mmioana, nice simple solution, but I'm running into an issue when the threshold is attempted to be calculated.

I enable the extension and the permissions banner doesn't show up, and I can immediately make an action. However when that action is completed the action sidebar shows an error that the threshold can't be calculated. This error goes away when I refresh the page.

https://github.com/user-attachments/assets/dde3ab37-26ee-45d0-b627-357b9f369743

Screenshot 2024-07-19 at 17 36 24

davecreaser (changes requested):
I like the approach, but see my comment for the errors I'm encountering when testing

iamsamgibbs:
@davecreaser The threshold is out of scope for this issue, it is covered by point 3 in the QoL issue #2658 - currently the threshold is relies on the contributors cache which needs to be refetched when the extension is installed (or when a member is assigned multi-sig permissions). **However**, @bassgeta is changing how the eligible signees are fetched as part of #2748 which would either resolve this issue or change how it needs to be resolved.

davecreaser:
@iamsamgibbs Thanks for the clarification! In that case, I'm happy with this, nice work @mmioana 📈

Nortsova (approved):
Thanks @mmioana for this PR 🤩 

Can confirm that there is no banner on MultiSig enabling action. 

And I can create motion with multiSig decision method:
image

iamsamgibbs (approved):
This is looking great @mmioana ! I've tested the flow as intended and it all works great.

I'm not sure this edge case is handled the most elegantly (although I'm not sure it is worth the amount of effort it would take to resolve it but thought it worth raising anyway!)

When testing it with metamask, if you approve the first transaction, then reject the second you get into this weird state where the extension is installed, but not enabled but the 'Enable permissions' banner never shows.

If you go back to the dashboard, refresh then navigate back to the extension page, then it will instantly trigger the 'Extension enable' transaction when the page loads - I think at this point it would be better to show the 'Enable permissions' banner rather than instantly triggering a transaction - as a user it is unclear why a transaction has started.

Also - if you navigate directly to the page (rather than coming from another page), then it seems the network client isn't available in time for things  to function as expected:

Screenshot 2024-07-22 at 16 03 08

After accepting the transaction, the extension is enabled correctly.

Here is a video of me testing this:

https://github.com/user-attachments/assets/cfcd44a4-47b7-4dec-bcba-a19c552da1ae

It probably isn't worth the effort of fixing it right now, it is probably very unlikely a user would accept the first transaction and reject the second (although it could happen if they lacked funds for the second transaction). And it is at least possible to finish enabling the extension if they get into this situation, so it doesn't completely break the app - even if the steps for this aren't immediately clear.

mmioana:
@iamsamgibbs thank you so much for your throughout testing 🚀 
I will create a new issue for this to investigate further this edge case

PR #2719, opened 2024-07-17
Feat/15967008 staged flow

CzarekDryl:
## Description

- Added flow for staged payment using permission and reputation method
- Issue - https://github.com/JoinColony/colonyCDapp/issues/2669

## Testing

* Install Staged payment extension and give this extension owner permission
* Create staged payment action and go to payment step (every step before payment should work the same as in payment builder action)


arrenv (changes requested):
@CzarekDryl Really great work tackling this feature. I appreciate all the work you have put in.

Adding in some feedback on things I have noticed:

1. I have noticed in this PR that the metadata does not update the recipient until the milestone is updated.
staged-name

2. Deprecating the Staged expenditure broke the app. I was not able to return to the page. I had to add translations to resolve it.
image

3. I see the claim delay column while loading a staged payment. It also shows up briefly with the full claim delay breaking the UI.
image
image

4. If you enter a letter in the amount field, the width and alignment breaks.
image

5. In a few cases, but not every time, I tried to create a Staged payment, and the app crashed with this error in console "TypeError: Cannot read properties of undefined (reading 'payouts')".
image

6. Nothing showed up in the Payment step. It should show the Make next payment button.
image

7. I received a permission error in console when trying to make all payments.
image

8. There is no option to make payments as the Payment creator in the payment step.
image

I could not test past this step.

CzarekDryl:
@arrenv 
1. Fixed
2. I guess this happened because the texts for the modal are missing
3. Fixed 
4. Fixed
5. I can't reproduce this, could you please give more information about this one?
6. Fixed
7. To make it work you need to give owner permission for this extension (I'm not sure why is that)
8. This will be handled in other PR

arrenv:
> @arrenv
> 5. I can't reproduce this, could you please give more information about this one?

Regarding 5, this happened quite a few times, I was just using the feature as normal. It is possible that the completed state is trying to load before the payouts object/array is available?

CzarekDryl:
@arrenv If it was problem with loading too early then now it should work fine, because I changed the action loading logic a bit

jakubcolony (changes requested):
Nice work on this PR @CzarekDryl, looks pretty functional and I was able to test releasing stages using both permissions and motions.

I think this can be now rebased on `feat/advanced-payments-ui` as #2513 has been merged.

While testing, I noticed the usual extension missing permissions banner is missing from the page. This is apparently a regression introduced in `feat/advanced-payments-ui` which I've fixed, so please take care when rebasing not to override it.
image

Also noticed some small UI issues:

- Long milestone name pushes the amount out of the boundary in the release modal:
image

- The colour of the "All remaining milestones" doesn't match Figma:
image
image

CzarekDryl:
@jakubcolony I think that `The colour of the "All remaining milestones" doesn't match Figma:` is a mistake in design because in all other cases it's blue. @arrenv Could you please confirm?

arrenv:
> @jakubcolony I think that `The colour of the "All remaining milestones" doesn't match Figma:` is a mistake in design because in all other cases it's blue. @arrenv Could you please confirm?

Just noting that this was confirmed to be missing in the design, via direct message.

arrenv (changes requested):
@CzarekDryl Thank you for your continued hard work!!

1. I'm still getting the extension permission banner that Jakub raised.

   image

2. I am also getting orange tabs as of the above screenshot.

3. I uninstalled the Staged payment extension but the action was still able to be interacted with, however, when trying to make the payment I failed but with no error in the UI. We should show validation in the UI as of the below design. 

   Additionally, if you try to complete a Staged payment that was created with a previously installed, but now uninstalled extension, you get a bunch of errors. What is happening on the contracts (thanks to Alex for clarifying), is that the Staged payment is being converted to a normal expenditure. I think the best user experience here is to show an error message remove transaction buttons and encourage cancelling the action.
   
   I have included a design here (let me know if this should be tackled in another PR): https://www.figma.com/design/5V8pr7iMwXsT9L3VAZsmUt/Colony-v5?node-id=26185-216290&t=SrEjADE0CuVbASHB-4

Requirements:
- Warning message should appear if the extension was installed, i.e. this is no longer a Staged payment.
- The "Cancel payment" button should open the Cancel action modal.
- The ability to release payments should be removed in the table and the flow panel.


   image

   image



4. The extension should be installed with only the "Arbitration" permission. This should also be visible when hovering over the custom permission pill in the extensions page.
image
image


5. There is a translation error for Staged payment action in the Transactions tab.
image

6. I should be able to use the "Staking" decision method to create a Staged payment.
image


7. Payment creator decision method should be an option for the "Make milestone payment" step.
image

8. This button copy should be "Make next payment"
image

9. When using the Reputation decision method when funding, loaders and timers appeared all over the place. Can we remove them from the Payment step, then can we use skeleton loaders instead where loaders are required.
image
image

CzarekDryl:
@arrenv It's ready to review

arrenv (changes requested):
Hey @CzarekDryl, very nice, thank you for the updates.

Some minor things that came up.

1. When validation was triggered for the milestone title, the width of the table was jumping.

[table-alignment.webm](https://github.com/user-attachments/assets/a0170939-5bce-4f5a-97a1-bf9890542efd)


Validation in Advanced payments table highlights the whole table and does not jump:
    image
    
Additionally, I also experienced this overflow issue in the Advanced payment table in this branch, which I don't experience on `master`.
    image
 
 
3. Regarding the last point from 1. this overflow issue happens in this amount field as well on Staged payments.
image

4. I used the Staking decision method, however, I did not get a modal to stake. It is also showing up as Permissions:
image

5. This error message only needs to show if the extension has been uninstalled. I currently have the extension deprecated, which should allow Staged payments to work as normal.
image
image

6. If the Staged payment is in the last step and the extension was uninstalled, it needs to show the other state we discussed, shown here - https://www.figma.com/design/5V8pr7iMwXsT9L3VAZsmUt/Colony-v5?node-id=26217-217823&t=A75UljR6BiwziTRn-4

    Currently it fails - 
    image

7. I created a Staged payment, I completely uninstalled the extension, then installed and enabled a new version, and the UI allowed me to try and complete the payment, but it fails. This needs to be handled like point 5 above.
image

8. I was in the funding step, and completely uninstalled the Staged payment extension, but it appeared like I could still interact with it. This needs to be handled with the cancel case that we went through, i.e. this one, which is relevant for every staged but the last staged if the extension has been uninstalled - https://www.figma.com/design/5V8pr7iMwXsT9L3VAZsmUt/Colony-v5?node-id=26217-217437&t=A75UljR6BiwziTRn-4

CzarekDryl:
@arrenv Table is jumping because of scrollbar appears

arrenv (changes requested):
@CzarekDryl Very nice!! Thank you for the updates. I have only been able to able to spot one issue at the moment, the rest look to be resolved.

1. When creating a new Staged payment with Staking, I get the loader triggering, after it has already loaded. The only way to fix it is to reload the page. I tried also with Permissions, and the loader showed as well, but resolved itself after a few seconds.

[loading-content.webm](https://github.com/user-attachments/assets/92d0fc76-93ad-4eb3-ab2b-3ebb3c2949e1)



### Testing other resolved fixes
Length and overflow issues all look to be resolved.

image

Staking now works correctly.

image

image

Staged Payment can be completed even with the extension deprecated.

When the extension has been uninstalled, for every step but the last one, cancel prompt shows correctly.

image

Message shows on final step if the extension is uninstalled.

image


We discussed the uninstalling and reinstalling issue would be handled in another PR.

jakubcolony (changes requested):
Thanks for addressing my feedback @CzarekDryl,

Long milestones/large amounts look okay now:
image

I was not able to release the payment without manually setting roles because the extension's permissions were missing, it looks like the required permissions were changed in https://github.com/JoinColony/colonyCDapp/commit/49662733fad05127fb93a47c6ec58b9891e55790#diff-dfe7158ed1985dc238d3f706656480752c1b0fbad533905398346e3e9fd4b590R542 - can you revert that change in this PR?

When releasing a payment, selecting decision method of Payment creator shows an empty note:
image

jakubcolony:
@CzarekDryl I have updated the base branch to ``feat/advanced-payments-ui-next`

CzarekDryl:
@jakubcolony Permission changes reverted and added description

arrenv (changes requested):
@CzarekDryl Very nice work. Thank you for working on all the updates.


1. I noticed if you use motions in the final step, you have an option to Make next payment even though the payment has already been finalized. It did not go away until you claimed.

image

2. I am still getting issues with the loading of the completed staged payment when using Staking decision method. I see an advanced payment with a broken claim delay column, then just a continuous loading.

[staking-decision.webm](https://github.com/user-attachments/assets/d11773da-e0d0-4fd4-aa55-a9308fdff2e9)

CzarekDryl:
@arrenv Please check button once again

CzarekDryl:
@arrenv Creating action using staking method should work fine now

jakubcolony (changes requested):
Nice work @CzarekDryl, looks like the full flow is nearly there.

I spotted some issues around releasing milestones using motions:
- Despite there still being unreleased milestones and the previous motion failing, the "Release next payment" button isn't displaying (which is inconsistent with how funding works):
image

- After selecting a motion that failed, the box says it's passed:
image

- Once I released all milestones using permissions, I can no longer see the previous motions:
image

- There is no loading state on the "Release next payment" button which means I can create multiple actions/motions to release the same milestone:

https://github.com/user-attachments/assets/d4f2f3e6-d2a5-4da6-a228-9bd70d50c77f

CzarekDryl:
@jakubcolony `After selecting a motion that failed, the box says it's passed:` this one was fixed in other [issue](https://github.com/JoinColony/colonyCDapp/issues/2974)

jakubcolony (changes requested):
@CzarekDryl Thanks for another round of changes, looks like it's nearly there, however I spotted a bug that prevents me from progressing a regular (non-staged) advanced payment:

image

Can confirm previous issues have been resolved:

- Missing "Release next payment button" ✅ 
image

- Previous motions being inaccessible after releasing using permissions ✅ 
image

CzarekDryl:
@jakubcolony Everything should be fine now with extension card
image
image

jakubcolony (approved):
Thanks for the fix @CzarekDryl, this looks good now 👍 

image

arrenv (approved):
@CzarekDryl Tested all scenarios I could think of and all looks great, nice work!!

PR #2721, opened 2024-07-17
Feat: Edit streams using motions

### LINKED ISSUE #2214: Streaming Payments API: Editing streams using motions
Parent issue: https://github.com/JoinColony/colonyCDapp/issues/2208

- New saga needs to be implemented
- Relevant block-ingestor handlers need to be implemented
- It should be possible to annotate the motion
- The proposed changes should be stored on the motion object (similar to edited expenditure slots)
- There should be a bidirectional link between motion and streaming payment objects (as with motions and expenditures)
- The motion will need to use multicall

iamsamgibbs:
[block-ingestor PR](https://github.com/JoinColony/block-ingestor/pull/256)

## Description

This PR introduces a new `editStreamingPaymentMotion` saga, and a block-ingestor handler. Pending streaming payment changes are stored on the motion object, and so are the pending streaming payment metadata changes.

This PR also adds changelogs to the streaming payment model, and the streaming payment metadata model to track edits whether they are made via permissions, or via a reputation motion.

A decent amount of the work is in the block-ingestor PR so please review that thoroughly too.

### Test edit using permissions

Step 1 - Install the streaming payment extension and voting reputation extension.
Step 2 - Create a new streaming payment using the temporary UI. Set the amount to `10`, select `Start one week from now` as the start time and `End three weeks from now` as the end time.

Screenshot 2024-07-21 at 15 21 51

Step 3 - Wait a few seconds for the payment to be created in the database, then click `Use latest streaming payment native ID`. Verify the values are as expected. Can also run this query to confirm the metadata changelog is empty: 
```
query MyQuery {
  getStreamingPayment(id: “{COLONY_ADDRESS}_{NATIVE_ID}”) {
    amount
    startTime
    endTime
    interval
    metadata {
      endCondition
      changelog {
        newEndCondition
        oldEndCondition
        transactionHash
      }
    }
  }
}
```

Screenshot 2024-07-21 at 15 22 14

Step 4 - Change the amount to `20`, select `Limit reached` as the end condition, then set the limit amount to `200`. Change the start time to `Start two weeks from now`, and the interval to `Daily`.
Step 5 - Ensure the relevant toggles are selected, then click `Edit`.

Screenshot 2024-07-21 at 15 22 42

Step 6 - Click refetch and rerun the query to ensure the values have updated as expected, and that the metadata changelog is now populated. Run this query to check the created action has the changelog.

```
query MyQuery {
  listColonyActions(filter: {type: {eq: EDIT_STREAMING_PAYMENT}}) {
    items {
      type
      streamingPayment {
        nativeId
      }
      streamingPaymentChanges {
        newValues {
          amount
          endTime
          interval
          startTime
        }
        oldValues {
          amount
          endTime
          interval
          startTime
        }
      }
    }
  }
}
```

Screenshot 2024-07-23 at 15 24 31

Screenshot 2024-07-23 at 15 24 35

### Test edit via motion

Step 7 - Refresh the page. Create a new streaming payment using the temporary UI. Set the amount to `10`, select `Start one week from now` as the start time and `End three weeks from now` as the end time. Step 8 - Wait a few seconds for the payment to be created in the database, then click `Use latest streaming payment native ID`.
Step 8 - Change the amount to `20`, select `Limit reached` as the end condition, then set the limit amount to `200`. Change the start time to `Start two weeks from now`, and the interval to `Daily`.
Step 9 - Ensure the relevant toggles are selected, then click `Edit via motion`. The motion link will be logged to the console, open this. Run this query to check the pending changes. Can also run the previous query to confirm no changes have yet been applied.

```
query MyQuery {
  getMotionByTransactionHash(
    transactionHash: “TX_HASH”
  ) {
    items {
      action {
        type
      }
      pendingStreamingPaymentMetadata {
        endCondition
        changelog {
          newEndCondition
          oldEndCondition
          transactionHash
        }
      }
      pendingStreamingPaymentChanges {
        amount
        endTime
        interval
        startTime
      }
      streamingPaymentId
      pendingStreamingPaymentMetadataId
      streamingPayment {
        id
      }
    }
  }
}
```

Screenshot 2024-07-21 at 15 24 22

Screenshot 2024-07-21 at 15 24 31

Screenshot 2024-07-21 at 15 24 42

Step 10 - Complete the motion flow and finalise the motion.
Step 11 - Rerun each of the queries and verify the changes have been applied to the streaming payment.

Screenshot 2024-07-23 at 15 28 56

Screenshot 2024-07-23 at 15 29 02

### Further testing

To properly put this PR through its paces, test with different combinations of changes.

For extra bonus testing points, check that cancelling the streaming payment creates a new cancel action is created in the database, and that no edit action is created by the cancellation.

## Diffs

**New stuff** ✨

* Edit streaming payment motion saga + block-ingestor handler

**Changes** 🏗

* Colony action model now has a streaming payment changelog
* Streaming payment metadata model now has a changelog

Resolves #2214


src/redux/sagas/utils/editStreamingPaymentMulticall.ts
```diff
@@ -0,0 +1,196 @@
...
+      pushSetEndTime(TIMESTAMP_IN_FUTURE);
+    }
+
+    if (hasStartTimeChanged) {
+      pushSetStartTime(realStartTimestamp);
+    }
+
+    if (hasEndTimeChanged) {
+      pushSetEndTime(realEndTimestamp);
+    }
```
rumzledz, line 176: Is it intentional to call `pushSetEndTime()` twice in this scope due to some race condition or something? Because if that's the case, was thinking maybe we can get away with initialising some var to undefined then finally call `pushSetEndTime` at the end to save us multiple calls, so like:

```js
const endTime = undefined;
const startTime = undefined;

if (hasEndTimeChanged) {
  endTime = TIMESTAMP_IN_FUTURE;
}

if (hasStartTimeChanged) {
  startTime = realStartTimestamp;
}

if (hasEndTimeChanged) {
  endTime = realEndTimestamp;
}

// Then finally
pushSetEndTime(endTime);
pushSetStartTime(startTime);
```

But I dunno, maybe there's something more to this than meets the eye 😄

rumzledz (commented):
Awesome work with this beast @iamsamgibbs ! I carefully followed your steps and got the expected results:

1. Set start date to one week from now and end date to 3 weeks from now:
Screenshot 2024-07-22 at 22 30 38
2. I can confirm that the changelogs are null
Screenshot 2024-07-22 at 22 32 25
3. I have set the form according to the steps and clicked the submit button
Screenshot 2024-07-22 at 22 36 20
4. I can confirm that the changelog fields are now populated
Screenshot 2024-07-22 at 22 37 09

rumzledz:
I can confirm that I can edit via motion and I can also finalise it:

Screenshot 2024-07-22 at 23 00 08
Screenshot 2024-07-22 at 23 01 47
Screenshot 2024-07-22 at 23 00 35

rumzledz:
I tried different combinations with motions and permissions and all seems to be working well @iamsamgibbs  👌

jakubcolony (changes requested):
Looks like a great piece of work again @iamsamgibbs, really nice to see you consider all scenarios including pending metadata.

Went through testing without issue 👍 

Only thing that stood out was the "placement" of the StreamingPayment's changelog which I think should live on the action model for consistency

src/redux/sagas/utils/editStreamingPaymentMulticall.ts
iamsamgibbs: It is intentional here and we want to call `setEndTime` twice as part of the same multicall.

The contract function doesn't allow for a new startTime that is after the current endTime, and also doesn't allow a new endTime that is before the current startTime. If just one or the other is true then we can just change the order in which we call `setStartTime` or `setEndTime`.

This section is to account for an edge case where both of these statements are true. To get around this, we set the endTime to the max value first, so that we can set the new startTime (which is after the current endTime) then finally set the actual endTime.

(Although now that I write this out, I can't think of a situation where it would be possible for both of these to be true at the same time...)

amplify/backend/api/colonycdapp/schema.graphql
```diff
@@ -3399,11 +3433,64 @@ type StreamingPayment @model {
...
   isWaived: Boolean
+  """
+  The motions created that relate to this streaming payment, if there are any
+  """
+  motions: [ColonyMotion]
+    @hasMany(indexName: "byStreamingPaymentId", fields: ["id"])
+  """
+  List of Streaming Payment changelog entries
+  """
+  changelog: [StreamingPaymentChangelog!]
```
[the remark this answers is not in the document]
iamsamgibbs: Ah okay, I only added it directly to the streaming payment rather than the action because #2213 states it doesn't require a corresponding action to be created - but if you're happy to have a colony action with it then I can go ahead and make that change. 👍

iamsamgibbs:
I've made the requested changes @rumzledz @jakubcolony  :)

rumzledz (approved):
It's still working as expected @iamsamgibbs ! ✨ Approving!

jakubcolony (approved):
The code looks good, thanks for making the changes 👍 

Editing still works as expected, the actions are showing correct changes:
image

Similarly with editing via permissions:
image

My only question would be whether it's easy enough to stop the metadata changelog being added when there's no change in the end condition - in the second screenshot I only edited the limit yet changelog was added with the same old and new value

But apart from that, really solid piece of work 💪

iamsamgibbs:
> My only question would be whether it's easy enough to stop the metadata changelog being added when there's no change in the end condition - in the second screenshot I only edited the limit yet changelog was added with the same old and new value

@jakubcolony I wasn't sure if it would be beneficial when fetching data from the UI to have a metadata changelog per action to avoid having to piece together the data from the previous changelog? If only the actual changes are being shown per edit then I guess it isn't necessary. I'll make this change. :)

iamsamgibbs:
Made that change, tiny new commit here and on the block-ingestor PR

jakubcolony:
> @jakubcolony I wasn't sure if it would be beneficial when fetching data from the UI to have a metadata changelog per action to avoid having to piece together the data from the previous changelog? If only the actual changes are being shown per edit then I guess it isn't necessary. I'll make this change. :)

Thank you, based on [these designs](https://www.figma.com/design/5V8pr7iMwXsT9L3VAZsmUt/Colony-v5?node-id=22702-224977&t=J6JotWBRGrQTJ9YY-4) only fields that changed need to be displayed so it's fine if we only create a changelog entry when there is an update to end condition

jakubcolony (approved):

Issue #2397, opened 2024-05-20
Flashing "Pending" button when rejecting first transaction.


arrenv:
Teamwork task: [Flashing "Pending" button when rejecting first transaction.](https://tw.auroracreation.com/#/tasks/15915835)

## Steps to reproduce

1. Ensure you are connected locally via a Metamask wallet. (Let me know if you are unsure how to do this)
2. Perform a transaction with multiple grouped transactions, e.g. staking with no activated tokens.
3. Reject the first transaction.
4. Notice the Pending button will be stuck in a flashing state.

## Expected behaviour

* The pending button should remain visible without flashing in this instance.
* It should disappear when the transaction has been cancelled.

## Actual behaviour

* The Pending button is stuck in a pending state.

![cancelled-tranasction](https://github.com/JoinColony/colonyCDapp/assets/33682027/2f132d88-8ee0-41ab-9675-f034341ca921)

PR #2723, opened 2024-07-17
fix: flashing pending button

joanna-pagepro:
## Description

Part of the fix for flashing "Pending" button when rejecting first transaction. 

## Testing

* Step 1. Ensure you are connected locally via a Metamask wallet. (Let me know if you are unsure how to do this)
* Step 2. Perform a transaction with multiple grouped transactions, e.g. staking with no activated tokens.
* Step 3. Reject the first transaction.
* Step 4. Check if the Pending button is stuck in a flashing state.

## Diffs

**New stuff** ✨

* 

**Changes** 🏗

* Checking if any transaction group failed, so that the pending state won't be applied

**Deletions** ⚰️

* 

## TODO

- [ ] Check if transactions that are after the failed one should go into the pending state. Right now, if the first one fails, the others can be in completed or pending states.
image

Contributes https://github.com/JoinColony/colonyCDapp/issues/2397

### DIFF
--- a/src/context/UserTransactionContext/UserTransactionContextProvider.tsx
+++ b/src/context/UserTransactionContext/UserTransactionContextProvider.tsx
@@ -178,6 +178,7 @@ const useGroupState = ({
       }, 3000);
     } else if (
       hasLatestGroupPendingTx &&
+      !someFailed &&
       groupState !== TransactionGroupStates.SomePending
     ) {
       if (!previousPending) {


melyndav (approved):
Hey @joanna-pagepro nice work with this fix. I tested this and wasn't able to replicate the flashing pending, looks like it's been fixed. Nice!

Export-1721223501881

iamsamgibbs (approved):
Looks good to me! Nice simple fix.

Nortsova (approved):
Thank you @joanna-pagepro ✨ Great work!

I can confirm that pending button is not flashing now: 
Jul-17-2024 23-04-38

PR #2717, opened 2024-07-17
feat: streaming payments extension page

joanna-pagepro:
## Description

Provides the ability to manage Streaming Payment Extension.

## Testing

* Step 1. Go to the extensions page
* Step 2. Click on Streaming Payments
* Step 3. Install the extension
* Step 4. Deprecate the extension
* Step 5. Uninstall the extension - please note that displaying the right messages in the modal will be handled in a separate issue.

## Diffs

**New stuff** ✨

* 

**Changes** 🏗

* The structure of the folders was updated to be the same as for the Staged Payments extension page (on a different branch)

**Deletions** ⚰️

* 

## TODO

- [ ] Update variables in Uninstall Modal to show correct messages

Resolves https://github.com/JoinColony/colonyCDapp/issues/2716

PR #2727, opened 2024-07-17
UI: Update colony extension detail page according to designs

### LINKED ISSUE #2487: [Multi-sig] feat: implement UI for customizing the thresholds
### Story

As a user configuring the multi-sig extension, I want to use the non temporary UI :D

### Description

The temporary UI is just that, temporary, we need to style it according to the designs.
Additionally (if possible), we need to move the `Save changes` button where it should be design wise.
_NOTE: [this PR](https://github.com/JoinColony/colonyCDapp/pull/2474) may contain the necessary changes for this to be possible_

### Design

[Figma link](The temporary UI is just that, temporary, we need)

image

mmioana:
Update colony extension detail page according to designs Rearranged extension tabs
Add ExtensionDetails to each tab
Add Extension carousel only to overview tab
Created context for save settings action
Styled settings page

## Description

The temporary UI is just that, temporary, we need to style it according to the designs.
Additionally (if possible), we need to move the Save changes button where it should be design wise.
NOTE: https://github.com/JoinColony/colonyCDapp/pull/2474 may contain the necessary changes for this to be possible
Figma link: https://www.figma.com/design/SaO0Zj6e7yGqLzjEsxqVhy/Permissions?node-id=21336-77636&t=a76ZwixZ88zrGV4p-1

## Testing

Go to colony Admin > Extensions > Multi-sig Permissions extension and check styling is according to designs and product specs. 
Install the extension and check the Settings tab.
Also, check other extensions now show at least the Overview tab

Resolves #2487


melyndav (changes requested):
Hey @mmioana nice start on this PR! Some feedback after testing:

1. The styling of the input dropdown should match the existing component in the design system. Particularly the font size in the dropdown and the font weight and colour of the selected menu item. 

image

 DS: https://www.figma.com/design/l1dOM5qiQYwF0ElvKDqqjg/Design-System---Colony-v3?node-id=1782-627&t=VPiZgFCMpAZtFhMv-4
 
 2. The CTA button should never be disabled but allow for the user to select and the UI to return the error validation state. We avoid disabling CTA's wherever possible in the app.
 
image

The error state should match the Figma design: https://www.figma.com/design/SaO0Zj6e7yGqLzjEsxqVhy/Permissions?node-id=21336-78426&t=90kZNys5YfZUZKuv-4


3. In the threshold input, I can just keep typing an unlimited amount until it throws an error, but then allows me to override the error and keep entering values. Recording to demonstrate: 

Export-1721320081063

The same behaviour also happens in these threshold inputs.
image


5. The submit button is enabled because of the input value from point # 4 above, but it doesn't allow me to submit (not showing an error validation). If I change the value, I then can submit. An error should be shown to the user. 

Export-1721318966420

6. These headings should be medium and not bold weights.
image


7. Can you please fix the header block height so the CTA showing does not change the content height when changing tabs. This should be global across all extensions if possible (or fixed for this extension and a sep issue raised if possible).

Export-1721319126595


8. Buttons in the deprecated state padding should match Figma and also for mobile as well. 

image


9. On mobile, there are padding issues between the page title and extension header.

image


10. On mobile, fields need to be stacked to handle long team names as per the Figma design. 

image


11. When I open a dropdown, the UI jump as a second scrollbar is triggered in my Firefox. 

Export-1721319832533

mmioana:
Hey @melyndav thank you for your prompt throughout review! 🌞  
I'm on my way to tackling these issues 🚀

mmioana:
Hey @melyndav 

Regarding point  **5. These headings should be medium and not bold weights as per Figma.**
350044821-c58f6e10-9f30-4a3c-a4b2-61551cfaf15f

I noticed in Figma one is medium weight and the other one is bold. Do we want to make them both medium or respect the Figma designs?
Screenshot 2024-07-22 at 09 03 21
Screenshot 2024-07-22 at 09 03 30


Regarding point **14. The permission notification should not show. The extension should get the right permission in installation and initialization.** - this will be solved with this PR #2694 

Regarding point **15. The re-enable modal should be styled to match the deprecated modal (with the correct padding and icon top left). This will need to happen across all extensions, so happy for this to be raised in a sep issue.** - I would propose to create a separate issue

src/components/frame/Extensions/pages/MultiSigPage/MultiSigPageSetup.tsx
```diff
@@ -234,125 +125,174 @@ const MultiSigPageSetup: FC<MultiSigPageSetupProps> = ({ extensionData }) => {
...
+              {thresholdType === MultiSigThresholdType.FIXED_THRESHOLD && (
+                <InputGroup
+                  {...register('globalThreshold', {
+                    required: true,
+                    min: 1,
+                    max: 99999,
+                    onChange: (e) => {
+                      handleThresholdValueChange(
+                        'globalThreshold',
+                        Number(e.target.value?.slice(-5)),
```
bassgeta, line 189: can we drop a comment on why we need to slice this value

src/components/frame/Extensions/pages/MultiSigPage/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+  };
+};
+
+export const getDomainThresholds = (
+  values: MultiSigSetupFormValues,
+  domainThresholdConfigs: DomainThresholdConfig[],
+  thresholdType: MultiSigThresholdType,
+) =>
+  domainThresholdConfigs.map((domain) => {
+    let threshold = 0;
```
bassgeta, line 44: I'd drop a comment saying we default to `INHERIT_FROM_COLONY` just to reduce potential head scratches in the future :D

bassgeta (changes requested):
A ton of changes, but most of them great! Good job on tackling this beast of an issue, both cosmetic and logic wise :D 
Just went over it codewise, left a few comments, nothing major, mostly some flags and requests for additional code comments.
After it's flagged as ready, I'll also rereview the UI part :) 
Overall, good job :muscle:

melyndav:
Hi @mmioana - to address your questions. 

1. Match the Figma designs, as these are SOT.

2. No problems, thanks for letting me know.

3. A separate issue is fine to have these addressed, thanks!

src/components/common/Extensions/SpecificSidePanel/SpecificSidePanel.tsx
```diff
@@ -44,11 +45,11 @@ const SpecificSidePanel: FC<SpecificSidePanelProps> = ({ extensionData }) => {
               </div>
               <div className="flex flex-col justify-start gap-y-2 md:flex-row md:flex-wrap">
                 <div className="flex flex-wrap gap-1">
-                  {statuses.map((status) => (
+                  {statuses.map((status, index) => (
                     <ExtensionStatusBadge
                       key={status}
                       mode={status}
-                      text={status}
+                      text={badgeMessages[index]}
```
rumzledz, line 52: I feel like it would be safer to just have a map of badge messages instead of using a state so like:


```ts
type IoannaStatus = 'installed' | 'enabled' | 'disabled' | 'deprecated';

const badgeMessages: Record<IoannaStatus, MessageDescriptor> = {
  // keys will be auto-completed at this point
  installed: {
    id: 'status.installed',
  },
  disabled: {
    id: 'status.disabled'
  },
  ...
}

const [statuses, setStatuses] = useState<IoannaStatus[]>([]);

// Update the statuses state, without worrying about the order:
setStatuses(state => [...state, 'disabled']);

// Then finally just
{statuses.map(status => (
  <ExtensionStatusBadge
    key={status}
    mode={status}
    text={formatMessage(badgeMessages[status])}
  />
)}
```

src/components/frame/Extensions/pages/MultiSigPage/hooks.ts
```diff
@@ -0,0 +1,161 @@
...
+      return rest;
+    });
+
+    setDomainThresholdConfigs(tempDomainThresholdConfigs);
+  }, [domains, multiSigConfig, setValue]);
+
+  useImperativeHandle(callback, () => ({
+    getValues: async () => {
+      await trigger(undefined, { shouldFocus: true });
+      if (!isValid) throw new Error('Error within form');
```
rumzledz, line 109: No need to use await since you’re not making use of the returned value. But looking this flow, it seems your intention is to validate all fields and react right away right? Maybe you want to assign a value in this case and make use of it so like:

```js
const isValid = await trigger(undefined, { shouldFocus: true }); // isValid is false if there’s at least one error in the form

// Then use it however you please, your majesty
if (!isValid) {
  // foobar
}
```

src/components/frame/Extensions/pages/MultiSigPage/hooks.ts
```diff
@@ -0,0 +1,161 @@
...
+            ? 0
+            : values.globalThreshold,
+        domainThresholds: getDomainThresholds(
+          values,
+          domainThresholdConfigs,
+          thresholdType,
+        ),
+      };
+    },
+  }));
```
rumzledz, line 126: Deps would be ideal for this hook, otherwise it will just keep on running on every render, like useEffect:

```js
useImperativeHandle(ref, () => {
  return {
    // boogie dance
  };
}, []); <<-- deps
```

src/components/frame/Extensions/pages/MultiSigPage/hooks.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  const handleGlobalThresholdTypeChange = (
+    newThresholdType: MultiSigThresholdType,
+  ) => {
+    setThresholdType(newThresholdType);
+  };
```
rumzledz, line 132: This function just seems to be doing what the state function does so might be better to just return it directly, and rename as `handleGlobalThresholdTypeChange`:

```js
return {
  setThresholdType: handleGlobalThresholdTypeChange
}
```

src/components/frame/Extensions/pages/ExtensionDetailsPage/ExtensionDetailsPage.tsx
```diff
@@ -33,6 +33,8 @@ import { getValidationSchema } from './validation.ts';
 const displayName = 'frame.Extensions.pages.ExtensionDetailsPage';
 
 const ExtensionDetailsPage: FC = () => {
+  // @TODO replace this with a proper use of context
+  const [activeTab, setActiveTab] = useState(0);
```
rumzledz, line 37: I'm wondering if this can benefit with a numeric enum so like:

```ts
enum ExtensionDetailsTab {
  Hi,
  Ioanna
}

const [activeTab, setActiveTab] = useState<ExtensionDetailsTab>(ExtensionDetailsTab.Hi);
```

From what I can remember, the library we use to handle tabs accepts numbers to control the tab index or something, hopefully numeric enums don't break that 👀

src/components/frame/Extensions/pages/ExtensionDetailsPage/utils.tsx
```diff
@@ -21,6 +21,11 @@ import { type SetStateFn } from '~types/index.ts';
 import { notNull } from '~utils/arrays/index.ts';
 import { addressHasRoles } from '~utils/checks/index.ts';
 
+export enum EXTENSION_TAB_ID {
```
rumzledz, line 24: Let's use Pascal case here:

```js
export enum ExtensionTabId {}
```

Looking at our codebase, I think we can use this case for constant variables 🤔

src/components/frame/Extensions/pages/MultiSigPage/hooks.ts
```diff
@@ -0,0 +1,161 @@
...
+
+      setValue(rest.name, threshold);
+
+      return rest;
+    });
+
+    setDomainThresholdConfigs(tempDomainThresholdConfigs);
+  }, [domains, multiSigConfig, setValue]);
+
+  useImperativeHandle(callback, () => ({
```
rumzledz, line 106: I think it would be better to refer to `callback` as `callbackRef` here, just for a bit of clarity since this is after all supposed to be a ref

And because we're already exploring the depths of refs here, if you have time, I think it would be great to give this hook a type so we know exactly what the name of the exposed functions are 😄

rumzledz (commented):
Awesome start to this @mmioana, I can see this is in Draft mode but because other people are dropping comments, I thought I'd review it as well 😄 

I must admit that I rarely ever find a use-case for the `useImperativeHandle` hook and I tend to stay away from it but looking at the proposed changes, I guess your use case makes sense and it's just doing something that's easily trackable/debuggable!

I left comments here and there, I hope you find them helpful! 🙏

src/components/frame/Extensions/pages/partials/InstallButton.tsx
```diff
@@ -58,6 +65,10 @@ const InstallButton = ({ extensionData }: InstallButtonProps) => {
         `/${colonyName}/${COLONY_EXTENSIONS_ROUTE}/${extensionData?.extensionId}/${COLONY_EXTENSION_SETUP_ROUTE}`,
       );
     }
+    // @TODO will need to do the same for Extension.VotingReputation
```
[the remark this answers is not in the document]
mmioana: As Joanna did some work regarding the extension tab navigation here https://github.com/JoinColony/colonyCDapp/blob/5c654be8274ffedc7e1be8f85f08ec4d79383a45/src/components/frame/Extensions/pages/Extensions/ExtensionDetailsPage/ExtensionDetailsPageContent/consts.ts
I didn't want to overlap too much with her work regarding the VotingReputation tab navigation

src/components/frame/Extensions/pages/ExtensionDetailsPage/ExtensionDetailsPage.tsx
mmioana: @rumzledz we would definitely benefit from that but I didn't want to overlap too much with Joanna's work here https://github.com/JoinColony/colonyCDapp/blob/5c654be8274ffedc7e1be8f85f08ec4d79383a45/src/components/frame/Extensions/pages/Extensions/ExtensionDetailsPage/types.ts

src/components/frame/Extensions/pages/MultiSigPage/hooks.ts
mmioana: @rumzledz even if the handler only has the responsibility for now to call the setThresholdType state handler, I would prefer to keep the separate handler as it makes things easier to change in the future if there is a need

bassgeta (approved):
Nice job navigating the sea of quirks code wise and ux wise, making that button submit the form was a herculean task (no joke) and you've handled it pretty nicely.
I'm usually not the biggest fan of `useImperativeHandle` but this is its use case for sure!
image
image
image

thresholds all update nicely and the changes are reflected after reload :100:

rdig (approved):
Wow this was a big components refactor. Nicely done Ioana seeing this all through to the end.

My testing was fairly pleasant and I couldn't find anything major being broken _(just some code nitpicks)_ but then again I came in at the end, after this PR already wend through a number of back and fourths

Screenshot from 2024-07-23 22-03-58
Screenshot from 2024-07-23 22-04-10
Screenshot from 2024-07-23 22-05-55
Screenshot from 2024-07-23 22-06-08

src/components/frame/Extensions/pages/ExtensionDetailsPage/partials/LazyConsensusSettingsTab.tsx
```diff
@@ -2,58 +2,71 @@ import { type Extension } from '@colony/colony-js';
...
-                <p>{title}</p>
-                <div>
-                  {value} {complementaryLabel === 'percent' ? '%' : 'Hours'}
+    <>
+      <div className="mb-9 md:hidden">
+        <ExtensionDetails extensionData={extensionData} />
+      </div>
+      <li>
+        {Object.keys(params)
+          .filter((key) => key !== '__typename')
```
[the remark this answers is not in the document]
mmioana: @rdig as Joanna did some refactoring for the Lazy Consensus and extensions tab navigation in this PR https://github.com/JoinColony/colonyCDapp/pull/2717 I would postpone any rework involving this specific extension

I would propose to create a separate issue or do you think we should try to cherry-pick from the [feat/streaming-payments-ui](https://github.com/JoinColony/colonyCDapp/tree/feat/streaming-payments-ui)?

src/components/v5/common/Pills/ExtensionStatusBadge/ExtensionStatusBadge.tsx
```diff
@@ -25,7 +25,10 @@ const ExtensionStatusBadge: FC<PropsWithChildren<PillsProps>> = ({
       'border border-gray-200 bg-base-white text-gray-900':
         mode === 'governance' || mode === 'payments',
       'bg-success-100 text-success-400':
-        mode === 'staking' || mode === 'enabled' || mode === 'new',
+        mode === 'staking' ||
+        mode === 'enabled' ||
+        mode === 'new' ||
+        mode === 'installed',
```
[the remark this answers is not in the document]
mmioana: @rdig I also feel we'd benefit more from using enums. However it seems to be a bigger refactoring task, so let's create a new issue?

src/components/common/Extensions/ExtensionItem/ExtensionItem.tsx
```diff
@@ -61,13 +59,15 @@ const ExtensionItem: FC<ExtensionItemProps> = ({
...
+          {button}
+        </div>
       </div>
-      {isMobile && (
+      {/* {isMobile && (
         <div className="ml-3.625rem mt-6 w-[calc(100%-3.625rem)] shrink-0 sm:ml-4 sm:mt-0 sm:w-auto">
           {button}
         </div>
-      )}
+      )} */}
```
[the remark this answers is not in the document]
mmioana: I just forgot to remove it 🙈

rumzledz (approved):
This is really brilliant stuff @mmioana ✨ The UI matches the Figma designs!

Screenshot 2024-07-24 at 12 37 02
Screenshot 2024-07-24 at 12 36 50
Screenshot 2024-07-24 at 12 32 32
Screenshot 2024-07-24 at 12 32 19

The only minor pick up, and it's really minor and I don't wanna make a change request for it and ruin the fun. The threshold field allows digits with decimal points and clicking the Save button doesn't do anything. But it's really minor stuff!

rumzledz (changes requested):
Apologies @mmioana, Mel just discussed the Threshold input field with me and it would be great to have it fixed in this PR so we won't ever have to look at it again. Everything else is brilliant!

mmioana:
@rumzledz the step property doesn't allow you to enter '.' values, however values with ',' were still allowed.
I added a onKeyDown event handler to prevent this from happening, if you can please re-check 🙏

rumzledz (approved):
I can confirm that the Threshold value doesn't accept decimal places now. Awesome work on this @mmioana 🛠️

melyndav (changes requested):
Hi @mmioana - thanks for your continued work on the PR with the UI fixes. 

Some minor pick-ups:

1. Default font weight in the inputs should be normal as per Figma designs and to keep consistency across the app with this component. 

As per Figma: 
image

---

2. From my previous comment (# 3) I can add more than 5 values into the threshold inputs: 

_In the threshold input, I can just keep typing an unlimited amount until it throws an error, but then allows me to override the error and keep entering values. The max should be set at 5 digits and an error shown when exceeded. Recording to demonstrate:_

Here I've added 8 numbers:
image

 
Same issue with the individual team thresholds:

image
 
 ---
 
 3. There is no error state for the above, so I am able to submit this value as mentioned in my previous comment.
 
 ---
 
 4. Can we please update the error copy to match Figma when the threshold is left empty before enabling. 
 
image

---

5. I see the 'save settings' button has been updated with the loading state, can this also be added to the enable button for consistency and avoid the spinner showing in the UI.

image

melyndav (approved):
Hi again @mmioana thanks for all the additional fixes, I think we can get this one merged. I will raise a few other minor issues in a separate issue. Thanks again for the work put into this new extension page. 🌼

bassgeta (approved):
Still working and looking good :sunglasses: 
Error messages do show up and look the part:
image
image
The enable button is also now a spinner:
image

Let's get this last huge chunk in and we're almost readty :rocket: 
side note, I like the `ButtonWithLoader` component, I think at some point it would be beneficial to unify our button components, but this is just nice use of component composition, good stuff, keep cooking :ok_hand:

PR #2748, opened 2024-07-19
[Multi-sig] feat: get eligible signees from the database instead of context

### LINKED ISSUE #2633: [Multi-sig] feat: implement a way of fetching members for each required role
### Story

When a user is creating a multi sig motion, they need to be able to see all colony members who have the necessary permissions to sign a multi sig motion.

### Description

Currently, we fetch this data from `MemberContext`, but this can be troublesome for larger colonies, which don't have all the members yet loaded. We need to fetch the data from the API if possible.
We need to modify [this hook](https://github.com/JoinColony/colonyCDapp/blob/4693db208d478a5cd734a665fb68c05c59e6d89f/src/hooks/multiSig/useEligibleSignees.ts) to fetch the data via graphql.
It should be able to somehow query similar to `getRoleByDomainAndColony`, but do it for each provided role.
All solutions are fair game, this is just a rough outline.

We should also return all the roles a user can sign with, so we can do the following for `Awaiting`

Image

- [ ] The notification banner pops up the same when there aren't enough permission holders in a domain
image
- [ ] The eligible signees widget returns all the permission holders on a pending multi sig motion

bassgeta:
## Description

This was a doozie :) I wanted to simplify our fetching of eligible signees to not use a 2d array anymore and it actually worked!

## Testing

Only 2 test cases to try out, no need to reset anything in between

### Can't create an action without enough permission holders
1. Install the MultiSig extension and give it a fixed threshold of 2
2. Refresh the page and try to create any action with Multi-Sig as the decision method
image
Manage reputation is also of special interest to us 
image
3. You shouldn't be able to create any multi sig motion, because `leela` is the only permission holder

On to the next test case we go!

### Multi sig widget works
1. Assign `amy` `arbitration` and `administration` Multisig roles
image
2. Assign `fry` `funding` and `architecture` Multisig roles
image
3. Refresh the page and create a simple payment motion
image
4. Verify that when hovering over `amy`'s `Awaiting 2` (the square for the 2 :') ) you get which roles she can approve with
image
5. Do the same for `fry`
image
6. If you're feeling like it (it's more of a sanity check), open up 2 more browsers, log in as `amy` and `fry` and approve this motion as each of them
7. Verify that the motion can be finalized, if you click on `Approval` in the stepper, 
image
the tooltips should show up correctly :)

### Some more test cases
* Motions requiring root permission holders (e.g. mint tokens), shouldn't be able to get created
image

* You can create motions that require `Architecture` permission holders e.g. creating a new team
image

* You can't create an award reputation multisig motion
image

* You can however create a reputation smite multisig motion
image

## Diffs

**New stuff** ✨

* `getEligibleSignees` async helper to fetch users from the API

**Changes** 🏗

* `useEligibleSignees` is now just a dumb hook wrapper around `getEligibleSignees`
* flattened `requiredRoles` to always be a 1d array when working with MultiSig
* `getAnyPermissionProofs` now uses a 1d array since MultiSig voting was it's only use case
* `voteOnMultiSig` saga now receives a 1d array or roles to vote with

Resolves #2633

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/ActionSidebarContent/partials/SidebarBanner.tsx
+++ b/src/components/v5/common/ActionSidebar/partials/ActionSidebarContent/partials/SidebarBanner.tsx
@@ -16,11 +16,9 @@ import NotificationBanner from '~v5/shared/NotificationBanner/index.ts';
 
 import {
   ACTION_TYPE_FIELD_NAME,
-  CREATED_IN_FIELD_NAME,
   DECISION_METHOD_FIELD_NAME,
 } from '../../../consts.ts';
 import { useIsFieldDisabled } from '../../hooks.ts';
-import { useHasEnoughMembersWithPermissions } from '../hooks.ts';
 
 const displayName =
   'v5.common.ActionSidebar.ActionSidebarContent.SidebarBanner';
@@ -33,12 +31,17 @@ const MSG = defineMessages({
   },
 });
 
-export const SidebarBanner: FC = () => {
+interface SidebarBannerProps {
+  hasEnoughMembersWithPermissions: boolean;
+}
+
+export const SidebarBanner: FC<SidebarBannerProps> = ({
+  hasEnoughMembersWithPermissions,
+}) => {
   const { watch } = useFormContext();
-  const [selectedAction, decisionMethod, createdIn] = watch([
+  const [selectedAction, decisionMethod] = watch([
     ACTION_TYPE_FIELD_NAME,
     DECISION_METHOD_FIELD_NAME,
-    CREATED_IN_FIELD_NAME,
   ]);
 
   const { installedExtensionsData } = useExtensionsData();
@@ -66,12 +69,6 @@ export const SidebarBanner: FC = () => {
 
   const showVersionUpToDateNotification =
     selectedAction === Action.UpgradeColonyVersion && !canUpgrade;
-
-  const hasEnoughMembersWithPermissions = useHasEnoughMembersWithPermissions({
-    decisionMethod,
-    selectedAction,
-    createdIn,
-  });
 
   return (
     <>
--- a/src/components/v5/common/ActionSidebar/partials/MultiSigSidebar/partials/MultiSigWidget/utils.ts
+++ b/src/components/v5/common/ActionSidebar/partials/MultiSigSidebar/partials/MultiSigWidget/utils.ts
@@ -3,7 +3,7 @@ import { isBefore, parseISO, subWeeks } from 'date-fns';
 
 import { MultiSigVote } from '~gql';
 import { type MultiSigUserSignature } from '~types/graphql.ts';
-import { type EligibleSignee } from '~types/multiSig.ts';
+import { type Threshold, type EligibleSignee } from '~types/multiSig.ts';
 
 import { type MultiSigSignee } from './types.ts';
 
@@ -110,4 +110,74 @@ export const getSignaturesPerRole = (
   });
 
   return { approvalsPerRole, rejectionsPerRole };
+};
+
+export const getNumberOfApprovals = (
+  approvalsPerRole: Record<number, MultiSigUserSignature[]>,
+  thresholdPerRole: Threshold,
+): number => {
+  const combinedApprovals = Object.entries(approvalsPerRole).reduce(
+    (approvalCount, [role, approvalsForRole]) => {
+      const thresholdForRole = thresholdPerRole ? thresholdPerRole[role] : 0;
+      return (
+        approvalCount + Math.min(approvalsForRole.length, thresholdForRole)
+      );
+    },
+    0,
+  );
+
+  return combinedApprovals;
+};
+
+export const getNumberOfRejections = (
+  rejectionsPerRole: Record<number, MultiSigUserSignature[]>,
+  thresholdPerRole: Threshold,
+): number => {
+  const combinedRejections = Object.entries(rejectionsPerRole).reduce(
+    (rejectionCount, [role, rejectionsForRole]) => {
+      const thresholdForRole = thresholdPerRole ? thresholdPerRole[role] : 0;
+      return (
+        rejectionCount + Math.min(rejectionsForRole.length, thresholdForRole)
+      );
+    },
+    0,
+  );
+
+  return combinedRejections;
+};
+
+export const getIsMultiSigExecutable = (
+  approvalsPerRole: Record<number, MultiSigUserSignature[]>,
+  thresholdPerRole: Threshold,
+): boolean => {
+  const isMultiSigExecutable =
+    Object.keys(approvalsPerRole).length > 0 &&
+    Object.keys(approvalsPerRole).every((role) => {
+      const approvals = approvalsPerRole[role]?.length || 0;
+      if (!thresholdPerRole || !thresholdPerRole[role]) {
+        return false;
+      }
+      const roleThreshold = thresholdPerRole[role];
+      return approvals >= roleThreshold;
+    });
+
+  return isMultiSigExecutable;
+};
+
+export const getIsMultiSigCancelable = (
+  rejectionsPerRole: Record<number, MultiSigUserSignature[]>,
+  thresholdPerRole: Threshold,
+): boolean => {
+  const isMultiSigCancelable =
+    Object.keys(rejectionsPerRole).length > 0 &&
+    Object.keys(rejectionsPerRole).every((role) => {
+      const rejections = rejectionsPerRole[role]?.length || 0;
+      if (!thresholdPerRole) {
+        return false;
+      }
+      const roleThreshold = thresholdPerRole[role] || 0;
+      return rejections >= roleThreshold;
+    });
+
+  return isMultiSigCancelable;
 };
--- a/src/components/v5/common/ActionSidebar/partials/MultiSigSidebar/partials/MultiSigWidget/utils.ts
+++ b/src/components/v5/common/ActionSidebar/partials/MultiSigSidebar/partials/MultiSigWidget/utils.ts
@@ -2,7 +2,8 @@ import { type ColonyRole } from '@colony/colony-js';
 import { isBefore, parseISO, subWeeks } from 'date-fns';
 
 import { MultiSigVote } from '~gql';
-import { type User, type MultiSigUserSignature } from '~types/graphql.ts';
+import { type MultiSigUserSignature } from '~types/graphql.ts';
+import { type EligibleSignee } from '~types/multiSig.ts';
 
 import { type MultiSigSignee } from './types.ts';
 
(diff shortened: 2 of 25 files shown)


src/utils/multiSig/getEligibleSignees.ts
```diff
@@ -0,0 +1,142 @@
...
+
+        userRoles.forEach((role) => {
+          if (!signeesPerRole[role]) {
+            signeesPerRole[role] = {};
+          }
+
+          signeesPerRole[role][userAddress] = {
+            userAddress,
+            // For some reason TS doesn't want to follow through with the type assumption here since it's a nested loop
+            user: userRoleEntry.targetUser as Partial<User>,
```
rumzledz, line 124: if Partial<User> works, it's probably because there's some nullable field that `user` expects to always have a value 🤷‍♂️

src/redux/types/actions/multiSig.ts
```diff
@@ -42,7 +42,7 @@ export type MultiSigActionTypes =
         colonyRoles: ColonyRoleFragment[];
         vote: MultiSigVote;
         domainId: number;
-        roles: ColonyRole[][];
+        roles: ColonyRole[];
```
rumzledz, line 45: Maybe you can use `RoleGroup` here as well while you're at it?

src/utils/multiSig/getEligibleSignees.ts
```diff
@@ -0,0 +1,142 @@
...
+      const response = await apolloClient.query<
+        GetMultiSigRolesForDomainQuery,
+        GetMultiSigRolesForDomainQueryVariables
+      >({
+        query: GetMultiSigRolesForDomainDocument,
+        variables: {
+          domainId: `${colonyAddress}_${permissionDomainId}`,
+          filter: queryFilter,
+        },
+        fetchPolicy: 'cache-first',
```
rumzledz, line 77: Might need to be on the lookout for this, stale data and whatnot

rumzledz (commented):
Great stuff with removing the 2D array stuff @bassgeta, one less problem to deal with 😄

src/utils/multiSig/getEligibleSignees.ts
bassgeta: huh this one is actually the default one, so I'll remove it, I think it should suffice, I will however try to invalidate queries after manage permissions for multisig roles is done

rumzledz: Invalidating queries would be indeed preferable 👌  At first glance, I thought this data gets updated often so I was gonna mention `cache-and-network` might be more suitable. But looking at it again, doesn't look like it's the case!

src/hooks/multiSig/useEligibleSignees.ts
```diff
@@ -1,104 +1,62 @@
...
-  const { eligibleSignees, uniqueEligibleSignees } =
-    getEligibleSignees(totalMembers);
-
-  const countPerRole: { [role: number]: number } = {};
-  Object.keys(eligibleSignees).forEach((role) => {
-    const numberOfUsers = Object.keys(eligibleSignees[role]).length;
-    countPerRole[role] = numberOfUsers;
-  });
+    fetchEligibleSignees();
+    // Since it's an array fetched from a function, it will always trigger this, so we "memo" it by stringifying it :)
```
davecreaser, line 54: 👏 this is pretty clever!

davecreaser (approved):
Great work @bassgeta! Nasty stuff made more simple for sure, looks really good.

Also, works like a charm, here's my testing below.

Before assigning permissions to Amy & Fry:
Screenshot 2024-07-23 at 11 38 34
Screenshot 2024-07-23 at 11 38 41
Screenshot 2024-07-23 at 11 38 07

And after:
Screenshot 2024-07-23 at 11 39 57
Screenshot 2024-07-23 at 11 40 35
Screenshot 2024-07-23 at 11 42 37
Screenshot 2024-07-23 at 11 42 48
Screenshot 2024-07-23 at 11 42 48
Screenshot 2024-07-23 at 11 42 54
Screenshot 2024-07-23 at 11 44 08

And the "further testing" results:
Screenshot 2024-07-23 at 11 46 28
Screenshot 2024-07-23 at 11 47 01
Screenshot 2024-07-23 at 11 47 20
Screenshot 2024-07-23 at 11 47 28

src/components/v5/common/ActionSidebar/partials/MultiSigSidebar/partials/MultiSigWidget/utils.ts
```diff
@@ -17,29 +18,27 @@ export const hasWeekPassed = (createdAt: string) => {
...
-  requiredRoles: ColonyRole[];
+  eligibleSignees: EligibleSignee[];
 }): MultiSigSignee[] => {
   const notSignedUsers: MultiSigSignee[] = eligibleSignees
     .filter((eligibleSignee) => {
       return !signatures.find(
-        (signature) => signature.userAddress === eligibleSignee?.walletAddress,
+        (signature) => signature.userAddress === eligibleSignee.userAddress,
       );
     })
```
mmioana, line 32: Would it make the code easier to follow if we create an array with the signature user addresses 
```
signatureUserAddresses = signatures.map(signature => signature.userAddress)
```

and then filter the eligibleSignees

```
 eligibleSignees
    .filter((eligibleSignee) => !signatureUserAddresses.includes(eligibleSignee.userAddress))
    .map(...)
```
?

iamsamgibbs (changes requested):
Nice job reworking all of this @bassgeta ! 🎉  It is definitely nicer getting the eligible signees like this.

Following your testing steps, all is working as expected:

Screenshot 2024-07-23 at 10 58 21

Screenshot 2024-07-23 at 11 02 26

I think this was working previously and is a regression though? So far the only vote is from a user with all 3 permissions, so the bar should show `1/2` up until a user with only 1 or 2 of the permissions votes at which point it should split to the `x/6`.

Screenshot 2024-07-23 at 11 02 43

I think the permissions banner shouldn't show here until a team is selected as it feels a bit misleading to the user (only in the case of manage permissions where the required role is different between root and sub-domains). Here I am logged in as a user with architecture multi-sig permission in Andromeda. When I select the multi-sig decision method, the `CreatedIn` field defaults to `General` so it shows me the permission banner which implies I can't create this action. If I switch the team to Andromeda (where I do have permissions) then the banner disappears.

https://github.com/user-attachments/assets/917dbab1-0694-4c8f-9cb7-5cde85c23d90

One other thought is how this banner will work for users on slower connections / if the query fails for any reason - if it was a much slower connection then there is potentially time to submit the action before the notice even shows. I don't know if we need a loading state or anything. (Probably an @arrenv question)

https://github.com/user-attachments/assets/3b4e96e2-5c13-48e7-9ef1-9e502bba004e

(It isn't a huge deal and maybe something we could address at a later point as it is only the notice about not having enough members to meet the threshold. The "you have no multi-sig permissions" still appears instantly.)

I've also left a comment about the change in the proofs breaking `ManagePermissionsInSubDomain` via permissions.

Other than these edge cases, nice work, I'm glad you were able to simplify things by a decent amount!

src/hooks/multiSig/useEligibleSignees.ts
```diff
@@ -1,104 +1,62 @@
...
+          domainId,
+        });
+        setSigneesResult(response);
+        setIsLoading(false);
+      } catch (error) {
+        console.warn('Error while fetching eligible signees', error);
+        setIsError(true);
+        setIsLoading(false);
       }
+    }
```
mmioana, line 51: I would refactor this to make use of the `finally` statement for setting the loading state
```
try {}
catch (error) {}
finally {
   setIsLoading(false);
}
```

src/utils/multiSig/getEligibleSignees.ts
```diff
@@ -0,0 +1,144 @@
...
+import { type EligibleSignee } from '~types/multiSig.ts';
+import { notMaybe } from '~utils/arrays/index.ts';
+
+const COLONY_ROLE_TO_FILTER_KEY = {
+  [ColonyRole.Root]: 'role_1',
+  [ColonyRole.Arbitration]: 'role_2',
+  [ColonyRole.Architecture]: 'role_3',
+  [ColonyRole.Funding]: 'role_5',
+  [ColonyRole.Administration]: 'role_6',
+};
```
mmioana, line 20: Would it help to type the `COLONY_ROLE_TO_FILTER_KEY` as
```
type SelectedColonyRoles = ColonyRole.Root | ColonyRole.Arbitration | ColonyRole.Architecture | ColonyRole.Funding | ColonyRole.Administration;

const COLONY_ROLE_TO_FILTER_KEY: Record<SelectedColonyRoles, keyof ColonyActionRoles> = {
  [ColonyRole.Root]: 'role_1',
  [ColonyRole.Arbitration]: 'role_2',
  [ColonyRole.Architecture]: 'role_3',
  [ColonyRole.Funding]: 'role_5',
  [ColonyRole.Administration]: 'role_6',
};
```
?

mmioana (approved):
Amazing work on this one @bassgeta 🎆 

Tested and can confirm it works as expected
Screenshot 2024-07-23 at 12 45 02
Screenshot 2024-07-23 at 12 45 25
Screenshot 2024-07-23 at 12 52 44
Screenshot 2024-07-23 at 12 52 59
Screenshot 2024-07-23 at 12 54 06
Screenshot 2024-07-23 at 12 54 21
Screenshot 2024-07-23 at 12 54 26
Screenshot 2024-07-23 at 12 54 32
Screenshot 2024-07-23 at 12 55 15
Screenshot 2024-07-23 at 12 55 58
Screenshot 2024-07-23 at 12 56 28
Screenshot 2024-07-23 at 12 56 57

Left a few code refactoring comments, but check if it makes sense to do them, otherwise please feel free to merge this beast 🚀

mmioana (changes requested):
Nice job on refactoring this @bassgeta 💪 

However, I noticed the error messages are no longer displayed, just the form button is disabled 😢 Is this as intended?
Screenshot 2024-07-23 at 17 37 21
Screenshot 2024-07-23 at 17 39 53
Screenshot 2024-07-23 at 17 45 58

The other actions seem to be working fine though 🚀 

Screenshot 2024-07-23 at 17 43 53
Screenshot 2024-07-23 at 17 44 02
Screenshot 2024-07-23 at 17 45 04
Screenshot 2024-07-23 at 17 46 34

src/components/v5/common/ActionSidebar/partials/ActionSidebarContent/partials/SidebarBanner.tsx
```diff
@@ -33,9 +32,15 @@ const MSG = defineMessages({
...
+
+export const SidebarBanner: FC<SidebarBannerProps> = ({
+  hasEnoughMembersWithPermissions,
+}) => {
   const { watch } = useFormContext();
-  const [selectedAction, decisionMethod, createdIn] = watch([
+  const [selectedAction, decisionMethod] = watch([
     ACTION_TYPE_FIELD_NAME,
     DECISION_METHOD_FIELD_NAME,
     CREATED_IN_FIELD_NAME,
```
iamsamgibbs, line 46: Can remove `CREATED_IN_FIELD_NAME`

mmioana (approved):
Awesome @bassgeta everything works now as expected ✨ 

Screenshot 2024-07-24 at 15 43 51
Screenshot 2024-07-24 at 15 44 06
Screenshot 2024-07-24 at 15 46 47
Screenshot 2024-07-24 at 15 47 45
Screenshot 2024-07-24 at 15 48 14
Screenshot 2024-07-24 at 15 48 51

iamsamgibbs (approved):
Nice job on this, still working as expected when following the testing steps:

Screenshot 2024-07-25 at 09 05 00

Screenshot 2024-07-25 at 09 05 56

Screenshot 2024-07-25 at 09 06 19

Screenshot 2024-07-25 at 09 07 09

However, I'm still think the loading is an issue. I think it maybe makes more sense to open it as a new issue (or add to the QoL issue) in order to get this merged, so I'm still approving the PR.

The "not enough members" notice does not show until after this has finished fetching. If I had filled out this form in order before selecting "multi-sig" then I would easily have enough time to submit the form before the message shows.

https://github.com/user-attachments/assets/805a67d6-1ba7-4ce6-b9f6-2461b5d9ce18

We could maybe add the threshold check as part of the form validation too?

Just noting as well that the other issues I previously noted are either invalid or being addressed in a separate issue.

bassgeta:
> Nice job on this, still working as expected when following the testing steps:
> Screenshot 2024-07-25 at 09 05 00 Screenshot 2024-07-25 at 09 05 56 Screenshot 2024-07-25 at 09 06 19 Screenshot 2024-07-25 at 09 07 09
> 
> However, I'm still think the loading is an issue. I think it maybe makes more sense to open it as a new issue (or add to the QoL issue) in order to get this merged, so I'm still approving the PR.
> 
> The "not enough members" notice does not show until after this has finished fetching. If I had filled out this form in order before selecting "multi-sig" then I would easily have enough time to submit the form before the message shows.
> Screen.Recording.2024-07-25.at.09.12.40.mov
> 
> We could maybe add the threshold check as part of the form validation too?
> 
> Just noting as well that the other issues I previously noted are either invalid or being addressed in a separate issue.

Thanks for the recheck, I think we can add this to the QoL issue, so we don't block this one too long + it's gonna be easier to look at just those changes too :D

PR #2753, opened 2024-07-22
Fix: Contributor reputation caching issues

### LINKED ISSUE #2315: Members cached reputation values don't update
## Steps to reproduce

1. Create a multicall transaction to award reputation using `emitDomainReputationReward`.
2. Pay attention to reputation updates in the contributors list and compare it to the value in the header.
3. Make a payment to the member in root or in a team.
4. Reputation won't update or appear in both instances.

Note: I can provide a reputation multicall script if required.

## Expected behaviour

* Reputation values in the contributor list should reflect the users actual reputation values.

## Actual behaviour

* Reputation is not correct for users in the contributor list. The value in the header is correct.
* No value is showing at all when filtering by a team, even after making a payment and when they should have reputation.

image

image

davecreaser:
## Description

- Fix an issue where there was a mismatch between the user hub reputation and the contributors page reputation values.

This issue stems from the fact that both reputation values are requested in different ways.

- The user hub uses the singular `getUserReputation` lambda to fetch the reputation directly from the network.
- The contributors page uses a search in the database to get all of the reputation values for the contributors of a colony.

Whilst this makes sense, it leads to the next issue, which is that the database needs to stay up to date with the network. The way this is implemented is by calling a lambda (`updateContributorsWithReputation`) every time the colony is loaded, inside the ColonyContextProvider. This lambda simply calls the network in the background to fetch the latest reputation data, and updates the database accordingly.

The problem that this PR addresses, is that the contributors page never actually responds to that database update, as it has already fetched the reputation on page load. It needs to fetch it again once the lambda has finished, if there was an update to the reputation values. This is accomplished using graphql subscriptions, reacting whenever the `lastUpdatedContributorsWithReputation` property of a colony has changed to refetch from the database.

## Testing

**Important**
Before testing this, you need to go to http://0.0.0.0:3001/reputation/monitor/toggle to ensure that reputation is being mined in your environment. More details [here](https://www.notion.so/colony/Reputation-in-development-c05b7a1f79fc4465a220c2806792e4a7) if you want to know why.

I developed this pretty much the whole time with my window split into two, with one browser logged into Leela on the left, and an incognito window logged into Amy on the right. I suggest the same for testing, since it makes it much quicker.

* Step 1. Set up the browsers as described above, and go to the contributors list on the `Amy` browser. Make a note of her reputation, both in the user hub button, and the contributors list.
Screenshot 2024-07-22 at 17 09 31

* Step 2. Make an action as Leela to either award/remove reputation for Amy, or make a simple payment to Amy.
Screenshot 2024-07-22 at 17 10 15

* Step 3. **Wait a few seconds** to give the reputation monitor time to process the changes. After a brief wait, refresh the page where Amy is logged in. You should see the new reputation in the user hub, and for a couple of seconds the old reputation in the contributors list. This value in the contributors list should then update to match the user hub when the lambda has finished processing the changes.
Screenshot 2024-07-22 at 17 10 49

**Note**
Whilst I am aware that this flicker of an incorrect value is not ideal, this will only ever happen if the user refreshes the page **while on** the contributors list. If they refresh the page anywhere else and _then_ navigate to the contributors list, then by the time they have reached this page the lambda should have finished and the reputation value should be correct already.

## Diffs

**New stuff** ✨

* Added a refetching mechanism to the MembersContextProvider to refetch the contributors if reputation has been updated.

**Changes** 🏗

* Changed the reputation update lambda to use the same formula for calculating percentages as the user hub button, to avoid the minor differences that were happening before (ie. one would show 33.04% while the other would show 33.05% in certain circumstances).

Resolves #2315


amplify/backend/function/updateContributorsWithReputation/src/utils.js
```diff
@@ -281,6 +282,29 @@ const updateReputationInDomain = async ({
   );
 };
 
+const calculatePercentageReputation = (userReputation, totalReputation) => {
```
davecreaser, line 285: I took this (and slightly edited it to work in the lambda) from `src/utils/reputation.ts` so that it was being calculated in a uniform way across the app. Before, there were minor differences leading to slightly different percentages being shown.

src/graphql/queries/colony.graphql
```diff
@@ -85,3 +85,9 @@ query GetColonyMemberInvite($id: ID!) {
     invitesRemaining
   }
 }
+
+subscription OnUpdateColony {
+  onUpdateColony {
+    ...Colony
+  }
+}
```
davecreaser, line 93: Big thanks to @chmanie for this... 🍺

amplify/backend/function/updateContributorsWithReputation/src/utils.js
```diff
@@ -281,6 +282,29 @@ const updateReputationInDomain = async ({
...
+  const totalReputationNumber = BigNumber.from(totalReputation);
+
+  const reputationSafeguard = BigNumber.from(100).pow(2);
+
+  if (
+    userReputationNumber.isZero() ||
+    totalReputationNumber.isZero() ||
+    userReputationNumber.mul(reputationSafeguard).lt(totalReputationNumber)
+  ) {
+    return 0;
```
rdig, line 297: Thank you for keeping the fallback here.

I've put this in a while back when I noticed colony rep percentage was not coming in during updates, breaking this logic and setting it to empty in the database

This ultimately broke the members/reputation query

src/graphql/queries/colony.graphql
rdig: This is actually the first subscription that's actually being used :tada:

rdig (approved):
Thank you so much for picking this up and fixing it.

I did notice it, and do agree that the flash that happens on the members page is less then ideal, however I don't think it's a big issue. If it becomes one, we can just fix it a later date.

Screenshot from 2024-07-22 21-53-05
Screenshot from 2024-07-22 22-03-13
Screenshot from 2024-07-22 22-03-35
Screenshot from 2024-07-22 22-04-43
Screenshot from 2024-07-22 22-05-16

amplify/backend/function/updateContributorsWithReputation/src/utils.js
```diff
@@ -281,6 +282,29 @@ const updateReputationInDomain = async ({
...
+  }
+
+  const reputation = userReputationNumber
+    .mul(reputationSafeguard)
+    .div(totalReputationNumber)
+    .toNumber();
+
+  return reputation / 10 ** 2;
```
mmioana, line 305: Is there a reason we don't divide back to `reputationSafeguard` value before the `.toNumber()` conversion?

mmioana (approved):
On another try I managed to also see the update on the member card (the members didn't disappear anymore)

Screenshot 2024-07-24 at 12 27 24
Screenshot 2024-07-24 at 12 27 18

jakubcolony (changes requested):
The fix works for me, very nice job 👍 

Left a comment about the colony fragment you're subscribing to, but otherwise all good

Before awarding:
Screenshot 2024-07-24 at 20 17 13
Screenshot 2024-07-24 at 20 15 56

After awarding:
Screenshot 2024-07-24 at 20 17 13

Screenshot 2024-07-24 at 20 17 17

src/graphql/queries/colony.graphql
```diff
@@ -85,3 +85,9 @@ query GetColonyMemberInvite($id: ID!) {
     invitesRemaining
   }
 }
+
+subscription OnUpdateColony {
+  onUpdateColony {
+    ...Colony
```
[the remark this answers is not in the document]
davecreaser: I've been trying for the last few hours to figure out a way to make it work without getting the full fragment, but I can't seem to get it to work. I think it needs to do this in order to refetch the contributors properly. When I took out the fragment and replaced it with just this:

```graphql
subscription OnUpdateColonyReputation {
  onUpdateColony {
    id
    lastUpdatedContributorsWithReputation
    reputation
  }
}
```

then the contributors were not updating properly when refetched. Their `reputation` items were updating, but not the `colonyReputationPercentage`. In order to get this to update I had to refetch the colony from the network (note: not from the cache, so I couldn't just use the refetchColony from the context provider as that has a 'cache-first' policy). And if I'm refetching the colony from the network like that anyway then I think it's better to just keep the fragment in the subscription.

How concerning is this fragment? And could you think of anything which may help? I'm a bit at a loss today but can take another look tomorrow if it's really important that it doesn't use the fragment.

amplify/backend/function/updateContributorsWithReputation/src/index.js
```diff
@@ -230,7 +231,14 @@ exports.handler = async (event) => {
...
 
-                console.log({ type, colonyReputationPercentage, contributorAddress, contributorRepDecimal, domainReputationPercentage, reputation })
+                console.log({
+                  type,
+                  colonyReputationPercentage,
+                  contributorAddress,
+                  contributorRepDecimal,
+                  domainReputationPercentage,
+                  reputation,
+                });
```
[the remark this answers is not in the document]
davecreaser: I didn't add them and assumed they were for debugging, should I remove them?

src/graphql/queries/colony.graphql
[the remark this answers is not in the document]
jakubcolony: @davecreaser I think there's a race condition between the subscription getting new data and contributor's `colonyReputationPercentage` actually being updated in the DB. 

I tried the lighter fragment you posted and got the same result of reputation percentage not being updated.

Then I added a 2s delay to the `searchColonyContributors` query refetch and it started returning the updated percentage value:

https://github.com/user-attachments/assets/923705c4-d324-4cb0-bd3b-3dde917e1052

The fact that `searchColonyContributors` is an OpenSearch powered query could explain this, posting updated data to OpenSearch is probably an asynchronous task. Fetching the bigger fragment took more time, so it coincidentally fixed the symptom. 

I added some logs to the lambda and could see the colony is updated (triggering the subscription) before the updated contributor is posted to OpenSearch:
image

My main concern with subscribing to the entire fragment is the fact that this subscription will be getting updates pretty frequently due to the way `updateContributorsWithReputation` works. A lot of redundant data would keep getting sent, and because some of the colony fragment fields are resolved by lambdas, it could also end up increasing the AWS bill.

Setting a timeout to refetch the search contributors query feels a bit hacky, but I couldn't think of another solution to account for the delay, other than reverting to traditional polling for that query.

amplify/backend/function/updateContributorsWithReputation/src/index.js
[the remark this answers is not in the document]
jakubcolony: They look like dev leftovers, I would remove them

jakubcolony (commented):
@davecreaser I think there's a race condition between the subscription getting new data and contributor's `colonyReputationPercentage` actually being updated in the DB. 

I tried the lighter fragment you posted and got the same result of reputation percentage not being updated.

Then I added a 2s delay to the `searchColonyContributors` query refetch and it started returning the updated percentage value:

https://github.com/user-attachments/assets/923705c4-d324-4cb0-bd3b-3dde917e1052

The fact that `searchColonyContributors` is an OpenSearch powered query could explain this, posting updated data to OpenSearch is probably an asynchronous task. Fetching the bigger fragment took more time, so it coincidentally fixed the symptom. 

I added some logs to the lambda and could see the colony is updated (triggering the subscription) before the updated contributor is posted to OpenSearch:
image

My main concern with subscribing to the entire fragment is the fact that this subscription will be getting updates pretty frequently due to the way `updateContributorsWithReputation` works. A lot of redundant data would keep getting sent, and because some of the colony fragment fields are resolved by lambdas, it could also end up increasing the AWS bill.

Setting a timeout to refetch the search contributors query feels a bit hacky, but I couldn't think of another solution to account for the delay, other than reverting to traditional polling for that query.

davecreaser:
Thanks Jakub for looking into it so in depth, very helpful! I think you're very much right about the race condition. I also considered polling at one point but i'm also not a huge fan of using that here since then it's also making a lot of requests to a search. I've implemented the timeout with a comment explaining it. I agree that I think it's the best way forwards here!

Ready for review again :)

jakubcolony (approved):
Works well now, thanks @davecreaser 👍 Nice shout with adding the comments to explain the reason behind timeouts

Once this goes into prod, we should probably do another check to make sure the 2s delay is enough

mmioana (approved):
LGTM @davecreaser 🎉 
Screenshot 2024-08-01 at 12 26 15
Screenshot 2024-08-01 at 12 26 53
Screenshot 2024-08-01 at 12 27 21
Screenshot 2024-08-01 at 12 27 43

rdig (approved):
Still looks good, and now that you've addressed all the other pending complaints, this is good to go.

Nicely done

PR #2756, opened 2024-07-23
Feat/2658 invalidate permissions query

mmioana:
## Description

- After users are given multi-sig permissions (including when the extension is installed and as a result of a manage permissions action / motion finalization), invalidate the permission holder query so we can immediately create multi-sig actions with them as potential signers

## Testing

Check members get multi-sig permissions after Multi-Sig extension is installed or after a Manage Permissions action without a page refresh.

* Step 1. Install Multi-Sig extension in Colony > Admin > Extension
Screenshot 2024-07-23 at 18 01 24

* Step 2. Without refreshing the page, check the Extension details section and click on `leela`
Screenshot 2024-07-23 at 18 07 09

* Step 3. Check `leela` already has Multi-Sig permissions
Screenshot 2024-07-23 at 18 07 15

* Step 4. Go to Colony > Members > Permissions
Screenshot 2024-07-23 at 18 09 26

* Step 5. Switch to the **Multi-sig Authority** tab and check `leela` is present in the list
Screenshot 2024-07-23 at 18 01 54

* Step 6. Assign `amy` some custom permissions using **Permissions** as decision method and **Takes actions via Multi-Sig** as authority
Screenshot 2024-07-23 at 18 02 30

* Step 7. Check `amy` appears in the **Multi-sig Authority** page without the need for refresh
Screenshot 2024-07-23 at 18 02 37

### Further testing scenario - 1
* Try setting the Multi-sig extension threshold to 3 
Screenshot 2024-07-24 at 16 38 45
* Give 3 users Owner permissions
* Try creating a Mint tokens action
Screenshot 2024-07-24 at 16 18 51
* Change Multi-sig extension threshold to 4
* Try creating a Mint tokens action - an error message should appear
Screenshot 2024-07-24 at 16 19 26

### Further testing scenario - 2
You can continue creating multiple Manage permissions actions and check the Members/Contributors/Permissions page update accordingly without the need for a page refresh

Contributes to #2658

### DIFF
--- a/src/redux/sagas/motions/managePermissionsMotion.ts
+++ b/src/redux/sagas/motions/managePermissionsMotion.ts
@@ -8,7 +8,6 @@ import { ADDRESS_ZERO } from '~constants/index.ts';
 import { type Action, ActionTypes, type AllActions } from '~redux/index.ts';
 import { Authority } from '~types/authority.ts';
 import { TRANSACTION_METHODS } from '~types/transactions.ts';
-import { clearContributorsAndRolesCache } from '~utils/members.ts';
 import { putError, takeFrom } from '~utils/saga/effects.ts';
 
 import {
@@ -335,8 +334,6 @@ function* managePermissionsMotion({
         state: { isRedirect: true },
       });
     }
-
-    yield clearContributorsAndRolesCache();
   } catch (caughtError) {
     yield putError(ActionTypes.MOTION_USER_ROLES_SET_ERROR, caughtError, meta);
   } finally {
--- a/src/redux/sagas/motions/managePermissionsMotion.ts
+++ b/src/redux/sagas/motions/managePermissionsMotion.ts
@@ -8,6 +8,7 @@ import { ADDRESS_ZERO } from '~constants/index.ts';
 import { type Action, ActionTypes, type AllActions } from '~redux/index.ts';
 import { Authority } from '~types/authority.ts';
 import { TRANSACTION_METHODS } from '~types/transactions.ts';
+import { clearContributorsAndRolesCache } from '~utils/members.ts';
 import { putError, takeFrom } from '~utils/saga/effects.ts';
 
 import {
@@ -334,6 +335,8 @@ function* managePermissionsMotion({
         state: { isRedirect: true },
       });
     }
+
+    yield clearContributorsAndRolesCache();
   } catch (caughtError) {
     yield putError(ActionTypes.MOTION_USER_ROLES_SET_ERROR, caughtError, meta);
   } finally {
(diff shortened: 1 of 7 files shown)


src/redux/sagas/motions/managePermissionsMotion.ts
```diff
@@ -334,6 +335,8 @@ function* managePermissionsMotion({
         state: { isRedirect: true },
       });
     }
+
+    yield clearContributorsAndRolesCache();
```
iamsamgibbs, line 339: I think this will clear the cache when the motion is created. We should do it when the motion is finalized, there are different sagas for finalizing a voting reputation motion or a multi-sig motion. :)

mmioana: Thank you for pointing this out @iamsamgibbs 🙏

iamsamgibbs (approved):
Great work on this @mmioana , tested and working like a charm!

Cache is getting cleared after installation.

https://github.com/user-attachments/assets/133543de-bc28-48e9-affe-a4274ee92fc7

Cache is getting cleared when assigning via permissions.

https://github.com/user-attachments/assets/9116456c-bb53-46d9-8f77-4ca76715744d

Cache is getting cleared when finalizing a multi-sig manage permissions motion.

https://github.com/user-attachments/assets/79814015-a3ff-44d8-8195-4f321ae7f7b3

I was unable to test reputation motions due to a bug unrelated to this PR, but it calls the same function as finalizing a multi-sig, so no reason it shouldn't work.

Screenshot 2024-07-25 at 09 49 53

When testing, I noticed you need to refresh after updating the threshold. This is something we should also clear the cache on (or whatever needs to be done to fix it). Happy for this to be added to the QoL issue, or tackled here if you would prefer.

https://github.com/user-attachments/assets/6479ad10-2442-4ca2-b478-f22d2a0b0a82

Completely unrelated, but I also noticed Leela inherits all the permissions from General down into all the sub-domains, whilst Amy doesn't - despite having basically the same permissions set. The only difference is that Leela has core permissions in general in addition to multi-sig permissions. Very likely related to point 20 on the QoL issue.

Screenshot 2024-07-25 at 09 39 10

Screenshot 2024-07-25 at 09 39 15

Screenshot 2024-07-25 at 09 39 20

Nice job! 🚀

bassgeta (approved):
Codewise it makes sense, buy boy do I dislike apollo's naming here, it makes it look so severe :sweat_smile: 
Immediately after installation `leela` has multisig roles
image
Permissions are there:
image
Amy immediately showed up (and my dev env got somehow borked)
image
Threshold of 1, mint tokens created:
image
Changed it to 2 and then it doesn't work anymore:
image

Nice work, let's get it merged!

Nortsova (approved):
Thanks for this PR @mmioana ! Now it works smoothly 🤩

Can confirm that leea has multisig permissions without reloading page:
image

Can confirm that Multi-Sig appears as option to create Motion with Multi-sig decision method:
image

Can confirm that Amy appears in a list after setting her permissions:

Jul-26-2024 12-45-45

Can confirm that threshold is giving an error
image

Note: threshold working strange and look like it update data only after second submit, or you need to reload instead of second submit. But I don't think that this is related to you PR ✨

PR #2757, opened 2024-07-23
Change undefined to empty string in AmountField component

arrenv:
## Description

This is a very simple PR to hopefully solve a nuisance that has consistently been occurring for me for quite some time. I'm not really sure if it is the right way to solve it, but it seems to make sense.

Basically, I am just assigning a value to a visible component instead of having the initial value as undefined. Validation still works, and the field seems to work as expected still.

## Testing

| Step | Expected |
| :--- | :-- |
| Create an action with an Amount field (e.g. Simple payment) and enter a value. | You should not get an error in console. |

## Changes
- Changed initial undefined value to empty string in `AmountField` component.

## Issue

- The first time entering a value in the Amount field of an action results in a console error "A component is changing an uncontrolled input to be controlled."
- Error can be repeated after refreshing the app.

image

Resolves N/A

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/AmountField/AmountField.tsx
+++ b/src/components/v5/common/ActionSidebar/partials/AmountField/AmountField.tsx
@@ -64,7 +64,7 @@ const AmountField: FC<AmountFieldProps> = ({
     dropdownRef,
     adjustInputWidth,
   } = useAmountField(tokenAddressController.value, maxWidth);
-  const [value, setValue] = useState<string | undefined>(
+  const [value, setValue] = useState<string>(
     field.value ? formatNumeral(field.value, formattingOptions) : '',
   );
 
--- a/src/components/v5/common/ActionSidebar/partials/AmountField/AmountField.tsx
+++ b/src/components/v5/common/ActionSidebar/partials/AmountField/AmountField.tsx
@@ -65,7 +65,7 @@ const AmountField: FC<AmountFieldProps> = ({
     adjustInputWidth,
   } = useAmountField(tokenAddressController.value, maxWidth);
   const [value, setValue] = useState<string | undefined>(
-    field.value ? formatNumeral(field.value, formattingOptions) : undefined,
+    field.value ? formatNumeral(field.value, formattingOptions) : '',
   );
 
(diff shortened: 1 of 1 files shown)


iamsamgibbs (approved):
Thanks for fixing this, it has been bugging me too. Doesn't seem to have introduced any new issues.

Screenshot 2024-07-23 at 16 16 56

Screenshot 2024-07-23 at 16 18 51

Compared to master:
Screenshot 2024-07-23 at 16 18 06

src/components/v5/common/ActionSidebar/partials/AmountField/AmountField.tsx
```diff
@@ -65,7 +65,7 @@ const AmountField: FC<AmountFieldProps> = ({
     adjustInputWidth,
   } = useAmountField(tokenAddressController.value, maxWidth);
   const [value, setValue] = useState<string | undefined>(
```
mmioana, line 67: Maybe we can also remove the `undefined` from the useState type?

mmioana (dismissed):
Thank you a lot @arrenv for solving this, I also noticed it since some time 👍 

On another branch I could see the error
Screenshot 2024-07-23 at 17 52 44

While on yours everything works smoothly
Screenshot 2024-07-23 at 17 53 38
Screenshot 2024-07-23 at 17 53 47

src/components/v5/common/ActionSidebar/partials/AmountField/AmountField.tsx
davecreaser: +1 to this!

davecreaser (approved):
Alt Text

Hurray! This has also been bothering me for a while, thanks for just doing it 😆 

Had a quick sanity check and all looks good.

Screenshot 2024-07-23 at 17 27 26

src/components/v5/common/ActionSidebar/partials/AmountField/AmountField.tsx
arrenv: Good pickup, thank you. Removed as well.

iamsamgibbs (approved):

davecreaser (approved):
Excellent, thanks for updating the type too. Let's get it in so we never see this error again 🤞

PR #2758, opened 2024-07-23
Fix: Check transaction creator on transaction update calls

davecreaser:
## Description

This PR simply updates the auth proxy hash, the real changes are [here](https://github.com/JoinColony/colonyCDappAuthProxy/pull/23)

Basically what we need to do is, ensure mutations to update a transaction only pass the auth proxy if the `from` parameter in the request is not only the same as the currently logged in user, but also the same as the `from` parameter of the actual transaction in the database. Give me (or @chmanie) a shout on Discord if you want more information on that!

## Testing

**Restart your stack so that the correct auth proxy version is running.**

This is a little tricky to test, but bear with me here...

* Step 1. Open the dev console in the network tab, log in as Leela, and make a Mint tokens action.

* Step 2. In the network tab, find the graphql request where the operation name is `CreateTransaction` and you're creating a mint tokens transaction, and find the transaction id.

Screenshot 2024-07-23 at 16 46 42

* Step 3. Go to localhost:20002, and run this query, replacing the id with the id that you found in the last step:
```graphql
query MyQuery {
  getTransaction(id: "ppaa1ncecr8-mintTokens") {
    from
    status
  }
}
```

Screenshot 2024-07-23 at 16 47 39

* Step 4. Go back to the network tab from step 2, and find the graphql request where the operation name is `UpdateTransaction`, and the transaction you're updating is the mint tokens transaction, setting it to the ready state. It should be only two or three requests below the one you found earlier.

Screenshot 2024-07-23 at 16 50 06

* Step 5. Right click on this request, and choose copy -> copy as cURL

Screenshot 2024-07-23 at 16 51 40

* Step 6. Paste this cURL request into your terminal. We're going to edit it a little bit, so don't send it just yet.

Screenshot 2024-07-23 at 16 54 15

* Step 7. In an incognito window, log in as someone else (I used Fry, since I always use Amy and I felt bad for Fry). Once logged in as this user, open the dev tools and navigate to the cookies. Find the cookie called `auth` and copy it's value.

Screenshot 2024-07-23 at 16 55 45

* Step 8. Paste this value into the request in your terminal, replacing the auth cookie from earlier.

Screenshot 2024-07-23 at 16 57 15

Screenshot 2024-07-23 at 16 58 23

* Step 9. Go back to your incognito tab, and copy the user's address in the hamburger menu.

Screenshot 2024-07-23 at 16 59 34

* Step 10. Paste this into the request, replacing the `from` variable in the graphql mutation.

Screenshot 2024-07-23 at 17 00 35

Screenshot 2024-07-23 at 17 01 37

* Step 11. Finally, send this cURL request. After all that, we are expecting it to fail! If it does, we're happy :)

Screenshot 2024-07-23 at 17 02 39

To double check that it did indeed not happen, run the query from step 3 again, and you should see that the transaction was not updated.

Screenshot 2024-07-23 at 16 47 39

**More in depth testing**

If you really want, try this with the auth proxy hash set to the latest commit on master, and see how the last step is different.

Another way to test this is to stop the auth proxy in your docker, and instead run it with `npm run dev` in your terminal. The logger is switched on there, so you'll have a lot more information about each request and whether or not it passes the auth checks.

## Diffs

**Changes** 🏗

* New auth proxy hash.

Resolves [colonyCDappAuthProxy#17](https://github.com/JoinColony/colonyCDappAuthProxy/issues/17)

### DIFF
--- a/docker/colony-cdapp-dev-env-auth-proxy
+++ b/docker/colony-cdapp-dev-env-auth-proxy
@@ -1,6 +1,6 @@
 FROM colony-cdapp-dev-env/base:latest
 
-ENV AUTH_PROXY_HASH=cdcbc5c7e01e3102282c7cd26ccf427b0c588849
+ENV AUTH_PROXY_HASH=0ca1fa78d55e2f2fdd9177fa526ae937855d7874
 
 # Add dependencies from the host
 # Note: these are listed individually so that if they change, they won't affect


rdig (dismissed):
Leaving the screenshots from the review here, since this is the PR where I tested this.

Yeah, basically everything is working as expected, I ran through the whole flow as you outlined it so all good.

Nice work and thank you for taking over this.

Screenshot from 2024-07-24 00-30-26
Screenshot from 2024-07-24 00-30-57
Screenshot from 2024-07-24 00-33-33
Screenshot from 2024-07-24 00-35-50
Screenshot from 2024-07-24 00-38-21
Screenshot from 2024-07-24 00-39-11
Screenshot from 2024-07-24 00-39-46

chmanie (commented):
For me this was always more of a theoretical issue than one that can actually be produced, so congrats for finding a way for testing this :1st_place_medal: 

I did not go through your steps, I'm just here to note that we should probably put this in draft until we have merged https://github.com/JoinColony/colonyCDappAuthProxy/pull/23.

That's because merging that PR will create a merge commit that should then probably be the actual commit hash being referenced here.

Thanks Dave for doing this!

chmanie (approved):
Great commit hash!

iamsamgibbs (approved):
Approving so you can get this merged as the auth-proxy change has already gone in!

PR #2737, opened 2024-07-18
Feat: Implement Multi-Sig extension was uninstalled UI

davecreaser:
## Description

- Implement the UI in the action sidebar which should be shown to users when they are accessing a multi sig action _after_ the multi sig extension has been uninstalled.
- Also implement this UI for Reputation Voting actions, since it wasn't there already.

[Link to designs here](https://www.figma.com/design/5V8pr7iMwXsT9L3VAZsmUt/Colony-v5?node-id=22857-221890&t=Eq8J9IE9YnZuNQ7r-4)

## Testing

* Install the multi sig extension
* Create a mint tokens multi sig action
* Uninstall the multi sig extension
* Go back to the mint token action, and check that you have a message in the sidebar explaining that the extension is uninstalled.
Screenshot 2024-07-18 at 16 23 05

* Repeat the above steps, but this time for Reputation Voting, not Multi Sig
Screenshot 2024-07-18 at 16 23 13

## Diffs

**New stuff** ✨

* New UninstalledMessage component, which is flexible and can be used for any extension.

**Changes** 🏗

* Ensure invalid motion statuses are displayed correctly.

Resolves #2712

PR #2759, opened 2024-07-23
fix: update list paddings, load all stakes regardless of colonies, update redirect to stakes

### LINKED ISSUE #2566: UserHub Stakes tab issues - padding, overflow, filtered and status pill issue
## Steps to reproduce

1. Ensure you have staked motions on your account.
2. Open up the UserHub Stakes tab.
3. Check for padding, overflow, and status pill issues.
4. Check that all claimable transactions are listed across all colonies.

## Expected behaviour

* Padding of items in the list should match the design.
* The items in the list should not overflow the container.
* The pills should show the correct status.
* The list of claimable motions should be visible across all colonies

## Actual behaviour

* Padding does not align with the design, and there
* The items in the list should not overflow the container.
* The pills should show the correct status.
* You only see staked actions that are in the current colony context.

image

image

Nortsova:
## Description

- [x] Padding of items in the list should match the design.
- [x] The items in the list should not overflow the container.
- [x] The pills should show the correct status.
- [x] The list of claimable motions should be visible across all colonies

## Testing

* Ensure that Reputation is installed

Step 1. Create actions with reputation decision method, **title should be long**
image
Step 2. Stake this motion
Step 3. Go to UserHub
image
Step 4. Create different motions on different stages (Claimable, Claimed, Finalizable, Pending)
image
Ensure that styles are correct here
Step 5. Go to another Colony (wayne) and open UserHub
image
Ensure that even tho Reputation extension is not installed, you can see stakes from all colonies
Step 6. Enable Reputation Extension for Wayne colony
Step 7. Create new Staking motion in Wayne colony
Step 8. Check UserHub to see all motions
image
Step 9. Click to motion from different colony and see that you will be redirected
Step 10. Uninstall Reputation Extension and go to UserHub (update: now Uninstalled became "Lost" status)
image
(Full uninstalled functionality done in #2737) 
Step 11. Create stakes, uninstall VotingReputation. Install it again. All not claimed stakes should have status "Lost".
image

Step 12. Create motions in different colonies and finalize them. You will see "Claimable" status
image
Step 13. Press the "Claim all" button and ensure that motions in both colonies claimed.
image
Step 14. Create motion and during pending and processing - open UserHub/Stakes tab 
image

It shouldn't blink. (Sam found this bug in this review - https://github.com/JoinColony/colonyCDapp/pull/2759#pullrequestreview-2201531690)

Step 15. Disconnect and connect wallet. Open UserHub - stakes should have correct statuses, not "Lost"

## Diffs

**New stuff** ✨

Bug fixed:
After Step 8 click on motion from other colonies. 
It was a bug with tx param, it was removed and the user was redirected to the colony page instead of motion page. Now it open the correct link

Resolves #2566


rdig:
Is this ready for review?

Nortsova:
> Is this ready for review?

No, it is in progress.

src/hooks/useNetworkMotionStatesAllColonies.ts
```diff
@@ -0,0 +1,189 @@
+import {
```
Nortsova, line 1: I would really appreciate for review this code, it kinda duplication for https://github.com/JoinColony/colonyCDapp/blob/master/src/hooks/useNetworkMotionStates.ts 
But I tried to do it for all user colonies, not only for the current one

iamsamgibbs (changes requested):
Nice start to this issue, it is definitely more complicated than it initially appeared! I think the estimate should definitely go up from 4 hours.

All the UI fixes look great, looking much tidier now!

Screenshot 2024-07-26 at 11 34 54

I did notice the "Claim all stakes" button does not appear on mobile - not sure if this is intended?

Screenshot 2024-07-26 at 12 19 58

There are a few issues I've noticed.

The stake amount should show using the native token for the colony it was created in. Currently it uses the token for the colony you have open.
 
Screenshot 2024-07-26 at 11 33 07

Switching to the correct colony is working nicely, great job with that!

https://github.com/user-attachments/assets/c458cca0-ef5c-4ea0-ba08-9ae995ba71ca

There is no loading spinner when you first open the userhub after a refresh:

https://github.com/user-attachments/assets/1fae1971-5e67-4ae5-aeb0-5ca88b1fdfc4

This is particularly noticeable on a slower connection - it also initially loads in with the wrong status:

https://github.com/user-attachments/assets/50b80ba8-9706-4669-a04b-a844817b6196

There is also this issue when opening the userhub straight after staking a motion. I think this only happens during the few seconds directly after staking a motion so this might not be a big deal and we could address it separately.

https://github.com/user-attachments/assets/c79b2757-1e7c-498d-84fc-7a2d3ce20152

After uninstalling the extension, all stakes are shown as "uninstalled".  I think for the stakes we know have already been claimed, we should still show those as "claimed" rather than setting it as "uninstalled". I personally think "uninstalled" should be shown as "lost" for the stakes that we know were unclaimed when the extension was uninstalled - but this might need checking with @arrenv 
 
Screenshot 2024-07-26 at 12 13 37

After re-installing the extension, new motions created with the new extension are seemingly using the same nativeId (or whatever is being called to check the status), as the old stakes (from the now uninstalled extension) are also being updated.

In this screenshot the bottom one is from the now uninstalled extension, the top one is from the newly installed extension.

Screenshot 2024-07-26 at 12 15 07

If you have unclaimed stakes across multiple colonies, then the "claim all stakes" button only works for the current colony you are in. It would be great to make this claim stakes across all colonies, or if this isn't possible it should be reworded "Claim all stakes in this colony" or something. The state also doesn't update until the userhub is closed and reopened.

https://github.com/user-attachments/assets/6d2f7c48-cb64-47f0-b926-b2d63c053364

Thank you for tackling this complicated issue though! It highlights some of the issues we may have when we make the userhub accessible outside of colonies which will be required as part of the crypto to fiat work.

src/graphql/queries/extensions.graphql
```diff
@@ -15,3 +15,39 @@ query GetExtensionInstallationsCount($id: ID!) {
     reputationWeighted
   }
 }
+
+query GetAllColoniesExtensions(
```
jakubcolony, line 19: Perhaps `GetJoinedColoniesExtensions` would better describe what it does?

src/graphql/queries/extensions.graphql
```diff
@@ -15,3 +15,39 @@ query GetExtensionInstallationsCount($id: ID!) {
...
+  $contributorAddress: ID!
+  $sortDirection: ModelSortDirection = ASC
+  $limit: Int = 100
+  $nextToken: String
+  $isWatching: Boolean
```
jakubcolony, line 24: It feels like this argument doesn't belong here anymore, should the query always set the filter to true instead?

src/hooks/useNetworkMotionStatesAllColonies.ts
```diff
@@ -0,0 +1,189 @@
...
+  return colonyExtensionsMap.filter(notNull);
+};
+
+const getVotingReputationAdress = (installedExtensions): string | undefined => {
+  const votingReputationExtension = installedExtensions.find(
+    (extension) => extension.extensionId === Extension.VotingReputation,
+  );
+
+  return votingReputationExtension?.address;
+};
```
jakubcolony, line 63: Is all this processing necessary if you only care about the hash and the address of each colony extensions?

src/hooks/useNetworkMotionStatesAllColonies.ts
jakubcolony: Good shout with a separate hook for all colonies. I think it can be simplified a bit further to roughly follow this flow:
- Get all installed extensions for joined colonies
- Select all voting reputation addresses (directly from the query data, you shouldn't need any mapping)
- Get network states for each of the motions
- Create a map between motion database IDs and states so that a flat map can be used

src/hooks/useJoinedColoniesWithExtensions.ts
```diff
@@ -0,0 +1,26 @@
+import { ADDRESS_ZERO } from '~constants/index.ts';
+import { useGetAllColoniesExtensionsQuery } from '~gql';
+import { notNull } from '~utils/arrays/index.ts';
+
+const sortByDate = <T extends { createdAt: string }>(a: T, b: T) =>
+  new Date(a.createdAt).getTime() - new Date(b.createdAt).getTime();
+
+const useJoinedColoniesWithExtensions = (userAddress?: string) => {
```
jakubcolony, line 8: This hook seems very specific to `useNetworkMotionStates`, I would either make it a helper or place it next to that hook to indicate it's not intended to be reusable

jakubcolony (changes requested):
Very nice effort with this pretty complex issue @Nortsova 

Reducing nesting of the mapping between colony addresses and motions and their states would help improve the overall readability. I think this could be achieved if you can instead create a map between database motion ID (it's unique across colonies and formatted as `chainId-votingRepExtnAddress_nativeMotionId`) and their corresponding states.

Please ensure you type the arguments of the functions you add as it reduces the risk of mistakes.

Nortsova:
I noticed an error that I can't manually repeat. It happen for this request:
```
const votingRepClient = VotingReputationFactory.connect(
     votingReputationAddress,
     ethersProvider as unknown as Provider,
 );
try {
 const motionState = await votingRepClient.getMotionState(nativeMotionId);
  statesMap.set(nativeMotionId, motionState);
 } catch (e) {
    console.log('error', e);
    statesMap.set(nativeMotionId, null);
 }
```

image

I also tried to do it with check `if fullWallet` but it didn't help

image

src/components/v5/common/ActionSidebar/ActionSidebar.tsx
```diff
@@ -102,7 +101,6 @@ const ActionSidebar: FC<PropsWithChildren<ActionSidebarProps>> = ({
   const isMobile = useMobile();
 
   useDisableBodyScroll(isActionSidebarOpen);
-  useRemoveTxParamOnClose();
```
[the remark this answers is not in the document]
Nortsova: thanks, this issue is fixed!

src/redux/types/actions/motion.ts
```diff
@@ -134,7 +134,11 @@ export type MotionActionTypes =
         userAddress: Address;
         colonyAddress: Address;
         extensionAddress: Address;
-        motionIds: string[];
+        motionIds: {
+          databaseMotionId: string;
+          colonyAddress: string;
+          extensionAddress: string;
+        }[];
```
iamsamgibbs, line 141: I wonder if there is a better name than `motionIds` now that this is an object

iamsamgibbs (changes requested):
Really great work on this @Nortsova ! 🎉 This definitely hasn't been an easy issue, so fantastic job on getting this to work so nicely!

Truncation looks good on both desktop and mobile:
 
Screenshot 2024-08-07 at 09 46 45

Screenshot 2024-08-07 at 09 47 06

The motion pills in the different states as viewed from another colony:

Screenshot 2024-08-07 at 09 54 23

Claiming all with stakes in different colonies works without a hitch. I also tested this from a completely new colony where the extension has never been installed before and was able to claim a stake from another colony via "claim all". 🎉 Nice work adjusting the saga to accommodate the different colony and extension addresses.

https://github.com/user-attachments/assets/ab125ae4-494b-4c90-991b-3f3c68846f40

Statuses show correctly after the extension has been uninstalled and reinstalled, and stakes open the correct transaction when clicked.

Screenshot 2024-08-07 at 09 58 59

I've only found two major issues. The first is if you have no voting reputation extension installed in a colony that has old motions, the stakes will fail to load:

https://github.com/user-attachments/assets/ac0231f9-91d1-432d-ae7f-deb92a0845bc

Once a new version of the extension is installed, they will load again. (This issue still happens if you switch to another colony where the extension is installed).

There is also this console warning which it would be great if you could look into and resolve. I'm not sure how this could affect other parts of the app if left as is.

Screenshot 2024-08-07 at 10 05 36

I'm guessing it has something to do with how `getContributorsByAddress` is being used to get all the colony extensions.

src/redux/types/actions/motion.ts
Nortsova: Good catch, @iamsamgibbs ! thanks!

src/hooks/useNetworkMotionStatesAllColonies.ts
```diff
@@ -0,0 +1,173 @@
...
+import { useEffect, useMemo, useState } from 'react';
+
+import { ADDRESS_ZERO, supportedExtensionsConfig } from '~constants/index.ts';
+import { useAppContext } from '~context/AppContext/AppContext.ts';
+import { useGetJoinedColoniesExtensionsQuery } from '~gql';
+import { notNull, notUndefined } from '~utils/arrays/index.ts';
+
+export type MotionStatesMap = Map<string, MotionState | null>;
+
+export type MotionStatesMapByColonies = Record<string, MotionStatesMap>;
```
jakubcolony, line 16: Looks unused

src/hooks/useNetworkMotionStatesAllColonies.ts
```diff
@@ -0,0 +1,173 @@
...
+
+export type RefetchMotionStates = (motionIdsToRefetch?: string[]) => void;
+
+export type MotionId = {
+  motionId: string;
+  colonyAddress: string;
+  databaseMotionId: string;
+};
```
jakubcolony, line 24: Can you improve the naming of this type?

src/hooks/useNetworkMotionStatesAllColonies.ts
```diff
@@ -0,0 +1,173 @@
...
+
+const getVotingReputationAddressByColony = (colony) => {
```
jakubcolony, line 26: Please type the argument of this function

src/hooks/useNetworkMotionStatesAllColonies.ts
```diff
@@ -0,0 +1,173 @@
...
+      };
+    }, {});
+  }, [joinedColoniesWithExtensions]);
+
+  const [loading, setLoading] = useState(false);
+
+  const [motionStatesMap, setMotionStatesMap] = useState<MotionStatesMap>(
+    new Map(),
+  );
+  useEffect(() => {
```
jakubcolony, line 96: Empty line between these two would improve read a bit better

src/components/common/Extensions/UserHub/partials/StakesTab/useStakesByFilterType.ts
```diff
@@ -35,23 +30,34 @@ export const useStakesByFilterType = () => {
     [data],
   );
 
-  const motionIds = useMemo(
+  const motionIdsMap = useMemo(
```
jakubcolony, line 33: This name should also be updated as the array represents motion stakes rather than motion IDs

src/hooks/useNetworkMotionStatesAllColonies.ts
```diff
@@ -0,0 +1,173 @@
...
+  return {
+    joinedColoniesWithExtensions,
+    loading,
+  };
+};
+
+/**
+ * Hook that accepts an array of motion IDs object and returns a map of motion IDs to their states
+ * Make sure to memoize the array of motion IDs to avoid infinite loops
+ */
```
jakubcolony, line 67: Please make sure to update the part about motion IDs

src/hooks/useNetworkMotionStatesAllColonies.ts
```diff
@@ -0,0 +1,173 @@
...
+const useNetworkMotionStatesAllColonies = (
+  motionIdsMap: MotionId[],
```
jakubcolony, line 69: Instead of a custom type, would passing entire `ColonyMotion` object from `userStake.action.motionData` make things simpler?

src/hooks/useNetworkMotionStatesAllColonies.ts
```diff
@@ -0,0 +1,173 @@
...
+    return joinedColoniesWithExtensions.reduce((prev, colony) => {
+      if (!colony) {
+        return prev;
+      }
+      return {
+        ...prev,
+        [colony.colonyAddress]: getVotingReputationAddressByColony(colony),
+      };
+    }, {});
+  }, [joinedColoniesWithExtensions]);
```
jakubcolony, line 89: Maybe moving the `useMemo` block to the `useJoinedColoniesWithExtensions` hook would make the body of this hook a bit lighter?

src/hooks/useNetworkMotionStatesAllColonies.ts
```diff
@@ -0,0 +1,173 @@
...
+      !signer
+    ) {
+      return;
+    }
+
+    const newMotionIds = motionIdsMap.filter((motion) => {
+      return !motionStatesMap.has(motion.databaseMotionId);
+    });
+    const deletedMotionIds = Array.from(motionStatesMap.keys()).filter(
+      (nativeMotionKey) =>
```
jakubcolony, line 117: Please update the naming to indicate if it is the native ID or database ID. It should be database ID as only those are unique between colonies

src/hooks/useNetworkMotionStatesAllColonies.ts
```diff
@@ -0,0 +1,173 @@
...
+    if (!newMotionIds.length && !deletedMotionIds.length) {
+      return;
+    }
+
+    const fetchMotionStates = async () => {
+      setLoading(true);
+      const statesMap = new Map(motionStatesMap);
+
+      await Promise.all(
+        newMotionIds.map(async (nativeMotion) => {
```
jakubcolony, line 130: I'm not sure about `nativeMotion`, why such name?

src/hooks/useNetworkMotionStatesAllColonies.ts
```diff
@@ -0,0 +1,173 @@
...
+            statesMap.set(motionStateKey, motionState);
+          } catch (e) {
+            console.error(e);
+            statesMap.set(motionStateKey, null);
+          }
+        }),
+      );
+      deletedMotionIds.forEach((nativeMotionId) =>
+        statesMap.delete(nativeMotionId),
+      );
```
jakubcolony, line 157: Please update naming here as well

src/components/common/Extensions/UserHub/partials/StakesTab/partials/StakeItem.tsx
```diff
@@ -1,33 +1,68 @@
...
+  const { colonyName: colonyNameUrl } = useParams();
+
+  const { data: stakeColonyData, loading: stakeColonyLoading } =
+    useGetFullColonyByAddressQuery({
+      variables: {
+        address: stake.action?.colonyAddress || '',
+      },
+      fetchPolicy: 'network-only',
+      nextFetchPolicy: 'cache-first',
+    });
```
jakubcolony, line 37: It's an expensive query to be making for each stake in the list, I think the best thing to do here would be to modify the `UserStake` model to include a reference to `colony` field. You can then extract just the native token and its decimals in one go.

You could go a step further if you wish and fetch the extensions in the UserStake fragment as well. The benefit of such approach is that it would also work for stakes in colonies that the user has subsequently left.

jakubcolony (changes requested):
Thank you for addressing my previous feedback @Nortsova, feels like it's not very far off.

I left a couple of comments, but overall you've done a great job so far. Nice work 💪

iamsamgibbs (changes requested):
Thanks for fixing the issues I raised! I've not seen the console warning and the code fixes for that make sense.

I've also tested the stake loading when there is no extension installed which is working now!

https://github.com/user-attachments/assets/7767e839-4913-4f31-b7af-1b8ac024fae0

Once you've fixed up the comments Jakub has left I'll drop an approval on this.

Unrelated to this PR, but the activity feed will need a similar fix to what you have done here as it tries to load the old motion data through the new extension:

https://github.com/user-attachments/assets/5a363a97-32cf-4344-8dbe-9554ecc9b6ec

It also doesn't support the new "Lost" pill and will just show "Unknown" so we'll need to update this at a later point.

src/graphql/fragments/colony.graphql
```diff
@@ -199,3 +199,25 @@ fragment JoinedColony on Colony {
...
+  extensions(filter: { isDeleted: { eq: false } }) {
+    items {
+      address: id
+      version
+      hash
+      isDeleted
+      isDeprecated
+      isInitialized
+    }
+  }
```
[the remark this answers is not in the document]
Nortsova: discussed with @jakubcolony today and decided left as it is

Nortsova:
Thank you @iamsamgibbs! This is a really important note. I created a ticket for Activity Feed - https://github.com/JoinColony/colonyCDapp/issues/2907

iamsamgibbs (approved):
Nice work on this! The code refactor makes sense. I've retested everything and all is still working! 👍 

Screenshot 2024-08-09 at 13 49 20
Screenshot 2024-08-09 at 13 53 22

jakubcolony (approved):
A lot of effort went into this PR so kudos to you @Nortsova.

I can see my stakes from across all colonies and navigate between them:

https://github.com/user-attachments/assets/a935398a-44b0-43fe-b82c-d4465996af68

After uninstalling voting rep, the status shows as Lost:
image

Finalizable and claimable statuses also work fine:
image

image

I left a couple of comments but they're nothing major, so thumbs up from me 👍 Very nice work 💯

rdig (dismissed):
Tested and everything works as expected, at least the part that you fixed in this PR. Very good work Anastasiia! :100: 

Screenshot from 2024-08-14 00-51-02
Screenshot from 2024-08-14 00-55-41
Screenshot from 2024-08-14 00-56-11
Screenshot from 2024-08-14 00-56-13

I did found something however, but something that's not really a issue related to this PR.

While a very long (multiple words) title shows up correctly, a title that contains a non-breakable word (no spaces) will break both the actions list UI as well as the action sidebar

Screenshot from 2024-08-14 00-50-24
Screenshot from 2024-08-14 00-50-41

I do want to note however, that it shows correctly in the Userhub where you fixed it in this PR

Screenshot from 2024-08-14 00-51-02

jakubcolony (approved):

mmioana (commented):
Wow, really impressive work @Nortsova 🥇 Thank you for tackling this issue 🤩 

Went through the testing steps and can confirm the required changes are in place
Screenshot 2024-08-14 at 10 15 55
Screenshot 2024-08-14 at 10 16 10
Screenshot 2024-08-14 at 10 18 50
Screenshot 2024-08-14 at 10 22 25
Screenshot 2024-08-14 at 10 23 24
Screenshot 2024-08-14 at 10 24 35
Screenshot 2024-08-14 at 10 25 08
Screenshot 2024-08-14 at 10 27 05

https://github.com/user-attachments/assets/90b9682e-bb02-4b1a-bb7f-480bae979401

Screenshot 2024-08-14 at 10 28 22
Screenshot 2024-08-14 at 10 28 34
Screenshot 2024-08-14 at 10 29 14
Screenshot 2024-08-14 at 10 29 22
Screenshot 2024-08-14 at 10 30 34

https://github.com/user-attachments/assets/7109b58d-cf6e-4db3-87f7-4b9f6b2025cd

However, I noticed claiming all stakes was not possible from the User Hub 😢 
Can you please check?

https://github.com/user-attachments/assets/76d6001d-b3f6-4bdb-9c48-716b82ae2341

Screenshot 2024-08-14 at 10 36 03

rdig (approved):
Working as expected after your changes. Re-tested and everything is there. Nicely done!

Screenshot from 2024-08-14 17-52-38
Screenshot from 2024-08-14 17-53-45
Screenshot from 2024-08-14 17-54-42
Screenshot from 2024-08-14 17-54-49

PR #2772, opened 2024-07-25
fix: missing translation for UserHub

### LINKED ISSUE #2564: Missing translation and metadata in UserHub Stakes tab
## Steps to reproduce

1. Create a Payment builder action.
2. Ensuring you have Reputation decision method enabled.
3. Fund the payment using the Reputation decision method.
4. Check the transaction in the Transactions tab of the UserHub.

## Expected behaviour

* The title should be transaction's custom title.
* The description should be the custom action title with a fallback to the action metadata.
* All details should be correct.
* It should either:
          * Link to the relevant Advanced payment, or
          * Should link to the "linked" motion view as of the design here - https://www.figma.com/design/5V8pr7iMwXsT9L3VAZsmUt/Colony-v5?node-id=24280-198587&t=21FkwrMm6ANcL8LC-4, or
          * Should not be linked and clickable.

## Actual behaviour

* You see a missing translation text.
* You see a generic metadata link.
* You can click on the transaction.

image

Nortsova:
## Description

| Before | After | 
|----------|----------|
| image  |  image
 |

## Testing
* Enable Reputation extension.

Step 1. Create Advanced Payment with Permissions decision method.
Step 2. Fund the payment using the Reputation decision method.
Step 3. Check the transaction in the Transactions tab of the UserHub. (check title and description to be correct)
image

Step 4. Create another simple motion (mint tokens etc) with Reputation and stake them
Step 5. Open UserHub and test all items to be clickable and open correct link

Resolves #2564


src/graphql/fragments/expenditures.graphql
```diff
@@ -15,6 +15,14 @@ fragment Expenditure on Expenditure {
       ...ExpenditureStage
     }
   }
+  actions(filter: { type: { eq: CREATE_EXPENDITURE } }) {
```
bassgeta, line 18: maybe to keep consistent with how other actions are fetched we can rename this field to `creatingActions`?

bassgeta (approved):
Nice little fix, tested it out and I can see the transaction in user hub
image
Clicking on it also redirects me to it:
image

Just dropped a minor comment, feel free to address it or not :)

iamsamgibbs (approved):
Nice work! Tested and working as expected. Both stakes open the correct transactions.

Screenshot 2024-07-29 at 10 56 55

jakubcolony (changes requested):
Nice work @Nortsova! 

While reviewing, I spotted an apparent bug with stake statuses which I fixed in #2803.

All looks good in your PR, the only thing I noticed is this requirement that seems to got missed:
> * Should link to the "linked" motion view as of the design here - https://www.figma.com/design/5V8pr7iMwXsT9L3VAZsmUt/Colony-v5?node-id=24280-198587&t=21FkwrMm6ANcL8LC-4, or

Clicking on the fund expenditure motion stake does not open the motion view, just the relevant payment:

https://github.com/user-attachments/assets/7f5976dd-a10f-4c4f-ac09-07a640b88a1e

arrenv:
That is a great pickup and idea @jakubcolony, was not how I was thinking it needed to be implemented, and would have expected it to just go to the current active step. So, your suggestion allows the user to go to the specific motion and claim their staked funds, which makes sense.

In saying that, if it is a lot of work to make it work as you have described, then I think it should be fine to leave it as it currently is. The user is still able to claim from their Userhub. @Nortsova may be able to comment on how much work could be involved.

jakubcolony (approved):
Thanks for clarification @arrenv. 

It is a more substantial amount of work to make the stake open the relevant expenditure motion, so I think it makes sense to keep it as a separate issue.

Regarding this PR, great work @Nortsova, LGTM 👍

Issue #2718, opened 2024-07-17
[ADV.PAYMENTS][STREAMING PAYMENT] As a Colony member, I can create STREAMING PAYMENT action and complete the entire STREAMING PAYMENT FLOW


AdrianDudko:
Teamwork task: [[ADV.PAYMENTS][STREAMING PAYMENT] As a Colony member, I can create STREAMING PAYMENT action and complete the entire STREAMING PAYMENT FLOW](https://tw.auroracreation.com/#/tasks/15967014)

## Related sub-issues/PR's

- [ ] Create an action (form) 
- [ ] Preview action 
- [ ] Streaming Payment Flow/Stepper
- [ ] Cancel
- [ ] Edit
- [ ] Multiple Decision Method Support

## Description
Provides the ability to Create a Streaming payment action and complete it.
Notion spec https://www.notion.so/colony/Streaming-Payments-Action-ab883324ab694de4b15d8db3e3c6d153
Handover doc https://www.notion.so/colony/Streaming-Payments-handover-92b9fad4b7c045e7bbbd3f4e494bd248

## Acceptance Criteria

- **General requirements**
    - Should be in alignment with desktop/tablet/mobile design from Designs.
    - Should use existing components and styles where possible for consistency.
    - All times and dates should be the displayed in users time zone (best guess).
        - As a fallback, it should show the time zone as UTC +0.
    - Streaming payment action type should not be visible if the extension is not installed.

- **Create a Streaming payment**
    - User can select the “Streaming Payment” option from the “Action type” options.
        - This should update the action creator with the relevant fields.
    - Fields
        - Stream from:
            - The team to stream the funds from.
            - Shows the team dropdown.
        - Recipient:
            - User is able to select or enter in the recipient for the stream.
            - This should show the full members list dropdown, with the ability to search and enter an address.
        - Starts:
            - User as the following options.
                - User can select “Start immediately”.
                - Users can use a date picker to select a start date and time for the stream.
            - Popup date picker is used for input.
                - Date picker includes the following:
                    - Show today's date as default.
                    - Quick month navigation arrows.
                    - Click on month and year to change in additional popup.
                    - Full month daily view.
[...]

PR #2780, opened 2024-07-26
feat: datepicker with time

joanna-pagepro:
## Description

Date and time picker component that will be used in the Advanced Payments form

Design:
https://www.figma.com/design/l1dOM5qiQYwF0ElvKDqqjg/Design-System---Colony-v3?node-id=1782-16199&t=EKLjparNxhFXSGKg-4

## Testing

The component is added temporarily to the `Simple Payment` action and will be removed from there before merging to the master.

* Step 1. Open the action sidebar and select the `Simple payment` action type
* Step 2. Check the last two fields (Starts and Ends)

## Diffs

**New stuff** ✨

* `DatepickerWithTime` component
* `TimeRow` component that will be used in `Action Sidebar`
* `TimeRowField` - `TimeRow` partial

**Changes** 🏗

* 

**Deletions** ⚰️

* 

## TODO


melyndav (changes requested):
Hey @joanna-pagepro really nice start on this component dev, I know it's a tricky one. so appreciate the work so far! A couple points of feedback:

1. When I trigger the custom date and time in the start option, the time input doesn't show in the default state and then pops in randomly once it's open.

image

Same with the end condition:

image
 

---

2. Is it possible to reduce the width of the component overall on desktop to align more with the Figma component? This will reduce the extra space to the left and the right of the date section. This is for both start and end variations. 

Figma dimensions:

image

---
 
 3. The apply and cancel CTA's should only show on mobile. On desktop, once a time is selected, the dropdown should close. This is for both modals. 
 
image

---

4. When I select the time input, the time dropdown isn't triggered in both start and end variations. 

image

---

5. The hover on the dates should be gray-100 to match the Figma component design. 

image

---

6. On mobile, can the whole dropdown be shown on the screen allowing the user to scroll down using tactile input? Currently, it has a fixed height with an internal scroll which makes it unusable. 

image


---

7. On mobile (from what I can see) the calendar section of the dropdown isn't full width of the container. You can see the extra padding on the right side of the dates. 

image

joanna-pagepro:
@arrenv @melyndav I used input type time in this datepicker, I've checked if it's supported on every modern browser - https://caniuse.com/mdn-html_elements_input_type_time and it is, but it doesn't work the same way. For example on Chrome you can see a dropdown and on Firefox, there's no dropdown and you can only type the time. Should we leave this input in this case? We can also use time select from https://reactdatepicker.com/#example-select-time-only 
image

joanna-pagepro:
@melyndav datepicker updated! Can you test it again?

melyndav (changes requested):
Hey @joanna-pagepro thanks for your continued work on this on. Some updates required:

1. Can we reduce the width on desktop/tablet to match the width of the design system component? Currently, it's wide and there is extra white space either side of the date area. 

---

2. When a user inputs a date, can we format the time as the values are entered?

Currently it looks like this when I type a time in and doesn't correct the formatting as I type, and I have to click out to get it to show correctly

image

---
 
 3. If I select today's date, the selected date stays on the previous date and doesn't change. Is it possible to have this update to avoid confusion?
 
 
image
 
 
 I also clicked out of the dropdown after selecting today's date (Aug 1) and it stayed as the previous date selected.
 
 
image

---

4. When I select a date, the dropdown closes and selects a time for me (12:00AM). It closes without me being able to select or input a time:

Export-1722524502334

---

5. On mobile, can the dates/days go full width of the container width to match the Figma design more closely? Currently on mobile there is extra space eitherside:

image

---

6. On both desktop and mobile, can the pre-selected time in the dropdown be styled with the medium font weight to match Figma?

image
 
 Figma with the weighted text:
 
 
image

joanna-pagepro:
@melyndav 
1. It's 328px, just like on a design
2. It's a different element. The one from the example is input type time which we got rid off because it wasn't working the same on every browser and the one that is used currently is also from this lib, but it's input type text and that's how it was implemented in the lib
3. I need more info about how it should work because there are a few cases like selecting the date, selecting time, selecting today's day, etc. and I don't know what is expected here. We don't have the `Cancel` and `Apply` buttons here, so I don't know when the date should be selected and datepicker closed and when you still want it to be opened and able to edit the date.
4. Same as no. 3
5. On mobile - which screen widths?
6. updated
7. I don't see a way to add a dropdown for the month using the current design and this is how it works on a different datepicker that we use for filtering activity table
8. updated

melyndav (changes requested):
Hey @joanna-pagepro! Thanks for the continued updates!

Here are my answers to your previous comment. And below this has some new updates. ☺️

1. Great, thanks for updating.

2. Okay, noted on the limitations. 

3. It should interact like the native react component. So I pick the date and then the time and the dropdown closes. 

4. As above.

5. This is at 425px - you can still see the big white spaces to the left and right of the calendar dates. 

image

6. Thanks.

7. Noted.

8. Noted.


### New comments 

A couple of other pick-ups:

1. In the default state when the dropdown is opened, today's date is highlighted but not selected. If I enter a time thinking the date is already selected and close the dropdown, nothing is entered into the field. 

Can we either a) remove this pre-selected date or b) pick up today's date as being pre-selected.

image

---

2. If I have a date selected and have added a time in, and I go back into the dropdown to change the time, once I've changed the time with a custom input value, and close out the modal, the time doesn't update. 

The time should update here since a value has been changed.

Export-1723031106681

joanna-pagepro:
@melyndav can you check if everything is ok now?

melyndav (approved):
Hey @joanna-pagepro really great work on this. I appreciate all your updates and communication throughout this work! Approved from my end 💯

arrenv (changes requested):
@joanna-pagepro Thank you for working on this. Just a couple of things I noticed.

1. Can the placeholder for "Starts" field instead be, "Select start".

2. At the moment it works so that as you make updates to dates and click out, the field is not updated. I was just testing the `Locale with time` and this seems like it works more naturally, as in no matter whether you change the date or time, it updates in the field, you can then click out, or it auto closes after clicking the time.

So, this is how it should update:
- If you click a date, it updates in the field. Dropdown stays open. If you click out, the date set should stay as the last set date.
- If you click a time and no date was pre-selected, it should close the dropdown with today's date.
- If you click a date and then time, it should close the box after clicking time. With the set date and time.
- If you have previously selected a date, then you open the dropdown again and select the time, it should close the dropdown with the new time and the previously selected date.

   [locale-time.webm](https://github.com/user-attachments/assets/45315e35-bb54-4f9a-85b7-bfe8ef695e8e)

joanna-pagepro:
@arrenv can you check if everything works fine now?

arrenv (approved):
@joanna-pagepro Thank you for the updates, that works much better now!

I will approve, although it was just the placeholder for the "Starts" field that instead needed to be "Select start".

joanna-pagepro:
@arrenv I will change it in my next PR with the form :)

PR #2789, opened 2024-07-26
Fix: Prevent payment actions with locked tokens

### LINKED ISSUE #2299: Prevent creating a payment with a locked token
## Steps to reproduce

1. Create two colonies with created and locked tokens.
2. Pay from one colony to the other.
3. Now in the receiving colony, create a payment with the locked token from the other colony.
4. The payment will fail, but with no reason shown in the UI.

## Expected behaviour

* We should prevent allowing payment with a locked token. Ideally, as this will also allow it to block motions and expenditures being created for payments with these tokens.
(or)
* At the very least, we should show a human readable error for when the transaction fails. Such as: "This token is locked, payments cannot be made."

## Actual behaviour

* You receive a transaction error when trying to make a payment with a locked token from another colony.

image

davecreaser:
## Description

It is important that users cannot make a payment with a locked token from anywhere that is not the colony which created it.

For example:
- User creates new colony `Davecol` with a new token `DaveCoin`. This coin is locked by default, until a user with the correct permissions makes an 'unlock token' action.
- Whilst the coin is locked, you should still be able to make payments with this coin wherever you want, as long as you are making that payment from within the colony `Davecol`.
- Lets say the user makes another colony `NewDavecol`, and this colony uses the already existing `DaveCoin` as it's token.
- It is completely fine to transfer funds from `Davecol` to `NewDaveCol`, even when the token is locked.
- It should _not_ be possible however, to transfer funds or make any type of payment using `DaveCoin` from within `NewDaveCol`, that is until a user in `DaveCol` unlocks the token.

All of this is currently working fine on the contracts side, and as it stands, if you try to make one of these payments, the UI will let you, but the transaction will fail. We want to stop users from even trying to make these types of payments, since it's not something they will be able to do. We also want to be helpful to the user and inform them why they can't do it.

This PR solves this by introducing a new graphql query and hook which can combine to give access to token locked states, so that the action forms can include this behaviour in their validation. Unfortunately, we don't store locked states on tokens directly in the db, and making that change (which was something I considered and explored) would be quite a large change requiring changes to schemas, ui, block-ingestor, and a migration. This approach keeps the current data structure intact and still manages to solve the problem in an at least slightly elegant way, in my opinion.

## Testing

* After running the create data script, you should be in a state where (as Leela) you're in Planex and Wayne colonies. Planex uses CREDS (locked) and Wayne uses GGG (locked). There should also be incoming funds from Wayne to Planex.
* Inside Planex, go to the incoming funds page, and claim the GGG funds which were sent from Wayne to Planex during the create data script process.
Screenshot 2024-07-26 at 16 21 50

* Still inside Planex, open the create new action form, and try to make a simple payment. Fill out the form as usual, but be sure to select GGG as the token you are paying with.
* You should not be able to submit the form, and should see an error informing you that the token is locked so you cannot make a payment.
Screenshot 2024-07-26 at 16 23 04

* Go to Wayne and try again to create a simple payment using GGG. It should work totally fine, since even though it is a locked token, GGG was created by Wayne.
Screenshot 2024-07-26 at 16 24 14

**(See further testing steps at this stage if you want to be thorough, or if you want to just do a quick test move onto the next step)**
* Still inside Wayne, create an unlock token action, and complete it.
Screenshot 2024-07-26 at 16 31 49

* Refresh the page, and go back to Planex. Try again to create a simple payment using GGG as before, only this time it should go through fine!
Screenshot 2024-07-26 at 16 33 48
Screenshot 2024-07-26 at 16 32 55

**Further testing**

If you'd like to test more in-depth (thank you!) then before unlocking the token there's a couple of other things you can try:
* Try making a transfer funds action in Planex using GGG (instead of simple payment). It should fail in the same way.
Screenshot 2024-07-26 at 16 25 06

* Try making an advanced payment action in Planex using GGG (instead of simple payment). It should fail in the same way.
Screenshot 2024-07-26 at 16 25 43

* Try making a new colony (run `node scripts/create-colony-url.js` to get a create url) and select existing token during the creation steps. Put in the address for the Gotham Guilder (GGG) - you can find this address on the dashboard of Wayne. Once the colony is created, transfer some funds to it from Wayne, then try all the payment types mentioned above inside that colony, using GGG. They should all fail, since this colony did not create the token, and it's still locked.
Screenshot 2024-07-26 at 16 26 20
Screenshot 2024-07-26 at 16 27 25
Screenshot 2024-07-26 at 16 30 01
Screenshot 2024-07-26 at 16 30 26
Screenshot 2024-07-26 at 16 31 02

## Diffs

**New stuff** ✨

* New graphql query `GetColonyTokensWithLockedStates`
* New hook `useTokenLockStates`

**Changes** 🏗

* Update the validation of all forms which create payment actions

Resolves #2299

### DIFF
--- a/src/graphql/queries/colony.graphql
+++ b/src/graphql/queries/colony.graphql
@@ -86,28 +86,12 @@ query GetColonyMemberInvite($id: ID!) {
   }
 }
 
-query GetColonyTokensWithLockedStates($name: String!) {
-  getColonyByName(name: $name) {
+query GetColonyTokenLockedState($nativeTokenId: ID!) {
+  getColoniesByNativeTokenId(nativeTokenId: $nativeTokenId, limit: 1) {
     items {
-      tokens {
-        items {
-          token {
-            colonies {
-              items {
-                colony {
-                  status {
-                    nativeToken {
-                      unlocked
-                    }
-                  }
-                  nativeTokenId
-                  name
-                }
-              }
-            }
-            name
-            id
-          }
+      status {
+        nativeToken {
+          unlocked
         }
       }
     }
--- a/src/graphql/queries/colony.graphql
+++ b/src/graphql/queries/colony.graphql
@@ -86,6 +86,34 @@ query GetColonyMemberInvite($id: ID!) {
   }
 }
 
+query GetColonyTokensWithLockedStates($name: String!) {
+  getColonyByName(name: $name) {
+    items {
+      tokens {
+        items {
+          token {
+            colonies {
+              items {
+                colony {
+                  status {
+                    nativeToken {
+                      unlocked
+                    }
+                  }
+                  nativeTokenId
+                  name
+                }
+              }
+            }
+            name
+            id
+          }
+        }
+      }
+    }
+  }
+}
+
 subscription OnUpdateColony {
   onUpdateColony {
     lastUpdatedContributorsWithReputation
--- a/src/hooks/useTokenLockStates.ts
+++ b/src/hooks/useTokenLockStates.ts
@@ -1,47 +1,51 @@
-import { useMemo } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 
 import { useColonyContext } from '~context/ColonyContext/ColonyContext.ts';
-import { type Token, useGetColonyTokensWithLockedStatesQuery } from '~gql';
+import { useGetColonyTokenLockedStateLazyQuery } from '~gql';
 
 const useTokenLockStates = (): Record<string, boolean> => {
   const {
-    colony: { name },
+    colony: { tokens },
   } = useColonyContext();
-  const { data } = useGetColonyTokensWithLockedStatesQuery({
-    variables: { name },
-  });
+  const [getColonyTokenLockedState] = useGetColonyTokenLockedStateLazyQuery();
+  const [tokenLockStatesMap, setTokenLockStatesMap] = useState({});
 
-  const tokenStatusMap = useMemo(() => {
-    if (!data) return {};
+  const fetchTokenStates = useCallback(async () => {
+    if (!tokens) {
+      return;
+    }
 
-    const tokens = data?.getColonyByName?.items?.[0]?.tokens?.items.flatMap(
-      (token) => token?.token,
+    const newTokenLockStatesMap = {};
+
+    const tokenStates = await Promise.all(
+      tokens.items.map(async (t) => {
+        if (!t) return undefined;
+
+        const { data } = await getColonyTokenLockedState({
+          variables: { nativeTokenId: t.token.tokenAddress },
+        });
+
+        const lockState =
+          data?.getColoniesByNativeTokenId?.items[0]?.status?.nativeToken
+            ?.unlocked;
+
+        return lockState;
+      }),
     );
 
-    return (tokens || []).reduce(
-      (map: Record<string, boolean>, token: Token) => {
-        if (!token) {
-          return map;
-        }
+    tokens.items.forEach((t, i) => {
+      if (!t || tokenStates[i] === undefined || tokenStates[i] === null) return;
+      newTokenLockStatesMap[t.token.tokenAddress] = tokenStates[i];
+    });
 
-        const colony = token.colonies?.items.find(
-          (colonyTokens) => colonyTokens?.colony.nativeTokenId === token.id,
-        );
+    setTokenLockStatesMap(newTokenLockStatesMap);
+  }, [getColonyTokenLockedState, tokens]);
 
-        if (!colony) {
-          return map;
-        }
+  useEffect(() => {
+    fetchTokenStates();
+  }, [fetchTokenStates]);
 
-        return {
-          ...map,
-          [token.id]: !!colony.colony.status?.nativeToken?.unlocked,
-        };
-      },
-      {},
-    );
-  }, [data]);
-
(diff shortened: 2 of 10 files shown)


src/hooks/useTokenLockStates.ts
```diff
@@ -0,0 +1,47 @@
...
+          (colonyTokens) => colonyTokens?.colony.nativeTokenId === token.id,
+        );
+
+        if (!colony) {
+          return map;
+        }
+
+        return {
+          ...map,
+          [token.id]: !!colony.colony.status?.nativeToken?.unlocked,
```
mmioana, line 37: Would it make sense to just return
```
{
          ...map,
          [token.id]: !!colony?.colony.status?.nativeToken?.unlocked,
}
```
?

mmioana (approved):
Awesome work on this one @davecreaser 🎉 

Tested all the use-cases and can confirm it works as expected
Screenshot 2024-07-30 at 09 21 29
Screenshot 2024-07-30 at 09 21 44
Screenshot 2024-07-30 at 09 22 35
Screenshot 2024-07-30 at 09 23 57
Screenshot 2024-07-30 at 09 24 44
Screenshot 2024-07-30 at 09 25 36
Screenshot 2024-07-30 at 09 37 39
Screenshot 2024-07-30 at 09 39 02
Screenshot 2024-07-30 at 09 39 53
Screenshot 2024-07-30 at 09 41 05
Screenshot 2024-07-30 at 09 41 21
Screenshot 2024-07-30 at 09 41 56

src/graphql/queries/colony.graphql
```diff
@@ -85,3 +85,31 @@ query GetColonyMemberInvite($id: ID!) {
     invitesRemaining
   }
 }
+
+query GetColonyTokensWithLockedStates($name: String!) {
+  getColonyByName(name: $name) {
```
bassgeta, line 90: we really do hide this data under 3 layers of nested `items`, good job getting this data.
I can only think of getting one level of nesting out is if you call `getColony` with its address instead of getting it by name, this will rid us of the first `items`

src/hooks/useTokenLockStates.ts
```diff
@@ -0,0 +1,47 @@
+import { useMemo } from 'react';
+
+import { useColonyContext } from '~context/ColonyContext/ColonyContext.ts';
+import { type Token, useGetColonyTokensWithLockedStatesQuery } from '~gql';
+
+const useTokenLockStates = (): Record<string, boolean> => {
+  const {
+    colony: { name },
+  } = useColonyContext();
+  const { data } = useGetColonyTokensWithLockedStatesQuery({
```
bassgeta, line 10: as discussed on Discord, it may be a bit more granular and performant to loop over all the tokens of a given colony, call `getColoniesByNativeTokenId` for each token in a colony and smash that into a state variable.

src/utils/tokens.ts
```diff
@@ -129,3 +130,18 @@ export const getNativeTokenByChainId = (chainId: string): Token => {
     ...selectedNetwork.nativeToken,
   };
 };
+
+export const shouldPreventPaymentsWithTokenInColony = (
+  tokenAddress: string,
+  colony: ColonyFragment,
+  tokenStatesMap: Record<string, boolean>,
```
bassgeta, line 137: I don't wanna be too nitpicky, but I think `tokenLockedStatesMap` would be a bit more informative.
Or something along the lines of how you named the hook `useTokenLockStates` :D

bassgeta (changes requested):
Overall, good changes, I love the hook and the helper which can just be reused everywhere, modular and simple :ok_hand: 
image
Can confirm the changes work, unlocking it also enables payments, so good job on untangling this tricky bug.

I've left 2 comments regarding fetching of the token's status, it could be more performant and readable to use `getColoniesByNativeTokenId` in the hook, and just go through all the tokens in a colony.
We can try it out and see how it reads/works, if it's gonna be bad, we can always use the query you wrote

src/hooks/useTokenLockStates.ts
davecreaser: This is the way!

davecreaser:
Thanks @bassgeta - it's ready for review again. Good thing you pointed that out, it's muuuuch cleaner now.

bassgeta (dismissed):
Nice going, code looks very readable now and stuff just works, what more can one ask for? 
Can't create a payment with a locked token
image
Then a payment in Wayne with GGG worked
image
And a payment in planex with GGG works too
image

nice fix, let's get it merged!

mmioana (dismissed):
Works as expected, awesome job @davecreaser 🎉 

Screenshot 2024-08-01 at 12 40 33
Screenshot 2024-08-01 at 12 42 15
Screenshot 2024-08-01 at 12 43 09
Screenshot 2024-08-01 at 12 44 02
Screenshot 2024-08-01 at 12 45 28

jakubcolony (approved):
Very nice work Dave, thanks for refactoring the query!

Everything works as expected:

Before unlocking:
- Simple payment
image

- Advanced payment
image

- Transfer funds
image

Can confirm it doesn't prevent those actions when using colony's own native token.

After unlocking:
- Simple payment
image

- Advanced payment
image

- Transfer funds
image

bassgeta (approved):
Reapproving after the rebase so we can get this badboi merged in :sunglasses:

rdig (approved):
Re-approving because of the rebase

PR #2794, opened 2024-07-29
Crypto-to-fiat: Resolve edge cases with liquidation addresses

### LINKED ISSUE #2793: Crypto-to-fiat: Resolve edge cases with liquidation addresses
## Specification

This issue is to resolve a few edge cases I came across with liquidation addresses.

### Multiple liquidation addresses

There will be situations where a user has multiple liquidation addresses and the app needs to be able to cope with that.

When updating bank account from EUR to USD, a new liquidation address has to be created. At the same time, Bridge API does not provide a way to delete the previous liquidation address.

The current implementation which picks the first liquidation address of a user does not account for this.

### Ensuring liquidation address exists

`checkKyc` handler ensures at least one liquidation address always exists. However, it doesn't check if the liquidation addresse is linked to the "active" bank account. The check needs to be improved.

- [x] New liquidation address is created when changing bank details from EUR > USD or vice-versa
- [x] Existing liquidation address is updated to be linked to the new bank account when updating details with the same currency
- [x] The `checkKyc` handler ensures there's a liquidation address for the current bank account
- [x] The liquidation address swapping logic selects the correct liquidation address based on the current bank account

jakubcolony:
## Description

See issue #2793 for details.

This PR resolves a number of potential issues with liquidation addresses:

1. Check KYC handler was checking for existence of any liquidation address, instead of one linked to the user's bank account. If it doesn't exist, it will create one.
2. When updating bank account details, a new bank account is created
   1. For accounts with the same currency, the existing liquidation address needed to be updated with the new account ID.
   2. For accounts with different currencies, a new liquidation address should be created.
3. As there can be multiple liquidation addresses that can change at any point (primarily when user updates their bank details), we cannot rely on the ones stored in our DB. For that, a new query was added to get liquidation address for a given user.

It also removes the `bridgeXYZQuery` lambda, as handling both queries and mutations in a single lambda allows for less code duplication and easier deployments.

**Note:** In this PR I'm also testing the possibility of splitting amplify schema into multiple files in hope to make it more manageable, considering the existing schema is already nearly 4k lines long.

## Testing

Restart your dev env - otherwise the separate schemas will not get picked up.

Issue 1:

1. Make sure you set the Posthog and Persona env variables and run `npm run prepare`. 
2. Start KYC flow (you don't need to finish it though).
3. Add USD bank account details (EUR accounts can't be tested in sandbox). You will also need to use a US address.

On next refetch the `CheckKYCStatus` query should return a liquidation address.

Issue 2:

1. Update bank details to another USD account.
2. Check that the liquidation address is still the same.

In production, if you updated bank details to EUR account, new liquidation address should be created. In sandbox however we can't really test subpoint 2). 

Issue 3

You can run the following query with your user address to confirm it returns the liquidation address:
```gql
query LiqAdd {
  bridgeGetUserLiquidationAddress(userAddress: "")
}
```

Resolves #2793


rdig (changes requested):
When trying to test this, loading the Crypto to fiat user profile tab, the `bridgeXYZMutation` crashes

Screenshot from 2024-08-02 15-02-51

So the KYC button doesn't show up:

Screenshot from 2024-08-02 15-02-53

I have all the `env` vars properly set, and have ran `npm run prepare` _(I've also tried setting the `apiKey` inside the `bridgeXYZMutation` to the sandbox key, but still no change)_

Screenshot from 2024-08-02 15-04-10
Screenshot from 2024-08-02 15-03-40

jakubcolony:
@rdig looks like the import path was off, the joys of vanilla JS development 😃

rdig (changes requested):
I keep running into issues. Sorry about that :(

I'm attempting to start the KYC process, but the mutation lambda errors out
Screenshot from 2024-08-06 18-45-23
Screenshot from 2024-08-06 18-45-19

Note that as far as I'm aware the sendbox key and url are correctly set:
Screenshot from 2024-08-06 18-46-16

jakubcolony:
@rdig I wondered if it's because of the obviously fake email you used but even with the same details, it works fine on my end. I am also connected to the sandbox API with the same key.


https://github.com/user-attachments/assets/aa0187a9-cfc7-4587-8081-64c4a536c02e

Let's catch up on this tomorrow

rdig:
Yeah, it's possible I'm doing something wrong with my setup.

rumzledz:
Hey @rdig just chiming in -- I've seen this happen to me when I came from a branch that had a different network hash. Just thought I'd double check if you've restarted your dev again on this branch? But knowing you, you probably already have!

rdig:
As I remember yes, I rand this fresh, but  if there are any doubts I can test again just to make sure my findings are not faulty

rumzledz (dismissed):
I've stuck with using a US address and a US bank account all seems to be working @jakubcolony ✨ 

I can prove that I updated it because the `id` field of the `bankAccount` object has been updated yet the `liquidationAddress` remained the same 👌 

same liquidation

rdig (approved):
Re-tested this _(live, with @jakubcolony)_, with proper US and EUR accounts and not it's working as expected.

Screenshot from 2024-08-14 14-06-41

PR #2799, opened 2024-07-29
Feat: Refetch extension data when updating thresholds

iamsamgibbs:
## Description

When we attempt to create a multi-sig action, we check the threshold settings from the extension data.

This PR adds an onSuccess function to the 'Save settings' button to refetch the extension data after a successful save. This means when we next check the thresholds we don't have stale data.

## Testing

Do not refresh the page at any step of testing

Step 1 - Install the extension (this should give Leela multi-sig permissions)
Step 2 - Open the action sidebar select create a mint tokens multi-sig action - there should be no error notices as the threshold is set to “majority” and you are the only user with multi-sig permissions
Step 3 - Close the sidebar, and change to a fixed threshold of 2. Click save. Now we are expecting an error notice as we do not have enough users with permissions to meet the threshold.

Screenshot 2024-07-29 at 11 38 08

Step 4 - Open the action sidebar again and select create a mint tokens action with the multi-sig decision method. We should have the error notice.

Screenshot 2024-07-29 at 11 39 55

If you want to further validate this is working, set the fixed threshold to 1 and the notice should be gone.

## Diffs

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

* Changes go here

**Deletions** ⚰️

* Deletions go here

## TODO

- [ ] Add TODOs

Contributes to #2658

### DIFF
--- a/src/components/frame/Extensions/pages/partials/SaveSettingsButton.tsx
+++ b/src/components/frame/Extensions/pages/partials/SaveSettingsButton.tsx
@@ -7,12 +7,16 @@ import ActionButton from '~v5/shared/Button/ActionButton.tsx';
 const displayName = 'pages.ExtensionDetailsPage.SaveSettingsButton';
 
 const SaveSettingsButton = () => {
-  const { isVisible, actionType, handleGetValues } = useContext(
-    ExtensionSaveSettingsContext,
-  );
+  const { isVisible, actionType, handleGetValues, handleOnSuccess } =
+    useContext(ExtensionSaveSettingsContext);
 
   return isVisible && !!actionType ? (
-    <ActionButton useTxLoader actionType={actionType} values={handleGetValues}>
+    <ActionButton
+      useTxLoader
+      actionType={actionType}
+      values={handleGetValues}
+      onSuccess={handleOnSuccess}
+    >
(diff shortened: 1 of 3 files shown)


mmioana (approved):
Nice and simple fix for this one, great job @iamsamgibbs 🎉 

Tested the steps and can confirm the extension data is re-fetched upon settings update
Screenshot 2024-07-29 at 19 44 39
Screenshot 2024-07-29 at 19 45 03
Screenshot 2024-07-29 at 19 45 27
Screenshot 2024-07-29 at 19 46 07

bassgeta (approved):
Nice and straightforward code changes, and it also works like a charm :sunglasses: 
Immediately after installation:
image
Threshold of 2:
image

Found a bug with domain threshold not being looked up correctly (will add it to the QoL ticket).
Let's get this one in the bag :sunglasses:

davecreaser (approved):
Nice one Sam, much nicer UX with this! Good simple solution too.

Tested and it's working great, also with changing the threshold down to 1 afterwards:
Screenshot 2024-07-31 at 10 12 01
Screenshot 2024-07-31 at 10 12 29
Screenshot 2024-07-31 at 10 12 50

PR #2800, opened 2024-07-29
[Multi-sig] feat: disable submit motion until we know if there are enough members, fetching optimisation

bassgeta:
## Description
When we were loading member data, we didn't disable submit, so users with a slow connection could technically submit the form before we even know if they can

https://github.com/user-attachments/assets/6cab29b7-5430-4bca-a43a-d130d72c61dc

## Testing

1. Install the Multi-Sig extension as `leela`
2. Give `amy` `Admin` Multi-Sig permissions in `General`
3. Refresh the page, open the dev tools on the `Network` tab (or use that GQL extension, whichever really), throttle the conenction to 3G
4. Try to create a mint tokens action, change all fields except changing the decision method to `Multi-Sig`, verify that nothing gets fetched
image
5. Choose `Multi-Sig` as the decision method, 2 things will happen in quick succession, verify both that:
a) The form submit is disabled at first with only 2 network requests going out
image
b) After loading is done, verify that the banner appears up top, submit stays disabled
image
6. Choose a different action type like `Create new team`, use `Multi-sig` as the decision method, after some loading, the button should change from being disabled to enabled
image

## Diffs

**New stuff** ✨

* `MultiSigMembersError` banner component which handles fetching if there are enough members

**Changes** 🏗

* `ActionSidebarContent` now keeps 2 state boolean values called `areMemberPermissionsLoading` and `canCreateAction` to use when deciding if form submit should be disabled

Contributes to #2658

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/ActionSidebarContent/ActionSidebarContent.tsx
+++ b/src/components/v5/common/ActionSidebar/partials/ActionSidebarContent/ActionSidebarContent.tsx
@@ -1,7 +1,7 @@
 import { useApolloClient } from '@apollo/client';
 import { WarningCircle } from '@phosphor-icons/react';
 import clsx from 'clsx';
-import React, { useEffect, type FC } from 'react';
+import React, { useEffect, type FC, useState } from 'react';
 import { useFormContext } from 'react-hook-form';
 import { useSelector } from 'react-redux';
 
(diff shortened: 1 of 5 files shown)


mmioana (approved):
Awesome work on this one @bassgeta 🎉 
I tested it and can confirm it works as expected (though there are 3 GraphQL requests being triggered after selecting Multi-Sig as decision method - GetCurrentExtensionVersion, GetMultiSigRolesForDomain and GetColonyExtension)

Screenshot 2024-07-30 at 08 18 53
Screenshot 2024-07-30 at 08 19 06
Screenshot 2024-07-30 at 08 19 14
Screenshot 2024-07-30 at 08 20 34
Screenshot 2024-07-30 at 08 20 54

iamsamgibbs (approved):
Nice job with this. Good job moving the error out into it's own component, much cleaner.

When there are not enough members:

https://github.com/user-attachments/assets/e8ebb291-1c43-4be6-ad44-088af9dd6261

And when there are:

https://github.com/user-attachments/assets/18c6a98c-4234-4748-955e-164bcd0d38da

davecreaser (approved):
Nice one @bassgeta, this is way nicer than before, a good improvement for sure!

Tested and it's working great:

https://github.com/user-attachments/assets/830f4ec1-af97-42cf-8dc0-f62e4be50ce9

Screenshot 2024-07-30 at 14 32 22

PR #2805, opened 2024-07-30
feat: add cancel to staged payment

CzarekDryl:
## Description

- Added canceling of staged payment action
- Issue https://github.com/JoinColony/colonyCDapp/issues/2669

## Testing

* Create staged payments action
* Cancel before or after review step using Permission

### DIFF
--- a/src/components/v5/common/CompletedAction/partials/PaymentBuilder/PaymentBuilder.tsx
+++ b/src/components/v5/common/CompletedAction/partials/PaymentBuilder/PaymentBuilder.tsx
@@ -180,6 +180,12 @@ const PaymentBuilder = ({ action }: PaymentBuilderProps) => {
           isLoading={!expenditure.metadata?.stages?.length}
           isPaymentStep={expenditureStep === ExpenditureStep.Payment}
         />
+        <CancelModal
+          isOpen={isCancelModalOpen}
+          expenditure={expenditure}
+          onClose={toggleCancelModalOff}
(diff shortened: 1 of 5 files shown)


bassgeta (approved):
Changes look sensible, nice and simple :ok_hand: 
image
image

Not sure if it's in the scope of this issue, but should we change this text to say something like "Member used permissions to cancel this action"
image

jakubcolony (changes requested):
Nice work @CzarekDryl!

Cancelling as creator:

https://github.com/user-attachments/assets/aab0a657-073e-4687-bc6a-126c92be2ca0

Cancelling payment created by another user:

https://github.com/user-attachments/assets/030813df-e88c-4e4c-835b-429a395b73be

I'm requesting changes as it seems possible to click on the "Fund" button in recording 1 after cancelling the payment (but before the updated status is fetched). I think this could re-use the existing `expectedStepKey` concept and disable the button while the updated status is loading.

Also a question to @arrenv regarding cancelling another user's payment:
- In recording 2, should there be an alternative copy in the cancel modal? It talks about the cancelling user not being penalised, it made me wonder if we want to change it to something along the lines of "payment creator will not be penalised"

CzarekDryl:
@jakubcolony I've disabled submit button after cancelling action

jakubcolony (approved):
Thanks for the changes @CzarekDryl, LGTM 👍

PR #2807, opened 2024-07-30
Chore: Add Eslint absolute import rule

davecreaser:
## Description

Add an eslint rule to enforce absolute imports with aliases. These are preferred over relative imports as they are easier to read and maintain. Since we use aliases to make imports simpler, the lint rule must also handle these, and use these when autofixing. We do allow sibling imports, and one level deep relative imports (parent folders).

I researched a few different lint rules which were available to do this, and also tried making a custom rule. Here's why I avoided certain rules:
- Making a custom rule was adding too much overhead, and while we could have a lot of control, it was too much work to maintain for such a small thing, especially when there's a lot of good options available.
- Using [this rule](https://github.com/import-js/eslint-plugin-import/blob/HEAD/docs/rules/no-relative-parent-imports.md) from the import plugin that we already use did throw the correct errors, but does not allow auto fixing, which would get very annoying very quickly.
- Using a simple and popular plugin such as [absolute-imports](https://www.npmjs.com/package/eslint-plugin-absolute-imports) or [no-relative-import-paths](https://www.npmjs.com/package/eslint-plugin-no-relative-import-paths) solved only half the problem. It fixed the relative imports, but didn't use aliases.
- Using a package like [eslint-plugin-import-alias](https://www.npmjs.com/package/eslint-plugin-import-alias) worked great, and was very lightweight and easy to audit/read the code. There were a few problems however. Firstly, I was concerned about the fact that the latest publish was six years ago, it seemed it wasn't being maintained. Secondly, you had to manually pass an object of aliases to the lint rule, which adds maintenance on our end.

After all the testing, I landed on [@limegrass/eslint-plugin-import-alias](https://www.npmjs.com/package/@limegrass/eslint-plugin-import-alias). Not only does it solve the issue, but it is also actively maintained (last publish 3 months ago) and has almost 7k weekly downloads. It also simply reads the aliases/paths from our tsconfig file, which makes it super simple to use. It even has a configurable depth, so we can allow sibling and parent imports as preferred.

## Testing

Shouldn't need much testing, just a basic check that the app still runs, but if you spot any fishy import changes then maybe check the part of the app which that file is used in.

## Diffs

**New stuff** ✨

* New eslint rule to enforce absolute imports where relevant, and the use of aliases.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -144,7 +144,7 @@ module.exports = {
     '@limegrass/import-alias/import-alias': [
       'error',
       {
-        relativeImportOverrides: [{ depth: 1 }],
+        relativeImportOverrides: [{ path: '.', depth: 1 }],
       },
     ],
 
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -26,6 +26,7 @@ module.exports = {
     'jsdoc',
     'react-hooks',
     'react-refresh',
+    '@limegrass/import-alias',
   ],
   overrides: [
     {
@@ -138,6 +139,12 @@ module.exports = {
           },
         ],
         warnOnUnassignedImports: true,
+      },
+    ],
+    '@limegrass/import-alias/import-alias': [
+      'error',
+      {
+        relativeImportOverrides: [{ depth: 1 }],
       },
     ],
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -118,6 +118,7 @@
         "@import-meta-env/unplugin": "^0.5.1",
         "@jackfranklin/test-data-bot": "^1.4.0",
         "@jest/globals": "^29.0.2",
+        "@limegrass/eslint-plugin-import-alias": "^1.4.1",
         "@storybook/addon-essentials": "^8.0.0",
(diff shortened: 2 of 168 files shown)


rumzledz (dismissed):
These changes, I like. This approval, you shall receive ✅ 

yoda-star-wars

rumzledz:
It's also handled via husky's pre-commit hook so the auto import resolution is automatically included in your commit. Awesome stuff ✨ 

lint-staged in action

rdig (approved):
Testing so far doesn't seem that this rule breaks anything, and running the various linters works as expected. 

Nicely done!

Screenshot from 2024-08-05 15-22-53

rumzledz (approved):
Still LGTM ✨ 

image

mmioana (dismissed):
Gave it a quick test, but seems nothing got broken 💯 

Screenshot 2024-08-14 at 17 37 59

Impressive work @davecreaser 🎉

rumzledz (approved):

rdig (approved):
Re-approved due to rebase

PR #2808, opened 2024-07-30
[Multi-sig] fix: use native domainId for getting domain from multisig extension params

bassgeta:
## Description

Long story short, somehow we changed the format of the domainId in multisig's installed params from `<COLONY_ADDRESS>_<DOMAIN_NATIVE_ID>` to just use it's native id. This means that all subdomain's just used the global threshold :D 

## Testing

1. Install Multi-Sig extension
2. Set the global threshold to 2, threshold in Andromeda to 1
image
3. Refresh the page and open the action sidebar
4. Verify that you can't create a `Manage reputation` action in General (since there aren't 2 permission holders there)
image
5. Verify that however, this would work in Andromeda
image

## Diffs

**Changes** 🏗

* `useDomainThreshold` now checks the nativeId instead of constructing the full id (also added a note to remove nullish coalescence once we merge #2800 

Contributes to #2658 (point 23)

### DIFF
--- a/src/hooks/multiSig/useDomainThreshold.ts
+++ b/src/hooks/multiSig/useDomainThreshold.ts
@@ -1,6 +1,5 @@
 import { type ColonyRole, Extension } from '@colony/colony-js';
 
-import { useColonyContext } from '~context/ColonyContext/ColonyContext.ts';
 import useExtensionData from '~hooks/useExtensionData.ts';
 import { type Threshold } from '~types/multiSig.ts';
 import { isInstalledExtensionData } from '~utils/extensions.ts';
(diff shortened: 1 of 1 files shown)


davecreaser (approved):
Nice catch, very simple, all good 🙃 If you can just remove the note since 2800 is removed then all good, but i'll pre approve so you don't need to chase me on that.

I tested that's it's all working, including making the andromeda action in general:
Screenshot 2024-07-30 at 14 49 34
Screenshot 2024-07-30 at 14 49 43
Screenshot 2024-07-30 at 14 49 54

iamsamgibbs (approved):
Good job! Nice simple fix.

Screenshot 2024-08-01 at 09 02 08

Screenshot 2024-08-01 at 09 07 34

mmioana (approved):
Such a simple fix @bassgeta nice 🎉 
Screenshot 2024-08-01 at 11 24 45
Screenshot 2024-08-01 at 11 26 21
Screenshot 2024-08-01 at 11 26 34

PR #2812, opened 2024-07-31
Hotfix for metatransaction checks

chmanie:
## Description

Metatransactions are now checked on a general basis (e.g. if the user has them enabled, they are being used). This enables the users to retry transactions as metatransactions and the other way round.

I removed a metatransaction check that was mainly used for the vesting contracts. If we still have methods that don't support metatransactions this needs to be reinstated (it wasn't used though as far as I can tell).

## Testing

1) Enable metatransactions for a user
2) Try to send a transaction

## Diffs

**Changes** 🏗

* Check whether user has metatransactions enabled on transaction sending not on creation

**Deletions** ⚰️

* Removed metatransaction property from transactions
* Removed forced non-metatransactions

### DIFF
--- a/amplify/backend/api/colonycdapp/schema.graphql
+++ b/amplify/backend/api/colonycdapp/schema.graphql
@@ -1812,10 +1812,6 @@ type Transaction @model {
       sortKeyFields: ["from"]
     )
   """
-  True if the transaction is a metatransaction
-  """
-  metatransaction: Boolean!
-  """
   A title to show in the UI
   """
   title: String # JSON string
--- a/scripts/create-data.js
+++ b/scripts/create-data.js
@@ -289,7 +289,6 @@ const addTxToDb = async ({
     methodContext: null,
     methodName,
     status: 'SUCCEEDED',
-    metatransaction: false,
     title: null,
     titleValues: null,
     params: JSON.stringify(params),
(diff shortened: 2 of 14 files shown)


jakubcolony (approved):
I'm a bit behind in terms of the original issue, however I tested a few actions and motions with metatx enabled, all of which seem to be working fine:

https://github.com/user-attachments/assets/d0d5bbcc-e74a-4829-bb47-71f5906b49ce


https://github.com/user-attachments/assets/008dffd1-8c90-42dd-bf9b-69ab1351d53f

iamsamgibbs (approved):
Changes look good to me! Tested creating new transactions with metatransactions turned on and off without issue.

I also tested creating a transaction using metatransactions, rejecting the transaction, then retried it from the userhub with metatransactions turned off without issue.

davecreaser (approved):
This looks to be working for me! Before I turned on meta transactions I was seeing an error of not enough gas in metamask, and afterwards it was a simple signature.

Code-wise I don't see anything I dislike either, all good! Nice one Chris 👏 

Screenshot 2024-07-31 at 12 03 04
Screenshot 2024-07-31 at 12 03 47
Screenshot 2024-07-31 at 12 03 59

PR #2814, opened 2024-07-31
Fix/2530 agreements pagination

### LINKED ISSUE #2530: Agreements missing on agreements page
## Steps to reproduce

Seemingly, no other way to reproduce but to create agreements and see if it happens again.
You can check the Betacolony here - https://app.colony.io/beta/agreements - and you will see only one agreement, which I have just created. But in the past another agreement has been created and the transaction can be loaded manually, but, does not appear in the list.

This is the working one: https://app.colony.io/beta/agreements?tx=0xd7b88d1fccfa50a1ba532c2ae9c66bc3226d2994343aef931e7bcd8a5ef69434

This is the missing agreement: https://app.colony.io/beta/agreements?tx=0xb68baa890a1a6a0eb451b13c759b0691831dd95436c9ceab58038b901b7001e2

## Expected behaviour

* All agreements that have passed the staking threshold, should appear on the agreements page.
* At the same time, we would like to ensure that Agreements appear in the activity feed.
* It would also be good if these worked off of subscriptions, so new ones show up without having to reload.

## Actual behaviour

* Agreements that have passed are not showing on the agreements page.

    #### Showing only one on the page
    image

  #### Missing one can be found and loaded manually
  image

mmioana:
## Description

**Expected behaviour**
All agreements that have passed the staking threshold, should appear on the agreements page. - _The problem was caused due to pagination issues, so this PR includes the mechanism to fetch all agreements_
At the same time, we would like to ensure that Agreements appear in the activity feed.
It would also be good if these worked off of subscriptions, so new ones show up without having to reload.

## Testing

TODO: Check all created agreements appear on the agreements page. 
Check agreements having a stake of 10% are showing up in the activity feed.
Currently the `QUERY_PAGE_SIZE = 20`, so if you want to check the pagination works properly, you can manually set it to `2`.

* Step 1. Install the **Reputation Weighted (Lazy Consensus method)** extension
* Step 2. Create 3 Agreements
Screenshot 2024-07-31 at 14 06 36
* Step 2.1. Support one with > 10%
* Step 2.2. Oppose one with > 10%
* Step 2.3. Support/Oppose one with < 10%
If you want to forward time please use 
`curl -X POST --data '{"jsonrpc":"2.0","method":"evm_increaseTime","params":[3600],"id":1}' localhost:8545`

* Step 3. Check activity feed shows the agreements with > 10% stake
* Step 4. Check Agreements page display all 3 agreements

Resolves #2530

### DIFF
--- a/src/components/frame/v5/pages/AgreementsPage/hooks.ts
+++ b/src/components/frame/v5/pages/AgreementsPage/hooks.ts
@@ -1,10 +1,12 @@
-import { useMemo } from 'react';
+import { useEffect, useMemo } from 'react';
+import { useSearchParams } from 'react-router-dom';
 
 import { useColonyContext } from '~context/ColonyContext/ColonyContext.ts';
 import {
   useGetColonyActionsQuery,
   ColonyActionType,
   ModelSortDirection,
+  useGetColonyActionQuery,
 } from '~gql';
 import {
   filterActionByMotionState,
@@ -12,6 +14,7 @@ import {
 } from '~hooks/useActivityFeed/helpers.ts';
 import useNetworkMotionStates from '~hooks/useNetworkMotionStates.ts';
 import { notNull } from '~utils/arrays/index.ts';
+import { isTransactionFormat } from '~utils/web3/index.ts';
 
 import { useFiltersContext } from './FiltersContext/FiltersContext.ts';
 
@@ -22,7 +25,7 @@ const useGetAllAgreements = () => {
     colony: { colonyAddress },
   } = useColonyContext();
 
-  const { data, loading, fetchMore } = useGetColonyActionsQuery({
+  const { data, loading, fetchMore, refetch } = useGetColonyActionsQuery({
     variables: {
       colonyAddress,
       filter: {
@@ -62,13 +65,34 @@ const useGetAllAgreements = () => {
   return {
     agreementsData,
     loading,
+    refetchAgreements: refetch,
   };
 };
 
+export const useGetCurrentOpenedAgreement = () => {
+  const [searchParams] = useSearchParams();
+  const transactionHash = searchParams.get('tx');
+  const isInvalidTx = !isTransactionFormat(transactionHash ?? undefined);
+  const { data: actionData } = useGetColonyActionQuery({
+    skip: isInvalidTx,
+    variables: {
+      transactionHash: transactionHash ?? '',
+    },
+  });
+  const action = actionData?.getColonyAction;
+  const motionData = action?.motionData;
+
+  return action?.type === ColonyActionType.CreateDecisionMotion
+    ? motionData
+    : null;
+};
+
 export const useGetAgreements = () => {
+  const currentAgreement = useGetCurrentOpenedAgreement();
+
   const { activeFilters, searchFilter } = useFiltersContext();
 
-  const { agreementsData, loading } = useGetAllAgreements();
+  const { agreementsData, refetchAgreements, loading } = useGetAllAgreements();
 
   const motionIds = useMemo(
     () =>
@@ -77,6 +101,17 @@ export const useGetAgreements = () => {
         .filter(Boolean) || [],
     [agreementsData],
   );
+
+  useEffect(() => {
+    if (
+      !loading &&
+      currentAgreement?.motionId &&
+      !motionIds.includes(currentAgreement?.motionId)
+    ) {
+      refetchAgreements();
+    }
+  }, [currentAgreement?.motionId, motionIds, loading, refetchAgreements]);
+
   const {
     motionStatesMap,
     loading: motionStatesLoading,
--- a/src/components/frame/v5/pages/AgreementsPage/hooks.ts
+++ b/src/components/frame/v5/pages/AgreementsPage/hooks.ts
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useMemo } from 'react';
+import { useMemo } from 'react';
 
 import { useColonyContext } from '~context/ColonyContext/ColonyContext.ts';
 import {
@@ -32,26 +32,32 @@ const useGetAllAgreements = () => {
       limit: QUERY_PAGE_SIZE,
     },
     fetchPolicy: 'network-only',
+    onCompleted: (newData) => {
+      if (newData?.getActionsByColony?.nextToken) {
+        fetchMore({
+          variables: { nextToken: newData?.getActionsByColony?.nextToken },
+          updateQuery: (prev, { fetchMoreResult }) => {
+            if (!fetchMoreResult) return prev;
+
+            // Here, combine the previous items with the newly fetched items
+            return {
+              ...prev,
+              getActionsByColony: {
+                ...prev.getActionsByColony,
+                items: [
+                  ...(prev.getActionsByColony?.items ?? []),
+                  ...(fetchMoreResult.getActionsByColony?.items ?? []),
+                ],
+                nextToken: fetchMoreResult?.getActionsByColony?.nextToken,
+              },
+            };
+          },
+        });
+      }
+    },
   });
 
   const agreementsData = data?.getActionsByColony?.items.filter(notNull);
-  const nextToken = data?.getActionsByColony?.nextToken;
-
-  const fetchNextPage = useCallback(
-    async (token) => {
-      await fetchMore({
-        variables: { nextToken: token },
-      });
-    },
-    [fetchMore],
-  );
-
-  useEffect(() => {
-    if (nextToken) {
-      fetchNextPage(nextToken);
-    }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [nextToken]);
 
   return {
     agreementsData,
--- a/src/components/frame/v5/pages/AgreementsPage/hooks.ts
+++ b/src/components/frame/v5/pages/AgreementsPage/hooks.ts
@@ -1,4 +1,4 @@
-import { useMemo } from 'react';
+import { useCallback, useEffect, useMemo } from 'react';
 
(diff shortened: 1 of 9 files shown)


src/components/v5/common/ActionSidebar/partials/Motions/Motions.tsx
```diff
@@ -55,10 +55,20 @@ const Motions: FC<MotionsProps> = ({ transactionId }) => {
...
+    /** Not using motionFinished as it might also include the NetworkMotionState.Finalizable */
+    if (
+      [NetworkMotionState.Finalized, NetworkMotionState.Failed].includes(
+        networkMotionState,
+      )
+    ) {
+      stopPollingForAction();
+    } else {
+      startPollingForAction();
+    }
```
rdig, line 67: So I don't think this works as you intended, as the list doesn't update properly when creating new agreements.

I'd say discard all of this, as it's really not part of this fix, and if you really want to fix this properly, use `subscriptions` instead of `queries`

src/components/frame/v5/pages/AgreementsPage/hooks.ts
```diff
@@ -15,23 +15,55 @@ import { notNull } from '~utils/arrays/index.ts';
...
+    },
+    [fetchMore],
+  );
+
+  useEffect(() => {
+    if (nextToken) {
+      fetchNextPage(nextToken);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [nextToken]);
```
rdig, line 54: There's a better way to do this, without `useEffect`. See this implementation: https://github.com/JoinColony/colonyCDapp/blob/master/src/context/TokenSelectContext/TokenSelectContextProvider.tsx#L27-L52

rdig (changes requested):
This works as expected, I even lowered the paging limit to prove this is behaving correctly.

Screenshot from 2024-08-01 20-39-31
Screenshot from 2024-08-01 20-45-53

I've only rejected this because I want you to refactor the way you do query pagination. Basically combining `fetchMore` with `onCompleted`, that way you can get rid of `useEffect`

Nice work!

src/components/frame/v5/pages/AgreementsPage/hooks.ts
mmioana: That is quite a nice implementation, thanks @rdig  🚀 On to updating mine

src/components/frame/v5/pages/AgreementsPage/hooks.ts
```diff
@@ -15,23 +15,61 @@ import { notNull } from '~utils/arrays/index.ts';
...
         type: { eq: ColonyActionType.CreateDecisionMotion },
       },
       sortDirection: ModelSortDirection.Desc,
+      limit: QUERY_PAGE_SIZE,
     },
     fetchPolicy: 'network-only',
+    onCompleted: (newData) => {
+      if (newData?.getActionsByColony?.nextToken) {
+        fetchMore({
+          variables: { nextToken: newData?.getActionsByColony?.nextToken },
```
rumzledz, line 38: I thought I'd just bring this up as well. And I'm not saying to do this in your PR. If the intention is to come up with a sort of infinite scrolling mechanism, we could either create our own hook that detects the scroll position of a container or use some library that can detect when a container reaches its bottom scroll threshold before firing the `fetchMore()` function.

And I say this because this `onCompleted` logic is set up to keep on calling the `GetColonyActionsQuery` until it fetches all the data, all in one go, which sort of defeats the purpose of pagination. You can prove this by setting `QUERY_PAGE_SIZE = 1` then open the GraphQL Network tab and filter the query by:
```
type":{"eq":"CREATE_DECISION_MOTION"}}
```
Screenshot 2024-08-02 at 12 53 43

You'll find that it will continue to query until all the data has been fetched, when `nextToken` becomes `null` essentially.

Our logic should be something like (this is future stuff, not in this PR):

```ts
<InfiniteScrollView onBottomReached={fetchMore}>
  {agreementItems.map(agreementItem => <AgreementItem agreementItem={agreementItem} />
</InfiniteScrollView>
```

rumzledz (commented):
Hey @mmioana this was a fun one to test 😄 I noticed a little glitch right after you've created an agreement. It momentarily appears in the Agreements page but then it disappears. Then if you reload the page, the new agreement is correctly shown
2530_GLITCH

If this is current behaviour on `master`, let me know and I'll approve this 👌 

The other comment I left is not a change request, more like a note for myself as well 😄

src/components/frame/v5/pages/AgreementsPage/hooks.ts
rdig: I half agree with this. Two things I want to point out:
- this is something for the far future, we have more pressing things to fix
- this needs to be a decision by product ultimately

But all in all I like your suggestion and support you coming up with more stuff like this in the future :tada:

mmioana:
@rdig @rumzledz I have added the refetch logic after a new agreement is created. 

The glitch you have noticed is due to the fact an agreement is saved as a draft until the create motion transaction is executed, after which the draft gets deleted, even if the motion data is not yet available. We could however save the draft only if there is an error while executing the createDecisionMotion, but would love to hear/read your opinions 😄

rdig (changes requested):
Re-reviewed this, and functionally it's all there. Nicely done Ioana!

The Draft "flash" issue is not really a issue after all, as it's only happening in DEV, where wallets sign things automatically for you.

In a real life environment, where you have to manually sign and confirm, the Draft gets created and doesn't disappear until the transaction gets executed and mined, which takes a while. 

So from that perspective, it's find, you don't have to change anything.

Screenshot from 2024-08-06 19-18-03

However, I did run into a issue happening inside `ActionSidebarContent` which in only happening here as I can't replicate it anywhere else. It's a warning, but something that I want you to investigate and fix.

Screenshot from 2024-08-06 19-13-10

Basically, to replicate the above, Just create a new agreement and just look at the console. It happens both with dev and metamask wallets.

Sorry I keep finding issues, but while you're here, might as well fix them :)

mmioana:
Thank you @rdig for your explanation, it is very valuable!

On my way to investigate the issue you found 🚀

iamsamgibbs:
Just a note to say I've just observed the same issue on #2817 so it is likely an issue on master:

Screenshot 2024-08-07 at 08 07 52

bassgeta (dismissed):
Code changes look good, everything makes sense just looking at it, so good job with that!
I do agree with the infinite scroll proposition, but it's out of scope for this issue + undefined product-wise :D
I get only 2 decision in the activity feed
image
With all of them on the agreements page:
image
Draft still gets displayed on top:
image
And disappears and shows up when it's created:
image
Nice one :ok_hand:

mmioana:
It seems Apollo throws that warning if you try to `refetchQueries` that are no longer active. I added a new helper function to check if `SearchActions` is active before trying to refetch

@rdig @bassgeta @rumzledz could you please give it another try?

rdig (dismissed):
I can still replicate this quite easily. 

Just create a new agreement and when the action renders, the warning is being thrown

Screenshot from 2024-08-07 23-28-22

_Note that I've pulled the latest changes just to ensure I'm not running stale code_

Screenshot from 2024-08-07 23-30-26


But since this is not necessarily related to this PR I feel less strongly that you fix it here.

So in an effort to move this along, I'm  good if you merge this without the fix. _(However I wouldn't mind if you actually fixed it)_

mmioana:
@rdig thank you for re-testing this! 

I tried creating now several agreements and couldn't get the warning any longer, so is there any specific flow you followed?

rdig:
Nope. Bog standard.

1. start dev env (including nuking node modules)
2. run create data script
3. start frontend
4. install voting reputation extension
5. create agreement

At the last step, once the action sidebar renders, I get the warning

mmioana:
@rdig I started a fresh environment, installed the extension in Planex
Went to agreements page -> create new agreement
Screenshot 2024-08-08 at 08 39 35
Went to dashboard -> create new agreement
Screenshot 2024-08-08 at 08 40 06
Went back to agreements page -> create new agreement
Screenshot 2024-08-08 at 08 40 33
Installed the extension in Wayne, went to agreements page -> create new agreement
Screenshot 2024-08-08 at 08 42 30

In none of these cases did I manage to reproduce again the warning 😢

iamsamgibbs (approved):
Nice work on this Ioana! Tested and agreements without stakes are not showing in the activity feed, but do show on the agreement page and widget:

https://github.com/user-attachments/assets/c99ab1ac-1f59-47ab-8403-8a3c3a83ceaf

I didn't encounter the console warning whilst testing either.

I did experience this issue with the timer not showing in the activity feed, could just be a local environment issue though:

https://github.com/user-attachments/assets/a523fb87-e5a7-4c04-aaa3-06b53913fa01

I also noticed the agreements in the activity feed are missing their custom titles:

Screenshot 2024-08-08 at 11 16 59

I think that would be nice to fix up here, but can also leave it for another issue if you would prefer. :)

rumzledz (dismissed):
I can confirm it works @mmioana thanks for this ✨ 

I have 2 agreements. 1 with 22% Yay, and 1 with 22% Nay. Both are showing up in the Latest Activity table:

yay nay

Agreements page is still working as expected

Screenshot 2024-08-08 at 11 35 52

mmioana:
I'll fix also the agreements titles, didn't notice that one @iamsamgibbs 👀

mmioana:
I added the custom agreement title in the ColonyActionsTable ✨ 
Can I please get another round of approvals @rdig @iamsamgibbs @rumzledz ? 🙏

rdig (approved):
All good now @mmioana. Nicely done!

Thanks for all your patience through all the rounds of reviews :100:

rumzledz (approved):

PR #2815, opened 2024-07-31
Add new auth proxy docker hash

### LINKED ISSUE #2813: Creating an action sometimes fails metadata creation
## Steps to reproduce

This is a race condition, so it won't always happen (I know, sucks).

1. Try to make any action like a simple payment.
2. Sometimes, it will look like the action was not complete, as the button will go back to it's basic state after pending, and it won't move to the action complete form.
3. When this happens, after refreshing, the action is there in the list (see video)

https://github.com/user-attachments/assets/3d8d193e-f5d9-4b9e-8dc9-5bba6c9aabc6

What is actually happening is that the auth proxy is failing a call to graphql because of a race condition where it is trying to fetch two things simultaneously and that leads to it sometimes failing. In this case, it's failing on the createActionMetadata mutation, so the metadata is never created, but the action is.

## Expected behaviour

* The action is made as normal, with proper metadata.

## Actual behaviour

* The action is made but the metadata isn't.

davecreaser:
## Description

- Add new auth proxy hash (for [this commit](https://github.com/JoinColony/colonyCDappAuthProxy/commit/f1d5e35271358946ec65e9a0f7d27fa66c2b878a))

## Testing

Just make sure your docker can still run with that hash!

## Diffs

**New stuff** ✨

* New auth proxy hash

Resolves #2813

### DIFF
--- a/docker/colony-cdapp-dev-env-auth-proxy
+++ b/docker/colony-cdapp-dev-env-auth-proxy
@@ -1,6 +1,6 @@
 FROM colony-cdapp-dev-env/base:latest
 
-ENV AUTH_PROXY_HASH=0ca1fa78d55e2f2fdd9177fa526ae937855d7874
+ENV AUTH_PROXY_HASH=f1d5e35271358946ec65e9a0f7d27fa66c2b878a
 
 # Add dependencies from the host
 # Note: these are listed individually so that if they change, they won't affect


chmanie (approved):
I approve this new hash.

iamsamgibbs (approved):

rumzledz (approved):

PR #2818, opened 2024-07-31
fix: add remaining tx group translations

chmanie:
## Description

When checking for the migrations that need to be done to add groupIds to all transactions, I discovered that some groups did not get proper treatment throughout the lifetime of #2127.

This PR fixes it.

## Testing

Create the following transactions:

1) cancelAndReclaimStake
2) cancelExpenditure
3) reclaimStake
4) cancelExpenditureViaArbitration
5) lockExpenditure
6) reclaimStake
7) deprecateExtension
8) installExtension
9) uninstallExtension
10) upgradeExtension

They should have a proper group key and also a translation for that group title and description.

## Diffs

**New stuff** ✨

- Added some missing field to `TRANSACTION_METHODS`
- Added some missing translations for these groups


rdig (approved):
All good here! 

Note that I've used some keys from here when running the migration in QA and Prod

Screenshot from 2024-08-01 00-17-17

iamsamgibbs (approved):
Nice one, all looks good to me!

PR #2833, opened 2024-08-02
Feat: Handle batch payment limit

### LINKED ISSUE #2250: Advanced Payments API: Determine the limit for CSV batch payments
## Specification

With CSV uploads, it is expected we will hit limit on transaction size at some point, after which the transactions will have to be split in order to be sent.

This issue is to establish such limit and if it's lower than the required 400 CSV rows, implement a work around.

iamsamgibbs:
## Description

When creating an advanced payment we hit a limit on the transaction size when creating a large number of payments in one go.

This PR adds a `chunkedMulticall` helper function to split large multicalls which are too big for a single transaction, into smaller chunks.

The `claimExpenditurePayouts` function has been refactored to make use of this. A limit of 78 payouts has been set as the multicall was failing with more than this.

The `createExpenditure` saga has been refactored to use the new helper. A limit of 164 payouts has been set, when creating multicalls with more than this the multicall function would work fine, but the resulting transaction receipt was too big to update the transaction in the db.

## Testing

Step 1 - Create an advanced payment action
Step 2 - Upload this csv with 400 rows (feel free to change the values if you like) [https://drive.google.com/file/d/1iMFNpRtMwyc3teh2mVPeBlRPEyZBRwFQ/view?usp=sharing](https://drive.google.com/file/d/1iMFNpRtMwyc3teh2mVPeBlRPEyZBRwFQ/view?usp=sharing)
Step 3 - Check you can go through the entire flow through to claiming the payouts

Further testing: Increase the chunkSize values in `claimExpenditurePayouts` and `createExpenditure` by 1 to verify they are the correct limit.

Note: I have noticed issues with very slow loading of all the transactions on the completed action screen (including in the action subtitle). I have also had issues when locking / funding the transaction `Transaction is not ready to send` - refreshing and trying the transaction again will resolve this. These are both out of scope for this PR.

## Diffs

**New stuff** ✨

* `chunkedMulticall` helper function 

**Changes** 🏗

* `claimExpenditurePayouts` refactored
* `createExpenditure` refactored
* CSV file upload now allows 402 rows to accommodate for the two header rows required by the template

Resolves #2250

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/forms/PaymentBuilderForm/partials/FileUploadModal/hooks.ts
+++ b/src/components/v5/common/ActionSidebar/partials/forms/PaymentBuilderForm/partials/FileUploadModal/hooks.ts
@@ -108,7 +108,8 @@ export const useUploadCSVFile = (
 
           const truncatedData = result.data.slice(dataDelimiterIndex + 2);
 
-          if (truncatedData.length > 400) {
+          // Max number of payments is 400, but the format requires two header rows
+          if (truncatedData.length > 402) {
             setError(DropzoneErrors.STRUCTURE);
             return;
           }
--- a/src/redux/sagas/expenditures/createExpenditure.ts
+++ b/src/redux/sagas/expenditures/createExpenditure.ts
@@ -4,6 +4,7 @@ import { takeEvery, fork, call, put } from 'redux-saga/effects';
 import { type ColonyManager } from '~context/index.ts';
 import { type Action, ActionTypes, type AllActions } from '~redux/index.ts';
 import { transactionSetParams } from '~state/transactionState.ts';
+import { type ExpenditurePayoutFieldValue } from '~types/expenditures.ts';
 import { TRANSACTION_METHODS } from '~types/transactions.ts';
 
 import {
@@ -24,6 +25,7 @@ import {
   createActionMetadataInDB,
   adjustPayoutsAddresses,
 } from '../utils/index.ts';
+import { chunkedMulticall } from '../utils/multicall.ts';
 
 export type CreateExpenditurePayload =
   Action<ActionTypes.EXPENDITURE_CREATE>['payload'];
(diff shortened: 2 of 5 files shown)


rdig (approved):
Works as expected! Nicely done!

Screenshot from 2024-08-05 15-02-06
Screenshot from 2024-08-05 15-02-58
Screenshot from 2024-08-05 15-02-16
Screenshot from 2024-08-05 15-02-29
Screenshot from 2024-08-05 15-05-57
Screenshot from 2024-08-05 15-06-03
Screenshot from 2024-08-05 15-06-18

I did run into the issue you mentioned about "transaction not ready to send", and indeed a page refresh fixed it:

Screenshot from 2024-08-05 15-03-12
Screenshot from 2024-08-05 15-05-06

rdig:
> Note: I have noticed issues with very slow loading of all the transactions on the completed action screen (including in the action subtitle). I have also had issues when locking / funding the transaction Transaction is not ready to send - refreshing and trying the transaction again will resolve this. These are both out of scope for this PR.

Ran into this issue quite consistently so we should ping @chmanie as he might know more, or be able to help :)

bassgeta (approved):
Code looks good, it's a nice util and pretty easy to use, nice one!
image
image
image
image

chmanie:
> > Note: I have noticed issues with very slow loading of all the transactions on the completed action screen (including in the action subtitle). I have also had issues when locking / funding the transaction Transaction is not ready to send - refreshing and trying the transaction again will resolve this. These are both out of scope for this PR.
> 
> Ran into this issue quite consistently so we should ping @chmanie as he might know more, or be able to help :)

Happy to help! How can I reproduce these issues?

jakubcolony (approved):
Clean and elegant solution to chunk multicall, great job @iamsamgibbs 

I changed all the claim delays to max value of 99999 and was able to create a payment. Initially, only the first 164 slots loaded but it seemed due to block-ingestor taking longer to process the events. Eventually all 400 slots showed up.

https://github.com/user-attachments/assets/9f498fab-9020-40d2-b7c5-e89c429a6957

Was also able to finalize the payment without issues:
image

iamsamgibbs:
@chmanie I don't think this is at all related to any of the transactions work you've been doing lately. I think as @jakubcolony suggested, the block-ingestor just takes a while to process the events.

The issue is: when making a large number of payments, it takes an entire 2 minutes for them to finish loading into the UI. I've recorded it here, but was also checking how long they take to show up in the database, so it does just seem like block-ingestor being slow:

https://github.com/user-attachments/assets/63f105c5-657d-44b9-852a-4f6877a3f53b

I'm not sure if there is anything we can do on block-ingestor to speed this up. From a UI point of view, at the end of the create expenditure saga, we could save the expected number of payments to an expenditureMetadata object, then query that on the completed action screen to at least show the correct number in the action subtitle and show the correct number of rows in the table with skeleton loaders until block-ingestor has finished processing.

But that is well outside the scope of this PR, so I'll merge it up!

PR #2866, opened 2024-08-05
Feat: Simple payment motion amount should exclude network fee

### LINKED ISSUE #2642: Payment amount value regression for motions
## Steps to reproduce

1. Create a new Simple payment action.
2. Enter an easy to track value, e.g. 1
3. Select the Reputation decision method (permissions works fine)
4. Create the payment.
5. Notice that on the created screen, the amount will now be 1.0101, however, it should be 1.

This was previously fixed in this PR - https://github.com/JoinColony/colonyCDapp/pull/1388

## Expected behaviour

* The amount should show the amount being paid to the recipient, excluding the network fee.
- It should use the live network fee while created or during the motion, then the payments stored network fee after successful payment.

## Actual behaviour

* The total payout shows with the amount being paid to the recipient in addition to the network fee.

image

iamsamgibbs:
[Related block-ingestor PR 262](https://github.com/JoinColony/block-ingestor/pull/262)

## Description

When a Simple Payment motion is created, the amount that is stored in the database includes the network fee.

This PR addresses an issue where the amount shown on the motion completed action screen and the activity feed includes the network fee.

During the period between a simple payment motion being created and the motion being finalized, it should display the total amount minus the **current** network fee. This means if the network fee increases or decreases before the motion is finalized, the amount on the completed action screen and activity feed is expected to change in line with the network fee.

Once the motion is finalized, the motion action is updated to reflect the network fee at the time the motion was finalized and the amount excluding the fee.

This PR adds a new `useAmountLessFee` hook which takes in the amount and networkFee from an action. If the networkFee is not null, we assume the amount is already excluding the network fee and return it. Otherwise, we get the current network fee and return the adjusted amount.

There is also a block-ingestor change to update the amount and network fee stored in the database when a simple payment motion is finalized.

## Testing

Step 1 - Install the reputation weighted voting extension.
Step 2 - Create a Simple Payment motion, use an easy to remember amount.
Step 3 - Check the amount in the completed action screen is correct.

Screenshot 2024-08-05 at 15 16 33

Step 4 - Stake the motion.
Step 5 - Refresh the dashboard and check the motion appears in the activity feed with the correct amount.

Screenshot 2024-08-05 at 15 18 50

Step 6 - Run the following mutation in GraphiQL to change the network fee stored in the database - pick any value between 0 - 100. (Note: Usually network fee changes are emitted by the chain, which is handled by block-ingestor to update the network fee stored in the database, but for the sake of testing, this is fine.)

```
mutation MyMutation {
  updateCurrentNetworkInverseFee(
    input: {id: "networkInverseFee", inverseFee: "50"}
  ) {
    id
    inverseFee
  }
}
```

Step 7 - Refresh the app, both the activity feed and the completed action screen should display a new value to account for the change in network fee.

Screenshot 2024-08-05 at 16 27 03

Step 8 - Complete the motion - the amount should not change after finalizing.

Screenshot 2024-08-05 at 16 27 54

Step 9 - Run the query again and set it back to 100.

```
mutation MyMutation {
  updateCurrentNetworkInverseFee(
    input: {id: "networkInverseFee", inverseFee: "100"}
  ) {
    id
    inverseFee
  }
}
```

Step 10 - Refresh the app, now that the motion has finalized the amount should not have changed.

Screenshot 2024-08-05 at 16 28 12

Further testing: Check the amount field has been unaffected by any changes when creating a simple payment action using the permissions decision method.

## Diffs

**New stuff** ✨

* `useAmountLessFee` hook

**Changes** 🏗

* Simple Payment completed action screen and activity feed descriptor uses `useAmountLessFee` hook

Resolves #2642


src/hooks/useAmountLessFee.ts
mmioana, line 1: As I believe it would provide better clarity, could we rename this hook `useAmountExcludingFee` ?

mmioana (approved):
Really cool work on this one @iamsamgibbs 🌟 

Tested creating a payment using `Reputation` (**ignore the action title**)
Screenshot 2024-08-06 at 14 28 44
Screenshot 2024-08-06 at 14 29 25
Screenshot 2024-08-06 at 14 29 45
Screenshot 2024-08-06 at 14 30 09
Screenshot 2024-08-06 at 14 31 16
Screenshot 2024-08-06 at 14 31 27
Screenshot 2024-08-06 at 14 31 40
Screenshot 2024-08-06 at 14 31 54

Also tested the simple payment action using `Permissions`
With inverseFee set to `100` 
Screenshot 2024-08-06 at 14 33 11
With inverseFee set to `50` 
Screenshot 2024-08-06 at 14 34 02
Screenshot 2024-08-06 at 14 34 25

bassgeta (approved):
Nice and clean :ok_hand: 
image
image
After updating the inverse fee:
image
Completed the motion:
image
After updating the fee back to 100:
image

src/hooks/useAmountLessFee.ts
iamsamgibbs: I think for now I will leave the naming as `lessFee` for consistencies sake as it is used in a few places across the block-ingestor - can always rename all uses at a later date if we want.

bassgeta (approved):
Reapproving the new hash :+1:

mmioana (approved):
Approve the new block ingestor hash ✅

PR #2867, opened 2024-08-05
Open UserHub on pending button click

### LINKED ISSUE #2834: Pending button no longer opens transaction tab
## Steps to reproduce

1. Create a new transaction.
2. While the pending button is visible, click on it.
3. Note that it now no longer opens up the Transactions tab.

## Expected behaviour

* The pending transaction button in the header, it should open the transactions tab in the Userhub.

## Actual behaviour

* The pending transaction button in the header click no longer works.

image

chmanie:
## Description

This PR tries to solve the UserHub opening/closing issues by just using a single state for the visibility of the UserHub. This sadly makes things a little more complicated but there are ambitions to try to solve these issues on the UX side of things (using an improved Modal flow).

For now this should do it. We got rid of the dual state and of the ominous `data-openhubifclicked` attribute by adding some state management wizardry. We basically set the tab that should open in the UserHub from outside and then reset it once it's visible.

## Testing
1. Try to have a pending TX (e.g. use MetaMask and don't confirm)
2. Click the pending TX button and hope that the UserHub opens, showing the transactions list
3. Click outside the UserHub, it should close
4. Also click some other buttons within the UserHub (e.g. Activate tokens) and check if it behaves like it does on `master`.

## Diffs

**Changes** 🏗

- Fix UserHub open state

Resolves #2834

### DIFF
--- a/src/components/common/Extensions/UserHubButton/UserHubButton.tsx
+++ b/src/components/common/Extensions/UserHubButton/UserHubButton.tsx
@@ -1,5 +1,5 @@
 import clsx from 'clsx';
-import React, { type FC, useState, useEffect } from 'react';
+import React, { type FC, useState, useEffect, useCallback } from 'react';
 import { usePopperTooltip } from 'react-popper-tooltip';
 import { useSearchParams } from 'react-router-dom';
 
@@ -12,7 +12,6 @@ import { useColonyContext } from '~context/ColonyContext/ColonyContext.ts';
 import { useTokensModalContext } from '~context/TokensModalContext/TokensModalContext.ts';
 import { TransactionStatus } from '~gql';
 import { useMobile } from '~hooks/index.ts';
-import useDetectClickOutside from '~hooks/useDetectClickOutside.ts';
 import useDisableBodyScroll from '~hooks/useDisableBodyScroll/index.ts';
 import usePrevious from '~hooks/usePrevious.ts';
 import { TX_SEARCH_PARAM } from '~routes';
@@ -26,19 +25,23 @@ import Button from '~v5/shared/Button/index.ts';
 import PopoverBase from '~v5/shared/PopoverBase/index.ts';
 import UserAvatar from '~v5/shared/UserAvatar/index.ts';
 
-import { UserHubTabs } from '../UserHub/types.ts';
+import { UserHubTab } from '../UserHub/types.ts';
 
 import { OPEN_USER_HUB_EVENT } from './consts.ts';
 
+interface Props {
+  openTab?: UserHubTab;
+  onOpen: () => void;
+}
+
 const displayName = 'common.Extensions.UserNavigation.partials.UserHubButton';
 
-const UserHubButton: FC = () => {
+const UserHubButton: FC<Props> = ({ openTab, onOpen }) => {
   const isMobile = useMobile();
   const {
     colony: { colonyAddress },
   } = useColonyContext();
   const { wallet, user } = useAppContext();
-  const [isUserHubOpen, setIsUserHubOpen] = useState(false);
   const { transactions } = useGroupedTransactions();
   const [prevGroupStatus, setPrevGroupStatus] = useState<
     TransactionStatus | undefined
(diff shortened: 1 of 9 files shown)


src/components/common/Extensions/UserHubButton/UserHubButton.tsx
```diff
@@ -52,21 +55,14 @@ const UserHubButton: FC = () => {
...
-  const ref = useDetectClickOutside({
-    onTriggered: (e) => {
-      // This stops the hub closing when clicking the pending button (which is outside)
-      if (!(e.target as HTMLElement)?.getAttribute('data-openhubifclicked')) {
-        setIsUserHubOpen(false);
-      } else {
-        setIsUserHubOpen(true);
-      }
-    },
-  });
```
iamsamgibbs, line 69: Good riddance 👋

src/components/frame/Extensions/layouts/ColonyLayout.tsx
```diff
@@ -5,13 +5,15 @@ import React, {
   type PropsWithChildren,
   useCallback,
   useEffect,
+  useState,
   // useState,
```
iamsamgibbs, line 9: This can go

iamsamgibbs (approved):
Nicely done Chris! This is much nicer to work with than before.

Tested the pending transaction:

https://github.com/user-attachments/assets/a5a1a962-483e-4d64-887f-b57e35111246

Also tested it on mobile and if a failed transaction still automatically opens the userhub which is working great:

https://github.com/user-attachments/assets/2e3f8ecd-cae4-41d7-a7e9-3c404ccd6239

This is the only issue that stood out to me, it is very slow closing the userhub after the activate / withdraw modal is opened:

https://github.com/user-attachments/assets/5d5d5b7c-e5d8-49b4-8864-56039b684090

But I just checked this on QA and it is behaving the same there - not sure if there is anything simple that can be done here to improve this, but can leave it for another time as it already exists on master.

Unrelated to this PR, but I also noticed this issue when closing the userhub on mobile if you have the action sidebar open:

https://github.com/user-attachments/assets/f06100de-5145-40ef-9db5-51ef2e8a4b3d

src/components/frame/Extensions/layouts/ColonyLayout.tsx
chmanie: Removed the comment and rebased, maybe you could re-approve?

iamsamgibbs (dismissed):

jakubcolony (changes requested):
Thanks for bringing some sanity to the UserHub @chmanie! 

The pending button now works flawlessly - I can click on any element of the button and it opens the transactions tab:

https://github.com/user-attachments/assets/1c824881-e504-4fad-b7e9-7bc1129af4f4

I noticed however the Completed button no longer triggers the UserHub, which is different to production:

https://github.com/user-attachments/assets/c51afd4d-9403-452e-a8ba-621787ba20c6


If you consider this a separate issue, ping me and I'll re-approve the pending button fix.

chmanie:
> I noticed however the Completed button no longer triggers the UserHub, which is different to production

Ah no worries, I'll fix this!

chmanie:
@jakubcolony sorted.

iamsamgibbs (approved):
Tested the fix, all good!

rdig (approved):
Works as advertised. Clicking on the pending pill brings up the correct transactions. Nicely done Chris!

[Screencast from 2024-08-07 23-52-19.webm](https://github.com/user-attachments/assets/1ff483ed-387d-4c7a-87b9-2ffc2769a2b5)

However I found two edge cases _(although they should be considered separatedly)_
- cancelling a action and not cancelling metamask will keep a pill  active, and if clicking on it will redirect you to an action that doesn't exist / 404 _(this is really edgy, I really don't think this will show up in a live environment ever)_
- cancelling metamask, or failing a transaction in some other way will add a failed transaction to the list. Clicking that failed transaction will bring you to a 404 page again. I think this can be easily replicated so I definitely think this needs to be fixed 

[Screencast from 2024-08-07 23-53-33.webm](https://github.com/user-attachments/assets/f22fa04e-875c-4996-a72c-1f72f073daa6)

chmanie:
@rdig The 404 thing is tracked by @arrenv somewhere I just couldn't find it. Feel free to open an issue for the first bug, that sounds like a bit of an explorational task

jakubcolony (approved):
The Completed button triggers the user hub now, nice one @chmanie 💯

PR #2871, opened 2024-08-06
Fix: Motion status showing as unknown

### LINKED ISSUE #2643: Motion status showing as "Unknown" in activity feed
## Steps to reproduce

This does not seem to happen in production, but it is happening in master, so would be good to avoid it going into production.
1. Ensure a colony has a motion.
2. Disconnect your wallet.
3. Connect your wallet again.
4. Check the activity feed and notice that all motions will have an "Unknown" status.
5. Reloading the page will show the correct status.

## Expected behaviour

* Motions should all show the correct status.

## Actual behaviour

* When first loading a colony, motions are showing with an "Unknown" status.

image

iamsamgibbs:
## Description

This PR fixes an issue on the local dev environment where motion status could not be properly fetched after disconnecting and reconnecting your wallet. Thanks to some digging with @rdig and @Nortsova the culprit was found.

## Testing

Step 1 - Install the voting reputation extension
Step 2 - Create a motion and stake to support it
Step 3 - Disconnect your wallet
Step 4 - Connect your wallet again
Step 5 - Check the motions in the activity feed and the userhub have the correct status

(On master they will have the "Unknown" status)

Test with both a metamask wallet and a dev wallet.

## Diffs

**Changes** 🏗

* Properly initialize signer for VotingReputationFactory

Resolves #2643

### DIFF
--- a/src/hooks/useNetworkMotionStates.ts
+++ b/src/hooks/useNetworkMotionStates.ts
@@ -1,5 +1,4 @@
 import { type MotionState, VotingReputationFactory } from '@colony/colony-js';
-import { type Provider } from '@ethersproject/providers';
 import { useEffect, useState } from 'react';
 
 import { useAppContext } from '~context/AppContext/AppContext.ts';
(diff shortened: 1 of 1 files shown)


Nortsova (approved):
Thank you @iamsamgibbs and @rdig for solving this issue. 

Can confirm that statuses shown correctly: 
Metamask wallet:
metamask-wallet
Dev wallet:
dev-wallet

jakubcolony (approved):
Your branch, after reconnecting ganache and metamask wallets:
image

image

`master`:
image

LGTM 👍

Issue #2862, opened 2024-08-05
Pending button styling breaks in Adv. payments modal


arrenv:
Teamwork task: [Fixes before split payment release](https://tw.auroracreation.com/#/tasks/16025768)

## Steps to reproduce

1. Create an advanced payment action.
2. Continue through the flow and Fund or Release the payment.
3. In the Funding or Release modal note the styling of the confirm button while pending.

## Expected behaviour

* The button styling should be consistent and as of the design.

## Actual behaviour

* The styling of the button breaks while the transaction is pending.

![image](https://github.com/user-attachments/assets/7fce0227-f03e-46e8-ab4a-82454b65ba62)

![button-change](https://github.com/user-attachments/assets/50925e57-0e1a-4faf-86ca-3164f5fe93f5)

melyndav:
Can this loading button be added for the cancel model within Advanced payments? There's no consistency with the modals in actions. 

<img width="670" alt="image" src="https://github.com/user-attachments/assets/a99d8895-bcf7-45bf-968a-ced647561abe">

Issue #2896, opened 2024-08-07
Permission banner on Staking advanced payment extension


arrenv:
Teamwork task: [Fixes before split payment release](https://tw.auroracreation.com/#/tasks/16025768)

## Steps to reproduce

1. Go to the extension page of a colony.
2. Go to the Staking advanced payments extension.
3. Install the extension.
5. Complete enabling.
6. Note the permissions banner.

## Expected behaviour

* The permission banner should not appear after successful installation.

## Actual behaviour

* When installing the Staking advanced payment extension, the permissions banner is visible after successful installation.
* It does not disappear until the app has been been reloaded.

![image](https://github.com/user-attachments/assets/607e79b4-7732-4c46-88bd-1b068d889803)

Issue #2890, opened 2024-08-06
Multiple `getExpenditure` queries fetched at the same time 


jakubcolony:
Teamwork task: [Fixes before split payment release](https://tw.auroracreation.com/#/tasks/16025768)

## Steps to reproduce

This issue can be reproduced on `feat/advanced-payments-ui` branch or in QA.

1. Create an advanced payment.
2. Open devtools network tab or GraphQL Network extension and observe the `GetExpenditure` queries.

## Expected behaviour

The expenditure is polled no more frequently than the current network's block time - 5 seconds in local dev.

## Actual behaviour

Multiple requests are either being sent at the same time or in very short succession:

https://github.com/user-attachments/assets/a86ac242-253d-4079-b50b-d1c2084de718

This leads to increased bandwidth consumption of the app (which may be of concern to mobile data users) and potentially increased AWS AppSync bills.

Issue #2879, opened 2024-08-06
Advanced payment funding motion, time spinner remained after motion failed


arrenv:
Teamwork task: [Fixes before split payment release](https://tw.auroracreation.com/#/tasks/16025768)

## Steps to reproduce

1. Create an advanced payment.
2. Take the advanced payment to the funding step and use the Reputation decision method.
3. In my case I fully staked and fully objected and took it to voting.
4. I then opposed the vote.
5. The motion outcome correctly failed.
6.  However, note that the timer loader was not removed.

## Expected behaviour

* The phase time spinner should not be visible if there are no active phases.

## Actual behaviour

* The phase time spinner remains even though the motion phases have ended. Even after reloading the page.

![image](https://github.com/user-attachments/assets/66171ff6-26ec-4d8f-b7a6-d2a46bf753c0)

PR #2902, opened 2024-08-08
fix: button width, hide timer, cache query, remove extension banner

CzarekDryl:
## Description

- Few fixes to unblock split payment
- It's best to go through each issue one by one

Resolves:
https://github.com/JoinColony/colonyCDapp/issues/2862
https://github.com/JoinColony/colonyCDapp/issues/2879
https://github.com/JoinColony/colonyCDapp/issues/2890
https://github.com/JoinColony/colonyCDapp/issues/2896

### DIFF
--- a/src/components/frame/Extensions/pages/Extensions/ExtensionDetailsPage/partials/ActionButtons/ActionButtons.tsx
+++ b/src/components/frame/Extensions/pages/Extensions/ExtensionDetailsPage/partials/ActionButtons/ActionButtons.tsx
@@ -26,6 +26,7 @@ const ActionButtons: FC<ActionButtonProps> = ({
   isSetupRoute,
   extensionStatusMode,
   extensionStatusText,
+  setIsEnabling,
 }) => {
   const { user } = useAppContext();
   const { colony } = useColonyContext();
@@ -78,6 +79,7 @@ const ActionButtons: FC<ActionButtonProps> = ({
         <InstallButton
           extensionData={extensionData}
           isVisible={isInstallButtonVisible}
+          setIsEnabling={setIsEnabling}
         />
         <EnableButton
           extensionData={extensionData}
--- a/src/components/frame/Extensions/pages/Extensions/ExtensionDetailsPage/partials/ActionButtons/types.ts
+++ b/src/components/frame/Extensions/pages/Extensions/ExtensionDetailsPage/partials/ActionButtons/types.ts
@@ -6,6 +6,7 @@ export interface ActionButtonProps {
   extensionData: AnyExtensionData;
   isSetupRoute: boolean;
   waitingForActionConfirmation: boolean;
+  setIsEnabling: SetStateFn<boolean>;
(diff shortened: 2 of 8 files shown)


davecreaser (approved):
Thanks for the fixes @CzarekDryl. Everything is looking alright to me, and I've tested each of the issues individually. Nice work 🎉 

The button width is fixed on the modals:
1_button_fixed

The loader is now not shown when it shouldn't be:
2_loader_not_shown

The graphql requests seem to be much more _relaxed_:
https://github.com/user-attachments/assets/43dd6726-e74c-4a4c-99df-6dce5d340ef4

And the permissions banner is no longer shown on the extensions page.
4_permissions_banner

Just one note: in the future could you please provide a bit more detail in the PR description? For example describing a little bit more in depth the changes that you've made and what exactly it fixes, and adding some testing steps. I realise that you added the links to the issues and that does help of course, but it makes it more easy to review the PR if it's consolidated in one place, with more in depth descriptions and testing steps. Thank you!

jakubcolony (changes requested):
Nice work @CzarekDryl, most of the issues got resolved, and I have some comments regarding the last one. 

#2862 ✅ 
image

#2879 I was not able to reproduce with both passed and failed vote outcome ✅ 
image
image

#2890 Not sure changing the fetch policy had much effect on the number of requests but after discussing with Arren I consider this resolved and to be followed up in #2900 ✅ 

#2896 The permissions needed banner should not have been removed, instead the cause of the issue needs to be resolved, which is most likely the fact that permissions are not updated in cache after they have been set for the staking extension. 

At the moment there's a regression of permissions banner not showing for staged expenditure even though permissions are indeed missing:
image

Feel free to ping me if you have any questions regarding the cache update.

CzarekDryl:
@jakubcolony Permission changes reverted and added description

jakubcolony (changes requested):
@CzarekDryl the banner still shows up for staking extension even though permissions have been set as part of the installation:
image

image

The `refetchColony` bit needs to be called after roles have been automatically set for staking extension, not after clicking on the banner (you shouldn't see it in the first place).

CzarekDryl:
@jakubcolony Now it uses refetchColony after installing

arrenv (changes requested):
@CzarekDryl Thank you for tackling these issues.

Is it possible to prevent the permissions banner from flashing up at all?

[permission-banner.webm](https://github.com/user-attachments/assets/fd76848b-2dd4-463b-a903-e3d02e5e979a)

CzarekDryl:
@arrenv Please check it now

jakubcolony (approved):
LGTM, thanks for your work @CzarekDryl 👍 

image

Can confirm there's no flickering after setting permissions.

arrenv (approved):
@CzarekDryl The installation all went smoothly and the extension shows up with the correct permissions. Thank you for the updates.

image


I ran into this issue though when doing an advanced payment with staking. Is this unrelated? Does it need a separate issue?

image

Issue #2840, opened 2024-08-02
Redo action does not include member


arrenv:
Teamwork task: [Redo action does not include member](https://tw.auroracreation.com/#/tasks/16019592)

## Steps to reproduce

1. Create a Simple payment action.
2. On the completed actions screen, click the kebab menu.
3. Click the "Redo action" option.
4. Note that all fields are copied, except the Member field.

## Expected behaviour

* The member field should also be copied over when redoing an action.
* Redo action should be visible in the activity feed kebab menu as well.
* Redo action should be available for all Advanced payment options, including Staged and Split.

## Actual behaviour

* The member field is not pre-filled when using the redo action option.
* The redo action option should also be available from the activity feed for payment related actions as of the original spec.
* Redo action is missing from Advanced payment actions, including Staged and Split.

![image](https://github.com/user-attachments/assets/98de3f23-9ac6-4d2b-9e30-c46b5fb7312c)

![image](https://github.com/user-attachments/assets/f509c6ac-b9f4-4441-9e0f-1a58849322af)

![image](https://github.com/user-attachments/assets/6f995e48-071d-4d5f-b05d-a7e1bd5e15cf)

AdrianDudko:
Closed by https://github.com/JoinColony/colonyCDapp/pull/2925

PR #2925, opened 2024-08-12
feat: added redo action to adv payments

CzarekDryl:
## Description

- The member field should also be copied over when redoing an simple payment action.
- Redo action should be visible in the activity feed kebab menu as well.
- Redo action should be available for all Advanced payment options, including Staged and Split.

## Testing

* Check if the action is 'copied' after selecting redo action option from meatball menu in activity table

## Diffs

**New stuff** ✨

* Added redo action option to activity table on homepage and activity page

Resolves https://github.com/JoinColony/colonyCDapp/issues/2840

### DIFF
--- a/src/components/v5/frame/ColonyHome/ColonyHome.tsx
+++ b/src/components/v5/frame/ColonyHome/ColonyHome.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 
 import FiltersContextProvider from '~common/ColonyActionsTable/FiltersContext/FiltersContextProvider.tsx';
 import ColonyActionsTable from '~common/ColonyActionsTable/index.ts';
@@ -15,6 +15,7 @@ import {
 import { formatText } from '~utils/intl.ts';
 import { setQueryParamOnUrl } from '~utils/urls.ts';
 // import TmpAdvancedPayments from '~v5/payments/TmpAdvancedPayments.tsx';
+import useGetActionData from '~v5/common/ActionSidebar/hooks/useGetActionData.ts';
 import Link from '~v5/shared/Link/index.ts';
 
 import Agreements from './partials/Agreements/index.ts';
@@ -32,6 +33,10 @@ const ColonyHome = () => {
   const isMobile = useMobile();
   const selectedDomain = useGetSelectedDomainFilter();
   const teamsBreadcrumbs = useCreateTeamBreadcrumbs();
+  const [selectedAction, setSelectedAction] = useState<string | undefined>(
+    undefined,
+  );
+  const { defaultValues } = useGetActionData(selectedAction || undefined);
 
   useSetPageBreadcrumbs(teamsBreadcrumbs);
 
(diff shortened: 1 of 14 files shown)


arrenv (changes requested):
@CzarekDryl The redo is working great, nice one!!

I noticed a couple of things, they may or may not be elated to this functionality, but raising here just in case it is.

1. If you have Split payment that is being re-done that has odd fractions, you get a long value that appears. I can understand that this is probably how this feature is able to work, but, it could be worth going the other way, so that you are spending less, rather then more then the entered value. NOTE: Removing this requirement to it's own issue here - https://github.com/JoinColony/colonyCDapp/issues/3012

[split-redo.webm](https://github.com/user-attachments/assets/8572cf70-5aad-49ea-b201-53aa2858c139)

2. If you redo an action then change the action type to another payment action, including simple payment, changing the amount field auto triggers the not enough token validation incorrectly.

[validation-redo.webm](https://github.com/user-attachments/assets/68ed14d8-ff56-4d20-b8ec-8317f85f7f94)

CzarekDryl:
@arrenv Everything should work fine now, please check it :)

arrenv (approved):
@CzarekDryl Looks to be working great, thank you!

arrenv:
Just want to note that not this PR, but this branch has re-introdced the width issue on the amount field if you type anything but a number in.

image

jakubcolony (changes requested):
@CzarekDryl please try to create a payment with the following values:
image

The app crashes whenever I try to redo such payment:

https://github.com/user-attachments/assets/7a07b126-27f0-43e1-a5c6-3de15af7b801

CzarekDryl:
@jakubcolony This should work as expected now.

https://github.com/user-attachments/assets/60cdaba3-1642-4857-963b-60061974c556

src/components/v5/frame/ColonyHome/ColonyHome.tsx
```diff
@@ -32,6 +33,10 @@ const ColonyHome = () => {
   const isMobile = useMobile();
   const selectedDomain = useGetSelectedDomainFilter();
   const teamsBreadcrumbs = useCreateTeamBreadcrumbs();
+  const [selectedAction, setSelectedAction] = useState<string | undefined>(
+    undefined,
+  );
+  const { defaultValues } = useGetActionData(selectedAction || undefined);
```
jakubcolony, line 39: Do we need two such states, in both `ColonyHome` and `ActivityPage` or could it live in `ColonyActionsTable` instead?

jakubcolony (approved):
Tested with different values including min/max allowed, permissions/staking decision method and split & staged payments, all of which seem to be working fine:

image

image

image

Nice work @CzarekDryl

src/components/v5/frame/ColonyHome/ColonyHome.tsx
CzarekDryl: @jakubcolony I tried to add it to the`ColonyActionsTable` but there was a problem with rerenders

Issue #2909, opened 2024-08-08
Multi-sig Extension has multiple grouped transactions


arrenv:
## Steps to reproduce

1. Ensure you have the Multi-sig extension uninstalled.
2. Click to install it.
3. Note that it will require 2 separate grouped transactions.

## Expected behaviour

* We should minimise the number of transactions as much as possible, such as by using multi-call. If we need to have dependent transactions, then they should be in one set of grouped transactions.

## Actual behaviour

* Installing the Multi-sig extension requires two separate grouped transactions

bassgeta:
There has been a lot of work already done on this [here](https://github.com/JoinColony/colonyCDapp/pull/2361/files#diff-c9b729ba88276c82efc4ab03b7967a5abdc06d7acf42db582f7b87df0bed2a26), so as agreed, we will be postponing this until we merge both branches into `master` and then realign all necessary extension installation flows.
To alleviate double transactions, we will be adding the "Permissions needed" banner back, done [here](https://github.com/JoinColony/colonyCDapp/pull/2926)

PR #2926, opened 2024-08-12
Fix: remove auto-enabling for MultisigPermissions extension in PermissionsNeededBanner

### LINKED ISSUE #2919: [Multi-sig] Transaction auto triggered when loading multi-sig ext. (without required permissions)
## Steps to reproduce

Note: I can understand why this is being done, but, it does not work if you don't have permissions and you don't get any feedback as to why.

1. Remove permissions from the Multi-sig extension.
2. Navigate to the Multi-sig extension page.
3. A transaction will automatically be triggered, with no context as to why or what it is for.

## Expected behaviour

* There should not be any auto triggered transactions when loading a page.
* You should see the missing permissions banner if there are permissions missing, that only users with the right permissions can trigger.

## Actual behaviour

* You get an auto triggered transaction when navigating to the Multi-sig extension page, which, does not have it's required permissions.
[auto-transaction.webm](https://github.com/user-attachments/assets/b2c073a2-2736-4d94-9c07-d3ed664fd59f)

* The errors if you visit the page if you don't have the required permissions.
image

* Result of switching connected wallet while on the page, without reloading. This also locked me into the account, as it would auto trigger the transaction and that failed transaction prevented me from disconnecting my wallet. I have to navigate away, reload the app, then disconnect the wallet.
image

mmioana:
Fix: remove auto-enabling for MultisigPermissions extension in PermissionsNeededBanner

## Description

> [!IMPORTANT]
> This PR only handles showing back the banner; enabling the permissions in a multi-transaction after the extension is installed will be tackled in issue [#2909](https://github.com/JoinColony/colonyCDapp/issues/2909)

You should see the missing permissions banner if there are permissions missing, that only users with the right permissions can trigger.

## Testing

TODO: Check the multi-sig permissions banner shows up after installing the Multi-Sig extension

* Step 1. Login as `leela` and go to Multi-Sig extension page
* Step 2. Install extension 
Screenshot 2024-08-12 at 15 37 35
* Step 3. Check permissions banner shows up but **don't** enable the permissions yet
Screenshot 2024-08-12 at 15 37 47
* Step 4. Login as `fry` and check permissions banner shows up, but the `Enable permissions` link doesn't show up
Screenshot 2024-08-12 at 15 38 05
* Step 5. Login as `leela` and enable the permissions for the extension
* Step 6. Check the banner doesn't show up
Screenshot 2024-08-12 at 15 38 23

## Diffs

**Changes** 🏗

* Permissions banner is back in place for Multi-Sig extension

Resolves #2919

### DIFF
--- a/src/components/frame/Extensions/pages/ExtensionDetailsPage/PermissionsNeededBanner.tsx
+++ b/src/components/frame/Extensions/pages/ExtensionDetailsPage/PermissionsNeededBanner.tsx
@@ -1,6 +1,6 @@
-import { ColonyRole, Extension, Id } from '@colony/colony-js';
+import { ColonyRole, Id } from '@colony/colony-js';
 import { CheckCircle, WarningCircle } from '@phosphor-icons/react';
-import React, { useEffect, useState } from 'react';
+import React, { useState } from 'react';
 import { FormattedMessage, defineMessages } from 'react-intl';
 
 import { useAppContext } from '~context/AppContext/AppContext.ts';
@@ -37,8 +37,6 @@ interface Props {
 }
 
 const PermissionsNeededBanner = ({ extensionData }: Props) => {
-  const shouldDisplay =
-    extensionData.extensionId !== Extension.MultisigPermissions;
   const { colony } = useColonyContext();
   const { user } = useAppContext();
   const { checkExtensionEnabled } = useCheckExtensionEnabled(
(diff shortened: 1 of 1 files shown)


bassgeta (approved):
Looking good, we can keep this code intact and let's handle enabling permissions automatically in a separate issue :D 
image
And after clicking enable:
image

let's go!

rumzledz (approved):
Clean and straightforward fix @mmioana ✨ 

Left is Leela, Right is Fry
ms enable

iamsamgibbs (approved):
Great work on this!

Screenshot 2024-08-21 at 09 20 25

As a user without permissions:

Screenshot 2024-08-21 at 09 20 46

PR #2927, opened 2024-08-12
feat: streaming payment form

joanna-pagepro:
## Description

Streaming payments form with validation

## Testing

* Step 1. Open the action sidebar and select the `Streaming payment` action type
* Step 2. If the extension is not installed, then you should see an error, and the submit button should be disabled, otherwise fill out the form and click the `Submit` button.

## Diffs

**New stuff** ✨

* Streaming Payment Form

**Changes** 🏗

* 

**Deletions** ⚰️

* 

## TODO

- 


jakubcolony (changes requested):
Nice start with streams @joanna-pagepro, code looks good but I have some feedback on the validation:

Extension installation validation ✅ 
image

----

Setting start date ✅ 
image
image

----
 Setting end date ❌ 
 - Despite the validation preventing from selecting a past date, I was able to select an hour in the past which is gonna fail on the contracts:
    
image

- This also affects "Start immediately option":
image


----

Interval/Amount per field ❌ 

- The validation message is in different units than the custom interval input and does not show units at all:
image

- There's no upper limit for custom interval:
image

----

Limit field ❌ 

- I think the limit field validation should trigger when different end condition is selected:
- 
https://github.com/user-attachments/assets/5683906f-3ef7-4685-9d6b-e76c6ab53153

arrenv:
> * There's no upper limit for custom interval:
> 
> image

The max limit for this field in the UI should be 5 digits. It is arbitrary but allows for ~274 years, which seems reasonable for now.

joanna-pagepro:
@jakubcolony @arrenv updated. Can you review it again?

jakubcolony (changes requested):
Thank you for resolving the previous issues @joanna-pagepro

I'm getting a validation error if I select "Hour" in "Amount per" dropdown - looks like the validation is conflicting with the available options:
image

- Selecting an end date in the past is no longer possible:
image

- Changing the end condition triggers validation as expected

- Custom interval is limited to 5 digits

arrenv (changes requested):
@joanna-pagepro Looks great, nice work.

I ran into the same issues as @jakubcolony just did, this also seems to apply if you use decimals in the Custom field.

There should be a limit of 5 decimal places on this field, so, the value can be less then 1 day.

image

joanna-pagepro:
@arrenv @jakubcolony I will update validation for hour value, but I found this in the docs: 
image
and it says that the minimum in the custom field is 1, so you can't use decimals there

joanna-pagepro:
@jakubcolony validation updated

arrenv:
> @arrenv @jakubcolony I will update validation for hour value, but I found this in the docs: image and it says that the minimum in the custom field is 1, so you can't use decimals there

I don't recall where that requirement of no less than 1 came from, I don't recall writing it. But happy to stick with it for now.

joanna-pagepro:
@arrenv error message updated

arrenv (approved):
@joanna-pagepro Thank you, looks all good on my end.

jakubcolony (approved):
Looks good to me, nice work @joanna-pagepro 👍 

image

Issue #2913, opened 2024-08-09
[Multi-sig] bug: mint tokens via multi sig has weird behaviour


bassgeta:
## Steps to reproduce

1. Create and finalize a mint tokens motion
2. Notice that the tokens don't immediately get claimed, instead they show up in the balances page as claimable
![image](https://github.com/user-attachments/assets/33ee6b02-52ed-4b53-8fbf-8adf59e45226)
![image](https://github.com/user-attachments/assets/9159012f-6007-41f6-acc5-e9732f583065)

Additionally if you create a mint tokens action via permissions, all the other mint token claims get automatically claimed.

## Expected behaviour

The tokens should get minted and immediately claimed.

## Actual behaviour

An outstanding claimable income is created.

Issue #2506, opened 2024-06-10
Mint tokens motion does not claim new tokens


arrenv:
## Steps to reproduce

1. Create a motion to mint tokens.
2. Go through the process to ensure it passes.
3. Finalize the motion.
4. Check the balance of the Colony, it would not have changed.
5. Check the incoming funds page, it will show the newly minted tokens, available to claim.

## Expected behaviour

* Mint tokens actions, regardless of decision method, should claim the tokens on execution.

## Actual behaviour

* Newly minted tokens are not claimed when using Reputation decision method.
![image](https://github.com/JoinColony/colonyCDapp/assets/33682027/e4e2d449-cda6-4f4b-964b-a1dcdb4c65f6)

   #### Minted tokens claimed with permissions
   ![image](https://github.com/JoinColony/colonyCDapp/assets/33682027/c0a151d1-4cf7-4b3c-a652-c03aaab2de70)

   #### Minted tokens with motion incoming funds
   ![image](https://github.com/JoinColony/colonyCDapp/assets/33682027/0a14cfa3-0542-45cf-9600-b4f59131e1c6)

jakubcolony:
This will most likely require a multicall motion, please ping me if you're planning on taking this issue to ensure it doesn't impact the expenditures multicall motions

jakubcolony:
Closing this issue in favour of #2913

PR #2932, opened 2024-08-13
Fix: claim colony funds after mint tokens motion is finalized

mmioana:
## Description

- Trigger the MOTION_CLAIM_MINT_TOKENS saga after clicking the Finalize button for a Mint tokens action
- Note: Created a new saga as we might want to re-work how we're claiming incoming funds - currently the block ingestor will claim all available funds within a colony for the given token

## Testing

TODO: Incoming funds resulted from a MINT_TOKENS motion are claimed and no longer appear as funds that need to be accepted. 
Screenshot 2024-08-13 at 12 37 00

* Step 1. Enable `Multi-sig Permissions` extension (no need to change the default settings)
* Step 2. Give `leela` Owner permissions to Take actions via Multi-sig
Screenshot 2024-08-13 at 12 46 13
* Step 3. Create a Mint Tokens with Multi-Sig decision method
Screenshot 2024-08-13 at 12 46 49
* Step 4. Finalize motion
Screenshot 2024-08-13 at 12 47 15
* Step 5. Check minted tokens are already claimed in the Incoming Funds page
Screenshot 2024-08-13 at 12 47 32
* Step 6. Enable `Reputation Weighted (Lazy Consensus method)` extension 
* Step 7. Choose `Testing governance` as extension settings and enable it
Screenshot 2024-08-13 at 12 48 47
* Step 8. Create Mint Tokens with Reputation decision method
* Step 9. Support motion by staking at least 10%
Screenshot 2024-08-13 at 12 50 10
* Step 10. Run `npm run forward-time 10`
* Step 11. Refresh page and finalize motion
Screenshot 2024-08-13 at 12 51 04
* Step 12. Check minted tokens are already claimed in the Incoming Funds page
Screenshot 2024-08-13 at 12 51 18

## Diffs

**New stuff** ✨

* MOTION_CLAIM_MINT_TOKENS redux saga

Resolves [#2913](https://github.com/JoinColony/colonyCDapp/issues/2913)


rdig (changes requested):
Very nice work! I've tested this and seems to be working as expected. Both Multisig and Motions will auto-claim the incoming funds

Screenshot from 2024-08-14 01-27-05
Screenshot from 2024-08-14 01-27-38
Screenshot from 2024-08-14 01-27-45
Screenshot from 2024-08-14 01-31-46
Screenshot from 2024-08-14 01-31-55

However I noticed a issue _(which might or might not be related to this)_. The colony balance doesn't update. No matter if it's via Multisig or via Motions, after a mint and automatic claim, the balance of the colony stays the same

Screenshot from 2024-08-14 01-29-37
Screenshot from 2024-08-14 01-29-49

This needs to be fixed, preferably here, as it's major problem if the balance gets out of date with the chain state.

mmioana:
Thank you @rdig for reviewing it 🙏 I'll check the issue regarding the balance update

mmioana:
Hey @rdig I tried to mint a larger amount of tokens and the balance seems to be updating

https://github.com/user-attachments/assets/03480b2e-cd6a-4c11-ad0a-d2af04651b3b


I'm wondering if the balance update was not visible only because there was a too small amount of tokens minted? 🤔

src/components/v5/common/ActionSidebar/partials/hooks.ts
```diff
@@ -81,3 +88,44 @@ export const useIsFieldDisabled = () => {
 
   return false;
 };
+
+export const useFinalizeSuccessCallback = (): FinalizeSuccessCallback => {
+  const { startPollingColonyData, stopPollingColonyData } = useColonyContext();
```
bassgeta, line 93: Dropped you a DM also, but I think we can do this without polling and it should work just fine :ok_hand:

bassgeta (changes requested):
Works like a charm, maybe we can remove polling so it's a bit less API spammy :D  
Other than that, balance before:
image
Finalized a mint tokens Multi-sig
image
And it's updated :sunglasses: 
image

jakubcolony (changes requested):
Nice start on the PR @mmioana!

I noticed this issue is overlapping with #2506. Ideally, the solution for both multi sig and "regular" motions would be to include the claim transaction in the motion's action. However, as far as I'm aware, it is not supported by multisig (and definitely not supported by regular motions). This is great as it gets us unblocked in both cases.

The only reason for requesting changes is the saga duplication, we've always had the ability to claim funds (e.g. the "Accept all funds" button in the Incoming Funds page) via the `CLAIM_TOKEN` redux action and it should be used here as well.

mmioana:
> Nice start on the PR @mmioana!
> 
> I noticed this issue is overlapping with #2506. Ideally, the solution for both multi sig and "regular" motions would be to include the claim transaction in the motion's action. However, as far as I'm aware, it is not supported by multisig (and definitely not supported by regular motions). This is great as it gets us unblocked in both cases.
> 
> The only reason for requesting changes is the saga duplication, we've always had the ability to claim funds (e.g. the "Accept all funds" button in the Incoming Funds page) via the `CLAIM_TOKEN` redux action and it should be used here as well.

Hey @jakubcolony thank you for noticing the overlapping issue 🌟 

The reason I decided to create the new saga was due to the fact we might want to re-work how we're claiming the incoming funds, as currently we're accepting all incoming funds and not only the ones related to a specific motion/action. 

However, this actually needs to be clarified with Product (will do so) and maybe then we can create a separate issue 🙌

bassgeta (approved):
Nice solution, just a handler and stuff **just** works :D
Automatic claim of my 111 CREDS multisig
image
image

Voting reputation:
image
image

mmioana:
@jakubcolony after discussing with Arren indeed it seems there is no issue in claiming all incoming funds for the same token in a colony. So I'll proceed with removing the duplicated saga 🙌

jakubcolony (approved):
> @jakubcolony after discussing with Arren indeed it seems there is no issue in claiming all incoming funds for the same token in a colony. So I'll proceed with removing the duplicated saga 🙌

Sounds good! It is actually impossible to "choose" funds you claim, you have to do it for the entire balance in given token.

Thanks for addressing my feedback, all looks good now 👍 

Before finalizing the motion:
image

image


After:
image

mmioana:
> > @jakubcolony after discussing with Arren indeed it seems there is no issue in claiming all incoming funds for the same token in a colony. So I'll proceed with removing the duplicated saga 🙌
> 
> Sounds good! It is actually impossible to "choose" funds you claim, you have to do it for the entire balance in given token.
> 
> Thanks for addressing my feedback, all looks good now 👍
> 
> Before finalizing the motion: image
> 
> image
> After: image

Thank you @jakubcolony for all the valuable feedback! 🌟

PR #2936, opened 2024-08-13
Fix: enforce transparent background for the tabs

### LINKED ISSUE #2839: Orange backgrounds are showing up on all tabs
## Steps to reproduce

This is happening on QA at the moment, it seems strange that it keeps coming back. You can see it on all tab pages, such as Members, account area.

## Expected behaviour

* The tabs should match the design.

## Actual behaviour

* The tabs background is currently showing as orange.

image

mmioana:
## Description

- In order to not encounter anymore the orange background for the tabs, this PR enforces a transparent background

## Testing

Please check the background for an active tab is transparent.

Places to check: 
- Extensions page
- User Account page
- Members page
- Permissions page
- User hub -> Stakes 

Resolves #2839

### DIFF
--- a/src/components/shared/Extensions/Tabs/Tabs.tsx
+++ b/src/components/shared/Extensions/Tabs/Tabs.tsx
@@ -38,7 +38,7 @@ const Tabs: FC<PropsWithChildren<TabsProps>> = ({
       {items.map(({ id, title, notificationNumber }) => (
         <Tab
           key={id}
-          className={clsx({
+          className={clsx('!bg-transparent', {
             '!font-semibold': id === activeTab,
             'text-gray-700': id !== activeTab,
           })}
--- a/src/styles/variables.css
+++ b/src/styles/variables.css
@@ -15,6 +15,7 @@
     77
   ); /* Used for colony name display (colony simplified) */
 
+  --rts-primary-color: transparent !important; /* Besides using the bg-transparent for a Tab item, we enforce rts to use transparent as primary color */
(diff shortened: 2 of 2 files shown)


bassgeta (approved):
Nice one, everything is transparent now!
I was only wondering if it would still work without the !important in variables.css or the !important in Tabs.tsx?
image
image
image
image
image

mmioana:
> Nice one, everything is transparent now! I was only wondering if it would still work without the !important in variables.css or the !important in Tabs.tsx? image image image image image

In the variables.css the `!important` is needed only due to the order in which the css file from the `react-tabs-scrollable` loads.
Also, I added the `!important` for the tailwind css class to overwrite some custom css that is applied to all buttons (and now I can not find the custom css file from where that styling comes from 🫠)

Nortsova (approved):
@mmioana thanks for this brilliant fix!

### Stakes:
| Before | After |
| ----- | ------ | 
| image | image |

### All members:
| Before | After |
| ----- | ------ | 
| image | image |

### Extension:
| Before | After |
| ----- | ------ | 
| image | image |

### Manage account:
| Before | After |
| ----- | ------ | 
| image | image |

### Permissions:
| Before | After |
| ----- | ------ | 
| image | image |

P.S. It was hard to test because this issue doesn't exist on the master branch. I switched to feat/advanced-payments-ui, installed all dependencies, then switched back to the master branch, and only then was I able to see the orange background. And this fix are perfectly solve it 🤩🥰

PR #2938, opened 2024-08-13
feat: streaming payment preview base

joanna-pagepro:
## Description

Streaming Payment action preview for flow with permissions decision method. 
Not included:
- `Pay funds` action.
- activity table action description update

## Testing

* Step 1. Create `Streaming payment` action using tmp form from dashboard page or streaming payment form when merged
* Step 2. Check action preview in the sidebar

## Diffs

**New stuff** ✨

* Streaming payment completed action components

**Changes** 🏗

* 

**Deletions** ⚰️

* 

## TODO

- [x] Action Sidebar status pill change
- [x] Mobile styles


src/components/v5/common/ActionSidebar/hooks/useGetStreamingPaymentData.ts
```diff
@@ -0,0 +1,63 @@
...
+) => {
+  const pollInterval = getSafePollingInterval();
+
+  const { data, refetch, loading, startPolling, stopPolling } =
+    useGetStreamingPaymentQuery({
+      variables: {
+        streamingPaymentId: streamingPaymentId || '',
+      },
+      skip: Number.isNaN(streamingPaymentId),
+      fetchPolicy: 'network-only',
```
jakubcolony, line 19: Perhaps `cache-and-network` would work better here, to avoid making network calls more often than polling interval

src/components/v5/common/CompletedAction/partials/StreamingPayment/partials/StreamingPaymentWidget/hooks.ts
```diff
@@ -0,0 +1,53 @@
...
+) => {
+  const {
+    colony: { tokens: colonyTokens, balances },
+  } = useColonyContext();
+  const { balancesByToken: expenditureBalances, loading } =
+    useColonyExpenditureBalances();
+
+  const tokensData = useMemo(
+    () =>
+      colonyTokens?.items.filter(notNull).map((item) => {
```
jakubcolony, line 22: You don't need to do mapping on all tokens, only on the one with `tokenAddress`

src/components/v5/common/CompletedAction/partials/StreamingPayment/partials/StreamingPaymentWidget/hooks.ts
```diff
@@ -0,0 +1,53 @@
...
+          domainId,
+        );
+
+        let totalBalance = colonyTokenBalance;
+        if (domainId) {
+          const expenditureTokenBalance =
+            expenditureBalances[item.token.tokenAddress] ?? 0;
+
+          totalBalance = totalBalance.add(expenditureTokenBalance);
+        }
```
jakubcolony, line 35: Balances held in expenditures should not be considered here, they cannot be used when claiming a stream

jakubcolony (changes requested):
Overall looks good, thanks for your work @joanna-pagepro

The stream only shows the "Not Started" status, that's fine if it's out of scope of this PR. Otherwise the view looks good:
image

I can confirm the claimable amount is updating, although I needed to refresh the page for that, I assume polling for stream updates will be implemented in the coming PRs.
image

joanna-pagepro:
@jakubcolony start date is the 7th of November and it's in the future, so the status is `Not started`. I don't understand why you can have anything to claim here right now. Can you explain why does it work this way? Are you using a different date or something on the backend side?

joanna-pagepro:
@jakubcolony I've updated my code and am now comparing start and end times with `blockTime`. Is it the correct way to get the right status?

jakubcolony:
@joanna-pagepro this is not current time but local blockchain time which is typically a few months ahead. And yes, the comparison should definitely be against the block time 👍

src/components/v5/common/CompletedAction/partials/StreamingPayment/partials/FinalizeWithPermissionsInfo/FinalizeWithPermissionsInfo.tsx
```diff
@@ -0,0 +1,131 @@
...
+  return (
+    <>
+      <FormattedDate value={date} day="numeric" month="short" year="numeric" />{' '}
+      {formatText(MSG.at)}{' '}
+      <FormattedDate value={date} hour="numeric" minute="numeric" />
+    </>
+  );
+};
+
+const FinalizeWithPermissionsInfo: FC<FinalizeWithPermissionsInfoProps> = ({
```
jakubcolony, line 65: This shouldn't be called FinalizeWithPermissionsInfo, it shows the stream creation

src/components/v5/common/CompletedAction/partials/StreamingPayment/partials/FinalizeWithPermissionsInfo/FinalizeWithPermissionsInfo.tsx
```diff
@@ -0,0 +1,131 @@
...
+    id: `${displayName}.yestardayAt`,
+    defaultMessage: 'Yesterday at',
+  },
+  at: {
+    id: `${displayName}.at`,
+    defaultMessage: 'at',
+  },
+});
+
+const formatDate = (value: string | undefined) => {
```
jakubcolony, line 31: This should be put into a util rather than duplicated

src/components/v5/common/CompletedAction/partials/StreamingPayment/partials/StreamingPaymentWidget/StreamingPaymentWidget.tsx
```diff
@@ -0,0 +1,286 @@
...
+  } = useActionSidebarContext();
+
+  const { amountAvailableToClaim, amountClaimedToDate } = useMemo(
+    () =>
+      getStreamingPaymentAmountsLeft(
+        streamingPayment,
+        Math.floor(blockTime ?? Date.now() / 1000),
+      ),
+    [blockTime, streamingPayment],
+  );
```
jakubcolony, line 84: Couldn't the existing hook be used?

jakubcolony (changes requested):
Thanks for your continued work @joanna-pagepro

A few things I spotted:

- There is no loading state of the Pay funds button:
image

- There is a UI flash every time I click on Pay funds:

https://github.com/user-attachments/assets/39b9c241-0f09-4914-83f9-4391b12fd439


----

Otherwise, the UI works well:

- Not Started stream:
image

- Active stream
image

- Ended stream:
image

src/components/v5/common/CompletedAction/partials/StreamingPayment/partials/StreamingPaymentWidget/StreamingPaymentWidget.tsx
joanna-pagepro: I don't think that it should be a hook. There's no functionality inside that requires this to be a hook. I need it to get the latest data every time `blockTime` is updated inside useEffect with counter to get the right `amountAvailableToClaim` and that's why it's in `useMemo`

jakubcolony: There are many other hooks that do not need to be hooks, yet we chose to make them such for consistency and readability reasons.

My concern is that we now have an identical fragment of logic in 2 different places and it's not clear which one to use going forward. If you intend to keep this as a function, please remove the hook

joanna-pagepro: Hook removed

src/components/v5/common/CompletedAction/partials/StreamingPayment/partials/StreamingPaymentWidget/StreamingPaymentWidget.tsx
```diff
@@ -0,0 +1,304 @@
...
+              ]),
+          {
+            key: '1',
+            content: (
+              <>
+                <div className="flex items-center justify-between gap-4">
+                  <h4 className="text-1">{formatText(MSG.streamingStatus)}</h4>
+                  <LoadingSkeleton
+                    isLoading={isLoading}
+                    className="h-5 w-14 rounded"
```
iamsamgibbs, line 222: This should match the height of the pill to prevent this jumping:

https://github.com/user-attachments/assets/6b6d8f0f-f539-4d64-8c34-23e66d50aee8

src/components/v5/common/CompletedAction/partials/StreamingPayment/consts.ts
```diff
@@ -0,0 +1,2 @@
+export const ONE_HOUR_IN_SECONDS = 3600;
+export const ONE_DAY_IN_SECONDS = ONE_HOUR_IN_SECONDS * 24;
```
iamsamgibbs, line 2: These exist in 2 places now, maybe we can just put these in `src/constants` as I imagine they'll be useful elsewhere too.

Screenshot 2024-09-05 at 11 27 08
Screenshot 2024-09-05 at 11 27 59

iamsamgibbs (changes requested):
Nice work on this so far, it's looking great!

Screenshot 2024-09-05 at 10 58 48

I noticed when creating payments via the temporary UI, the start / end would be correct, but not when creating via the action sidebar:

Screenshot 2024-09-05 at 11 06 22

Here is another example using "start immediately" rather than a fixed date - I'm guessing this is using the block time as it is putting the start date back in April.
 
Screenshot 2024-09-05 at 11 11 07

Screenshot 2024-09-05 at 11 11 23

Seems like an issue with the action sidebar form rather than the changes in this PR though as the ones created via the temporary UI are fine.

One issue I came across - this payment never updated to "Ended", it just stayed as "Active" once the limit was reached:

Screenshot 2024-09-05 at 11 16 44

I think there is a rounding error when calculating the limit:

Screenshot 2024-09-05 at 11 14 47

Screenshot 2024-09-05 at 11 17 55

I also noticed this issue with the UserPopover - not sure if it is related to the changes in this PR, but I can see there are some changes to those files:

https://github.com/user-attachments/assets/98ffc632-1d91-486b-83d6-b60f620e64d5

Also this issue with the activity list which I'm assuming is out of scope for this issue:

 
Screenshot 2024-09-05 at 11 21 20

joanna-pagepro:
@iamsamgibbs I'm just displaying the limit value, just like the amount, so I guess that something is wrong somewhere else.
I'm still working on the actions list, but other issues should be resolved.

joanna-pagepro:
@iamsamgibbs action description is now updated as well.

@AdrianDudko @arrenv I guess that we will need a separate task for filtering streaming payments by status on activity page.

iamsamgibbs (changes requested):
Thanks for the changes on this @joanna-pagepro ! I'm still experiencing a few issues though.

The custom title isn't getting picked up on the completed action (it was working in the screenshots from my previous review). The amount interval doesn't seem to be getting set correctly either. This should be 1 CRED per hour, but at the rate it is going up it looks more like 1 CRED per second.

https://github.com/user-attachments/assets/975b9757-0487-4729-8e61-8ecd6a1d7d71

The custom titles are also missing from the activity feed:

Screenshot 2024-09-09 at 09 09 06

This is looking better, but there is still something here making the elements jump around when they finish loading:

https://github.com/user-attachments/assets/f7be8f2e-cb78-450a-8ac2-1842d6b0a178

iamsamgibbs:
Just a note on the title. Seems like it is not getting created in the database:

Screenshot 2024-09-12 at 10 44 05

Think we need one of these in the saga:

```
    if (customActionTitle) {
      yield createActionMetadataInDB(txHash, customActionTitle);
    }
```

I'm guessing it got accidentally removed in a rebase as it was previously working.

iamsamgibbs (changes requested):
I'm still having a few issues with this unfortunately @joanna-pagepro 

Every payment seems to be starting in 1970, regardless of the selected option:

Screenshot 2024-09-12 at 10 41 34

The claim amounts are calculating correctly, so I'm guessing this is either a UI issue, or something isn't saving to the database quite right, but the value getting submitted in the form in the transaction seems to be correct.

The limit reached pill is correctly updating now 🎉

https://github.com/user-attachments/assets/84bd9148-f617-4da0-86c9-e809ca47e8f6

However, when loading a payment, the pills initially show active before switching to limit reached. They should show a loading spinner or no pill at all until the correct state is calculated / fetched.

https://github.com/user-attachments/assets/7d449845-6b40-4e8f-b028-df9ac7f6a2b3

This is still jumping a little bit, but it's such a minor difference that I think it is okay at this point.

https://github.com/user-attachments/assets/13b52289-f110-4afa-98ef-c5671e9878a4

This user popover bug is still here:

https://github.com/user-attachments/assets/13831017-897a-4228-9d28-4aaa7f9d44e7

I don't know if this is too much of an edge case, but payments are not automatically updating from "not started" to "active". The payment in this recording should have started at 11:00 - but it requires me to close and re-open the action for the pills to update and the claim value to start changing.

https://github.com/user-attachments/assets/fb7517f6-700d-4c3e-aecf-e1f2f9611b54

joanna-pagepro:
@iamsamgibbs and now? :)
Everything updated (but title)

jakubcolony:
@joanna-pagepro please rebase, the changes to set action title when creating a streaming payment should be in.

iamsamgibbs (changes requested):
Great work, thank you for fixing so many of these issues! The only remaining issue (other than the title, which will hopefully be fixed by the rebase!) is just showing the incorrect status when loading the action. "Active" is briefly shown before "Limit reached".

https://github.com/user-attachments/assets/5e8bf7f6-0537-4188-9f96-7ba4d7178674

Otherwise, the issues with start time are fixed:

Screenshot 2024-09-17 at 13 07 10

Loading jumping issue is resolved:

https://github.com/user-attachments/assets/51621f0c-3a3a-484d-b9e7-528f05c613eb

User popover issue is resolved:

Screenshot 2024-09-17 at 13 13 41

And status correctly changes to "Active" when a payment starts:

https://github.com/user-attachments/assets/394be4fe-ca13-4838-b515-d18493dba91f

jakubcolony (changes requested):
Left a few comments but otherwise this looks good, nice work:

image

image

image


I also managed to track down the bug with `ethersProvider.getBlock`, for which there's issue #3113

src/components/v5/common/ActionSidebar/hooks/useGetStreamingPaymentData.ts
```diff
@@ -0,0 +1,119 @@
...
+
+      setPaymentStatus(status);
+      setAmounts({ amountAvailableToClaim, amountClaimedToDate });
+    },
+    [amounts.amountAvailableToClaim, streamingPayment],
+  );
+
+  useEffect(() => {
+    if (streamingPayment && !loading) {
+      const currentTimestamp = Math.floor(blockTime ?? Date.now() / 1000);
```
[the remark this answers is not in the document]
joanna-pagepro: I've added another useEffect and called `fetchCurrentBlockTime` inside

joanna-pagepro:
@jakubcolony @iamsamgibbs can you review again?

jakubcolony (approved):
LGTM

joanna-pagepro:
@iamsamgibbs :)

iamsamgibbs (approved):
Nice job with this @joanna-pagepro Thank you for all the fixes!

https://github.com/user-attachments/assets/b9bcf883-9d81-43d6-8640-809a432474cd

Thank you for adding the "Limit reached" pill to the activity feed too, just noting that the filters don't work for the status or the activity type, but that is an issue for another time!

Screenshot 2024-09-19 at 08 56 09

arrenv (commented):
Late to the party, but all looks good, thank you @joanna-pagepro

PR #2940, opened 2024-08-13
Fix: Return the correct list of actions based on the `createdAt` filter

### LINKED ISSUE #2792: Filtering by date in activity feed should be OR
## Steps to reproduce

1. Ensure you have actions in a colony that cover multiple time periods as of the filter. i.e. Past hour, Past 24 hours.
2. Check 

## Expected behaviour

* Selecting multiple date filters should be an OR filter so the action could have happened in the "Past hour" OR "Past 24 hours".

## Actual behaviour

* Selecting multiple date filters is an AND filter which is impossible as an action can't match multiple dates.

filtering-date

rumzledz:
## Description

We have the following predefined date filters:
- past hour
- past 24 hours
- past week
- past month
- past year

The filter wasn't working properly because our logic was basically:

```js
case dateFilter.pastHour: return data;
case dateFilter.pastDay: return data;
case dateFilter.pastWeek return data;
case dateFilter.pastMonth: return data;
case dateFilter.pastYear: return data; 
```

So if you applied the past hour filter + the past year filter, it will first enter the `case` block for `pastHour`. If there are actions in the past hour, it will return those actions without ever having the chance to check the `case` block for `pastYear`.

But luckily, we can easily re-order the checks and start with the most all-encompassing range:

```js
case dateFilter.pastYear: return data;
case dateFilter.pastMonth: return data;
case dateFilter.pastWeek return data;
case dateFilter.pastDay return data;
case dateFilter.pastHour return data;
```

Technically speaking, checking for `pastHour` data when you've already selected `pastYear` data is irrelevant because `pastYear` already covers `pastHour`. 

## Actionable Item

Create a separate issue to use a radio button when setting a date filter as per Arren's suggestion, because as I have mentioned in the Description section and how @Nortsova has observed, it's quite redundant to have multiple selected dates set. As a business rule, the app is only meant to use a custom date range if it's applied even though you have applied other predefined dates. And setting `pastHour` and `pastYear` at the same time will only use `pastYear` anyway.

| Before (date: Aug 13 2024) | After (date: Aug 13 2024) |
| ---- | ---- |
| Screenshot 2024-08-13 at 15 29 04 | Screenshot 2024-08-13 at 15 28 37 |

## Testing

1. Create a Simple Payment
2. Go to the Activity page http://localhost:9091/planex/activity
3. Filter the date by past hour
4. Verify that you can see the Simple Payment action you just made
5. Keep track its date field, this represents the block time
6. Click this Action
7. When the Completed Action component comes up, copy the transaction hash from the URL bar
Screenshot 2024-08-13 at 15 39 43
8. Open the [GraphQL Playground](http://localhost:20002/)
9. Run the following query to update its `createdAt` date to some time within the past 24 hours

> [!WARNING]
> You might not experience it but sometimes, I find that it takes a couple of minutes for the update to be picked up in the app so if your transaction isn't showing up immediately after you've mutated the `createdAt` date, give it like 3-5mins.

> [!INFO]
> On my machine, the current block time is April 20 2023 13:15. Adjust your `createdAt` value accordingly

```gql
mutation UpdateCreatedAt {
  updateColonyAction(
    input: {
      // replace this with the hash you just copied
      id: "0xc611ea533df3da3be76fab8338158f9eb51744e9cfdb80be02d3c68b6ad09743",
      createdAt: "2023-04-20T12:50:47.000Z"
    }
  ) {
    createdAt
  }
}
```
10. Verify that you can see the Transaction if you turned on the `Past 24 hours` filter:
11. This is the part where I need to apologise because you'll have to do the same steps for the following filters ❤️ :
- Past week
- Past month
- Past year

Resolves #2792

### DIFF
--- a/docker/colony-cdapp-dev-env-auth-proxy
+++ b/docker/colony-cdapp-dev-env-auth-proxy
@@ -1,6 +1,6 @@
 FROM colony-cdapp-dev-env/base:latest
 
-ENV AUTH_PROXY_HASH=646d498db30e0d3cbd1a62a4d522260624e6340e
+ENV AUTH_PROXY_HASH=f1d5e35271358946ec65e9a0f7d27fa66c2b878a
 
 # Add dependencies from the host
 # Note: these are listed individually so that if they change, they won't affect
--- a/src/components/common/ColonyActionsTable/FiltersContext/FiltersContextProvider.tsx
+++ b/src/components/common/ColonyActionsTable/FiltersContext/FiltersContextProvider.tsx
@@ -12,6 +12,7 @@ import {
   type ActivityFeedFilters,
   type ActivityDecisionMethod,
 } from '~hooks/useActivityFeed/types.ts';
+import useCurrentBlockTime from '~hooks/useCurrentBlockTime.ts';
 import { type AnyActionType } from '~types/actions.ts';
 import { type MotionState } from '~utils/colonyMotions.ts';
 
@@ -36,6 +37,7 @@ const FiltersContextProvider: FC<PropsWithChildren> = ({ children }) => {
     pastYear: false,
     custom: undefined,
   });
+  const { dateFromCurrentBlockTime } = useCurrentBlockTime();
 
   const handleActionTypesFilterChange = useCallback(
     (event: React.ChangeEvent<HTMLInputElement>) => {
@@ -115,7 +117,9 @@ const FiltersContextProvider: FC<PropsWithChildren> = ({ children }) => {
   );
 
   const activeFilters: ActivityFeedFilters = useMemo(() => {
-    const date = getDateFilter(dateFilters);
+    const date = dateFromCurrentBlockTime
+      ? getDateFilter(dateFilters, dateFromCurrentBlockTime)
+      : null;
     const actionTypes = actionTypesFilters.reduce<AnyActionType[]>(
       (result, actionType) => {
         const apiActionTypes = ACTION_TYPE_TO_API_ACTION_TYPES_MAP[actionType];
@@ -142,6 +146,7 @@ const FiltersContextProvider: FC<PropsWithChildren> = ({ children }) => {
     };
   }, [
     actionTypesFilters,
+    dateFromCurrentBlockTime,
     dateFilters,
     decisionMethod,
     motionStates,
--- a/src/components/common/ColonyActionsTable/utils.ts
+++ b/src/components/common/ColonyActionsTable/utils.ts
@@ -46,47 +46,17 @@ export const makeLoadingRows = (pageSize: number): ActivityFeedColonyAction[] =>
 
 export const getDateFilter = (
   dateFilter: DateOptions | undefined,
+  dateFromCurrentBlockTime: Date,
 ): Pick<ActivityFeedFilters, 'dateFrom' | 'dateTo'> | undefined => {
   if (!dateFilter) {
     return undefined;
   }
 
-  const now = new Date();
   const baseFilter = {
-    dateTo: now,
+    dateTo: dateFromCurrentBlockTime,
   };
 
   switch (true) {
-    case dateFilter.pastHour: {
-      return {
-        dateFrom: sub(now, { hours: 1 }),
-        ...baseFilter,
-      };
-    }
-    case dateFilter.pastDay: {
-      return {
-        dateFrom: sub(now, { days: 1 }),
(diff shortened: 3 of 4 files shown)


src/components/common/ColonyActionsTable/partials/ActionsTableFilters/partials/filters/DateFilters/DateFilters.tsx
```diff
@@ -52,6 +53,10 @@ const DateFilters: FC = () => {
             '!fixed !inset-auto !left-1/2 !top-1/2 w-full !-translate-x-1/2 !-translate-y-1/2':
               isMobile,
           })}
+          // I can't just use Object.values().some(Boolean) since I really need to specifically check if it's true, and not if it's truthy
+          disabled={Object.values(dateFilters).find(
+            (dateFilter) => dateFilter === true,
+          )}
```
Nortsova, line 59: please review **disabled** flow again (I left a detailed comment above). thanks!

Nortsova (changes requested):
Hey @rumzledz ! Thanks for that brilliant fix! ✨

I can confirm that filtering works great:

My time is: 15 August, 16:00

Pay leela 49.5 CREDS by leela: Today at 2:50 PM

TEST ACTIVITY ACTION (I will update that): Today at 3:50 PM

Past hour:
image

Past 24 hours:
image

**Updated**: TEST ACTIVITY ACTION: 13 Aug 2024 at 3:50 PM
image

Past week:
image

**Updated**: TEST ACTIVITY ACTION: 03 Aug 2024 at 3:50 PM
image

Past month:
image

**Updated**: TEST ACTIVITY ACTION: 03 Jun 2024 at 3:50 PM

image

Custom: 01 Jun 2024 - 08 Jun 2024

image


Custom: 01 Jun 2024 - 15 Aug 2024
image

Custom: 10 Jul 2024 - 15 Aug 2024

image


However, there are a couple of bugs were found:
It is not obvious why I can't click a custom range because it looks the same disabled and not disabled:
Custom range disabled:
image
Custom range not disabled:
image

There is no way to choose a predefined range if I already choose a custom one (I can't clear date-picker):
image

My proposal here is to enable everything and clean state:
Click on predefined - custom range cleaned up
Click on the custom range - predefined checkboxes cleaned up

There is also no reason to give multiple choices to the user since only the highest selected option will apply:
image

Totally unrelated to this PR, but an interesting one:
date-picker-bug

I know that some of these bugs were already present, so feel free to pick whichever ones you consider important 😃

src/components/common/ColonyActionsTable/partials/ActionsTableFilters/partials/filters/DateFilters/DateFilters.tsx
rumzledz: Hey @Nortsova all your points are super valid! 🙌  I spoke to @arrenv and basically we'll have to create a separate issue to address all those things you've mentioned as they are currently on production as well

As a follow-up, I've also decided to remove the disabling of the fields because we might end up using radio buttons for the date filters, as per Arren's suggestions.

Thank you for reviewing this!

I think things to note now that the disable logics are removed: When a custom date range is entered, it will be prioritised. This is production behaviour.

I'll be adding this in the description as well 😄

Nortsova (approved):
Thanks @rumzledz for considering my feedback! Looks great ✨

rdig (dismissed):
I like the approach for this fix, and I'm actually relieved this didn't devolve into us having to refactor the component.

Nicely done!

Screenshot from 2024-08-15 21-26-22
Screenshot from 2024-08-15 21-26-27

bassgeta (dismissed):
Love this, so simple and it just works, nice job turning it on its head, it makes much more sense for filtering to be done this way :ok_hand: 
image
FIltering works for all dates like they should but there's something I've noticed which we could rework (this is most likely just an improvement for localhost as in prod the dates won't be far apart).

Instead of doing 
`const now = new Date()`

we could be using the current block's timestamp
I just want to cc @jakubcolony if this is a correct assumption from my end :D

bassgeta (approved):
Nicely done, thanks for aligning it with the current block time!
image

All the others still work as intended, so let's merge this fix!

rdig (approved):
Re-approved since it got dismissed. Still good to go!

Nortsova (approved):
Thanks for updating time to block time! Works perfectly! 👏

PR #2945, opened 2024-08-13
Fix: Disable click for pending transactions in Userhub

### LINKED ISSUE #2738: Clickable UserHub transactions before completed transaction
## Steps to reproduce

1. You will need to test using Metamask in local dev (let me know if you are unsure how to do this).
2. Create a new action but don't confirm the transaction.
3. Click on the "Pending" button in the header.
4. The transaction tab will open.
5. Click on the item for the current transaction.
6. You will load up a non existent transaction.

## Expected behaviour

* The item in the transaction tab for incomplete transactions with no transaction hash should not be clickable.

## Actual behaviour

* Incomplete transaction items in the transaction tab of the Userhub are clickable but take you to a 404.

image

rumzledz:
## Description

This just simply disables click for pending transactions in the UserHub

disable pending click

## Testing

1. Make a Simple Payment
2. As soon as the Pending button comes up, click it
3. The UserHub should come up and and you should automatically land on the Transactions tab
4. Click your Transaction while it's in a Pending state
5. Verify that you are not able to
6. Verify that the click is successful only when the Transaction exits the Pending state

Resolves #2738


rumzledz:
> However I did notice the not found action is shown for a split second after clicking the succeeded transaction. Is there anything we can do about it?

Aw man 😕 All right I'll see what I can do thanks for spotting this! 👌

rumzledz:
@mmioana can you please try this again when you get the chance? Thank you 😄

src/components/common/Extensions/UserHub/partials/TransactionsTab/partials/GroupedTransaction.tsx
```diff
@@ -93,7 +94,8 @@ const GroupedTransaction: FC<GroupedTransactionProps> = ({
     values.group?.key &&
     !GROUP_KEYS_WHICH_CANNOT_LINK.includes(
       values.group.key as TRANSACTION_METHODS,
-    );
+    ) &&
+    status === TransactionStatus.Succeeded;
```
Nortsova, line 99: Nice fix, but in terms of the single responsibility principle, wouldn’t it be better to move this check to where it’s used and just pass the prop as `isClickable={status === TransactionStatus.Succeeded}`?

Nortsova (approved):
Hey @rumzledz , it works great! Thanks for this fix and for the 'not found' fix as well!

pending-not-clickable

src/components/common/Extensions/UserHub/partials/TransactionsTab/partials/GroupedTransaction.tsx
rumzledz: Hey @Nortsova thanks for reviewing! I actually have multiple versions of this. I had a version where I wanted to move it outside but it doesn't make sense to just move this single boolean outside, you'd have to deal with the entire lot because why stop at `status === TransactionStatus.Succeeded `? 😂
```js
const canLinkToAction =
  isClickable &&
  values.group?.key &&
  !GROUP_KEYS_WHICH_CANNOT_LINK.includes(
    values.group.key as TRANSACTION_METHODS,
  )
```
I also have another version where this component is stripped down to the point that it only has a few hooks. Would have to move around stuff outside the component etc. etc. If we really wanted to do proper SRP, you'd also have to group up every logical piece of responsibility in the component like the Header, the bit that handles the accordion-like behaviour etc.  because simply moving the boolean isn't gonna cut it. It could easily turn into a rabbithole 🐰 and I've been trying my best not to overdo it in my PRs lately 😂  Maybe someday we are free enough to audit our components and see which ones to refactor 👌

rdig (approved):
All good. I can confirm I can click a successful action, yet cannot cannot click a link to an action if the transaction is in a pending or failed state

Screenshot from 2024-08-21 21-07-04
Screenshot from 2024-08-21 21-16-57
Screenshot from 2024-08-21 21-17-11

Nicely done!

bassgeta (approved):
Nice work :ok_hand: Connected with metamask and tried to click it, nothing happens, as intended :sunglasses: 
image

PR #2966, opened 2024-08-16
UI: Update the Colony Switcher UI

### LINKED ISSUE #2885: [New navigation] Update the Colony Switcher UI
## Description

We have a very basic version of the Colony Switcher at the moment. This issue aims to make it follow the Figma designs.

## Acceptance Criteria

* [ ] The current hover state for Colony items is wrong so it should be fixed: [Figma designs](https://www.figma.com/design/uKxVl441Wnn02SNiOA24De/Navigation?node-id=4537-9201&t=Qz678dDXjP3Dwgfy-4)

Image

* [ ] Radio buttons to indicate what Colony has been selected is implemented: [Figma designs](https://www.figma.com/design/uKxVl441Wnn02SNiOA24De/Navigation?node-id=3930-18031&t=Qz678dDXjP3Dwgfy-4)

Image

* [ ] Mobile View is implemented: [Figma designs](https://www.figma.com/design/uKxVl441Wnn02SNiOA24De/Navigation?node-id=3931-20270&t=Qz678dDXjP3Dwgfy-4)

Image

rumzledz:
## Description

This PR updates the UI of the Colony Switcher.

Screenshot 2024-08-16 at 18 19 12

Screenshot 2024-08-16 at 18 19 27

## Testing

> [!NOTE]
> The Create new colony button won't work. It will be addressed in a different PR.

1. Connect Dev Wallet 1 so you can view Colonies on the Colony Switcher
2. Verify the UI matches the [Figma designs](https://www.figma.com/design/uKxVl441Wnn02SNiOA24De/Navigation?node-id=4537-9201&t=Qz678dDXjP3Dwgfy-4)
3. Set the browser width to 1024px
4. Verify the UI matches the [Figma designs](https://www.figma.com/design/uKxVl441Wnn02SNiOA24De/Navigation?node-id=3931-20270&t=Qz678dDXjP3Dwgfy-4)

Resolves #2885

### DIFF
--- a/src/components/v5/shared/Navigation/ColonySwitcher/partials/ColonySwitcherAvatar.tsx
+++ b/src/components/v5/shared/Navigation/ColonySwitcher/partials/ColonySwitcherAvatar.tsx
@@ -3,7 +3,6 @@ import React from 'react';
 
 import { useColonyContext } from '~context/ColonyContext/ColonyContext.ts';
 import { usePageLayoutContext } from '~context/PageLayoutContext/PageLayoutContext.ts';
-import { useTablet } from '~hooks';
 import ColonyIcon from '~icons/ColonyIcon.tsx';
 import Avatar from '~v5/shared/Avatar/index.ts';
 
@@ -12,8 +11,6 @@ export const ColonySwitcherAvatar = () => {
 
   const { showTabletColonyPicker } = usePageLayoutContext();
 
-  const isTablet = useTablet();
-
   const colonyAvatarSrc = colonyContext?.colony?.metadata?.avatar
     ? colonyContext.colony.metadata?.thumbnail ??
       colonyContext.colony.metadata?.avatar
@@ -21,10 +18,12 @@ export const ColonySwitcherAvatar = () => {
 
   return (
     <div
-      className={clsx('rounded-full border-2 border-base-white', {
-        'border-blue-400': showTabletColonyPicker && isTablet,
-        'border-gray-900': !colonyContext && !isTablet,
-      })}
+      className={clsx(
+        'flex-shrink-0 rounded-full shadow-[0_0_0_4px] shadow-transparent transition-shadow duration-[250] md:shadow-none',
+        {
+          '!shadow-blue-400': showTabletColonyPicker,
+        },
+      )}
     >
       {colonyContext ? (
         <Avatar
--- a/src/components/v5/shared/Navigation/ColonySwitcher/partials/JoinedColonyItem/JoinedColonyItem.tsx
+++ b/src/components/v5/shared/Navigation/ColonySwitcher/partials/JoinedColonyItem/JoinedColonyItem.tsx
@@ -3,6 +3,7 @@ import React from 'react';
 import { capitalizeFirstLetter } from '~utils/strings.ts';
 import ColonyAvatar from '~v5/shared/ColonyAvatar/index.ts';
 
+import { ActiveColonyIndicator } from './partials/ActiveColonyIndicator.tsx';
 import { type JoinedColonyItemProps } from './types.ts';
 
 const displayName =
@@ -14,14 +15,15 @@ const JoinedColonyItem = ({
   name,
   colonyAddress,
   tokenSymbol,
+  isActiveColony,
 }: JoinedColonyItemProps) => {
   return (
     <button
       type="button"
-      className="group flex h-[37px] flex-row items-center justify-start gap-3 rounded-lg p-2 hover:bg-gray-50"
-      onClick={() => onClick(name)}
+      className="group flex h-[50px] w-full flex-row items-center justify-start gap-3 rounded px-6 py-8 hover:bg-gray-50 md:px-2 md:py-0"
+      onClick={() => !isActiveColony && onClick(name)}
     >
-      <section className="w-8">
+      <div className="w-8">
         <ColonyAvatar
           colonyAddress={colonyAddress}
           className="w-8 flex-shrink-0"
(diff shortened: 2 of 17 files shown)


src/components/v5/shared/Navigation/ColonySwitcher/partials/ColonySwitcherAvatar.tsx
```diff
@@ -2,30 +2,19 @@ import clsx from 'clsx';
 import React from 'react';
 
 import { useColonyContext } from '~context/ColonyContext/ColonyContext.ts';
-import { usePageLayoutContext } from '~context/PageLayoutContext/PageLayoutContext.ts';
-import { useTablet } from '~hooks';
 import ColonyIcon from '~icons/ColonyIcon.tsx';
 import Avatar from '~v5/shared/Avatar/index.ts';
 
-export const ColonySwitcherAvatar = () => {
+export const ColonySwitcherAvatar = ({ className }: { className?: string }) => {
```
davecreaser, line 8: Just a personal preference thing here, but if this className is being passed just to differentiate between mobile and desktop, could we not just handle these class names inside the ColonySwitcherAvatar component itself? Otherwise it looks like I can style the ColonySwticherAvatar however I want, even though it should be quite specific and self contained.

src/components/v5/shared/Navigation/ColonySwitcher/partials/JoinedColonyItem/JoinedColonyItem.tsx
```diff
@@ -14,11 +15,12 @@ const JoinedColonyItem = ({
   name,
   colonyAddress,
   tokenSymbol,
+  isActiveColony,
 }: JoinedColonyItemProps) => {
   return (
     <button
       type="button"
-      className="group flex h-[37px] flex-row items-center justify-start gap-3 rounded-lg p-2 hover:bg-gray-50"
+      className="group flex h-[50px] w-full flex-row items-center justify-start gap-3 rounded-lg px-4 py-8 hover:bg-gray-50 md:px-2 md:py-0"
```
davecreaser, line 23: Pretty minor, but since testing this PR is all about eagle design eyes, this className should have `rounded` instead of `rounded-lg` I believe!

davecreaser (changes requested):
Looking good @rumzledz!

Just a couple of minor things in the code. One of them is definitely just a personal opinion so feel free to ignore it or tell me to get lost. The other is a minor difference from Figma.

Also, I spotted that you can still select a colony which you are currently in, leading to some loading state flickering. It would be better if the currently active colony was disabled or did not do anything on click to avoid this. See the video below:

https://github.com/user-attachments/assets/b34655a4-ec3c-44a2-902a-257237c069e2

src/components/v5/shared/Navigation/ColonySwitcher/partials/JoinedColonyItem/JoinedColonyItem.tsx
rumzledz: You're totally right! I shall update 👌

rumzledz:
Thanks for reviewing @davecreaser ! Can you please have a look again when you're free? Thank you! 😄

davecreaser (approved):
Thanks for the changes! Looking good 🎉 

Screenshot 2024-08-19 at 17 25 01

bassgeta (approved):
Nice work on this, stuff is starting to look real pretty :tada: On desktop, everything looks as it should :heavy_check_mark: 
image
Mobile is also legit:
image
Also checked Firefox and works like a charm there too.
:ship:

src/components/v5/shared/Navigation/ColonySwitcher/partials/JoinedColonyItem/JoinedColonyItem.tsx
```diff
@@ -41,6 +43,9 @@ const JoinedColonyItem = ({
           {tokenSymbol}
         </p>
       </section>
+      <section className="flex flex-1 flex-row items-center justify-end">
+        <ActiveColonyIndicator isActive={isActiveColony} />
+      </section>
```
iamsamgibbs, line 48: Seems weird that we are using sections inside of a button, why not just divs?

iamsamgibbs (changes requested):
Nice job Romeo! This is looking great! It'll be great to see this when it has been rebased to include the "no connected wallet" and "no joined colonies" states.

Screenshot 2024-08-22 at 08 21 26

Only one minor issue, I think it would be nicer if the switcher had a max-height to restrict it to the height of the page and scrolled when there are too many colonies to fit on the page. Currently it just extends off the bottom of the page, which on it's own isn't great, but it is also showing cut off content on the main page (notice the activity feed gets cut off)

https://github.com/user-attachments/assets/15db263a-d8dd-40f0-aae5-5bb111b0105e

Mobile works fine with many colonies though:
 
Screenshot 2024-08-22 at 08 35 25

I noticed this one other bug which is tangentially related - the chevrons get pushed out of place on a long colony name:

Screenshot 2024-08-22 at 08 25 01

This is otherwise looking great though!

iamsamgibbs (approved):
Thanks for fixing the issue with the long colony name, and making the list scrollable:

Screenshot 2024-08-22 at 14 59 50

I'm not sure this behaviour is ideal when scrolling the navigation menu, but it is a definite edge case so happy to approve it as is unless you'd rather fix it here and I'll give it another review!

https://github.com/user-attachments/assets/2edf4725-c35d-48ba-acd4-371ae894df2b

rumzledz:
Hey @iamsamgibbs , thanks for your continued efforts in finding all the nuances of this component. Regarding:

> I'm not sure this behaviour is ideal when scrolling the navigation menu, but it is a definite edge case so happy to approve it as is unless you'd rather fix it here and I'll give it another review

This is unfortunately inherent to the popper hook. You can also see this behaviour with the other components in the page header which uses this hook:

popper

The popover just inherently follows its reference element and when the reference element goes outside the view, it follows it. It's visually uglier with the Colony Switcher component because it's seemingly more "unexpected" since the rest of the page remains as-is but because the Colony Switcher is the one being scrolled outside, the popover follows it.

The most straightforward "fix" I can think of is to keep the ColonySwitcher component in its place and only have the Action/Route buttons scrollable. That way, the popover also remains in place. I'll try to negotiate this with product. Currently on production or master, this Sidebar is broken anyway on a very short height viewport, like the sidenav buttons overflow 😂

PR #2970, opened 2024-08-19
Disable multi sig extension install when colony version is too low

### LINKED ISSUE #2918: [Multi-sig] Add notices for incorrect colony version
## Steps to reproduce

1. Ensure you have a colony on version 15 or lower.
2. Install Multi-sig extension.
3. Create a new Manage permission action to award Multi-sig permission.
4. The action will fail for no discernable reason.

## Expected behaviour

* The user should be made aware that they cannot use multi-sig until they upgrade their colony.
* We should show a validation message when selecting the Authority type when assigning permissions.
* We should show an error banner on the multi-sig extensions page.

image

[Figma link for extension banner](https://www.figma.com/design/6qONTE1uD9K4yoIbfFzidg/Extensions?node-id=2187-15952&t=vw4fNCGsF36Bp7Ow-4)

image

## Actual behaviour

* Trying to assign multi-sig permissions to users on colony network version 15 or lower fails with no UI errors.

image

davecreaser:
## Description

- Disable the install button of the multisig extension when the colony version is lower than the minimum version (16).
- Set this functionality up for use in other extensions in the future.

## Testing

Before you can test this properly, you need to deploy colony network version 15 in hardhat, and also run the create data script with the previous version. To do this, stop your environment completely, and follow the below steps:

1. Run `npm run dev` as usual.
2. Once this is up and running, start a hardhat console by running `npm run hardhat`
3. Run the following commands in the hardhat console:

```
deployIMWSS = require('./scripts/deployOldUpgradeableVersion.js').deployColonyVersionIMWSS
cn = await artifacts.require("IColonyNetwork").at("0x777760996135F0791E2e1a74aFAa060711197777")
await deployIMWSS(cn)
```

4. Once this is finished (can take a while) run `await cn.getColonyVersionResolver(15)` to ensure version 15 is deployed.
5. Now, you'll create data, but you need to make sure you create data using the previous version. To do this, run `node scripts/create-data.js --usePreviousColonyVersion`
6. Run `npm run frontend` as usual.

Now, you're ready to test!

* Step 1. Load up the extensions page
Screenshot 2024-08-19 at 16 47 47

* Step 2. Go into the multi-sig extension details page. You should see the warning banner, and the install button should be disabled.
Screenshot 2024-08-19 at 16 48 07

* Step 3. Click on the Upgrade Colony link in the warning banner. The action sidebar should open with Upgrade Colony as the action type.
Screenshot 2024-08-19 at 16 48 42

## Diffs

**Changes** 🏗

* Disabled to extension install button if the colony version is lower than required for that extension

Resolves #2918

### DIFF
--- a/src/components/frame/Extensions/pages/ExtensionDetailsPage/ExtensionTopRow.tsx
+++ b/src/components/frame/Extensions/pages/ExtensionDetailsPage/ExtensionTopRow.tsx
@@ -1,4 +1,11 @@
-import { Extension, Id } from '@colony/colony-js';
+import {
+  type ColonyVersion,
+  Extension,
+  type ExtensionVersion,
+  Id,
+  isExtensionCompatible,
+  getExtensionLowestCompatibleColonyVersion,
+} from '@colony/colony-js';
 import { WarningCircle } from '@phosphor-icons/react';
 import React, { useCallback, type FC } from 'react';
 import { defineMessages } from 'react-intl';
@@ -82,6 +89,17 @@ const ExtensionsTopRow: FC<ExtensionsTopRowProps> = ({
       address: address || '',
     });
 
+  const extensionCompatible = isExtensionCompatible(
+    Extension.MultisigPermissions,
+    extensionData.availableVersion as ExtensionVersion,
+    colony.version as ColonyVersion,
+  );
+
+  const minimumColonyVersion = getExtensionLowestCompatibleColonyVersion(
+    Extension.MultisigPermissions,
+    extensionData.availableVersion as ExtensionVersion,
+  );
+
   const handleUpgradeColony = useCallback(
     () =>
       toggleActionSidebarOn({
@@ -92,7 +110,7 @@ const ExtensionsTopRow: FC<ExtensionsTopRowProps> = ({
 
   return (
     <>
-      {(extensionData.neededColonyVersion || 0) > colony.version && (
+      {!extensionCompatible && (
         <div className="pb-6">
           <NotificationBanner
             status="error"
@@ -104,7 +122,7 @@ const ExtensionsTopRow: FC<ExtensionsTopRowProps> = ({
             }
           >
             {formatText(MSG.colonyVersionTooLow, {
-              minimumColonyVersion: extensionData.neededColonyVersion,
+              minimumColonyVersion,
             })}
           </NotificationBanner>
         </div>
--- a/src/components/frame/Extensions/pages/ExtensionDetailsPage/ExtensionTopRow.tsx
+++ b/src/components/frame/Extensions/pages/ExtensionDetailsPage/ExtensionTopRow.tsx
@@ -1,11 +1,16 @@
 import { Extension, Id } from '@colony/colony-js';
-import React, { type FC } from 'react';
+import { WarningCircle } from '@phosphor-icons/react';
+import React, { useCallback, type FC } from 'react';
 import { defineMessages } from 'react-intl';
 
+import { Action } from '~constants/actions.ts';
+import { useActionSidebarContext } from '~context/ActionSidebarContext/ActionSidebarContext.ts';
 import { useColonyContext } from '~context/ColonyContext/ColonyContext.ts';
 import { type AnyExtensionData } from '~types/extensions.ts';
 import { addressHasRoles } from '~utils/checks/index.ts';
 import { formatText } from '~utils/intl.ts';
+import { ACTION_TYPE_FIELD_NAME } from '~v5/common/ActionSidebar/consts.ts';
+import NotificationBanner from '~v5/shared/NotificationBanner/index.ts';
 
 import ActionButtons from '../partials/ActionButtons.tsx';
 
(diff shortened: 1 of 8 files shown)


src/components/frame/Extensions/pages/ExtensionDetailsPage/ExtensionTopRow.tsx
```diff
@@ -65,8 +82,33 @@ const ExtensionsTopRow: FC<ExtensionsTopRowProps> = ({
...
     <>
+      {(extensionData.neededColonyVersion || 0) > colony.version && (
+        <div className="pb-6">
+          <NotificationBanner
+            status="error"
+            icon={WarningCircle}
+            callToAction={
+              <button type="button" onClick={handleUpgradeColony}>
+                {formatText(MSG.upgradeColony)}
+              </button>
```
rumzledz, line 103: At the moment, the underline is being overlapped by the bottom parts of the button copy:

| Current implementation | Figma design |
| ---- | ---- |
| Screenshot 2024-08-20 at 05 55 16 | Screenshot 2024-08-20 at 05 55 27 |

If you just add the following `callToActionClassName` value, it should be good to go!

```ts
<NotificationBanner
  status="error"
  icon={WarningCircle}
  callToAction={
    <button type="button" onClick={handleUpgradeColony}>
      {formatText(MSG.upgradeColony)}
    </button>
  }
  callToActionClassName="underline-offset-2" // <<-- Add me :)
>
  {formatText(MSG.colonyVersionTooLow, {
    minimumColonyVersion: extensionData.neededColonyVersion,
  })}
</NotificationBanner>
```

**Result**
image

rumzledz (changes requested):
Great great work on this one @davecreaser ! I deployed an older colony version using the the code you gave in the Test Description and I was able to bring up the Extensions banner and I can confirm that clicking the Upgrade Colony button brings up the correct Action Form 🔥 

upgrade to 16

I did leave a wee change request because I felt like it would be good to just deal with it here since it's a straightforward fix

src/components/frame/Extensions/pages/ExtensionDetailsPage/ExtensionTopRow.tsx
rumzledz: **Figma link:** https://www.figma.com/design/6qONTE1uD9K4yoIbfFzidg/Extensions?node-id=2187-15468&t=weswhiL8GdhTqFqW-4

bassgeta (approved):
Good stuff :sunglasses: Thanks for the instructions on how to get an old version deployed, it would be nice to write that down somewhere handy.
Also nice catch with disabling the click if the button is disabled, since technically one could tinker in the UI and enable the button.
image
:ship:

src/components/frame/Extensions/pages/ExtensionDetailsPage/ExtensionTopRow.tsx
davecreaser: Nice catch, thanks @rumzledz. I've added it globally to the notification banner call to action after checking the designs, seems like it should always have an offset of 2px.

src/components/v5/shared/NotificationBanner/NotificationBanner.tsx
```diff
@@ -54,7 +54,7 @@ const NotificationBanner: FC<NotificationBannerProps> = ({
         {callToAction ? (
           <div
             className={clsx(
-              'flex-shrink-0 text-xs font-medium underline md:hover:no-underline',
+              'flex-shrink-0 text-xs font-medium underline underline-offset-2 md:hover:no-underline',
```
rumzledz, line 57: Cleannn ✨

rumzledz (approved):
Just saw the push @davecreaser and can confirm it now matches the Figma design:
image
LGTM! ✅

Nortsova (approved):
Thanks for this fix @davecreaser 👏

I can confirm that I can't to install Multi-sig with old colony version:
image
I was able to install Multi-sig after upgrading colony version: 
image

Not related to your PR but still interesting UI bug: (to be honest I don't know how it should work in this case)
image

src/constants/extensions.ts
```diff
@@ -585,6 +585,7 @@ export const supportedExtensionsConfig: ExtensionConfig[] = [
       ColonyRole.Architecture,
       ColonyRole.Funding,
     ],
+    neededColonyVersion: 16,
```
chmanie, line 588: We are already maintaining this for all extensions and all colony versions: https://github.com/JoinColony/colonyJS/blob/main/packages/core/src/versions/helpers.ts#L47

I think we should have that in one single place as opposed to copied everywhere.
Let me know if I can help integrating this.

chmanie (changes requested):

src/constants/extensions.ts
davecreaser: Oh nice! Yeah that's much better. I see multisig isn't there though, what's the best way of going about this? If we added multisig there is it easy to get it into this PR, or is it worth creating a cleanup ticket to add that to colonyJS and update this part of the cdapp?

davecreaser: Finally got it to work :) Thanks for all your help with it today!

chmanie:
Great job adding and using the ColonyJS API 👏

PR #2972, opened 2024-08-20
[Multi-sig] feat: realign how permissions, created in and domain thresholds work

### LINKED ISSUE #2768: [Multi-sig] Manage Permissions can only be created by users with permissions in `root`
Previously, the Manage Permissions saga was built with the understanding that `root` permissions are required to assign permissions in the `root` domain, and `root` or `architecture` permissions are required to assign permissions in a sub-domain.

The missing piece from this is that `root` or `architecture` permissions are required in **the parent domain** of the sub-domain you are assigning permissions in. Currently the UI will let you attempt to create this action if you have permissions in the sub-domain you are assigning permissions in, but the contract call will fail.

As it currently stands we only have one parent domain, so this issue is to ensure the following is true about the Manage Permissions action:
- Assigning permissions in the `root` domain is only possible as a user with `root` permissions
- Assigning permissions in a sub-domain, is only possible as a user with either `root` or `architecture` permissions in the `root` domain.
- Assigning permissions in a sub-domain is not possible if you only have permissions in the sub-domain and don't have permissions in the `root` domain.

- [ ] The "missing permissions" error banner should show as expected.
- [ ] The "submit" button should be disabled as expected.
- [ ] The saga should be reworked as necessary.
- [ ] The voting reputation motion saga should be checked (Presumably the encoded action should be created in the root domain too - not sure if this is currently the case)
- [ ] The multi-sig saga should be checked. The permissions for this are slightly different. `root` is required when assigning roles in the `root` domain - the same as permissions. But only `architecture` can be used when assigning roles in a sub-domain, users with `root` permissions are unable to vote or create the action in a sub-domain. (Check what the contracts allow for when managing permissions in a sub-domain if the action is created in the sub-domain)

This should be done on the `feat/multi-sig` branch due to the permissions refactor on this branch.

Additionally, a possibly related issue: If you create a multi-sig motion for managing permissions for a user in a subdomain, but create the motion in `General`, you cannot finalize it.

Additional related fixes and reworks here are:
- Multi-sig motions do not have a concept of createdIn domain, all the users in the impacted domain can vote. That means that if you create a motion that moves funds from a domain, all the permission holders in that domain are eligible. With voting reputation you could create that in General, but the contracts don't enforce this, so we shouldn't either. We should remove created in for all multisig motions, they are created in the affected domain
- If a domain has majority threshold for the threshold type, we don't need to count all the permission holders from the parent domain into the threshold (they however are potential signees)

bassgeta:
## Description

This PR solves a lot of unclear stuff regarding multisig permissions, such as
- if there are no members in a domain with majority approval, the minimum threshold is 1
- the domain threshold for majority approval is calculated from the permission holders in that domain
- permission proofs for multi-sig are always checked for the domain the action is taking in
- no more created in for multi-sig
- permissions and multi-sig validation for assigning permissions to a user in a subdomain

Turns out, if you create multi-sig actions for managing permissions in a subdomain, the extension will automatically create the motion in the parent domain, so we didn't need any extra shennanigans except getting the proofs for a user for that domain (seems counter intuitive, I know).
There's just a small bug left on network, that the manage permissions motion in a subdomain gets created in that subdomain, so I've implemented a helper/bandage `getDomainIdForActionType` in order fetch eligible signees from the root domain for this motion type only.

Luckily it isn't massive, but there's a lot of changes which may bork everything :')

## Testing

Now there's a lot to test here, so let's just go step by step.
In all the cases you'll need at least 2 browsers, let's use `leela` and `amy`.
While doing all of this, keep note that the `Created in` row doesn't appear in any completed multisig actions.

### Architecture in a subdomain
1. As `leela` give `amy` `Architecture` permissions in `Andromeda`
image
2. As `amy` try to create a manage permissions action in a subdomain. The UI should be disabled
image

### Multi-sig shennanigans

Before we get started, let's do some basic config we will reuse for all test cases

1. Install the multi-sig extension
2. Set the global threshold to 2
image
3. Set the threshold in `Rocinante` to `Majority approval` and a fixed threshold of ` to `Ascendant`
image
4. Give `leela` `Owner` Multi-sig permissions in `General`
image

Now we're ready to test stuff!

#### Threshold changes
1. As `leela` try to create a Simple payment multi-sig motion as `leela` from `General`. You should see the "Not enough members" error as the global threshold is 2 and we do not have 2 people with `Payer` permissions there.
image
3. Try to create the same motion in `Ascendant`, you should be able to create and finalize it, because the domain has a fixed threshold of 1.
image
4. Set the global multisig threshold to `Majority approval`. A simple payment motion can now be created in `Rocinante`, because the majority is 1, and the global threshold doesn't override it
image
5. Set the global multisig threshold back to 2

#### Architecture permissions and manage permissions shennanigans
1. As `leela` give `fry` Architecture multisig permissions in `Andromeda`
2. As `leela` try to create a Manage permissions multisig motion to give a user any permissions in `Andromeda`. Since `fry` has permissions only in `Andromeda` and not in `General`, we can't create the action
image
3. As `leela` give `amy` `Architecture` in `General`
4. Try to create the same multisig motion, you should be able to create it now
image
5. Verify that  there are no voting buttons on the motion as `fry`
image
6. Reject the motion as `amy`, it should work seamlessly
image
7. Remove the vote
image
8. Vote for the motion and finalize it, the user should receive the roles
image

#### Manage reputation still working?
1. As `leela` give `amy` `Arbitration` multi-sig permissions in `Andromeda`
image
2. As `leela` try to create an award reputation multi-sig motion, it should be disabled since we just have 1 owner.
image
3a. TAs `leela` try to create a remove reputation motion in `Daedalus`, it shouldn't be able to get created since there aren't enough members
image
3b. As `leela` try to create a remove reputation multi-sig motion in `Andromeda`
image
4. Sign it and finalize it as `amy`, it should finalize successfully
image
5. As `amy` try to give `amy` `Owner` multi-sig permissions in `General`. It should error out nicely
image
6. As `leela` give `amy` `Root` multi-sig permissions in `General` (I've pasted a wrong screenshot saying owner here, use 
image
7. As `leela`  create an `Award reputation` multi-sig motion
image
8. As `amy` vote for it and finalize it, it should all go through smoothly
image

#### What about simple payments?
1. As `leela` try to create a simple payment multi-sig motion from `Daedalus`, since it has a threshold of 2, you shouldn't be able to create it
image
2. As `leela`, assign `amy` the `Payer` multi-sig permission bundle in `Daedalus`
image
3. As `leela` create the simple payment multi-sig motion from `Daedalus` again, it should work because we have `amy` with `Payer` in that domain + `leela` in root
image
4. As `amy` sign the motion and finalize it, it should work seamlessly
5. An additional check is to assign `fry` the `Funding` multi-sig role in `Daedalus`
image
6. As `leela` create the simple payment multi-sig motion from `Daedalus` again, `fry` should show up as a potential signee with 1 role
image
7. You should be able to sign it as `fry` ( a third browser appears!) too (no need to try out finalization)
image

## Diffs

**Changes** 🏗

* the majority domain threshold is a minimum of 1, calculated from users in the target domain
* fixed priority of domain thresholds to custom fixed domain threshold > global threshold > majority approval
* `useEligibleSignees` receives an array of domainIds, stringifies it and parses it
* we calculate the `requiredRoles` for signing just once and pass them down
* multisig motions no longer have the createdIn field (not even in completed action views)
* realigned manage permissions, move funds, payment and manage reputation permission proof checks for multi-sig

Resolves #2768


src/components/v5/common/ActionSidebar/partials/MultiSigSidebar/partials/MultiSigWidget/utils.ts
```diff
@@ -163,3 +163,18 @@ export const getIsMultiSigExecutable = (
...
+): number => {
+  // @NOTE this will find the parent domain, but as it stands we just return root
+  if (
+    actionType === ColonyActionType.SetUserRolesMultisig &&
+    Number(multiSigDomainId) !== Id.RootDomain
+  ) {
+    return Id.RootDomain;
+  }
+
+  return Number(multiSigDomainId);
```
mmioana, line 179: Could we extract `Number(multiSigDomainId)` into a constant?

src/hooks/multiSig/useEligibleSignees.ts
```diff
@@ -32,18 +32,21 @@ export const useEligibleSignees = ({
...
-  // Since it's an array fetched from a function, we "memo" this by stringify it to prevent it always triggering the useEffect
+  // Since they are both arrays fetched from a function, we "memo" this by stringifying them to prevent always triggering the useEffect
   const requiredRolesString = JSON.stringify(requiredRoles);
+  const domainIdsString = JSON.stringify(domainIds);
 
   useEffect(() => {
     async function fetchEligibleSignees() {
       const parsedRequiredRoles = JSON.parse(requiredRolesString);
+      const parsedDomainIds = JSON.parse(domainIdsString);
+
```
mmioana, line 43: Given this `useEffect` gets triggered each time `requiredRolesString` or `domainIdsString` changes, we should have access to the latest values from which we created these strings (`requiredRoles` and `domainIds`) so we could directly use those here, without the need to parse the stringified value

mmioana (approved):
Wow really cool work @bassgeta and a lot of testing steps 👏🥇 

However, given the priority is to firstly take into account if a domain has a specific threshold configured, I think this testing step should be updated from 
_As leela try to create a simple payment multi-sig motion in Rocinante. Since the majority there is 1, BUT the global threshold is 2, the motion can't be created_
to
_As leela try to create a simple payment multi-sig motion in Rocinante. Since the majority there is 1, the global threshold is not considered, thus the motion CAN be created_

Everything else looks good from my side 👍 

Screenshot 2024-08-26 at 13 14 41
Screenshot 2024-08-26 at 13 17 58
Screenshot 2024-08-26 at 13 18 30
Screenshot 2024-08-26 at 13 19 08
Screenshot 2024-08-26 at 13 19 52
Screenshot 2024-08-26 at 13 21 01
Screenshot 2024-08-26 at 13 21 12
Screenshot 2024-08-26 at 13 22 12
Screenshot 2024-08-26 at 13 23 02
Screenshot 2024-08-26 at 13 23 18
Screenshot 2024-08-26 at 13 23 49
Screenshot 2024-08-26 at 13 24 23
Screenshot 2024-08-26 at 13 24 30
Screenshot 2024-08-26 at 13 24 44
Screenshot 2024-08-26 at 13 26 03
Screenshot 2024-08-26 at 13 26 59
Screenshot 2024-08-26 at 13 28 35
Screenshot 2024-08-26 at 13 28 58
Screenshot 2024-08-26 at 13 30 58
Screenshot 2024-08-26 at 13 31 23
Screenshot 2024-08-26 at 13 31 36
Screenshot 2024-08-26 at 13 32 20
Screenshot 2024-08-26 at 13 32 29
Screenshot 2024-08-26 at 13 33 45
Screenshot 2024-08-26 at 13 34 15
Screenshot 2024-08-26 at 13 34 46
Screenshot 2024-08-26 at 13 34 58
Screenshot 2024-08-26 at 13 35 17
Screenshot 2024-08-26 at 13 35 38
Screenshot 2024-08-26 at 13 36 07
Screenshot 2024-08-26 at 13 36 47
Screenshot 2024-08-26 at 13 37 58
Screenshot 2024-08-26 at 13 38 36
Screenshot 2024-08-26 at 13 39 17
Screenshot 2024-08-26 at 13 39 51
Screenshot 2024-08-26 at 13 40 35
Screenshot 2024-08-26 at 13 41 29
Screenshot 2024-08-26 at 13 41 49
Screenshot 2024-08-26 at 13 42 06
Screenshot 2024-08-26 at 13 45 31
Screenshot 2024-08-26 at 13 46 13
Screenshot 2024-08-26 at 13 46 49

Also, left 2 small refactoring comments, please feel free to decide whether it's worth doing them 🙏

src/components/v5/common/ActionSidebar/partials/ActionSidebarContent/partials/MultiSigMembersError/MultiSigMembersError.tsx
```diff
@@ -27,13 +30,29 @@ export const MultiSigMembersError: FC<MultiSigMembersErrorProps> = ({
...
+  const domainId = useMemo(() => {
+    // for both of these actions you need roles in the parent domain
+    if (
+      selectedAction === Action.TransferFunds ||
+      selectedAction === Action.ManagePermissions
+    ) {
+      return Id.RootDomain;
+    }
+
+    return fromDomain ?? team ?? Id.RootDomain;
```
iamsamgibbs, line 48: I'm not sure if `Id.RootDomain` if no team is selected is ideal. This means if there are insufficient members in the root domain the "Not enough members" message will be shown before selecting a team, even if there are sub-domains where the threshold is met.

This should instead default to `undefined` if it doesn't break `useHasEnoughMembersWithPermissions`. Or prevent this message from showing until `fromDomain` has a value.

src/redux/sagas/actions/managePermissions.ts
```diff
@@ -147,7 +147,7 @@ function* managePermissionsAction({
     const requiredRoles =
       domainId === Id.RootDomain
         ? PERMISSIONS_NEEDED_FOR_ACTION.ManagePermissionsInRootDomain
-        : PERMISSIONS_NEEDED_FOR_ACTION.ManagePermissionsInSubDomain;
+        : PERMISSIONS_NEEDED_FOR_ACTION.ManagePermissionsInSubDomainViaMultiSig;
```
iamsamgibbs, line 150: This is the action saga, so `ManagePermissionsInSubDomain` should be correct.

iamsamgibbs (changes requested):
Nice job reworking all these changes - lots of good stuff in here!

Followed the testing steps without issue, I have left a couple of comments though.

I think this error message shows too soon:
 
Screenshot 2024-08-27 at 12 02 00

It would be great to either show this at this point if we can confirm none of the domains have sufficient members, otherwise it should wait until a team is selected before showing this message. Currently it just checks for sufficient members in root domain if no team is selected.

Didn't realise the global fixed threshold would still apply to sub-domains that have "Majority approval" set - I think that could probably be communicated better in the UI as it wouldn't have been clear to me without the explanation here.

src/components/v5/common/ActionSidebar/partials/ActionSidebarContent/partials/MultiSigMembersError/MultiSigMembersError.tsx
bassgeta: Isn't this how it functioned before? `createdIn` was always set to the root team by default, that's why this just worked. But I agree that we shouldn't show the error until we check the domain, I'll try and handle this here, thanks for the heads up!

bassgeta: As agreed on Discord, we're gonna be changing the error message to:
`There are not enough members in this or higher teams with the required permissions to approve this action. Select a different team, or assign more users with the required multi-sig permissions.` and implemented [here](https://github.com/JoinColony/colonyCDapp/blob/d267e29e1dccde7c0804568baf99ceba8c009093/src/i18n/en.json#L1091)

iamsamgibbs (approved):
Thanks for the changes @bassgeta Looks good to me now! 🎉

rdig (approved):
Good to go

Issue #2952, opened 2024-08-14
Crypto-to-fiat: Calculate receive amount in user hub tab


jakubcolony:
### Story

As a user, I want to see the fees that will be applied to my withdrawal and the final amount I will receive.

<img width="431" alt="image" src="https://github.com/user-attachments/assets/29c55b95-f38e-4037-be82-b9b48d2e4505">

### Description

This issue is to add logic to calculate the final amount user will receive. This amount is subject to the following fees:
- Gateway fee (AKA Developer fee, see [Bridge docs](https://apidocs.bridge.xyz/docs/developer-fees)) - a new handler will need to be added to bridgeXYZMutation lambda
- Wire fee - fixed fee dependent on transaction method (https://apidocs.bridge.xyz/docs/transaction-costs)

Issue #2665, opened 2024-07-05
Hackathon: New user hub tab UI/UX Updates


melyndav:
## Description

As part of the hackathon, the following updates are required to the new user hub functionality. As I can't complete a transaction, I'm unable to see the completed states etc so will test these at a later date. 

- [ ] When selecting the Max option, the amount doesn't correspond with the balance shown.

<img width="438" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/1ec560e5-797e-4dbe-b6ef-7fb55cef9406">

- [ ] Hover state is missing from the Max text button (should be text-blue-300).
<img width="58" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/7372272d-dafa-408a-9ec1-7a47fea9c425">

- [ ] Font weight of both values should match Figma

<img width="166" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/e4004b59-e87c-4906-a50f-b91648972caf">

- [ ] Correct currency icons are missing for both USDC and EU etc.

<img width="76" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/04a5d5b1-29b6-43c7-9813-124658259a6e">
 
- [ ] All validation variables is missing

<img width="434" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/09d5116e-982c-4b3a-ab6e-b625eb876155">

- [ ] Tooltips are missing on the transaction fee tables

<img width="36" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/4cad01c1-749b-4b96-9c2e-1330c880c47b">

- [ ] Transaction fee text should be a lighter gray token to match Figma

<img width="127" alt="image" src="https://github.com/JoinColony/colonyCDapp/assets/82127904/b638e7ba-fc04-4d87-b762-795a47b4b9e3">

jakubcolony:
> I'm sure this is a known issue, but there is no checking if the validation steps have been completed as I have not completed validation but can go straight into using the feature.

@melyndav I extracted the above into its own issue - #2953

rdig:
@mmioana can we get an estimate on this for the remaining work you're doing in #2980 ?

PR #2980, opened 2024-08-22
Fix: User hub UI/UX updates for crypto to fiat tab

mmioana:
## Description

**As part of the hackathon, the following updates are required to the new user hub functionality**

- [ ] When selecting the Max option, the amount doesn't correspond with the balance shown.

image

- [ ] Hover state is missing from the Max text button (should be text-blue-300).
image

- [ ] Font weight of both values should match Figma

image

- [ ] Correct currency icons are missing for both USDC and EU etc.

image
 
- [ ] All validation variables is missing

image

- [ ] Tooltips are missing on the transaction fee tables

image

- [ ] Transaction fee text should be a lighter gray token to match Figma

image
 

[Figma link](https://www.figma.com/design/C2grfQysdsYXz0j4rADR6K/Crypto-to-Fiat?node-id=4496-126047&t=oXY5LcTCkwrBygr0-0)

## Testing

TODO: Please check the above mentioned changes are in place (or any other Figma mismatch)

> [!IMPORTANT]
> The values for the summary card will be updated in another PR related to [#2952](https://github.com/JoinColony/colonyCDapp/issues/2952)

Resolves [#2665](https://github.com/JoinColony/colonyCDapp/issues/2665)


iamsamgibbs (changes requested):
Nice job with this! Thank you for tidying up the rushed hackathon code!

Screenshot 2024-08-22 at 11 27 35

Tooltips look great:

Screenshot 2024-08-22 at 11 27 44

Screenshot 2024-08-22 at 11 27 40

Also looking good on dark mode:

Screenshot 2024-08-22 at 11 28 04

The only requirement missing from the issue is the max button not setting the correct value (I think you can just remove the decimals value from the Numeral component in the Withdraw Card so that it shows the full balance)

Screenshot 2024-08-22 at 11 28 22

It also doesn't correctly format the number when it is set via the max button, it should be comma separated.

src/components/common/Extensions/UserHub/partials/CryptoToFiatTab/partials/ReceiveCard.tsx
```diff
@@ -32,25 +44,38 @@ const ReceiveCard: FC<ReceiveCardProps> = ({
...
-  const { error } = getFieldState(name);
+  const {
+    setValue,
+    formState: { errors },
+  } = useFormContext();
+  const amountValue = useWatch({ name });
+  const error = get(errors, name)?.message as Message | undefined;
 
-  const amountValue = getValues(name);
+  const { currency: selectedCurrency } = useCurrencyContext();
```
iamsamgibbs, line 56: The selected currency needs to come from the details the user has entered in the crypto to fiat user profile section rather than the currency context

iamsamgibbs (changes requested):
Great job fixing up the 'max' button! It is working as intended now!

Screenshot 2024-08-23 at 14 00 38

Although I realised whilst re-reviewing this that the 'receive' currency should only ever be in EUR/USD dependant on the bank details entered when the user signs up to crypto to fiat rather than the currency selected in the currency context. (I'm not sure if that detail is currently stored in another context or not, whether it is a simple change to make here? Otherwise that may increase the scope so I think just leaving the currency hardcoded here to avoid confusion might be preferred - let me know what you think! :) )

src/components/common/Extensions/UserHub/partials/CryptoToFiatTab/partials/ReceiveCard.tsx
mmioana: Hey @iamsamgibbs I already did this change into this PR https://github.com/JoinColony/colonyCDapp/pull/2988 as I realised it wasn't properly set and shouldn't have used the useCurrencyContext()

iamsamgibbs (approved):
Just seen that you have fixed this on #2988 - so approving here!

bassgeta (approved):
The devil's in the details, really nice job tackling this bunch of different issues and formatting values like they should.
It's kinda lame that there aren't many options to format a value jsut for display purposes, but keeping it as a number under the hood. Now we have to do double parsing, but hey that's life in React right now :D

Max works
image
Hover:
image
The icons are changed and the font weights are in line with Figma :heavy_check_mark: 
Validation:
image
Tooltips:
image
And the fee texts are also a lighter shade of gray, LGTM :sunglasses:

mmioana:
I rebased this branch and I have lost your approvals @bassgeta @iamsamgibbs 😢 
Could you please give it another try?

iamsamgibbs (changes requested):
Given this another test after the rebase and looks good!

Screenshot 2024-08-27 at 11 22 57

Although, now that #2956 has been merged in, I've spotted this hover state issue with the withdraw / receive cards. When disabled, the border shouldn't change on hover:

Screenshot 2024-08-27 at 11 22 07

I think it makes sense to fix this up here as it will be a tiny change and there are already changes to the withdraw / receive cards.

mmioana:
@iamsamgibbs thank you for noticing this issue regarding the hovering state! I'm on to solving it 🚀

iamsamgibbs (approved):
Great job, thanks for the border hover fix (and all the other disabled state fixes you've added!)

https://github.com/user-attachments/assets/8b04c56c-4ed0-402c-9f12-6d4cec8e3653

bassgeta (approved):
Everything still working and looking good after the rebase, thanks for also pulling out that form field to an uppercase constant!
Additionally, the hover states also work now:
image

Let's get this nice functionality merged, only the balance awaits :sunglasses:

rumzledz (approved):
Merge it @mmioana 🔥

mmioana:
@rdig I tested it now and can confirm we get the USD conversion and the max value updates properly
Have you set up the coin gecko env vars and run `npm run prepare`?

Screenshot 2024-08-29 at 12 39 22

https://github.com/user-attachments/assets/d79ee706-0359-4a44-99e8-83574346e9eb

rdig:
I believe I do, but will re-test to make sure.

PR #2992, opened 2024-08-23
Fix: Intermittent "not ready to send" transaction errors

### LINKED ISSUE #2957: Advanced payments: Intermittent "not ready to send" transaction errors
## Steps to reproduce

This is rather tricky to reproduce, I could not identify a pattern, and the issue is usually gone after a page refresh.

Following these steps eventually reproduces it for me:
1. Create advanced payment
2. Lock the payment
3. If previous transaction succeeds, fund the payment

Lock and fund transactions intermittently fail with `Transaction ${id} is not ready to send.` error in the console.

## Suggestions (optional)

Most likely something has changed in the sagas after the refactor that advanced payments sagas do not follow properly.

All sagas in `expenditures` should be updated with the fix.

iamsamgibbs:
## Description

Occasionally the fundingExpenditure saga would fail with a "not ready to send" transaction error.

This was occurring because the transaction status was set to `pending` rather than `ready`.

The incorrect transaction status was being caused by `yield put(transactionPending(fundMulticall.id))` which sets the transaction status to pending, **however** it doesn't wait for the mutation to complete before proceeding. This meant on some occasions the `pending` mutation would complete **after** the `ready` mutation even though it was called first.

This PR removes `yield put(transactionPending(fundMulticall.id))` as it was being set to `ready` straight away after anyway, so setting it to `pending` first was unnecessary.

I have opened a separate PR on master (#2994) to make `yield put(transactionPending(transactionId))` actually wait for the mutation to complete before proceeding which will address any potential edge cases on other sagas where it is used.

## Testing

Step 1: Create an advanced payment action (I found the bug was happening more consistency with a larger number of payments, so feel free to import this CSV: https://drive.google.com/file/d/1h70lE_I_6Aa3L0cDRG1f9mjlzrvkBJfd/view?usp=sharing)
Step 2: Proceed through the rest of the flow to the funding stage. Fund using the permissions decision method.

The transaction should complete without issue - there is no need to complete the rest of the flow.

Whereas, on `feat/advanced-payments-ui` the funding step will fail on occassion:

Screenshot 2024-08-23 at 12 42 36

## Diffs

**Deletions** ⚰️

* `yield put(transactionPending(fundMulticall.id))` removed from fundingExpenditure saga

Resolves #2957

### DIFF
--- a/src/redux/sagas/expenditures/fundExpenditure.ts
+++ b/src/redux/sagas/expenditures/fundExpenditure.ts
@@ -1,7 +1,6 @@
 import { ClientType } from '@colony/colony-js';
 import { fork, put, takeEvery } from 'redux-saga/effects';
 
-import { transactionPending } from '~redux/actionCreators/index.ts';
 import { type Action, ActionTypes, type AllActions } from '~redux/index.ts';
 import { TRANSACTION_METHODS } from '~types/transactions.ts';
 
@@ -117,8 +116,6 @@ function* fundExpenditure({
         ActionTypes.TRANSACTION_CREATED,
       );
     }
-
-    yield put(transactionPending(fundMulticall.id));
 
     yield initiateTransaction(fundMulticall.id);
 


chmanie (approved):
Nice fix. Yes, this was most likely a leftover of all of the tx rebases that I did. This particular line of code shouldn't be anywhere, actually.

Nortsova (approved):
Hey @iamsamgibbs ! thanks for this fix! works great!

`feat/advanced-payments-ui` behaviour:
image

current behaviour:
image


Note: here is not related to this PR thing that I noticed, recipient items renders too long, maybe we can investigate and optimize rendering here: 
items rendering issue

I think something like `react-virtualized` could solve this bug easily, but need more investigation on it

iamsamgibbs:
@Nortsova I also noticed the same rendering issues, these aren't occurring on master, there is also a `feat/advanced-payments-ui-next` branch, so I'm assuming there is a fix on one of these other branches that just hasn't been applied here yet.

PR #3018, opened 2024-08-28
feat: user hub updated with streams

joanna-pagepro:
## Description

- [] TODO write description

## Testing

* Step 1. E.g. add reputation
* Step 3. E.g. Create motion

## Diffs

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

* Changes go here

**Deletions** ⚰️

* Deletions go here

## TODO

- [ ] Add TODOs


src/components/common/Extensions/UserHub/partials/BalanceTab/partials/BalanceInfoRow/BalanceInfoRow.tsx
```diff
@@ -0,0 +1,159 @@
...
+              value={tokenBalanceData?.activeBalance ?? 0}
+              decimals={tokenDecimals}
+              suffix={` ${multiLineTextEllipsis(nativeToken?.symbol ?? 'CLNY', 5)}`}
+              appearance={{ size: 'small' }}
+            />
+          ),
+        },
+        {
+          key: '3',
+          label: formatText(MSG.activeFunds),
```
iamsamgibbs, line 117: Should be MSG.stakedFunds

src/components/common/Extensions/UserHub/partials/BalanceTab/partials/StreamsInfoRow/StreamsInfoRow.tsx
```diff
@@ -0,0 +1,186 @@
...
+  const getTotalFunds = useCallback(
+    async (items: StreamingPaymentItems) => {
+      const { totalAvailable, totalClaimed } = await calculateTotalsFromStreams(
+        {
+          streamingPayments: items,
+          currentTimestamp: Math.floor(blockTime ?? Date.now() / 1000),
+          currency,
+          colony,
+        },
+      );
```
iamsamgibbs, line 98: I'm not sure this works as it assumes all streaming payments are from the current colony. It seems like the `useGetStreamingPaymentsQuery` gets all the streaming payments for that user from every colony, so we should instead group the streaming payments by colony before passing them to this function - or something like that, there may be a better way to do it.

iamsamgibbs (changes requested):
Nice work on this so far! I've left a few comments.

The UI looks good and I can see the number ticking up here:

https://github.com/user-attachments/assets/a19c1c50-8ee3-4a49-b055-8ffffd9751da

This is probably a personal preference thing, but the numbers all jumping into place when opening the balance tab is a bit distracting and might be better to just be static when it loads and only animate on update:

https://github.com/user-attachments/assets/7d33eda5-6556-4756-a167-d0858965c93a

Reputation tab also looks good:

Screenshot 2024-09-04 at 09 54 27

On mobile I wasn't sure if these buttons should be stacked or not, it seems like the figma mobile design might need updating as there are a few differences.

Screenshot 2024-09-04 at 11 14 54

Figma:
Screenshot 2024-09-04 at 11 15 09

One thing I wanted to check as it isn't clear from this PR or the original issue - is the 'total claimed' and 'available to claim' supposed to be for specifically within that colony? The "Your overview in Wayne Enterprises" to me suggests it should be for just this colony.

Screenshot 2024-09-04 at 11 13 10

Currently it shows the same value in any colony and is fetching streaming payments for all colonies I believe - see the comment I left regarding this.

joanna-pagepro:
@iamsamgibbs thanks for your review!
I've updated everything and added a filter to get streaming payments for the current colony. I've also added a few more todo comments because those imports will be needed when part of the code will be uncommented.

iamsamgibbs (approved):
Nice work! Thank you for making those changes and adding the comment re imports.

Tested with a streaming payment in planex:

Screenshot 2024-09-05 at 12 19 36

And without in Wayne:

Screenshot 2024-09-05 at 12 19 27

jakubcolony (changes requested):
Thanks for your PR @joanna-pagepro, I left some code comments for you to look at.



image

I was not able to test the UI using action form as it's still affected by the bug raised in #2938 regarding wrong units the start/end time is set in, but using the testing UI I can confirm the amounts look correct, and they update every second:

https://github.com/user-attachments/assets/9aaa6c48-8689-47c8-a61e-503c3e359c4f

arrenv (changes requested):
Thank you @joanna-pagepro for your work and your patience. Nice work, I have noticed a couple of things.

1. After creating a streaming payment, the UI did not update, I had to reload the app. Can we have it update without requiring an upload.

2. The value does not appear to be aligned correctly on my end.

    image

3. I am not able to test completely as the claim is not working for me. I am getting a permissions error.

    image

src/components/common/Extensions/UserHub/partials/BalanceTab/partials/StreamsInfoRow/StreamsInfoRow.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+  const getTotalFunds = useCallback(
+    async (items: StreamingPaymentItems) => {
+      const { totalAvailable, totalClaimed, isActive } =
+        await calculateTotalsFromStreams({
+          streamingPayments: items,
+          currentTimestamp: Math.floor(blockTime ?? Date.now() / 1000),
+          currency,
+          colony,
+        });
```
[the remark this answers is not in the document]
jakubcolony: That's exactly what I meant, thank you

jakubcolony (approved):
Code-wise all looks good, nice work. I can confirm I'm getting the same issues Arren reported, I was only able to claim by manually assigning permissions.

joanna-pagepro:
@jakubcolony I think that there's a problem when installing the extension because the enable action is not called if there are no `initializationParams` and if I understand it correctly, the extension is installed without correct permissions. Can you help me with this one?

iamsamgibbs:
Ah, seems like this is likely related to #3027 (I'm assuming it made it's way onto a parent branch somewhere further up the chain) - looks like I mistakenly assumed having no `initializationParams` meant it shouldn't be automatically enabled. Should be an easy fix to just check for `neededColonyPermissions` and still make the `setUserRoles` transaction even if there is no initialise step. I'll put up a fix for this today.

joanna-pagepro:
@arrenv can you check 1. and 2. again?

iamsamgibbs:
Actually on further inspection, the changes in #3027 don't appear to be on this branch, so won't be the cause of any issues here. Apologies for any confusion!

arrenv (changes requested):
@joanna-pagepro Thank you for taking a look and addressing the issues.

1. Number of decimals should match the spec to have up to 5 decimal places, this is the only exception to the fiat currency number formatting being fixed to 2 decimal places.

    image

2. The stream rate did not update properly until reloading again.

    [streaming-reload.webm](https://github.com/user-attachments/assets/15617cde-de4f-4844-823d-ec156f60535d)

joanna-pagepro:
@arrenv can you check again? Also, what do you mean in 1? Should there be 5 decimal places instead of 2?

arrenv:
> @arrenv can you check again? Also, what do you mean in 1? Should there be 5 decimal places instead of 2?

Yes, it should show up to 5 decimal places, it is influenced by the rate. Just as of my example I had sent (https://codesandbox.io/p/sandbox/bold-feather-xfh4q3).

For example:
- if the rate is `0.0001` per second, then we should show 4 decimal places.
- if the rate is `0.01` per second, then we should show 2 decimal places.
- if the rate is `0.01457` per second, then we should show 5 decimal places.

joanna-pagepro:
@arrenv decimal places updated. I've also changed the animation in action details view

arrenv (approved):
@joanna-pagepro Thank you for the updates. All looks good now.

image


image


I still need to reload the app to see the new stream, although, the same applies to the activity feed. So, I think this needs to be tackled in another issue.

PR #3027, opened 2024-08-28
Feat: Add extension install and enable saga

### LINKED ISSUE #2876: Staking advanced payments installation transactions
## Steps to reproduce

1. Go to the extension page of a colony.
2. Go to the Staking advanced payments extension.
3. Install and enable the extension.
4. Note the transactions required to install and enable.

## Expected behaviour

* Transactions should be multi-called as much as possible.
* There should not be two separate transaction groups.

## Actual behaviour

* There are two separate groups of transactions required to install and enable the transaction.

[staking-ext-install.webm](https://github.com/user-attachments/assets/c6328bbd-2a4a-4c61-b0d4-ac37288e89bb)

iamsamgibbs:
## Description

The "Staking Advanced Payments" extension should be automatically initialised and assigned permissions when it is installed. Currently this is achieved by calling the `extensionEnable` saga after the `extensionInstall` saga successfully completes. This does work, but means there are two separate transaction groups to install then initialise and enable which is not ideal (see screenshot).

Screenshot 2024-08-29 at 10 25 16

This PR replaces the `extensionInstall` saga with a new `extensionInstallAndEnable` saga. It uses the `enabledAutomaticallyAfterInstall` extension data property to determine whether the extension should be enabled or not as part of the saga.

The new saga uses grouped transactions as opposed to multicall. This is a deliberate decision as the initialise transaction requires the extension client context rather than the colony client context, and the setUserRoles transaction requires the extension address which is not available until after the install transaction has completed. The transactions will now appear within the same group.

Screenshot 2024-08-29 at 10 15 36

## Testing

* Step 1 - Switch to a metamask wallet (message me if you need help setting this up)
* Step 2 - Deprecate the "Staking Advanced Payments" extension
* Step 3 - Uninstall the "Staking Advanced Payments" extension
* Step 4 - Install the "Staking Advanced Payments" extension

There should be three pending transactions.
 
Screenshot 2024-08-29 at 10 15 00

Confirm these transactions, after the final transaction the tab will automatically switch to the settings tab. The extension will be initialised and enabled, you should not see the enable button or missing permissions banner at any point. (It can take a little while for the permissions to get assigned before being redirected).

Screenshot 2024-08-29 at 10 20 15

Check the transactions in the user hub, the transactions should all be part of the 'Add extension to Colony' group.

Screenshot 2024-08-29 at 10 15 36

Test that the installation, initialising and enabling has not been broken for any of the other extensions - notably the reputation weighted voting extension which has a separate initialising step. (Note: The Staking Advanced Payment extension is currently the only one which is enabled when it is installed.)

## Diffs

**New stuff** ✨

* Added `extensionInstallAndEnable` saga

**Changes** 🏗

* On a successful install, wait for the extension to receive it's permissions to avoid showing "enable permissions" banner

**Deletions** ⚰️

* Removed `extensionInstall` saga

Resolves #2876


src/components/frame/Extensions/pages/Extensions/ExtensionDetailsPage/partials/InstallButton.tsx
```diff
@@ -112,70 +72,32 @@ const InstallButton = ({
...
         setActiveTab(ExtensionDetailsPageTabId.Settings);
-      } catch {
-        toast.error(
+      } else {
+        await waitForDbAfterExtensionAction({
+          method: ExtensionMethods.INSTALL,
+          refetchExtensionData,
+        });
+
+        setIsPolling(false);
```
rumzledz, line 89: Looking at `setIsPolling(false)`, I think it's being called in all if-else blocks and in both the `catch ` and `try` blocks. Maybe we can just add this inside `finally`? But at the same time, maybe there's a reason why these have to be called at specific parts of the flow

rumzledz (approved):
Awesome work on this one @iamsamgibbs 🔥 

I can confirm that there are three pending transactions when installing the Staking Advanced Payments saga and that it automatically switches to the Extension settings tab after it's done:



I also tried installing another extension (Reputation Weighted) and all is well 👌

mmioana (approved):
Nailed another one @iamsamgibbs 🥇 

Tested and can confirm the `Staking Advanced Payments` extension gets installed and enabled in a grouped transaction
Screenshot 2024-09-02 at 10 36 27
Screenshot 2024-09-02 at 10 36 41

https://github.com/user-attachments/assets/9d32f478-2b9c-4e62-b002-77dfbea4bc9f

Screenshot 2024-09-02 at 10 37 41

Code-wise looks good from my side 👍

bassgeta (approved):
Very nice, code changes look sensible and clean!
As for the functionality, I can confirm that installing the extension works:
1. Uninstalled
image
2. Then installing it:
image
image
However, I'm not sure if your changes are at fault, installing the voting reputation extension seemingly does nothing, after a refresh I get into this state where I need to manually enable it:
image

Not using `Metamask` helps, but maybe there's something broken on my end since I also get this banner a lot
image

could be a problem on my machine since 2 other reviewers had no issues, so I'm approving it nonetheless :)

Issue #3012, opened 2024-08-27
Split payment fixes to formatting and rounding


arrenv:
Teamwork task: [Split payment fixes to formatting and rounding](https://tw.auroracreation.com/#/tasks/16048170)

## Steps to reproduce

1. Create a Split payment action.
3. Split up the payment so that it results in a decimal percent share. e.g. by thirds.
2. Use an amount that should split equally by thirds e.g. 2343.
4. Note comma used to represent the decimal split.
5. Execute the transaction.
6. On the complete view, click the kebab icon in the top right of the action screen.
7. Click the "Redo action" button.
8. Note that the values are higher than the originally set 2343.

I can understand that this is likely rounding due to the fractions being used, however, the rounding is not required with an equal split. Also, it should not be higher then what the user wants to spend.

## Expected behaviour

* We should use dots for representing decimal values.
* Total amounts when using rounding should not be more then initially intended to be spent.
* Redoing an action should show the same values used for the original action.

## Actual behaviour

* There is a comma used to split the decimal value, which is inconsistent with the rest of the values.
* Total amounts is higher then the originally intended value.
* Redoing an action shows the true values being used.

### Percent values are using commas and dots
![image](https://github.com/user-attachments/assets/a7599560-6ae7-44f6-b76d-405d51a1055e)

### Rounding of values is higher than available balance
[split-redo.webm](https://github.com/user-attachments/assets/18002fdf-7464-4541-8baf-4016606bf43a)

PR #3042, opened 2024-08-30
fix: split payment table ui

CzarekDryl:
## Description

- Equal distribution method now uses fractions to calculate amount instead of percentages
- Changed decimals to dots everywhere in payment table
- Total amounts is not higher then the originally intended value
image

## Testing

* Create a split payment and play with it

(Resolves | Contributes to) https://github.com/JoinColony/colonyCDapp/issues/3012

### DIFF
--- a/src/components/shared/Numeral/Numeral.tsx
+++ b/src/components/shared/Numeral/Numeral.tsx
@@ -13,11 +13,11 @@ import type Decimal from 'decimal.js';
 
 import styles from './Numeral.module.css';
 
+const displayName = 'Numeral';
+
 // needed for capitalized abbreviations
 numbro.registerLanguage(numbroLanguage);
 numbro.setLanguage('en-GB');
-
-const displayName = 'Numeral';
 
 export type NumeralValue = string | number | BigNumber | Decimal;
 
--- a/src/components/v5/common/ActionSidebar/partials/forms/SplitPaymentForm/partials/SplitPaymentPayoutsTotal/SplitPaymentPayoutsTotal.tsx
+++ b/src/components/v5/common/ActionSidebar/partials/forms/SplitPaymentForm/partials/SplitPaymentPayoutsTotal/SplitPaymentPayoutsTotal.tsx
@@ -13,6 +13,7 @@ const SplitPaymentPayoutsTotal: FC<SplitPaymentPayoutsTotalProps> = ({
   token,
   convertToWEI,
   className,
+  value,
 }) => {
   const summedAmount =
     useMemo(() => {
@@ -44,7 +45,7 @@ const SplitPaymentPayoutsTotal: FC<SplitPaymentPayoutsTotalProps> = ({
         'flex items-center gap-3 text-gray-900 text-1',
       )}
     >
-      <Numeral value={summedAmount} decimals={token.decimals} />
+      <Numeral value={value || summedAmount} decimals={token.decimals} />
       <div className="flex items-center gap-1">
         <TokenAvatar
           tokenAddress={token.tokenAddress}
--- a/src/components/v5/common/ActionSidebar/partials/forms/SplitPaymentForm/partials/SplitPaymentPercentField/SplitPaymentPercentField.tsx
+++ b/src/components/v5/common/ActionSidebar/partials/forms/SplitPaymentForm/partials/SplitPaymentPercentField/SplitPaymentPercentField.tsx
@@ -26,7 +26,8 @@ const SplitPaymentPercentField: FC<SplitPaymentPercentFieldProps> = ({
         }
       }}
       wrapperClassName="flex-row flex text-md"
-      type="number"
+      type="text"
+      shouldAllowOnlyNumbers
       mode="secondary"
       placeholder={formatText({ id: 'actionSidebar.enterValue' })}
       suffix="%"
--- a/src/components/v5/common/ActionSidebar/partials/forms/SplitPaymentForm/partials/SplitPaymentRecipientsField/hooks.tsx
+++ b/src/components/v5/common/ActionSidebar/partials/forms/SplitPaymentForm/partials/SplitPaymentRecipientsField/hooks.tsx
@@ -1,5 +1,6 @@
 import { createColumnHelper, type ColumnDef } from '@tanstack/react-table';
 import Decimal from 'decimal.js';
+import moveDecimal from 'move-decimal-point';
 import React, { useMemo, useCallback, useEffect } from 'react';
 import { type FieldValues, type UseFieldArrayReturn } from 'react-hook-form';
 
@@ -31,6 +32,7 @@ export const useRecipientsFieldTableColumns = ({
   amount,
   fieldArrayMethods: { update },
   disabled,
+  distributionMethod,
 }: {
   name: string;
   token: Token;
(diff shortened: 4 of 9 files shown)


src/components/shared/Numeral/Numeral.tsx
mmioana, line 1: I would propose to create a new component `NumeralWithDots` which encapsulated the usage of `Numeral` and passes the formatted values, instead of adding the `shouldUseDots` property to the existing `Numeral`.

Hence, we'll have
```
export const NumeralWithDots = ({formattedValue, ...rest}: Props) =>
  <Numeral 
      {...rest} 
      formattedValue={typeof formattedValue === 'string' ? formattedValue.replace(',', '.') : formattedValue} 
  />
```

src/components/v5/common/ActionSidebar/partials/forms/SplitPaymentForm/partials/SplitPaymentPayoutsTotal/SplitPaymentPayoutsTotal.tsx
```diff
@@ -44,7 +44,7 @@ const SplitPaymentPayoutsTotal: FC<SplitPaymentPayoutsTotalProps> = ({
         'flex items-center gap-3 text-gray-900 text-1',
       )}
     >
-      <Numeral value={summedAmount} decimals={token.decimals} />
+      <Numeral value={summedAmount} decimals={token.decimals} shouldUseDots />
```
mmioana, line 47: And use the above defined component here 
`<NumeralWithDots value={summedAmount} decimals={token.decimals} />`

mmioana (changes requested):
Amazing start on this one @CzarekDryl 👏 

However, noticed when there is a single recipient in the payments table, the amount still contains the `,` instead of a `.`
Screenshot 2024-09-02 at 11 26 18

When adding more recipients the values are correctly formatted
Screenshot 2024-09-02 at 11 27 21
Screenshot 2024-09-02 at 11 28 10
Screenshot 2024-09-02 at 11 28 22

And another UX point, I noticed the form's amount field still uses `,` for separating decimals, but then in the table we use the `.`
Can this be a consistency issue @arrenv @melyndav ?

CzarekDryl:
@iamsamgibbs Redo action was merged to other branch, so I think we will need another issue for this case. Everything else should work as expected now
image
image

iamsamgibbs (changes requested):
Thanks for the changes, this is looking a lot better!

The comment regarding the "redo action" makes sense and can be handled in a separate PR - lets make sure it doesn't get lost though.

The "Total amount" field is working much better now:

Screenshot 2024-09-17 at 11 17 16

Although I did find when the total is exactly `1000` that the comma doesn't show, only once you go to `1,001`:

Screenshot 2024-09-17 at 11 17 27

Screenshot 2024-09-17 at 11 19 06

CzarekDryl:
@iamsamgibbs As we talked on discord. In equal distribution method I'm passing value from amount field instead of calculating it from table rows.
image

src/components/v5/common/ActionSidebar/partials/forms/SplitPaymentForm/partials/SplitPaymentPercentField/SplitPaymentPercentField.tsx
```diff
@@ -26,7 +26,8 @@ const SplitPaymentPercentField: FC<SplitPaymentPercentFieldProps> = ({
         }
       }}
       wrapperClassName="flex-row flex text-md"
-      type="number"
+      type="text"
+      shouldAllowNumbers
```
jakubcolony, line 30: What is the difference between that and `type="number"`?

src/components/v5/common/Fields/InputBase/FormInputBase.tsx
```diff
@@ -47,7 +48,14 @@ const FormInputBase: FC<FormInputBaseProps> = ({
...
           onChange(Number.isNaN(valueAsNumber) ? 0 : valueAsNumber);
         } else {
-          onChange(inputValue);
+          onChange(
+            shouldAllowNumbers
+              ? inputValue
+                  .replace(/[^0-9.]/g, '')
+                  .replace(/(\..*?)\..*/g, '$1')
+                  .replace(/^0[^.]/, '0')
+              : inputValue,
```
jakubcolony, line 57: Rather than allowing numbers, isn't this disallowing non-numbers? Setting `shouldAllowNumbers` to false still allows them

src/components/v5/common/ActionSidebar/partials/forms/SplitPaymentForm/partials/SplitPaymentRecipientsField/hooks.tsx
```diff
@@ -161,24 +170,28 @@ export const useRecipientsFieldTableColumns = ({
...
+                    {distributionMethod === SplitPaymentDistributionType.Equal
+                      ? '100%'
+                      : `${Number(
+                          dataRef.current
+                            ?.reduce(
+                              (acc, _, index) =>
+                                Number(acc) + Number(getPercentValue(index)),
+                              0,
+                            )
+                            .toFixed(4),
```
jakubcolony, line 185: this needs a fallback if `dataRef.current` is undefined

jakubcolony (changes requested):
@CzarekDryl #2925 also got merged into `feat/advanced-payments-ui-next`, so has this not been rebased that the "redo" action is missing?

Otherwise nice start with the fixes!

The only problem I spotted is that selecting "Equal" distribution type shows `undefined` with bunch of zeros as the total:
image

Everything else looks good:

image


image

image

src/components/v5/common/ActionSidebar/partials/forms/SplitPaymentForm/partials/SplitPaymentPercentField/SplitPaymentPercentField.tsx
CzarekDryl: Input with `type="number"` doesn't allow to type dots

src/components/v5/common/Fields/InputBase/FormInputBase.tsx
CzarekDryl: Renamed to `shouldAllowOnlyNumbers`

CzarekDryl:
@jakubcolony @iamsamgibbs I've rebased this branch so redo action is now available in split payment

iamsamgibbs (changes requested):
Thanks for making that change, this looks good now!

Screenshot 2024-09-27 at 12 07 52

I've noticed an issue with redoing the action unfortunately! The total doesn't get set properly:

https://github.com/user-attachments/assets/948a8e25-a530-42a4-bcc0-7ff9007da085

Screenshot 2024-09-27 at 12 11 16

CzarekDryl:
@iamsamgibbs Redo action should work fine now

https://github.com/user-attachments/assets/453dc314-d4e0-4cea-9e4e-3552eddc0650

iamsamgibbs (changes requested):
Still having rounding issues I'm afraid @CzarekDryl 

https://github.com/user-attachments/assets/dbcd35d6-9b1d-4887-970b-13243ee6c73b

https://github.com/user-attachments/assets/4eb01aaf-152b-44fc-a7c7-a5eb6579730a

PR #3045, opened 2024-08-30
Feat: Create Magicbell user and notifications data for user

### LINKED ISSUE #3004: Notifications: Create Magicbell user
### Story

As a *user*, I need a *user created for me in Magicbell*, so that *the Magicbell system can send me notifications in the Colony App*

### Description

Each user in our system should have a corresponding user in the magicbell system, so that we can send notifications to them. The magicbell user must have a unique user id, which in our case can just be the wallet address.

When the app loads / users log in, we should check if the user has a notifications data object with a Magicbell unique user ID. If not, then it means that they have not yet had a Magicbell user made. In this case, we should create the user, and then once that has finished, update the user in our system to include notifications data with their Magicbell unique user ID.

- [ ] Users have Magicbell users created for them if needed when the app loads / they log in.

davecreaser:
## Description

- In order to keep track of user's notification settings and preferences, we need to store all of this notifications data inside our database. On top of this, to be able to send notifications via Magicbell, we need to create users in their system via the API.
- All of this should be done once for each user, so when the app loads or the user logs in, we need to check if the user has notifications data already saved, and if not, then do so.
- This needs to be done inside a lambda, as we will be calling the Magicbell API using a key and a secret, which cannot be done browser side for obvious security reasons.

This PR covers the creation of this lambda which makes the API call and the notifications data object in our db, as well as a schema update to include notifications data as a new model, and also a creation of a new NotificationsContext, which will wrap the app and handle all of the logic.

There is also an update to the auth proxy, covered in [this pr](https://github.com/JoinColony/colonyCDappAuthProxy/pull/27) which just adds the lambda as an allowed mutation.

## Testing

We need to test that this creates users in Magicbell, and also creates the notifications data object in our db.

There is some setup to do first however:
1. Ask me for env variables that you need to add in a `.env` file to the new lambda. These will be the api key and api secret for Magicbell.
2. Ask me to invite you to the Magicbell project, so you can see the users being created.

Now you're good to test!

* Step 1. Check that there's no users made in Magicbell, or at least, that the user you are planning to test with is not there. You can do this by going to the Magicbell dashboard of the test project here `https://app.magicbell.com/projects/8931/users` and looking for the wallet address of the user.

Here's what the users section of the dashboard looks like when it's empty:

Screenshot 2024-08-30 at 17 06 50

If the user you want to test with exists already, you can delete it with this curl:

```curl
curl --location --request DELETE 'https://api.magicbell.com/users/external_id:<WALLET_ADDRESS>' \
--header 'X-MAGICBELL-API-KEY: <API_KEY>' \
--header 'X-MAGICBELL-API-SECRET: <API_SECRET>'
```

* Step 2. Check that the notifications data doesn't exist for the user already, by querying our db in `localhost:20002`:

Screenshot 2024-08-30 at 17 11 28

```graphql
query MyQuery {
  getUserByAddress(id: "<WALLET_ADDRESS>") {
    items {
      notificationsData {
        magicbellUserId
      }
      profile {
        displayName
      }
      id
    }
  }
  listNotificationsData {
    items {
      magicbellUserId
      id
      userId
    }
  }
}
```

If it does already exist, you can remove it with:

```graphql
mutation MyMutation {
  deleteNotificationsData(input: {id: "<WALLET_ADDRESS>"}) {
    id
  }
}
```

* Step 3. Once you have a clear state, all you need to do is load up the app, and log in as a user! You should see this mutation happen in the network tab `CreateUserNotificationsData`:

Screenshot 2024-08-30 at 17 14 36

If you then check the users dashboard of Magicbell to ensure that the user was created there, and run the query in our db again to check that the user notification data was created:

Screenshot 2024-08-30 at 17 20 46

Screenshot 2024-08-30 at 17 21 31

* Step 4. Refresh the app, and ensure that the `CreateUserNotificationsData` mutation is not called again (since the notifications data already exists, it doesn't need to be called again).

* Step 5. It would be useful for other testers if you could then clean up the Magicbell users list by deleting the user that you have made:

```curl
curl --location --request DELETE 'https://api.magicbell.com/users/external_id:<WALLET_ADDRESS>' \
--header 'X-MAGICBELL-API-KEY: <API_KEY>' \
--header 'X-MAGICBELL-API-SECRET: <API_SECRET>'
```

## Diffs

**New stuff** ✨

* New lamdba `CreateUserNotificationsData`
* New context `NotificationsContext`
* New model in db `NotificationsData`

**Changes** 🏗

* Changed auth proxy hash

Resolves #3004


amplify/backend/function/createUserNotificationsData/src/index.js
```diff
@@ -0,0 +1,85 @@
...
+    body: JSON.stringify({
+      user: {
+        external_id: walletAddress,
+      },
+    }),
+  });
+
+  const data = await response.json();
+
+  if (data.errors) {
```
rumzledz, line 54: Will it be safer to wrap this fetch in a `try/catch` block instead?

amplify/backend/function/createUserNotificationsData/src/utils.js
```diff
@@ -0,0 +1,39 @@
...
+  let body;
+  let response;
+
+  try {
+    response = await fetch(url, options);
+    body = await response.json();
+    return body;
+  } catch (error) {
+    /*
+     * Something went wrong... obviously 🦆
```
rumzledz, line 30: 🦆

src/context/NotificationsContext/NotificationsContextProvider.tsx
```diff
@@ -0,0 +1,40 @@
...
+const NotificationsContextProvider = ({
+  children,
+}: {
+  children: ReactNode;
+}) => {
+  const { updateUser, user } = useAppContext();
+  const [createUserNotificationsData] =
+    useCreateUserNotificationsDataMutation();
+
+  useEffect(() => {
```
rumzledz, line 18: I wonder if it's all right to perform this effect inside the AppContextProvider instead since that seems to handle user update operations as well. I see the value with decoupling the logic for notifications though but I wonder if it's worth it since there's technically nothing to consume from this Provider

But then again, if we do have something to consume in the future then totally worth keeping it the way it is

rumzledz (commented):
Hey @davecreaser, no change requests on my side! Just a bunch of questions about the schemas and lambdas to help me better understand this flow. I am a lambda noob after all 😂 Also left a question about the Notification provider

src/context/NotificationsContext/NotificationsContextProvider.tsx
mmioana: @rumzledz I think it totally makes sense to keep this provider as the idea would be in the future to display the notifications in the user hub

mmioana (approved):
You smashed this one @davecreaser 💪 👏 

Followed your testing steps and can confirm it works as expected and the magic bell user id gets inserted in the DB
Screenshot 2024-09-02 at 12 34 57
Screenshot 2024-09-02 at 12 35 05
Screenshot 2024-09-02 at 12 36 15
Screenshot 2024-09-02 at 12 36 31
Screenshot 2024-09-02 at 12 36 38
Upon page refresh data doesn't get re-created
Screenshot 2024-09-02 at 12 37 05

And cleaned the users from magic bell to enable smooth future testing 😁 
Screenshot 2024-09-02 at 12 37 50

src/context/NotificationsContext/NotificationsContextProvider.tsx
rumzledz: Oh right thanks @mmioana , in that case I shall use this as we can all forget about everything else I said 😂 
"But then again, if we do have something to consume in the future then totally worth keeping it the way it is"

amplify/backend/api/colonycdapp/schema/schema.graphql
```diff
@@ -1855,6 +1871,24 @@ type User @model {
...
+}
+
+"""
+Holds the notifications data for the user, such as their unique Magicbell user id, and their notifications preferences.
+"""
+type NotificationsData @model {
+  """
+  Unique identifier for the user
+  """
+  userId: ID!
```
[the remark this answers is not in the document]
davecreaser: I think it's not necessary to add an index to the userId, since it won't be queried by that (as far as I can see atm). I do wonder about changing this schema to instead be like this though:

```graphql
"""
Holds the notifications data for the user, such as their unique Magicbell user id, and their notifications preferences.
"""
type NotificationsData @model {
  """
  Unique identifier for the user
  """
  userId: ID! @primaryKey
  """
  Unique identifier for the user in Magicbell
  """
  magicbellUserId: ID
}
```

Since the id, userId and magicbellUserId are all likely to be the wallet address, so we can streamline it to just use the userId as the primary key (also since this model will _probably_ only ever be queried with the user). I think it's best for now to leave it as it is though, and as this feature develops over time we can revisit this and see if it's still fit for purpose, or if any indexing is needed.

src/context/NotificationsContext/NotificationsContextProvider.tsx
davecreaser: Yep, that's exactly it, this notifications provider is going grow in the future. The plan is for it to be an abstraction so we don't have magicbell code throughout the app, but instead have it only in one place. That way if we ever need to change from using magicbell to a different notifications provider (or even build one in-house) then we only have to change it in this context, not everywhere notifications are used.

amplify/backend/function/createUserNotificationsData/src/index.js
davecreaser: Good point, it's probably a good idea to do both! I'll put it in now.

amplify/backend/api/colonycdapp/schema/schema.graphql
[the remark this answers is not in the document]
rumzledz: All right that makes sense! Cheers @davecreaser

bassgeta (approved):
Nice one, I like the new context for all notifications shennanigans, and the changes are very straight forward and easy to build upon.
Upon opening `planex` the mutation fired:
image
and the data is also there
image
There's no call after a refresh:
image

And I also cleaned up after myself :sunglasses:

rdig (approved):
Tested with user `amy` and it created the user as expected. Nicely done.

I've left some comments above which I think we should seriously look into as I want our local dev environment to be on par with whatever we run in production.

Testing:

Initial:
Screenshot from 2024-09-02 18-50-25
Created user:
Screenshot from 2024-09-02 18-54-19
Screenshot from 2024-09-02 18-54-23
Screenshot from 2024-09-02 18-54-27
Cleaned up user after testing:
Screenshot from 2024-09-02 18-49-16

amplify/backend/api/colonycdapp/schema/schema.graphql
```diff
@@ -956,6 +966,12 @@ type Mutation {
   createUniqueUser(input: CreateUniqueUserInput): User
     @function(name: "createUniqueUser-${env}")
   """
+  Create notification data for a user
+  """
+  createUserNotificationsData(
+    input: CreateUserNotificationsDataInput
```
jakubcolony, line 972: should input be marked as required?

amplify/backend/function/createUserNotificationsData/src/event.json
```diff
@@ -0,0 +1,5 @@
+{
```
jakubcolony, line 1: this file can be deleted

amplify/backend/function/createUserNotificationsData/src/index.js
```diff
@@ -0,0 +1,89 @@
...
+const { graphqlRequest } = require('./utils');
+
+const { createNotificationsData } = require('./graphql');
+
+const MAGICBELL_USERS_URL = 'https://api.magicbell.com/users';
+
+let apiKey = 'da2-fakeApiId123456';
+let graphqlURL = 'http://localhost:20002/graphql';
+let magicbellApiKey = process.env.MAGICBELL_API_KEY;
+let magicbellApiSecret = process.env.MAGICBELL_API_SECRET;
```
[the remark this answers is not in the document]
jakubcolony: The way to make it work would be to place an `.env` file in the lambda's directory and use something like dotenv to load the values

amplify/backend/function/createUserNotificationsData/src/index.js
```diff
@@ -0,0 +1,89 @@
...
+    const data = await response.json();
+
+    if (data.errors) {
+      throw new Error();
+    }
+  } catch (error) {
+    throw new Error(
+      `Could not create Magicbell user with wallet address "${walletAddress}"`,
+    );
+  }
```
jakubcolony, line 62: Please consider logging the error out as it will make QA/prod debugging much easier

amplify/backend/api/colonycdapp/schema/schema.graphql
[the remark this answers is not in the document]
jakubcolony: @davecreaser how about simply using wallet address as the ID? That way it's deterministic, and you can utilise the auto-generated queries if needed.

amplify/backend/function/createUserNotificationsData/src/utils.js
```diff
@@ -0,0 +1,39 @@
+const fetch = require('cross-fetch');
+
+const TokenType = {
+  COLONY: 'COLONY',
+  ERC20: 'ERC20',
+};
```
jakubcolony, line 6: Is this used anywhere?

amplify/backend/function/createUserNotificationsData/src/index.js
```diff
@@ -0,0 +1,89 @@
+const fetch = require('cross-fetch');
```
jakubcolony, line 1: Are you planning to have a single lambda handling all calls to Magicbell or multiple ones?

jakubcolony (commented):
Solid piece of work @davecreaser! 🔨 

Left a couple of comments but overall great start on notifications!

amplify/backend/function/createUserNotificationsData/src/index.js
davecreaser: There will be two separate ones, this one is needed since it uses the secret key. When we send notifications it will be done via the block ingestor so won't be in a lambda. For fetching notifications, we'll use the built in Magicbell React context, and we can optionally pass that an HMAC code which we create using the userID and our secret key, so computing that will be done in a lambda.

amplify/backend/function/createUserNotificationsData/src/index.js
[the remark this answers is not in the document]
davecreaser: It looks like it's working fine without the dotenv package, but yeah I have a `.env` file in the lambda directory locally. I'll merge with it as is and leave a note for myself to clean this up / take a look at how we can mock this locally if needed.

amplify/backend/function/createUserNotificationsData/src/index.js
davecreaser: It will be logged like this in the amplify container anyway (tested it out just to make sure), and it looks like all the other lambdas also just throw in the same way.

amplify/backend/function/createUserNotificationsData/src/utils.js
davecreaser: Good catch, just a left over copy / paste relic 🙃

PR #3052, opened 2024-09-03
Fix: Advanced payments UI not updating after successful motion stake

### LINKED ISSUE #2882: Advanced payment funding motion, UI did not update despite Staking succeeding
## Steps to reproduce

1. Create an Advanced payment or any other action with Reputation decision method.
2. If doing an Advanced payment take the advanced payment to the funding step and use the Reputation decision method.
3. I fully staked in support.

Note: This was tested on QA, but did not happen every time, but happened more than once.
Note: I just found this to also happen on a Simple payment motion as well on production.

## Expected behaviour

* The UI should update after successfully staking.
* You should not get locked into the Support or Oppose tabs if you switch during a transaction.

## Actual behaviour

* The UI did not update after successfully staking, I had to reload the app in order to see the status of the motion change.
* The screenshot shows the staking succeeding in the transactions tab, but the staking widget with no update.
* I also ran into a previous issue where I could switch from "Support" to "Oppose" options while the transactions where in progress, and "Stake" button was disabled. Resulting in me not being able to stake again until reopening the action.

image

image

[locked-staking-voting.webm](https://github.com/user-attachments/assets/e466c85e-76e6-414d-bc81-ee0ca4cd09b0)

iamsamgibbs:
## Description

This PR fixes an issue with the funding motion for an advanced payment getting stuck on the staking step.

Screenshot 2024-09-03 at 13 29 03

The issue was occurring because the PaymentBuilderWidget would re-render whilst polling for motion updates, which would cause the `transform` function to change resulting in the `useAyncFunction` promiseListener to get unsubscribed.

This PR memoizes the `transform` function to prevent it changing on these re-renders unless the actual values change.

Whilst working on this I also noticed similar behaviour on the voting and reveal steps so have memoized those `transform` functions too.

This PR also disables the "Oppose" and "Support" buttons whilst the form is submitting.

## Testing

Follow these steps on `feat/advanced-payments-ui-next` if you'd like to replicate the original issue - it is also easier to replicate using the metamask wallet rather than a developer wallet.

Step 1 - Install and enable the reputation weighted voting extension
Step 2 - Create an advanced payment action using permissions
Step 3 - Confirm details for the advanced payment
Step 4 - Fund using the reputation decision method
Step 5 - Support the motion and click "Stake" - if you are on `feat/advanced-payments-ui-next` this is where the bug should occur whilst confirming the transactions (it doesn't always occur so may take a few attempts to replicate - vote to oppose the motion to make it easier to repeat).
Step 6 - Whilst the stake saga is processing, the "Oppose" and "Support" buttons should be disabled.

Screenshot 2024-09-03 at 13 25 58

Step 7 - Follow through the rest of the motion to confirm the vote and reveal steps work as expected.

## Diffs

**Changes** 🏗

* Memoized `transform` functions for staking, voting and revealing
* Disabled "Oppose" and "Support" buttons whilst form is submitting

Resolves #2882

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/Motions/steps/RevealStep/hooks.ts
+++ b/src/components/v5/common/ActionSidebar/partials/Motions/steps/RevealStep/hooks.ts
@@ -78,11 +78,15 @@ export const useRevealStep = ({
 
   const { vote, hasUserVoted, userVoteRevealed, setUserVoteRevealed } =
     useRevealWidgetUpdate(voterRecord || [], stopPollingAction);
-  const transform = mapPayload(() => ({
-    colonyAddress,
-    userAddress: user?.walletAddress ?? '',
-    motionId: BigNumber.from(motionId),
-  }));
+  const transform = useMemo(
+    () =>
+      mapPayload(() => ({
+        colonyAddress,
+        userAddress: user?.walletAddress ?? '',
+        motionId: BigNumber.from(motionId),
+      })),
+    [colonyAddress, user?.walletAddress, motionId],
+  );
 
   const handleSuccess: OnSuccess<Record<string, number>> = (_, { reset }) => {
     reset();
--- a/src/components/v5/common/ActionSidebar/partials/Motions/steps/StakingStep/partials/StakingForm/StakingForm.tsx
+++ b/src/components/v5/common/ActionSidebar/partials/Motions/steps/StakingStep/partials/StakingForm/StakingForm.tsx
@@ -132,7 +132,7 @@ const StakingForm: FC<StakingFormProps> = ({
                             checked && !disabled && isDarkMode,
                         }),
                       icon: ThumbsDown,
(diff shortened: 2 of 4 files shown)


rdig (approved):
All good here, tested with dev wallet and Metamask and it seems that now the states update accordingly.

Nicely done

Screenshot from 2024-09-03 18-22-58
Screenshot from 2024-09-03 18-24-20
Screenshot from 2024-09-03 18-25-13
Screenshot from 2024-09-03 18-26-10
Screenshot from 2024-09-03 18-26-26

jakubcolony (approved):
I rebased this branch on `feat/advanced-payments-ui` as we're looking to include it in the upcoming QA release.

The original issue would only reproduce for me when I staked the second motion after failing the first one.

This time, due to the buttons being disabled, I was able to stake both sides as expected:

image

Very nice work @iamsamgibbs 💪

PR #3058, opened 2024-09-03
[Notifications] Feat: Add empty notifications list to UserHub

### LINKED ISSUE #3006: Notifications: Empty notifications list
### Story

As a *user*, I want to *see when my notifications list is empty*, so that *I know that I have no notifications*

** TODO: Fill out below. Remember to handle notifications feature off state **

### Description

When the user opens the userhub, there should be a new tab called Notifications, which when clicked will open the Notifications list. When the user has no notifications, this list should show an empty state.

If the user has turned notifications off in their user settings (app-wide, not colony specific) then this tab should not show.

This ticket does not cover showing notifications, or showing the 'unread alert' with a number on the notifications tab button; it only covers the empty state (all screen sizes).

### Design

[Figma link](https://www.figma.com/design/0Fi3AaDlr3LAXBfsOXBGyk/User-Account?node-id=6754-9623&t=DbRYLaxphOdTpK4N-4)

Image

- [ ] The user can see an empty notifications list as per the designs.
- [ ] The notifications tab does not show when the user has turned off notifications app wide.

iamsamgibbs:
## Description

This PR adds the notifications tab to UserHub with the empty notifications list state.

[Figma link](https://www.figma.com/design/0Fi3AaDlr3LAXBfsOXBGyk/User-Account?node-id=6754-9623&t=DbRYLaxphOdTpK4N-4)

Screenshot 2024-09-03 at 17 30 27

Screenshot 2024-09-03 at 17 30 19

## Testing

Simply load up a colony and open the UserHub. The notifications tab should be visible and show the empty state. Test on all screen widths.

The tab should also not be visible when the notifications service is disabled by the user. This has not yet been implemented, but to test it change this line in `src/components/common/Extensions/UserHub/UserHub.tsx`

```
const notificationsServiceIsEnabled = true; // change to false
```

## Diffs

**New stuff** ✨

* Added notifications tab to userhub

**Changes** 🏗

* Re-ordered tabs to match figma

Resolves #3006

### DIFF
--- a/src/components/common/Extensions/UserHub/partials/NotificationsTab/NotificationsTab.tsx
+++ b/src/components/common/Extensions/UserHub/partials/NotificationsTab/NotificationsTab.tsx
@@ -1,0 +1,42 @@
+import { Binoculars } from '@phosphor-icons/react';
+import React from 'react';
+import { defineMessages } from 'react-intl';
+
+import { formatText } from '~utils/intl.ts';
+import EmptyContent from '~v5/common/EmptyContent/EmptyContent.tsx';
+
+const displayName = 'common.Extensions.UserHub.partials.NotificationsTab';
+
+const MSG = defineMessages({
+  notifications: {
+    id: `${displayName}.notifications`,
+    defaultMessage: 'Notifications',
+  },
+});
+
+const NotificationsTab = () => {
+  const isEmpty = true;
+
+  return (
+    <div className="h-full px-6 pb-6 pt-6 sm:pb-2">
+      <p className="heading-5">{formatText(MSG.notifications)}</p>
+      <div className="flex h-full flex-col justify-center pt-4 sm:h-auto sm:justify-normal">
+        {isEmpty && (
+          <>
+            <EmptyContent
+              title={{ id: 'empty.content.title.notifications' }}
+              description={{ id: 'empty.content.subtitle.notifications' }}
+              icon={Binoculars}
+              className="sm:pt-[100px]"
+            />
+            <div className="h-[25%] sm:h-0" />
+          </>
+        )}
+      </div>
+    </div>
+  );
+};
+
+NotificationsTab.displayName = displayName;
+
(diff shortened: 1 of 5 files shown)


src/components/common/Extensions/UserHub/partials/NotificationsTab/NotificationsTab.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const isEmpty = true;
+
+  return (
+    <div className="h-full px-6 pb-6 pt-6 sm:pb-2">
+      <p className="heading-5">{formatText(MSG.notifications)}</p>
+      <div className="flex h-full flex-col justify-center pt-4 sm:h-auto sm:justify-normal">
+        {isEmpty && (
+          <>
+            <EmptyContent
+              title={{ id: 'empty.content.title.notifications' }}
```
bassgeta, line 27: maybe this could be locally defined and we passed it in already i18nd via `formatText(MSG.noNotifications)` or something like that

src/components/common/Extensions/UserHub/partials/NotificationsTab/NotificationsTab.tsx
```diff
@@ -0,0 +1,42 @@
...
+              description={{ id: 'empty.content.subtitle.notifications' }}
```
bassgeta, line 28: this could also be locally defined and sent over already rendered

bassgeta (approved):
Looking good, dropped 2 comments if we want to define messages locally instead (`EmptyContent` supports both ids or rendered text)
Other than that, really nice, it's gonna be easy to plug in actual components here!

bassgeta (approved):
Reapproving just in case, thanks for the changes!

Nortsova (approved):
Hey @iamsamgibbs thanks for the PR ✨ 

Empty notifications looks good:
image

No notification tab if notificationsServiceIsEnabled - false:
image

PR #3063, opened 2024-09-04
[Notifications] Feat: Show unread notifications alert

### LINKED ISSUE #3008: Notifications: Show unread notifications alert
### Story

As a *user receiving notifications*, I want to *see a visual indicator that I have unread notifications*, so that *I know I should check my notifications list*

### Description

The user should not have to open their userhub notifications list to know that they have unread notifications - they should instead see a simple visual indicator on the userhub button itself, as well as an indicator with a number on the userhub Notifications tab button to see when and how many unread notifications they have.

### Design

[Figma link](https://www.figma.com/design/0Fi3AaDlr3LAXBfsOXBGyk/User-Account?node-id=6401-22917&t=DbRYLaxphOdTpK4N-4)

Image

- [ ] The indicator is shown on the userhub button
- [ ] The indicator is shown on the notifications tab

### Suggestions (optional)

* The `useNotifications` hook likely provides a number of unread notifications. Alternatively the `MagicbellProvider` context might have it somewhere. Check the Magicbell docs for the best way to do this, ideally without fetching all of the notifications and checking for how many of them are unread.

iamsamgibbs:
## Description

Note: It probably makes sense to test #3066 straight after this one as it is such a tiny additional change.

This PR adds unread notifications alerts to the UserHub button and UserHub notifications tab. It also adds the MagicBell provider to get easy access to the unread notifications count.

## Testing

Add `MAGICBELL_API_KEY="<API_KEY>"` to your `.env.local.secrets` file.

Add `MAGICBELL_API_KEY="<API_KEY>"` and `MAGICBELL_API_SECRET="<API_SECRET>"` to `amplify/backend/function/createUserNotificationsData/.env` if you don't already have it.

Log with a dev wallet where the user has unread notifications (at the time of writing this is true for Leela - dev wallet 1)
Check the unread notification alert appears on the UserHub when closed, then on the notifications tab. (If there are no notifications, you can create a notification via the UI [https://app.magicbell.com/projects/8931/compose](https://app.magicbell.com/projects/8931/compose) - set the user's wallet address as the recipient).

Screenshot 2024-09-09 at 11 22 28

Screenshot 2024-09-09 at 11 22 37

## Diffs

**New stuff** ✨

* MagicBell provider added

**Changes** 🏗

* UserHub now shows unread notifications alert

Resolves #3008

### DIFF
--- a/src/components/common/Extensions/UserHub/UserHub.tsx
+++ b/src/components/common/Extensions/UserHub/UserHub.tsx
@@ -9,6 +9,7 @@ import Select from '~v5/common/Fields/Select/index.ts';
 import TitleLabel from '~v5/shared/TitleLabel/index.ts';
 
 import { tabList } from './consts.ts';
+import CountBadge from './partials/CountBadge.tsx';
 import CryptoToFiatTab from './partials/CryptoToFiatTab/CryptoToFiatTab.tsx';
 import NotificationsTab from './partials/NotificationsTab/NotificationsTab.tsx';
 import ReputationTab from './partials/ReputationTab/index.ts';
@@ -45,6 +46,8 @@ const UserHub: FC<Props> = ({ initialOpenTab = UserHubTab.Balance }) => {
 
   // @TODO: get from notifications context
   const notificationsServiceIsEnabled = true;
+  // @TODO: Get from context
+  const unreadCount = 1;
 
   const filteredTabList = tabList.filter((tabItem) => {
     const isFeatureFlagEnabled =
@@ -120,7 +123,10 @@ const UserHub: FC<Props> = ({ initialOpenTab = UserHubTab.Balance }) => {
                           'font-medium': selectedTab === id,
                         })}
                       >
-                        <span className="mr-2 flex shrink-0">
+                        <span className="relative mr-2 flex shrink-0">
+                          {id === UserHubTab.Notifications && (
+                            <CountBadge count={unreadCount} />
+                          )}
                           <Icon size={14} />
                         </span>
                         {formatText(label)}
--- a/src/components/common/Extensions/UserHub/UserHub.tsx
+++ b/src/components/common/Extensions/UserHub/UserHub.tsx
@@ -1,3 +1,4 @@
+import { useBell } from '@magicbell/react-headless';
 import clsx from 'clsx';
 import React, { type FC, useState, useContext } from 'react';
 import { defineMessages } from 'react-intl';
@@ -46,8 +47,8 @@ const UserHub: FC<Props> = ({ initialOpenTab = UserHubTab.Balance }) => {
 
   // @TODO: get from notifications context
   const notificationsServiceIsEnabled = true;
-  // @TODO: Get from context
-  const unreadCount = 1;
+
+  const { unreadCount } = useBell() || {};
 
   const filteredTabList = tabList.filter((tabItem) => {
     const isFeatureFlagEnabled =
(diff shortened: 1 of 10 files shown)


src/components/v5/shared/Button/Button.tsx
```diff
@@ -112,6 +113,9 @@ const Button = forwardRef<HTMLButtonElement, PropsWithChildren<ButtonProps>>(
             >
               {children}
             </ButtonContent>
+            {showNotificationDot && (
```
bassgeta, line 116: what if we add a new prop called `rightContent` or something like that and we just compose this dot in from where we need to? And we can also pull it into a reusable component called `NotificationDot`

src/context/NotificationsContext/NotificationsContextProvider.tsx
```diff
@@ -30,10 +31,27 @@ const NotificationsContextProvider = ({
...
+
+  if (!user?.notificationsData?.magicbellUserId) {
+    return (
+      <NotificationsContext.Provider value={value}>
+        {children}
+      </NotificationsContext.Provider>
+    );
+  }
+
   return (
```
bassgeta, line 46: Nice and straight forward branching, love to see it :100:

bassgeta (approved):
Left a comment we can address regarding composition, since the `Button` component is already pretty bloated. But not a dealbreaker for me in any way!
Other than that, this is legit, thanks for the context return branching, it just makes sense implementation wise to have constantly render but wrap around in different context providers depending on the state, slick!
The unread counter is there:
image
Mobile:
image
Also doesn't break if we have a 3 digit notification count:
image

src/components/v5/shared/Button/Button.tsx
iamsamgibbs: Realised I can just include it in the {children} so I have removed this!

iamsamgibbs:
The comment on triple digits reminded me there is a note on the spec to cap this at 99 so I have also implemented that.

src/components/common/Extensions/UserHub/UserHub.tsx
```diff
@@ -120,7 +124,10 @@ const UserHub: FC<Props> = ({ initialOpenTab = UserHubTab.Balance }) => {
                           'font-medium': selectedTab === id,
                         })}
                       >
-                        <span className="mr-2 flex shrink-0">
+                        <span className="relative mr-2 flex shrink-0">
+                          {id === UserHubTab.Notifications && (
+                            <CountBadge count={unreadCount} maximum={99} />
+                          )}
```
Nortsova, line 130: This might just be a matter of preference, but I prefer to keep the UI and logic separate. In my opinion, it would be better to move this part of the code:
```
if (!count || count < 1) {
    return null;
  }
```
into this component, so that CountBadge always renders something.

it would look like this:
```suggestion
                          {id === UserHubTab.Notifications && (!unreadCount || unreadCount < 1) && (
                            <CountBadge count={unreadCount} maximum={99} />
                          )}
```

Nortsova (approved):
Hey @iamsamgibbs, thank you for the great PR!

I sent a notification via https://app.magicbell.com/projects/8931/compose , and now there's one more here 👏 :
image

Mobile view:
image

One thing to consider: would it be more understandable for the user if we added a notification pill to the dropdown as well?
image

I left a small comment, but please bear in mind that it's just a suggestion. :)

src/components/common/Extensions/UserHub/UserHub.tsx
iamsamgibbs: I think I'd prefer to keep it inside the component as I think `!unreadCount || unreadCount < 1` (or an equivalent) would end up being used to wrap every use of this component.

This pattern also exists in a bunch of other components which return null at the top of the component based on a conditional check.

PR #3065, opened 2024-09-05
[Dashboard] Feat: Update recent activity table

### LINKED ISSUE #2850: [Dashboard] Update Recent activity table
### Story

-

### Description

Given the dashboard redesign, also the Recent activity table should be updated according to designs.
The empty state containing only the `New colony creation` action will be covered in [#3061](https://github.com/JoinColony/colonyCDapp/issues/3061)

- [ ] The Recent activity table is updated according to designs

[Figma link](https://www.figma.com/design/gac6xVgGNCAge5Bia933dp/Dashboard?node-id=4859-36367&t=3YmH0OCdhMXflaYC-0)
[Figma link empty state](https://www.figma.com/design/gac6xVgGNCAge5Bia933dp/Dashboard?node-id=4867-40639&t=3YmH0OCdhMXflaYC-0)

iamsamgibbs:
## Description

This PR makes changes to the activity table on the dashboard to reflect the new figma designs.

[Figma link](https://www.figma.com/design/gac6xVgGNCAge5Bia933dp/Dashboard?node-id=4859-36367&t=3YmH0OCdhMXflaYC-0)

Screenshot 2024-09-10 at 10 22 12

Screenshot 2024-09-10 at 10 22 23

It was very tempting to separate out the dashboard activity feed from the main activity feed (and I did actually start to do this before realising how much work it would be), but I instead opted to add a few new props to the `useActionsTableProps` hook and the `Table` component. This probably isn't ideal but both the hook and the table component are in need of a refactor at some point and this didn't feel like the right place to do it.

The newly added props are used to remove the horizontal padding, remove the table header and border, hide the user avatar.

Note: The empty state for the table is not included in this PR and is covered by a separate issue.

## Testing

Spin up the environment and run the create-data script.

Check the activity feed on the dashboard matches the figma design on all screen widths.

Check the main activity feed on the activity page is unaltered.

## Diffs

**Changes** 🏗

* Dashboard activity feed changed to match new figma designs
* New props added to `useActionsTableProps` hook and the `Table` component
* KebabMenu and MeatBallMenu now use `text-gray-600`

Resolves #2850

### DIFF
--- a/src/components/common/ColonyActionsTable/RecentActivityTable.tsx
+++ b/src/components/common/ColonyActionsTable/RecentActivityTable.tsx
@@ -1,0 +1,46 @@
+import React, { type FC } from 'react';
+
+import useActionsCount from '~hooks/useActionsCount.ts';
+import useGetSelectedDomainFilter from '~hooks/useGetSelectedDomainFilter.tsx';
+import { type ColonyAction } from '~types/graphql.ts';
+import Table from '~v5/common/Table/index.ts';
+
+import { useActionsTableProps } from './hooks/useActionsTableProps.tsx';
+import { type ColonyActionsTableProps } from './types.ts';
+
+const displayName = 'common.RecentActivityTable';
+
+const RecentActivityTable: FC<ColonyActionsTableProps> = ({ ...props }) => {
+  const selectedDomain = useGetSelectedDomainFilter();
+  const nativeDomainId = selectedDomain?.nativeId;
+
+  const { actionsCount: totalActions, loading: loadingActionsCount } =
+    useActionsCount({
+      domainId: nativeDomainId,
+    });
+
+  const pageCount = Math.ceil(totalActions / (props.pageSize ?? 10));
+
+  const { tableProps, renderSubComponent } = useActionsTableProps({
+    ...props,
+    showUserAvatar: false,
+    hasHorizontalPadding: false,
+    showTotalPagesNumber: !loadingActionsCount,
+    isRecentActivityVariant: true,
+  });
+
+  return (
+    <Table<ColonyAction>
+      {...tableProps}
+      showTableHead={false}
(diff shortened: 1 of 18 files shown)


iamsamgibbs:
Thanks for the review @mmioana - good catch with this!

This is happening on master, although we don't see it as often as the re-render on resize issue hides it:

https://github.com/user-attachments/assets/64406adb-0647-4f59-92b9-eee51910258e

I've had a quick look at this and think it might be a bit of a pain to change, so I'd propose opening a new bug issue to resolve this if you agree?

mmioana:
> Thanks for the review @mmioana - good catch with this!
> 
> This is happening on master, although we don't see it as often as the re-render on resize issue hides it:
> 
>  Screen.Recording.2024-09-10.at.16.30.50.mov 
> I've had a quick look at this and think it might be a bit of a pain to change, so I'd propose opening a new bug issue to resolve this if you agree?

Yes, totally agree, let's open a new issue for it 💯

mmioana (approved):

rdig (approved):
Tested all breakpoints and all seem to be rendering fine, even the obnoxiously small ones _(which shouldn't really count)_

Nicely done Sam!

Screenshot from 2024-09-12 12-42-52
Screenshot from 2024-09-12 12-43-33
Screenshot from 2024-09-12 12-43-50
Screenshot from 2024-09-12 12-44-11

bassgeta (approved):
Yeah this component most definitely needs a refactor, but good job adding additional things onto it, it doesn't really look that bad, totally sensible!
But the `Table` refactor was a hot topic at the beginning of the year and it was too much of a pain to fix properly :D 
UI wise, everything works perfect:
image
image
And the latest activity table is intact:
image
Also looks nice on mobile :ok_hand: 
image

PR #3073, opened 2024-09-11
Stop bridge mock container running as part of the dev env

jakubcolony:
## Description

For the most part of the development, the bridge mock API is not required and it only increases the build time and the resource usage of the dev environment. 

This PR removes the `mocking` container from running as part of `npm run dev`. Instead, it can be started by running `npm run dev:all`. The latter has been traditionally used to run the dev env with the `safe` container, however, I found that to have some issues and having discussed with Raul, I commented the safe bits out.

## Testing

Should be fairly straightforward:
- `npm run dev` should not start the `mocking` container 
- `npm run dev:all` should start all the dev containers, including `mocking`.

You can verify both by looking at the console or Docker Desktop containers list.


iamsamgibbs (changes requested):
Nice quality of life change here @jakubcolony !

I had no issue when testing with `npm run dev` - no mocking logs in the console, not running in docker, was still able to interact with the app, performed a test action without issue.

Screenshot 2024-09-11 at 18 51 59

Screenshot 2024-09-11 at 18 52 05

But with `npm run dev:all` I'm getting this error (even after removing the docker containers):

Screenshot 2024-09-11 at 18 56 48

rdig (changes requested):
The normal "dev" scripts works flawlessly, I even tried the C2F page to make sure it loads without issue, just in case

Screenshot from 2024-09-12 11-35-49
Screenshot from 2024-09-12 11-36-59

I ran into an error however, different from Sam's though, which is that apparently the block ingestor can't access the network assets resources, hence it can't get the network address

Screenshot from 2024-09-12 11-42-29

Don't really know for sure if this is a problem with your change or not, but might as well look into it.

jakubcolony:
@iamsamgibbs I tried a few times and haven't run into that issue. Could you check which Docker version you're using? 

Apparently this was reported as a bug and fixed earlier this year: https://github.com/docker/compose/issues/11544

rdig (approved):
Re-tested this and everything working fine. Glad to have this in.

Screenshot from 2024-10-10 17-46-45
Screenshot from 2024-10-10 17-50-38
Screenshot from 2024-10-10 17-51-15
Screenshot from 2024-10-10 17-51-37

iamsamgibbs:
> @iamsamgibbs I tried a few times and haven't run into that issue. Could you check which Docker version you're using?
> 
> Apparently this was reported as a bug and fixed earlier this year: [docker/compose#11544](https://github.com/docker/compose/issues/11544)

I think you're right that this is a docker issue, I've stayed on an old version for ages as one of the newer versions had known issues a while back. Will update docker tomorrow and give this a test!

iamsamgibbs (approved):
All good after updating docker!

With `npm run dev`:

Screenshot 2024-10-11 at 11 20 59
Screenshot 2024-10-11 at 11 22 33
Screenshot 2024-10-11 at 11 22 50

With `npm run dev:all`:

Screenshot 2024-10-11 at 11 27 04

Great change, lets get it merged!

mmioana (approved):
Nice fix @jakubcolony 

With `npm run dev` the `mocking` container is not running
Screenshot 2024-10-16 at 11 49 28
Screenshot 2024-10-16 at 11 49 43

With `npm run dev:all` the `mocking` container is running
Screenshot 2024-10-16 at 11 53 16
Screenshot 2024-10-16 at 11 52 56

rdig:
All good here in the end. I'm merging it!

Issue #2238, opened 2024-04-15
Payment builder - Edit action - After Locked state - Action Creator path with Permissions


AdrianDudko:
Teamwork task: [Payment builder - Edit action - After Locked state - Action Creator path with Permissions](https://tw.auroracreation.com/#/tasks/15873666) 

## Related sub-issues/PR's

- [ ]

## Acceptance criteria

**General requirements**
- User can Edit an action After Locked State, which means that action can be edited at 
	- Funding step
	- Release step
	- Payment step
- Changes can be made an unlimited number of times. 

**Kebab menu**
- Kebab menu with Edit/Cancel/Copy:
	- Edit button ,In the locked state,
		- Visible to the payment creator.     
		- Visible to members who hold the Payer level permissions or above or `Arbitration` permission on the contracts.    
		- Visible to all members if the Lazy consensus extension is enabled.     
		- It is not visible otherwise.
		- Clicking starts the edit mode.

**Edit mode**
- Dispaly Yellow border & Edit Pill iin top left corner
	- Pill: tooltip on hover
- Kebab menu should be hidden
- Buttons in Stepper should be disabled
	- for example "Fund' button in Funding Step
- Updates the read-only fields to be editable for the following fields.
	- Recipients
	- Amounts and tokens
	- Claim delays
	- All other fields should remain read only.
- Clicking \"Change payment\" opens the Change payment modal
	- button should disabled until the user makes changes
- Clicking "Cancel edit mode" opens the Exit Edit Mode modal

**Exit Edit Mode modal**
- Clicking "Yes, confirm exit" turns off the the Edit Mode and ignore any changes
- Clicking "Return to edit mode" gets user back to edit mode

**Change payment modal**
- similar to the Cancel modal used for Cancelling Before Locked State
- user can use Dropdown to choose to execute Changes
	- Reputation: if Reputation extension is installed
		-  **this Decision Method is not supported yet an is not in scope of this task**
	- Permissions:  members who hold the Payer level permissions or above or `Arbitration` permission on the contracts.    

**Change step in Stepper**
-  if the user Edits the action in the Funding or Release step, the Changes step appears in Stepper before the step in which the edit was made. 
	-  Changes shows the Widget with information about who made Changes to the action, what Permissions were used and number of changes counter. 
- if the user Edits the actions in Payment step, the Changes step appears in Stepper before the step in which the edit was made. 
[...]

PR #3075, opened 2024-09-11
feat: add edit to payment builder action

CzarekDryl:
## Description

This PR adds the ability to edit actions after the locked state during the Funding, Release steps. Editing in Payment step is not possible for now.

- Edit Mode:
  - Displays yellow border and Edit Pill.
  - Disables stepper buttons and hides Kebab menu.
  - Makes Recipients, Amounts, and Claim Delays editable.
  - Change payment button opens modal, disabled until changes are made.

- Change Payment Modal:
  - Dropdown for execution options (Reputation and Permissions).

- Changes Step in Stepper:
  - Changes pill appears after edits in Funding, Release step.
  - Shows details on edits, permissions, and a change counter.

- If there is funding needed after editing, new pill should show after changes one

## Testing

* Install Lazy Consensus extension to test funding with motion
* Create advanced payment action and lock it
* Play around with funding and editing

[Design](https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-v5?type=design&node-id=4102-376576&mode=design&t=lVqi0Mno6dqgrCs8-4)

Resolves [https://github.com/JoinColony/colonyCDapp/issues/2238](https://github.com/JoinColony/colonyCDapp/issues/2238)


arrenv (changes requested):
@CzarekDryl Nice one, a lot of work here, thank you for tackling it. I spent some time testing and came across a few things:

1. Change payment modal is showing two rows of buttons as well as a red button. There should only be the Confirm changes button and it should not be red.
image

2. Clicking on the "Confirm changes" button did not do anything other than close the modal.

3. Making a change in the Review stage showed the changes step after it, which I was not able to click on. While in the Review stage, the changes pill should appear before review and be clickable so that you can see who made changes, when and how (i.e. as the Payment creator), and be able to handle multiple.
image

4. There are some formatting issues with the edit mode, the recipients change to being unverified and the inputs change width, also, is there anyway to make the yellow border coming in not impact the content width?

    #### Read only
    image

    #### Edit mode
    image

5. Confirm details button should be disabled while in edit mode, and while the edit transaction is happening in order to prevent a clash.
image

6. It seems the changes count and change history is not working fully it seems to only count the amount changes, in the following screenshot I had edited the following, a total of 5 changes but it only shows 2 changes, changes included:
    - 1st Row: Amount was changed from 2 to 1, claim delay was changed from 1 to 0.
    - 2nd Row: Recipient was changed from amy to leela, amount was changed from 3 to 1, claim delay was changed from 2 to 0.
   
    image

7. Trying to fund the payment crashed the app with the following error and I could not continue testing any further.
image

CzarekDryl:
@arrenv 

1. Fixed
2. Fixed
3. and 5. in this issue I'm taking care of editing after the locked state
4. Which browser are you testing this on? On my end, the gaps aren't that large — it works like when creating actions
image

6. It's working fine for me
image
7. Fixed
image

PR #3076, opened 2024-09-11
Feat: Load advanced payments rows incrementally

### LINKED ISSUE #2903: Advanced payments with a large number of payments are slow to load
## Steps to reproduce

Step 1 -  Create an advanced payments action
Step 2 - Upload this csv with 400 rows [https://drive.google.com/file/d/1iMFNpRtMwyc3teh2mVPeBlRPEyZBRwFQ/view?usp=sharing](https://drive.google.com/file/d/1iMFNpRtMwyc3teh2mVPeBlRPEyZBRwFQ/view?usp=sharing)
Step 3 - Click "Create payment"

## Expected behaviour

* The completed action screen shows all 400 payments in both the table and the action subtitle.

## Actual behaviour

* The completed action screen only shows the first 164 payments in the table, then takes a further 2 minutes before all 400 payments finally show.
* The action subtitle starts at 164 payments and slowly increases as more payments are processed.

https://github.com/user-attachments/assets/003534bf-6175-4033-a77d-9c8eeab84a2d

## Suggestions (optional)

This seems to be due to how block-ingestor processes the payments. Investigate whether anything we can done here to speed things up - I'm unsure how possible this is.

Alternatively, at the end of the create expenditure saga, we could save the expected number of payments to an expenditureMetadata object, then query that on the completed action screen to at least show the correct number in the action subtitle and show the correct number of rows in the table with skeleton loaders until block-ingestor has finished processing.

iamsamgibbs:
## Description

This PR addresses an issue where advanced payments with a large number of payments (over 164 rows) are slow to load on the completed action screen.

This issue occurs because payments with over 164 rows are split into multicall transactions. This means when they are picked up by the block-ingestor, it processes them in batches of 164. This resulted in the first 164 rows loading on the completed action screen, then waiting for the next batch to be processed before updating to 328, etc. This gave the false impression only the first 164 rows had gone through unless you waited on the screen log enough for the remaining payments to be processed.

This PR improves this user experience by setting an `expectedNumberOfPayouts` on the `expenditureMetadata` model which is then used on the completed action screen to show the correct number of payouts in the subtitle and table footer, and to show the correct number of rows in the table.

Payments are still processed in batches, so after the first 164 rows are loaded, the next 164 will still appear to be loading in the table. I looked into how these are processed in the block-ingestor and I don't think there are any sensible changes we can make to process these faster.

(See the original issue #2903 for a video of the original issue)

---

This PR also adds `expectedNumberOfTokens` in a similar way to show the correct number of tokens in the subtitle.

It also addresses a bug with the activity feed where the description would show the number of unique recipients rather than the number of payments.

## Testing

Step 1 -  Create an advanced payments action
Step 2 - Upload this csv with 400 rows [https://drive.google.com/file/d/1iMFNpRtMwyc3teh2mVPeBlRPEyZBRwFQ/view?usp=sharing](https://drive.google.com/file/d/1iMFNpRtMwyc3teh2mVPeBlRPEyZBRwFQ/view?usp=sharing)
Step 3 - Click "Create payment"

The completed action screen should show the correct number of payments in the subtitle, and the table footer, and the table should have the correct number of rows.

The action in the activity feed should show the correct number of payments, even whilst they are still being processed by the block-ingestor / loading into the table.

Further testing:
* Test that the completed action still shows correctly for payments with fewer than 164 rows. (Note: For payments with fewer than 8 rows, the total number of payments is not shown in the table footer.)
* Test the correct number of tokens shows when creating payments with more than 1 token.

## Diffs

**New stuff** ✨

* `expectedNumberOfPayouts` and `expectedNumberOfTokens` added to `expenditureMetadata` model

**Changes** 🏗

* Advanced payments completed action screen now uses `expectedNumberOfPayouts` and `expectedNumberOfTokens` to show correct values whilst block-ingestor processes payments.
* Activity feed now shows number of payments rather than unique recipients

Resolves #2903 

Example of how it should look:

https://github.com/user-attachments/assets/b51f3121-d7d3-4110-be7b-eb451f8d9c74


amplify/backend/api/colonycdapp/schema/schema.graphql
```diff
@@ -3751,6 +3751,8 @@ type ExpenditureMetadata @model {
   id: ID! # Self-managed, formatted as colonyId_nativeId
   fundFromDomainNativeId: Int!
   stages: [ExpenditureStage!]
+  expectedNumberOfPayouts: Int
+  expectedNumberOfTokens: Int
```
iamsamgibbs, line 3755: Not marked as required so that old expenditures are unaffected. Uses of expectedNumberOfPayouts will fallback to expenditure.slots.length. Uses of expectedNumberOfTokens will fallback to a set of the payouts array mapped to include only the token address.

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/PaymentBuilderTable/PaymentBuilderTable.tsx
```diff
@@ -178,38 +180,71 @@ const PaymentBuilderTable: FC<PaymentBuilderTableProps> = ({
...
+        amount: '0',
+        tokenAddress: '',
+        isClaimed: false,
+        id: items.length + index + 1, // Calculate the next available item.id
+        isLoading: true,
+      }),
+    );
+
+    return [...populatedItems, ...placeholderItems];
+  }, [expendituresGlobalClaimDelay, items]);
```
iamsamgibbs, line 240: Add items.payouts here?

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/PaymentBuilderTable/PaymentBuilderTable.tsx
```diff
@@ -86,7 +89,7 @@ const useGetPaymentBuilderColumns = ({
         footer: hasMoreThanOneToken
```
iamsamgibbs, line 89: Check hasMoreThanOneToken, might need updating with expectedNumberOfTokens

src/components/common/ColonyActions/helpers/mapItemToMessageFormat.tsx
```diff
@@ -267,14 +268,9 @@ export const mapColonyActionToExpectedFormat = ({
         notMaybe(actionData.metadata?.customTitle),
       ),
     [ActionTitleMessageKeys.Members]: actionData.members?.length || 0,
-    [ActionTitleMessageKeys.RecipientsNumber]: new Set(
-      expenditureData?.slots.map((slot) => slot.recipientAddress),
-    ).size,
```
iamsamgibbs, line 272: RecipientsNumber should show the number of payments, not unique recipients.

rumzledz (commented):
This is pretty awesome @iamsamgibbs ! 🔥 Using the CSV file you've supplied, I can confirm that the total number of rows are showing correctly on the Completed Action component and on the Latest Activity table even though the block ingestor is still processing the data:



Then eventually, it shows the payment total:

Screenshot 2024-09-13 at 00 55 33

There's a bit in the test steps that says: "Note: For payments with fewer than 7 rows, the total number of payments is not shown in the table footer."

Maybe I'm not understanding it correctly but I've added exactly 7 rows but the total is still not showing up:

7 rows

I also left a wee comment about the loader component and that's pretty much it!

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/PaymentBuilderTable/PaymentBuilderTable.tsx
```diff
@@ -116,7 +119,11 @@ const useGetPaymentBuilderColumns = ({
         cell: ({ row }) => {
           const formattedHours = convertPeriodToHours(row.original.claimDelay);
 
-          return !isDataLoading ? (
+          return row.original.isLoading ? (
+            <div className="flex w-[4rem] items-center">
+              <div className="h-4 w-full overflow-hidden rounded skeleton" />
+            </div>
```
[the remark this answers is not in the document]
iamsamgibbs: Good suggestion! I've updated this table to use the `LoadingSkeleton` component.

iamsamgibbs:
Yes my bad, should be fewer than 8 - I've updated the description!

mmioana (approved):
Such a pleasure looking over you changes @iamsamgibbs 🥇 

Importing your csv
https://github.com/user-attachments/assets/a3e827d0-8cdc-4c9b-8547-0cde9bf9bdb9

Slowly slowly the payments appeared

https://github.com/user-attachments/assets/d4721070-ef59-4759-b78e-2f265e6cee30

The action in the activity list showed the correct number of payments

Screenshot 2024-09-13 at 13 04 55

Having 8 payments added and then removing a row

https://github.com/user-attachments/assets/6663eb43-40e2-40c7-b291-56e93cc3aa9d

After unlocking the token from `wayne`, created an advanced payment with 2 different tokens
Screenshot 2024-09-13 at 13 14 23
Screenshot 2024-09-13 at 13 14 28

The only issue I have noticed was when uploading the large csv with 400 entries, as I forgot to fill the `Decision method` and `Fund from` fields, after correcting the form, I noticed the error input was still shown
Screenshot 2024-09-13 at 13 03 02
However, maybe it makes sense to be treated in a separate issue 👍

iamsamgibbs:
Thank you for the review @mmioana ! :) I recall the issue you mention being an issue whilst this form was first being developed. It is related to how validation works on this particular action. Due to the potential number of rows, error validation was running very slowly when the was running per change. The validation for this action is instead only updating on submit. I agree this isn't ideal as it results in misleading UI, perhaps a good middle ground would be to have validation for all fields except the table be performed on change, and the table changes be performed on submit, but I don't know how possible this is and would definitely be a lot of work warranting it's own PR!

bassgeta (approved):
Yup pretty legit changes, and nicely unified the loading states to use `LoadingSkeleton`, dope to see that!
Uploading a CSV works and generates a ton of rows
image
The subtitle is correct and 400 lines of loading are displayed when creating an action :exploding_head: 
image
image

And finally, it shows up after some time:
image

rumzledz (approved):
Same results are last time @iamsamgibbs and can confirm 8 is indeed the minimum to show the total number of rows. Approved! ✅

PR #3080, opened 2024-09-12
[Dashboard] Weekly/monthly funds trend

### LINKED ISSUE #2858: [Dashboard] Weekly/monthly funds trend
### Story

- As a USER, I want to see at a glance what the funds trend is.

### Description

Given the focus on financial actions and tools, we want to display the total funds, payments and incoming funds trend on a weekly/monthly basis.
For numbers larger than 5 digits using 99,999%+.

## Implementation

Relies on the queries for the monthly chart - `fetchDomainBalance` - please query for values having `TimeframeType.WEEKLY` and `TimeframeType.MONTHLY` as timeframeType and `1` as timeframePeriod
We need to implement a cron lambda that computes the payments and income on a weekly/monthly basis and we store them in a cache table. 
Following the steps described here https://www.notion.so/colony/Getting-Started-with-CDapp-Lambda-Functions-82295a2197314973a57bd7c3c0a0759e
We need to configure the cron expressions in order to trigger it either every Monday or every start of month at 1AM - this can be later decided - so we need to add in the `backend-config.json` where the new lambda gets added 
```
{
  "trigger": {
    "cloudwatchRule": [
      {
        "ruleName": "weeklyMondayRule",
        "scheduleExpression": "cron(0 1 ? * 2 *)"
      },
      {
        "ruleName": "monthlyFirstDayRule",
        "scheduleExpression": "cron(0 1 1 * ? *)"
      }
    ]
  }
}
```

If there is no item present for a colony and domain for the previous week/month, we need to compute the values for both the previous week/month.
If there is an item present already, then we just need to assign the current value to the previous one and compute a new current value.

**Maybe it doesn't even matter to store the current value as this should be computed real time based on the income and payments values. However the previous week/month needs to be cached**

- [ ] Dashboard page Total funds card displays the weekly trend
- [ ] Dashboard page Total in and out card displays the monthly trend

[Figma link](https://www.figma.com/design/gac6xVgGNCAge5Bia933dp/Dashboard?node-id=4878-89693&t=SNW5sWiyJk5omQMX-0)

mmioana:
Feat: Add cron lambda to cache all needed previous domain balances 

## Description

- This PR adds a cron lambda to cache all previous domain-level and colony-level balances for both the previous last 30 days and the previous total amount from the start of last week. Given the time at which the lambda is triggered, we have no idea what the selected currency is and we want to provide a uniform solution, the cached value will be in `USDC`

- These values will be used for the increase/decrease trends in these cards, however this PR fetches the value only for the `Total in and out` card

Property 1=Completed
Streaming widget

## Testing

TODO: Given you can't really test the cron lambda runs properly locally, as there is no event to trigger it, we need to make use of a manual query

* Step 1. First things first, run your dev environment
* Step 2. Run the `create-data` script
* Step 3. For ease of testing (kind of) please save the following code in a new file (you can name it `.diff`)
```
diff --git a/amplify/backend/api/colonycdapp/schema/schema.graphql b/amplify/backend/api/colonycdapp/schema/schema.graphql
index fa83b314f..4b99b0358 100644
--- a/amplify/backend/api/colonycdapp/schema/schema.graphql
+++ b/amplify/backend/api/colonycdapp/schema/schema.graphql
@@ -3373,6 +3373,7 @@ type ColonyAction @model @searchable {
   Approved tokens impacted by the action (used for manage tokens)
   """
   approvedTokenChanges: ApprovedTokenChanges
+  updatedAt: AWSDateTime!
 }
 
 """
diff --git a/src/graphql/generated.ts b/src/graphql/generated.ts
index 8d2d7e683..ef56997c2 100644
--- a/src/graphql/generated.ts
+++ b/src/graphql/generated.ts
@@ -1456,6 +1456,7 @@ export type CreateColonyActionInput = {
   toPotId?: InputMaybe<Scalars['Int']>;
   tokenAddress?: InputMaybe<Scalars['ID']>;
   type: ColonyActionType;
+  updatedAt?: InputMaybe<Scalars['AWSDateTime']>;
 };
 
 export type CreateColonyActionMetadataInput = {
@@ -2856,6 +2857,7 @@ export type ModelColonyActionConditionInput = {
   toPotId?: InputMaybe<ModelIntInput>;
   tokenAddress?: InputMaybe<ModelIdInput>;
   type?: InputMaybe<ModelColonyActionTypeInput>;
+  updatedAt?: InputMaybe<ModelStringInput>;
 };
 
 export type ModelColonyActionConnection = {
@@ -2901,6 +2903,7 @@ export type ModelColonyActionFilterInput = {
   toPotId?: InputMaybe<ModelIntInput>;
   tokenAddress?: InputMaybe<ModelIdInput>;
   type?: InputMaybe<ModelColonyActionTypeInput>;
+  updatedAt?: InputMaybe<ModelStringInput>;
 };
 
 export type ModelColonyActionMetadataConditionInput = {
@@ -4167,6 +4170,7 @@ export type ModelSubscriptionColonyActionFilterInput = {
   toPotId?: InputMaybe<ModelSubscriptionIntInput>;
   tokenAddress?: InputMaybe<ModelSubscriptionIdInput>;
   type?: InputMaybe<ModelSubscriptionStringInput>;
+  updatedAt?: InputMaybe<ModelSubscriptionStringInput>;
 };
 
 export type ModelSubscriptionColonyActionMetadataFilterInput = {
@@ -9358,6 +9362,7 @@ export type UpdateColonyActionInput = {
   toPotId?: InputMaybe<Scalars['Int']>;
   tokenAddress?: InputMaybe<Scalars['ID']>;
   type?: InputMaybe<ColonyActionType>;
+  updatedAt?: InputMaybe<Scalars['AWSDateTime']>;
 };
 
 export type UpdateColonyActionMetadataInput = {

```

The reason for this patch is to be able to manipulate the  `updatedAt` field for a colony action

* Step 4. Run `git apply .diff`
* Step 5. Now the interesting part - let's create some historical testing data 👨‍🦳 
Go to http://localhost:20002 and run the following mutation
```
mutation MyMutation {
  createColonyAction(
    input: {
      amount: "230000000000000000000" # any amount would be fine, though for this action to have a visible result use a value higher than 10^18
      showInActionsList: true
      initiatorAddress: "<LEELA_WALLET_ADDRESS_HERE>"
      blockNumber: 977
      colonyId: "<COLONY_ADDRESS_HERE>"
      fromDomainId: "<DOMAIN_ID_HERE>" # or just use <COLONY_ADDRESS_HERE_1>
      rootHash: "0x0000000000000000000000000000000000000000000000000000000000000000"
      toDomainId: "<ANOTHER_DOMAIN_ID_HERE>" # or just use <COLONY_ADDRESS_HERE_2>
      tokenAddress: "<TOKEN_ADDRESS_HERE>"
      type: MOVE_FUNDS
      updatedAt: "2024-08-16T14:55:07.825Z" # update this to any date previous to 30 days ago
    }
  ) {
    amount
    blockNumber
    colonyId
    createdAt
    fromDomainId
    id
    recipientAddress
    rootHash
    toDomainId
    tokenAddress
    type
    updatedAt
  }
}
```
* Step 6. Run a few mutations in a couple of domains to get some testing data 
* Step 7. Now it's time to cache the values from the past so run the following query
```
query MyQuery {
  cacheAllDomainBalance {
    success
  }
}
```
* Step 8. You can test the cache has been created by running this query
```
query MyQuery {
  listCacheTotalBalances(
    filter: {domainId: {eq: "0xD4c521eE2C1EF1325FF354C4DEd6cB9c3d62a6ec_1"}}
  ) {
    items {
      colonyAddress
      date
      domainId
      id
      timeframePeriod
      timeframeType
      totalUSDC
      totalUSDCIn
      totalUSDCOut
    }
  }
}
```
* Step 9. Now please go to http://localhost:9091/planex and check if the trends are showing up for the `Total in and out` card

Property 1=Completed

To be sure, the values are properly computed check the `getValuesTrend` method
The formula is
If current === 0
`-100`
if current !== 0 and previous === 0
`100` 
if previous > 0 and current > 0
`(previous - current) * 100 / previous` - special thanks to @Nortsova 🥇 

Further testing: please be as creative as possible

## Diffs

**New stuff** ✨

* The wonderful `cacheDomainBalance` lambda
* `useCurrencyHistoricalConversionRate` hook to compute the USDC value to the selected currency based on the date at which the historical balance was computed
* `getValuesTrend` utils
* `GetCachedDomainBalance` query

**Changes** 🏗

* Add `previousValue` property to `TitledSection` component

## TODO

- [ ] Still need to test the cron lambda gets triggered based on provided schedule - but probably will tackle it in another issue

Resolves #2858


src/hooks/useCurrencyConversionRate.ts
```diff
@@ -0,0 +1,43 @@
...
+  tokenSymbol: string;
+  conversionDenomination: CoinGeckoSupportedCurrencies;
+}
+
+interface CurrencyConversionRate {
+  conversionRate?: number;
+  lastUpdatedAt?: Date;
+}
+
+export const useCurrencyConversionRate = ({
```
mmioana, line 19: Need to update this to take historical data into account is date is provided

bassgeta (changes requested):
D A M N
took me a while to get through everything, but this stuff is actually quite straight forward. Nice job with the caching per day, basically saving us a ton of compute time for subsequent requests :+1: 
Left just some minor "cosmetic" comments which we can tackle, but overall a really great job with this!
So I can confirm that batch processing happens when triggering this lambda:
image
image
image
Filtering over teams works like a charm too:
image
image
It also doesn't break on mobile:
image

So I think we're good with this implementation, we probably need to run it against large sets of data to see how many resources it's gonna hog :D

amplify/backend/api/colonycdapp/schema/balance.graphql
```diff
@@ -93,4 +114,47 @@ type TokenExchangeRate @model {
...
+  colonyAddress: String! @index(name: "byColonyAddress", queryField: "cacheTotalBalanceByColonyAddress", sortKeyFields: ["date"] )
+  """
+  Domain id within a colony
+  """
+  domainId: String
+  """
+  The timeframe type (day/week/month) for which we want to cache the in/out token balance
+  Defaults to TimeframeType.DAILY
+  """
+  timeframeType: TimeframeType
```
jakubcolony, line 139: Should this be nullable?

jakubcolony (changes requested):
You've nailed it @mmioana! 🔨 

I left a couple comments but overall the code reads very nicely.

I've noticed some cache entries have `domainId` set to an empty string, if that's expected, should they be `null` instead?

image

Not sure if this is relevant to this PR, so I'll let you decide how to handle it, when quickly switching between teams, the value for the previously selected team replaces the current one once it loads:

https://github.com/user-attachments/assets/415ab1e1-fdf1-4df1-835d-55c32cd66ce9

Is this expected? The first screenshot shows General's payments at £2,377 yet when All teams is selected, the total is £0:
image
image

mmioana:
Thank you for your throughout review @jakubcolony 🙌 💯 
I did made some of the suggested changes and fixed the problem with the wrong data being shown, could you please re-check?

src/components/v5/frame/ColonyHome/partials/TotalInOutBalance/partials/TitledSection.tsx
```diff
@@ -45,9 +51,26 @@ export const TitledSection: FC<TitledSectionProps> = ({
...
+          </LoadingSkeleton>
+          <LoadingSkeleton
+            isLoading={isLoading}
+            className="h-4 w-[45px] rounded"
+          >
+            <div className="flex flex-row items-center gap-0.5 text-xs font-medium text-blue-400">
+              <ArrowUp
+                size={10}
+                transform={`rotate(${trend.isIncrease ? '0' : '180'})`}
+              />
```
Nortsova, line 69: In `FundsCardsTotalDescription.tsx` we have another logic for arrow up/down. Don't you want to move that arrow in a separate component, so we can reuse it in FundsCardsTotalDescription and here?

amplify/backend/api/colonycdapp/schema/balance.graphql
```diff
@@ -93,4 +114,47 @@ type TokenExchangeRate @model {
...
+  """
+  totalUSDCIn: String
+  """
+  The total out balance in USDC
+  """
+  totalUSDCOut: String
+  """
+  The total balance (totalIn - totalOut) in USDC
+  """
+  totalUSDC: String
```
Nortsova, line 159: just a random question: is it okay that we use USDC as a constant here? shouldn't it be something like "totalBase" + "baseCurrency" so we can change it in the future? Or it is fine to have permanent conversion to USDC?

amplify/backend/function/fetchDomainBalance/src/utils.js
```diff
@@ -26,44 +27,51 @@ const getActionWithFinalizedDate = (action) => ({
   finalizedDate: getActionFinalizedDate(action)
 })
 
-const getPeriodFromNow = (timeframePeriod, timeframeType) => {
+const getPeriodFromNow = (timeframePeriod, timeframeType, timeframePeriodEndDate) => {
```
Nortsova, line 30: sorry, could you please explain why this is called "fromNow"? thank you ❤️

Nortsova (commented):
Thank you, @mmioana! ✨ You've done an enormous amount of work!

I can confirm that trends works good with historical data: 
image

I left a couple of small comments to this PR. Thank you again for your work! 🙌

amplify/backend/function/fetchDomainBalance/src/utils.js
Nortsova: And another thing, would it be better to refactor arguments?
like this 
```suggestion
const getPeriodFromNow = (timeframeType, timeframePeriod, timeframePeriodEndDate) => {
```
or
```suggestion
const getPeriodFromNow = ({ timeframeType, timeframePeriod, timeframePeriodEndDate }) => {
```

rumzledz (approved):
@mmioana went beast mode on this beast ticket 🔥 I paired up with Ioanna over a call to go through the test steps and can confirm that changes are working as expected based on the test steps 💯 

I moved 230 CREDS from General to Andromeda and the bar charts are still working as expected:

Screenshot 2024-09-25 at 20 55 22

I updated the cache and the previous 30 days trend value has been updated correctly as well:

Screenshot 2024-09-25 at 20 51 35

Nortsova (approved):
Thank you @mmioana for your fast changes! Everything work as expected:

image
image
image

Really appreciate your work  🙌 you are a superhero 🦸‍♀️

mmioana:
Will merge this to avoid rebase conflicts ✅

Issue #2883, opened 2024-08-06
Buggy motion voting flow 


arrenv:
## Steps to reproduce

1. Create a new action with the Reputation decision method.
2. Fully stake and oppose the motion to take it to vote.
3. Vote in support.
4. Note that you need to sign twice, despite the pending button showing only one.
5. Note that the UI did not update automatically, I had to reload the app to update the status.
6. After reloading not that the UI does not inform you of your vote. It shows a Support button and the change vote button.
7. Click on the "Support" button then click the "Change Vote" button.
8. Complete the transactions.
9. Your vote will change and you will see the correct state.
10. Now, without clicking the "Oppose" button, click the "Change Vote" button.
11. Note that nothing will happen, but there are no errors or validation.

## Expected behaviour

* The UI should accurately show the number of wallet signatures required.
* The UI should update automatically when a transaction as succeeded that should update the status of the vote and motion.
* The voting widget should show the users vote and the text explaining they can change their vote.
* We should remove the single option after voting and when wanting to change votes. The UI, should just handle that when the user clicks "Change Vote" it should change to the opposite of their current vote. As if the below design.

![image](https://github.com/user-attachments/assets/ea861351-b90a-40b3-9d91-0178c7c3da40)

## Actual behaviour

* There are two wallet signatures required, but the pending button only shows 1.
* Sometimes voting does not update the UI automatically.
* There is missing content after the initial vote (this did not seem to happen every time).
* There are no errors or validation if you don't select the only option when changing vote.

[voting-flow.webm](https://github.com/user-attachments/assets/41a4f020-0107-410a-a0bc-d4073068a5c8)

iamsamgibbs:
Just a note that the UI issues have been resolved by #3108

The incorrect pending count will be addressed in a separate PR

iamsamgibbs:
I've spent some more time looking into the issue with the "Pending" button showing the incorrect number. I think this will be quite complex to fix and doesn't feel like a high priority issue right now. I suggest creating a new issue for this which we can fix at a later point. Thoughts? @arrenv @rdig 

To summarise the issue and potential fix:

The "Pending" button does not show the "correct" count because the first metamask popup is a **signature** rather than a **transaction**. The first number in the "Pending" button is determined by the number of transactions in a group that do not have the "Success" status. The second number is the total number of transactions in the group.

Ideally the signature would also show in the UserHub transactions list and could be retried if the signature was rejected.

[...]

PR #3108, opened 2024-09-16
Fix: Motion change vote

iamsamgibbs:
## Description

This PR addresses a few issues with the motion voting step.

On occasion, the UI would not update after a successful transaction to show the status of the vote. This PR resolves this by memoizing the transform function.

This PR also removes the single option after voting, and instead just handles the vote change when clicking the "Change vote" button.

The "Change vote" button uses the `ActionButton` component, which I have added a `loadingBehavior` prop to which allows for the button to be disabled when submitting to match how the "Submit vote" button works.

The "Oppose" and "Support" vote buttons are now disabled whilst the form is submitting.

(Note: The pending count issue described in the original issue will be handled in a separate PR.)

## Testing

* Step 1 - Install the reputation voting extension
* Step 2 - Create a motion and stake it
* Step 3 - Vote to either "Oppose" or "Suppose" the motion

The "Oppose" and "Support" buttons should be disabled whilst submitting. The UI should automatically update after the transaction.

* Step 4 - Click "Change vote" to change your vote (without needing to first select an option).

Notice the "Change vote" button will be disabled whilst submitting.

https://github.com/user-attachments/assets/4e5aba4d-bac9-4342-a41b-63676941bdf0

## Diffs

**Changes** 🏗

* Voting transform function is now memoized
* Voting options are no longer shown after initial vote
* "Change vote" button now uses `ActionButton` component to submit payload with changed vote
* "Oppose" and "Support" vote buttons are now disabled whilst the form is submitting.

Contributes to #2883

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/Motions/steps/VotingStep/hooks.tsx
+++ b/src/components/v5/common/ActionSidebar/partials/Motions/steps/VotingStep/hooks.tsx
@@ -1,6 +1,6 @@
 import { Extension } from '@colony/colony-js';
 import { BigNumber } from 'ethers';
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import { number, object } from 'yup';
 
 import MemberReputation from '~common/Extensions/UserNavigation/partials/MemberReputation/index.ts';
@@ -15,6 +15,7 @@ import { type InstalledExtensionData } from '~types/extensions.ts';
 import { type VoterRecord } from '~types/graphql.ts';
 import { type MotionAction } from '~types/motions.ts';
 import { mapPayload } from '~utils/actions.ts';
+import { MotionVote } from '~utils/colonyMotions.ts';
 import { formatText } from '~utils/intl.ts';
 
 import { type DescriptionListItem } from './partials/DescriptionList/types.ts';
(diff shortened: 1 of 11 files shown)


src/components/v5/common/ActionSidebar/partials/Motions/steps/VotingStep/hooks.tsx
```diff
@@ -116,31 +120,67 @@ export const useVotingStep = ({
...
+    setCurrentUserVote(vote);
     setHasUserVoted(true);
     reset();
     startPollingAction();
   };
 
   const validationSchema = object()
     .shape({
-      vote: number().required(),
+      vote: number().oneOf([MotionVote.Nay, MotionVote.Yay]).required(),
```
rdig, line 161: :+1:

rdig (approved):
All good here. Nicely done! :100: 

Screenshot from 2024-09-17 08-48-42
Screenshot from 2024-09-17 08-49-14

Here's the demo of how the controls interact and their disabled states


https://github.com/user-attachments/assets/4d4f939a-5014-4336-a64b-2988ac246b41

src/components/v5/shared/Button/types.ts
```diff
@@ -97,6 +97,11 @@ export interface CloseButtonProps extends CommonButtonProps {
   iconSize?: number;
 }
 
+export enum LoadingBehavior {
```
bassgeta, line 100: :100:

bassgeta (approved):
Nice and sensible changes, love the new loading behaviour variable!
Feels like the motion widget code is a bit more hardened now and it feels snappier with all the loading stages.
Disabled buttons when the transaction is pending:
image
Disabled state when voting:
image

PR #3122, opened 2024-09-18
[Dashboard] fix: rename fundsClaims to fundsClaimData because of the index bug

bassgeta:
## Description

This is bizzare, but for some reason using a plural breaks the index since the attribute ends with an s?!
See [this issue](https://github.com/aws-amplify/amplify-category-api/issues/321) for example
This is going to need a migration script too

## Diffs

**Changes** 🏗

* renamed `fundsClaims` to `fundsClaimData`

Contributes to #2858


jakubcolony (approved):
Not the first time amplify catches us out with easter eggs like that, nice work identifying the issue! 

There might be a chance you can get away without a migration, the only way funds claims are linked to a colony is via `colonyFundsClaimsId` which seems to have kept its name after your changes:
image

 
image

mmioana (approved):
Thank you for identifying and fixing this nasty issue @bassgeta 
Let's get it merged 💪 

Screenshot 2024-09-19 at 10 10 36

iamsamgibbs (approved):
Nice job fixing this weird issue! Query is still working as expected:

Screenshot 2024-09-19 at 09 51 52

src/graphql/fragments/colony.graphql
```diff
@@ -28,7 +28,7 @@ fragment Colony on Colony {
   balances {
     ...ColonyBalances
   }
-  fundsClaims {
```
rdig, line 31: We can always do `fundsClaimz` :) 

Or even worse, `fundsClaim5`

rdig (approved):
All good here. Nice fix. And sorry you had to deal with AWS bullshit :hankey: 

Screenshot from 2024-09-19 11-58-17

PR #3124, opened 2024-09-18
[Dashboard] feat: breadcrumbs

### LINKED ISSUE #3087: [Dashboard] Breadcrumbs
### Description

The breadcrumb on the dashboard screen should show as “Dashboard” only.

**Before**
Image

**After**
Image
Image

## Implementation
Use 
```
const pathSections = location.pathname.split('/');
const breadcrumbs = pathSections?.slice(2);  // this will be [''] if you are on colony level
```
This is just an idea, please feel free to be as creative as you want 💡 

- [ ] Selecting a route should update the displayed breadcrumb
- [ ] If it is a nested route, all partial routes are displayed as breadcrumbs
- [ ] Hovering over a breadcrumb item should have the correct styling
- [ ] Clicking on a breadcrumb item should navigate to the related route

[Figma link](https://www.figma.com/design/gac6xVgGNCAge5Bia933dp/Dashboard?node-id=4858-9978&node-type=frame&t=IJ0893EIblyb233u-0)
[Figma component](https://www.figma.com/design/gac6xVgGNCAge5Bia933dp/Dashboard?node-id=4878-89693&node-type=frame&t=HxDBCcAqAILoneMc-0)

bassgeta:
## Description

This PR introduces a new `BreadcrumbsContext` instead of reusing the `PageHeaderContext` which will be removed soon.
The context enables us to toggle showing of the breadcrumbs and what the first item should be (this is due to the colony route being just `/:colonyName` and it's hard to route match otherwise)

## Testing

Spin up the app and see how the breadcrumbs work.

1. Verify that under the colony routes, the first item is always `Dashboard`
image
2. All the other pages should display the URL they are on but in breadcrumb mode
image
3. Pages with more tabs also have the tab under the breadcrumbs
image
4. Extension pages get the correct breadcrumbs
image
5. They don't appear on the colony creation page, landing page and the profile creation page
6. The user account page breadcrumbs link to the landing page and change according to which tab is selected
image

## Diffs

**New stuff** ✨

* `BreadcrumbsContext` to easily control display of breadcrumbs

**Changes** 🏗

* The `Breadcrumbs` component is now leaner and URL aware, also has its own utils to override URL path segments into i18n strings

Resolves #3087

### DIFF
--- a/src/components/frame/LandingPage/LandingPage.tsx
+++ b/src/components/frame/LandingPage/LandingPage.tsx
@@ -5,7 +5,7 @@ import { useNavigate } from 'react-router-dom';
 
 import InvitationBlock from '~common/InvitationBlock/index.ts';
 import { useAppContext } from '~context/AppContext/AppContext.ts';
-import { usePageHeadingContext } from '~context/PageHeadingContext/PageHeadingContext.ts';
+import { useBreadcrumbsContext } from '~context/BreadcrumbsContext/BreadcrumbsContext.ts';
 import { MainLayout } from '~frame/Extensions/layouts/index.ts';
 import LoadingTemplate from '~frame/LoadingTemplate/index.ts';
 import ColonyIcon from '~icons/ColonyIcon.tsx';
@@ -100,16 +100,12 @@ const LandingPage = () => {
   const navigate = useNavigate();
   const { user, connectWallet, wallet, walletConnecting, userLoading } =
     useAppContext();
-  const { setBreadcrumbs } = usePageHeadingContext();
+  const { setShouldShowBreadcrumbs } = useBreadcrumbsContext();
 
   useEffect(() => {
-    setBreadcrumbs([
-      {
-        key: 'landing-page',
-        label: 'Colony App',
-      },
-    ]);
-  }, [setBreadcrumbs]);
+    setShouldShowBreadcrumbs(false);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
 
   if (userLoading || walletConnecting) {
     return <LoadingTemplate />;
--- a/src/components/frame/v5/pages/UserProfilePage/UserProfilePage.tsx
+++ b/src/components/frame/v5/pages/UserProfilePage/UserProfilePage.tsx
@@ -1,12 +1,11 @@
 import React, { type FC, useEffect } from 'react';
+import { defineMessages } from 'react-intl';
 import { Navigate, Outlet, useLocation, useNavigate } from 'react-router-dom';
 
+import { useBreadcrumbsContext } from '~context/BreadcrumbsContext/BreadcrumbsContext.ts';
 import { useFeatureFlagsContext } from '~context/FeatureFlagsContext/FeatureFlagsContext.ts';
 import { FeatureFlag } from '~context/FeatureFlagsContext/types.ts';
-import {
-  usePageHeadingContext,
-  useSetPageHeadingTitle,
-} from '~context/PageHeadingContext/PageHeadingContext.ts';
+import { useSetPageHeadingTitle } from '~context/PageHeadingContext/PageHeadingContext.ts';
 import {
   USER_CRYPTO_TO_FIAT_ROUTE,
   USER_ADVANCED_ROUTE,
@@ -19,6 +18,13 @@ import { formatText } from '~utils/intl.ts';
 import { TabId } from './types.ts';
 
 const displayName = 'v5.pages.UserProfilePage';
+
+const MSG = defineMessages({
+  homeBreadcrumbLabel: {
+    id: `${displayName}.homeBreadcrumbLabel`,
+    defaultMessage: 'Colony',
+  },
+});
 
 const tabRoutes = [
   {
@@ -58,7 +64,8 @@ const UserProfilePage: FC = () => {
   const { pathname } = useLocation();
   const featureFlags = useFeatureFlagsContext();
 
-  const { setBreadcrumbs } = usePageHeadingContext();
+  const { setRootBreadcrumbItem, setShouldShowBreadcrumbs, resetBreadcrumbs } =
+    useBreadcrumbsContext();
 
   const allowedTabRoutes = tabRoutes.filter((route) =>
     route.featureFlag
@@ -73,14 +80,17 @@ const UserProfilePage: FC = () => {
   )?.id;
 
   useEffect(() => {
-    setBreadcrumbs([
-      {
-        key: 'landing-page',
-        label: 'Colony App',
-        href: '/',
-      },
-    ]);
-  }, [setBreadcrumbs]);
+    setRootBreadcrumbItem({
+      link: '',
+      label: formatText(MSG.homeBreadcrumbLabel),
+    });
+    setShouldShowBreadcrumbs(true);
+
+    return () => {
+      resetBreadcrumbs();
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
 
   useSetPageHeadingTitle(formatText({ id: 'userProfile.title' }));
 
--- a/src/components/v5/shared/Breadcrumbs/Breadcrumbs.tsx
+++ b/src/components/v5/shared/Breadcrumbs/Breadcrumbs.tsx
@@ -1,42 +1,61 @@
+import { CaretRight } from '@phosphor-icons/react';
 import clsx from 'clsx';
 import React, { type FC } from 'react';
+import { useLocation } from 'react-router-dom';
+
+import { useBreadcrumbsContext } from '~context/BreadcrumbsContext/BreadcrumbsContext.ts';
 
 import Link from '../Link/index.ts';
 
-import { type BreadcrumbsItem, type BreadcrumbsProps } from './types.ts';
+import { type BreadcrumbsProps } from './types.ts';
+import { getBreadcrumbItemName } from './utils.ts';
 
 const displayName = 'v5.Breadcrumbs';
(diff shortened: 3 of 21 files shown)


src/routes/ColonyRoute.tsx
```diff
@@ -53,6 +56,15 @@ const ColonyRoute = () => {
     walletAddress: user?.walletAddress,
   });
 
+  useEffect(() => {
+    setRootBreadcrumbItem({ link: `/${colonyName}`, label: 'Dashboard' });
```
mmioana, line 60: Maybe we should use a MessageDescriptor here and not a hard-coded value

src/components/frame/v5/pages/UserProfilePage/UserProfilePage.tsx
```diff
@@ -73,14 +72,14 @@ const UserProfilePage: FC = () => {
...
   useEffect(() => {
-    setBreadcrumbs([
-      {
-        key: 'landing-page',
-        label: 'Colony App',
-        href: '/',
-      },
-    ]);
-  }, [setBreadcrumbs]);
+    setRootBreadcrumbItem({ link: '', label: 'Colony' });
```
mmioana, line 75: Maybe we should use a MessageDescriptor here and not a hard-coded value

src/components/v5/shared/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,47 +1,64 @@
...
 
-  return <h5>{item.label}</h5>;
-};
+  const getBreadcrumbLink = (index: number) => {
+    const rootLinkPrefix = rootBreadcrumbItem
+      ? `${rootBreadcrumbItem.link}/`
+      : '';
+
+    return `${rootLinkPrefix}${breadcrumbItems.slice(0, index + 1).join('/')}`;
+  };
```
mmioana, line 31: I think if we want to cover the paths from the user account, we should refactor like this

```suggestion
  const pathSections = location.pathname.split('/');
  const initialPath = pathSections.slice(0, 2).join('/');
  const breadcrumbItems = pathSections.slice(2);

  const getBreadcrumbLink = (index: number) => {
    return `${initialPath}/${breadcrumbItems.slice(0, index + 1).join('/')}`;
  };
```

src/components/v5/shared/Breadcrumbs/consts.ts
```diff
@@ -0,0 +1,17 @@
...
+    return {
+      ...names,
+      [extensionConfig.extensionId]: extensionConfig.name,
+    };
+  },
+  {},
+);
+
+export const breadcrumbItemNames: Record<string, MessageDescriptor> = {
+  ...extensionsBreadcrumbNames,
```
mmioana, line 16: Maybe it would make sense to actually add all the route constants here and add a custom translation for each possible breadcrumb at least as a `TODO` for now, otherwise we end up with this edge cases

Screenshot 2024-09-20 at 10 23 42

Screenshot 2024-09-20 at 10 24 38

Though for these 2 I would suggest to tackle them now 🙏

mmioana (changes requested):
From my side looks like a pretty nice and simple solution 👏 


Screenshot 2024-09-20 at 10 22 56
Screenshot 2024-09-20 at 10 23 01
Screenshot 2024-09-20 at 10 23 04
Screenshot 2024-09-20 at 10 23 08
Screenshot 2024-09-20 at 10 23 20
Screenshot 2024-09-20 at 10 23 24
Screenshot 2024-09-20 at 10 23 30
Screenshot 2024-09-20 at 10 23 42
Screenshot 2024-09-20 at 10 23 53
Screenshot 2024-09-20 at 10 23 55
Screenshot 2024-09-20 at 10 24 31
Screenshot 2024-09-20 at 10 24 38

There is an issue however if you click on the `Dashboard` item in the navigation that a `>` appears, probably due to the fact there is an extra `/` in the url, so maybe we need to filter out the path slices that are empty

https://github.com/user-attachments/assets/caaeecad-7979-4448-b39a-4330f27a81c1


Also left a small enhancement that would be nice to tackle now 🙏 

Other than that, awesome work @bassgeta ⭐

mmioana (approved):
Awesome work @bassgeta re-tested and it all works perfectly 👌 


https://github.com/user-attachments/assets/3d7eb24c-3aad-478f-8ba8-b4554507790d

rdig (changes requested):
Very nice work here! I tested all the places you highlighted, and while most is working as expected, there's places we need to fix and/or improve.

I'll first start with everything that works:

Screenshot from 2024-09-23 10-11-56
Screenshot from 2024-09-23 10-12-04
Screenshot from 2024-09-23 10-12-10
Screenshot from 2024-09-23 10-12-15
Screenshot from 2024-09-23 10-12-19
Screenshot from 2024-09-23 10-12-42
Screenshot from 2024-09-23 10-12-46

And now I'll highlight things that need improvement.

Landing page, shows a link to the "Colony"

Screenshot from 2024-09-23 10-04-44
Screenshot from 2024-09-23 10-04-49
Screenshot from 2024-09-23 10-04-52

Members page, only highlights the sidebar navigation, if the "All Teams" tab is selected, if "Contributors" is selected, the highlight goes away

Screenshot from 2024-09-23 10-12-25
Screenshot from 2024-09-23 10-12-35
Screenshot from 2024-09-23 10-12-37

The user profile's base breadcrumb is called "Colony", which shouldn't be since you're redirecting the landing page. If this is the "spec", then it needs to be addressed and this problem highlighted

Screenshot from 2024-09-23 10-12-55

mmioana:
@rdig 

Regarding the 1st problem you spotted, I'm wondering if this issue is related more to the new navigation feature rather than this one?

And for the 2nd problem, these are the specs `If the user is outside of a colony, the Root level is set as Colony and the first level breadcrumb is the corresponding page title/user location.`, but will confirm this is the expected behaviour

src/components/frame/LandingPage/LandingPage.tsx
```diff
@@ -100,16 +105,21 @@ const LandingPage = () => {
...
+      link: '',
+      label: formatText(MSG.homeBreadcrumbLabel),
+    });
+    setShouldShowBreadcrumbs(true);
+
+    return () => {
+      resetBreadcrumbs();
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
mmioana, line 122: I think we only need to `setShouldShowBreadcrumbs(false)` here based on the specs

mmioana:
@rdig @bassgeta

The first level breadcrumb is called Colony and will lead to the landing page is the intended behaviour for the `User profile` pages. 

No breadcrumbs will be shown on the `landing page`. The user will see a list of joined/created colonies which they can then select if they leave the user account via the breadcrumb mentioned above. 

So @bassgeta if you can just hide toggle off the breadcrumbs for the `LandingPage`

iamsamgibbs (commented):
Nice job, this is looking great!

Screenshot 2024-09-23 at 09 37 01

Screenshot 2024-09-23 at 09 46 49

I've noticed a couple of issues this has highlighted, but I believe they are all on the nav feature branch rather than issues that have been introduced here.

This issue with the sticky header on mobile. I'm guessing the breadcrumbs shouldn't be sticky on mobile, it looks messy. (It also includes the page title on some pages).

Screenshot 2024-09-23 at 09 50 24

However, it seems like this is on the new nav branch:

Screenshot 2024-09-23 at 10 00 04

And doesn't exist on master:

Screenshot 2024-09-23 at 10 01 05

So maybe it should be tackled on the nav feature branch.

Also noticed the same issues with the navigation that Raul mentioned, but these are unrelated to the breadcrumbs and should maybe be handled in the nav feature branch too?

Screenshot 2024-09-23 at 09 37 43
Screenshot 2024-09-23 at 09 41 19

Screenshot 2024-09-23 at 09 41 48

Not sure which of these issues are best tackled here, but the breadcrumbs work itself looks good, so happy to approve here. Also just seen the discord message regarding the breadcrumb on the landing page and think that should be tackled first.

src/components/frame/LandingPage/LandingPage.tsx
```diff
@@ -100,16 +100,12 @@ const LandingPage = () => {
...
-    setBreadcrumbs([
-      {
-        key: 'landing-page',
-        label: 'Colony App',
-      },
-    ]);
-  }, [setBreadcrumbs]);
+    setShouldShowBreadcrumbs(false);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
mmioana, line 108: Thank you for this change 💪

mmioana (approved):
Everything works as expected now 💪 thank you for picking this issue up and doing all these changes  🙌 
https://github.com/user-attachments/assets/4291b202-a3ef-4e67-8993-2eb42423d5f1

iamsamgibbs (approved):
Looking good now, breadcrumbs are not appearing on landing page:

Screenshot 2024-09-24 at 11 32 46

And still working as expected elsewhere:

Screenshot 2024-09-24 at 11 32 52

Approving on the assumption that the other raised issues are being handled elsewhere. :)

rdig (approved):
All good. 

No breadcrumbs on the landing page anymore.

Screenshot from 2024-09-25 10-35-11

The tabs on the various pages still don't highlight the left sidebar navigation item, but, as it was pointed out, that's a issue with the Nav Refactor, rather than this specific PR

Screenshot from 2024-09-25 10-36-13
Screenshot from 2024-09-25 10-36-17

Nice work here Kerry! :1st_place_medal:

Issue #3067, opened 2024-09-10
[Dashboard] Update the total in/out values computed by the lambda


mmioana:
Given we use the expenditures and not only the actions/motions to compute the total in/out values, we need to make sure

- [ ] the expenditure do not include the simply payments, otherwise we have an overlap with the existing actions
- [ ] make sure the expenditure payout is claimed 
- [ ] the expenditure funding can be from multiple domains, but as the claimed amount can't be decided from which domain it happened, we need to consider each funding domain contributed equally - @arrenv does it make sense to consider it like this?

### Implementation
Lambda to be updated `fetchDomainBalance`
Consider using this query to check the expenditures values when implementing this change

```
query GetDomainExpenditures {
  listExpenditures(filter: {nativeDomainId: { eq: _DOMAIN_ID_HERE_ }, status: { eq: FINALIZED }}) {
    items {
      id
      finalizedAt
      slots {
        claimDelay
        id
        payouts {
          amount
          isClaimed
          networkFee
          tokenAddress
        }
      }
      fundingActions: actions(
        filter: { type: { eq: MOVE_FUNDS } }
      ) {
        items {
          fromDomain {
            id
          }
          toDomain {
            id
          }
        }
      }
    }
  }
}
```

mmioana:
Depends on https://github.com/JoinColony/colonyCDapp/issues/2844

PR #3132, opened 2024-09-19
[Dashboard] Update the total in/out values computed by the lambda

mmioana:
## Description

This PR addresses updating the income and payments values correctly based on the income funds,`PAYMENT` and `MOVE_FUNDS` action types and the expenditures both at colony level and domain level.

## Testing

TODO: What needs to be tested here is that the income and payments values correspond to the financial actions happening within a colony or domain.
Easiest place to spot the changes included in this PR is the `Total in and out` card.

Screenshot 2024-09-25 at 12 09 36
Screenshot 2024-09-25 at 12 09 47

> [!IMPORTANT]
> Switching fast between the domain filters could show up the values from another domain.
> This issue is addressed in [another PR](https://github.com/JoinColony/colonyCDapp/pull/3080)

* Step 1. Please run the `create-data` script
* Step 2. Leave the `All teams` filter on
* Step 3. Check the income section of the card matches the value from the `Incoming funds` page
http://localhost:9091/planex/incoming

Screenshot 2024-09-25 at 12 15 29

* Step 4. Then the payments section should include all payments done within the colony 
* Step 5. To further test this, please create a new `Advanced payment` action

Screenshot 2024-09-25 at 12 19 56

* Step 6. The payments section should update (might need a page refresh)
* Step 7. You can compare the first `Total` card value from above with the `Income` - `Payments` values. They should match 🎉 

Screenshot 2024-09-25 at 12 22 44

* Step 8. Now select another domain. For eg. `Andromeda` and check the initial `Income` and `Payments` values

Screenshot 2024-09-25 at 12 24 04

* Step 9. Create some payments or transfer funds actions and check if the values are updating properly 🤩 

**Further testing**
* Step 1. Run the `create-colony-url` and setup a fresh new colony

Screenshot 2024-09-25 at 15 08 12
* Step 2. Mint some tokens 

Screenshot 2024-09-25 at 15 08 34
* Step 3. Check the total and income sections

Screenshot 2024-09-25 at 15 08 39

* Step 4. Create a new domain

Screenshot 2024-09-25 at 15 10 35

* Step 5. Transfer some funds from `General` to the new domain

Screenshot 2024-09-25 at 15 11 39

* Step 6. Check the `Total in and out` and `Total` cards for both `General` and the new domain

Screenshot 2024-09-25 at 15 12 02
Screenshot 2024-09-25 at 15 12 07
Screenshot 2024-09-25 at 15 12 12

* Step 6. Create a payment from the new domain
* Step 7. Check the `Total in and out` and `Total` cards for both `General` and the new domain
* Step 8. You can get a bit more funky and transfer some funds from one of `planex` domain to one of the new colony's domains

## Diffs

**New stuff** ✨

* `getAllDomainExpenditures` query
* `getFormattedActions` helper
* https://github.com/JoinColony/colonyCDapp/pull/3145 is cherry-picked on this PR

**Changes** 🏗

* Update `getExpendituresData` to take into account the colony level filtering
* No domain filtering for `getActionsData` at colony level
* Refactoring of `groupBalanceByPeriod`

Resolves [#3067](https://github.com/JoinColony/colonyCDapp/issues/3067)

### DIFF
--- a/amplify/backend/function/fetchDomainBalance/src/api/graphql/schemas.js
+++ b/amplify/backend/function/fetchDomainBalance/src/api/graphql/schemas.js
@@ -90,15 +90,16 @@ module.exports = {
     }
   `,
   getColonyExpenditures: /* GraphQL */ `
-    query GetDomainExpenditures(
+    query GetColonyExpenditures(
       $colonyAddress: ID!
       $nextToken: String
       $limit: Int
     ) {
-      listExpenditures(
+      getExpendituresByColony(
+        colonyId: $colonyAddress
         nextToken: $nextToken
         limit: $limit
-        filter: { colonyId: { eq: $colonyAddress }, status: { eq: FINALIZED } }
+        filter: { status: { eq: FINALIZED } }
       ) {
         items {
           id
--- a/amplify/backend/function/fetchDomainBalance/src/api/graphql/schemas.js
+++ b/amplify/backend/function/fetchDomainBalance/src/api/graphql/schemas.js
@@ -89,7 +89,7 @@ module.exports = {
       }
     }
   `,
-  getAllDomainExpenditures: /* GraphQL */ `
+  getColonyExpenditures: /* GraphQL */ `
     query GetDomainExpenditures(
       $colonyAddress: ID!
       $nextToken: String
@@ -114,7 +114,7 @@ module.exports = {
               tokenAddress
             }
           }
-          fundingActions: actions(
+          createExpenditureActions: actions(
             filter: { type: { eq: CREATE_EXPENDITURE } }
           ) {
             items {
@@ -155,7 +155,7 @@ module.exports = {
               tokenAddress
             }
           }
-          fundingActions: actions(
+          createExpenditureActions: actions(
             filter: { type: { eq: CREATE_EXPENDITURE } }
           ) {
             items {
--- a/amplify/backend/function/fetchDomainBalance/src/api/graphql/schemas.js
+++ b/amplify/backend/function/fetchDomainBalance/src/api/graphql/schemas.js
@@ -119,6 +119,18 @@ module.exports = {
               tokenAddress
             }
           }
+          fundingActions: actions(filter: { type: { eq: MOVE_FUNDS } }) {
+            items {
+              fromDomain {
+                id
+              }
+              toDomain {
+                id
+              }
+              amount
+              id
+            }
+          }
         }
       }
     }
--- a/amplify/backend/function/fetchDomainBalance/src/api/graphql/schemas.js
+++ b/amplify/backend/function/fetchDomainBalance/src/api/graphql/schemas.js
@@ -89,6 +89,42 @@ module.exports = {
       }
     }
   `,
+  getAllDomainExpenditures: /* GraphQL */ `
+    query GetDomainExpenditures(
+      $colonyAddress: ID!
+      $nextToken: String
+      $limit: Int
+    ) {
+      listExpenditures(
+        nextToken: $nextToken
+        limit: $limit
+        filter: { colonyId: { eq: $colonyAddress }, status: { eq: FINALIZED } }
+      ) {
+        items {
+          id
+          finalizedAt
+          createdAt
+          slots {
+            claimDelay
+            id
+            payouts {
+              amount
+              isClaimed
+              networkFee
+              tokenAddress
+            }
+          }
+          fundingActions: actions(
+            filter: { type: { eq: CREATE_EXPENDITURE } }
+          ) {
+            items {
+              id
+            }
+          }
+        }
+      }
+    }
+  `,
   getDomainExpenditures: /* GraphQL */ `
     query GetDomainExpenditures(
       $colonyAddress: ID!
@@ -119,15 +155,10 @@ module.exports = {
               tokenAddress
             }
           }
-          fundingActions: actions(filter: { type: { eq: MOVE_FUNDS } }) {
+          fundingActions: actions(
+            filter: { type: { eq: CREATE_EXPENDITURE } }
+          ) {
             items {
-              fromDomain {
-                id
-              }
-              toDomain {
-                id
-              }
-              amount
               id
             }
           }
--- a/amplify/backend/function/fetchDomainBalance/src/services/actions.js
+++ b/amplify/backend/function/fetchDomainBalance/src/services/actions.js
@@ -6,7 +6,7 @@ const {
   getTokensDecimalsFor,
 } = require('../api/graphql/operations');
 const {
-  getActionWithFinalizedDate,
+  getFormattedActions,
   getFormattedIncomingFunds,
   getFormattedExpenditures,
   getTokenAddressesFromExpenditures,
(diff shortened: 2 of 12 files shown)


amplify/backend/function/fetchDomainBalance/src/services/actions.js
```diff
@@ -30,17 +30,22 @@ const getInOutActions = async (colonyAddress, parentDomainId) => {
...
   const tokensDecimals = await getTokensDecimalsFor(expendituresTokenAddresses);
 
   // Getting all remaining financial actions not included in incoming funds or expenditures
   let actions = await getAllActions(colonyAddress, parentDomainId);
 
+  console.log(filteredExpenditures.length);
+
   return [
-    ...getFormattedIncomingFunds(incomingFunds, parentDomainId),
+    // ...getFormattedIncomingFunds(incomingFunds, parentDomainId),
```
mmioana, line 46: This will need to be uncommented

amplify/backend/function/fetchDomainBalance/src/services/actions.js
```diff
@@ -30,17 +30,22 @@ const getInOutActions = async (colonyAddress, parentDomainId) => {
...
     ...actions.map((action) => getActionWithFinalizedDate(action)),
-    ...getFormattedExpenditures(expenditures, parentDomainId, tokensDecimals),
+    // ...getFormattedExpenditures(filteredExpenditures, parentDomainId, tokensDecimals),
```
mmioana, line 48: This will need to be uncommented

amplify/backend/function/fetchDomainBalance/src/api/graphql/schemas.js
```diff
@@ -89,6 +89,42 @@ module.exports = {
       }
     }
   `,
+  getAllDomainExpenditures: /* GraphQL */ `
```
mmioana, line 92: Maybe we should rename this to `getColonyExpenditures` or `getColonyDomainsExpenditures`

amplify/backend/function/fetchDomainBalance/src/services/tokens.js
```diff
@@ -1,38 +1,54 @@
...
-        .sub(formattedNetworkFee)
-        .div(tokenScaleFactor)
-        .mul(normalizedTokenExchangeRate)
-        .div(BigNumber.from(10).pow(exchangeRateDecimals));
-}
+  return (
+    BigNumber.from(amount)
+      // @TODO: networkFee needs to be add when calculating outgoing expenditures as it is added on top of the amount
+      // However, I am not sure if this is the case with every other action being passed here
+      // Maybe in some cases networkFee needs to be subtracted?
```
mmioana, line 21: Totally agree with this comment, though we should consider adding the `networkFee` or `-networkFee` at the actions/funds/expenditure level for better clarity 😁

amplify/backend/function/fetchDomainBalance/src/utils.js
```diff
@@ -117,12 +117,15 @@ const buildAPIEndpoint = (url, queryParams) => {
...
-const getFormattedIncomingFunds = (incomingFunds, parentDomainId) =>
+const getFormattedIncomingFunds = (incomingFunds, domainId) =>
   incomingFunds.map((incomingFund) => ({
     amount: incomingFund.amount,
     finalizedDate: incomingFund.updatedAt,
     token: incomingFund.token,
-    toDomainId: parentDomainId,
+    // @TODO: There will only ever be incomingFunds if "All teams" or the "Root domain" is selected
+    // So this value will only ever be undefined or 1
+    // Not sure if it is actually needed?
```
mmioana, line 127: It's only needed to be able to properly add it to the ` balance[period].out` or ` balance[period].in` array

amplify/backend/function/fetchDomainBalance/src/api/graphql/schemas.js
```diff
@@ -119,15 +155,10 @@ module.exports = {
               tokenAddress
             }
           }
-          fundingActions: actions(filter: { type: { eq: MOVE_FUNDS } }) {
+          fundingActions: actions(
```
mmioana, line 158: Maybe we should rename this into `createExpenditureActions` 
And maybe also leave a comment on this line that we'll take into account also the `MOVE_FUNDS` actions needed for expenditures in case there are not enough funds in the domain where the expenditure was created 
`let actions = await getAllActions(colonyAddress, domainId);`

src/redux/sagas/utils/proofs.ts
```diff
@@ -506,3 +524,58 @@ export function* getMultiPermissionProofs({
...
+  requiredColonyRole: ColonyRole;
+  permissionAddress: string;
+  isMultiSig?: boolean;
+}
+
+// Returns the permission proofs from the domain where the user has been assigned the permissions
+// If the user has permissions directly in the requiredDomainId those will be returned first
+// However, if the user has inherited permissions in the requiredDomainId,
+// the permission proofs will be returned from the domain from which they are inherited
+export const getSinglePermissionProofsFromSourceDomain = async ({
```
bassgeta, line 542: I know there's probably a reason for this, but at first glance, why would this be any different to how `getSinglePermissionProofsLocal` already works like?
That one tries to get permissions for the passed domainId, and if it doesn't find them it goes to the parent (root) domain and does it again for the parent domain

bassgeta (approved):
Damn this one handles ALL the edge cases doesn't it :D Good job both of you on this one, it's a ton of head scratching and I imagine quite some console.log("HERE???") moments, so big props :muscle: 

Dropped 2 comments, one that may improve our querying performance and the other is more of a question, feel free to reply and resolve if I completely misfired with the `getPermissionProofs` comment 

ANYHOW, the testing steps:
Incoming funds :heavy_check_mark: 
image
image

Paid 1000 CREDS to `leela`
image
the payments increased (@mmioana I am guessing that the network fee needs to be included and this is fine? I can confirm that my network fee was 10 CREDS here)
image

Andromeda initial state:
image
Pay 80 from Andromeda:
image
Transfer 100 from General to Andromeda:
image

Andromeda state is correct :tada: 
image

New colony testing time :clock1: 
image
Minted 5000 tokens :heavy_check_mark: 
image
image
Moved 100 tokens to the new team:
image
General :heavy_check_mark: 
image
New team :heavy_check_mark: 
image
Paid 15 tokens out from the new domain:
image
works :heavy_check_mark: 
image


this is definitely making good use of Github's storage with all the screenshots :sunglasses:

src/redux/sagas/utils/proofs.ts
iamsamgibbs: This is actually part of the cherry-picked commit from #3145 so maybe better discussed over there, but the gist is that `getSinglePermissionProofsLocal` returns the permission proofs from the first domain you have permissions in, whereas this new function will return the permission proofs from the actual domain you have the permissions in. For example, you are funding an expenditure in Serenity, you have been assigned the necessary permissions in the root domain so you inherit them in Serenity. `getSinglePermissionProofsLocal` will return the permission proofs for Serenity, `getSinglePermissionProofsFromSourceDomain` will return the permission proofs for the root domain.

bassgeta: ahhh gotcha, such an edge case, but in that case it makes sense :D and I would really prefer not to add another flag to that function, so you were spot on with decoupling it to a new function!
But it's still bugging me that it almost feels like `getSinglePermissionProofs` should function in the way you've described 
image
but I guess the difference is with the `idx` return value?

Nortsova (approved):
Hey @mmioana ! Amazing work! I tested it and it works as expected:

Before:
image
After:
image
Andromeda Moving funds also was updated: 
image

After minting a token for the new colony value is updated even without refresh:
image

Transfer funds from General to SubDomain:
image
image
image

Payment to leela from test domain:
image

Great work 💪🙌

PR #2908, opened 2024-08-08
fix: do not allow user to use app without accepting metamask sign-in

Nortsova:
## Description

Work in progress! 

- [] TODO write description

## Testing

TODO: High level overview of what reviewers are testing with your pr. 

TODO: Followed by step-by-step instructions for testing your branch so reviewers can easily jump straight in and start testing. 

* Step 1. E.g. add reputation
* Step 2. E.g. install Governance extension
* Step 3. E.g. Create motion

## Diffs

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

* Changes go here

**Deletions** ⚰️

* Deletions go here

## TODO

- [ ] Add TODOs

Resolves #2555

PR #3133, opened 2024-09-19
fix: remove getSigner spread as it is no longer needed and getSigner already included in the RetryRpcProvider class

### LINKED ISSUE #3113: `useCurrentBlockTime` hook is broken in dev
## Steps to reproduce

The issue is most notable in streaming payments, however it should be also reproducible on `master`:

1. Create an advanced payment.
2. See errors in the console:

image

Looks like `ethersProvider` is missing `getBlock` method after changes from https://github.com/JoinColony/colonyCDapp/commit/fec0d0a6e9dd5d609dd3d7c1e9e5b0a55fefef8f.

Nortsova:
## Description
I pair-programmed this with @jakubcolony, and we agreed on the solution with @rdig 🙌 Thank you guys for helping!

The issue was that we encountered an error when creating advanced payments:
image

The history of this bug is quite long. It seems @rdig implemented the `getSigner` method 9 months ago to make the Provider work. However, 7 months ago, @chmanie added this method to the Provider class itself, and by using the spread operator, we unintentionally lost a couple of methods from the Provider prototype (such as `getBlock`).

In my PR for the Metamask refactor (https://github.com/JoinColony/colonyCDapp/pull/2908), I removed the usage of the getBasicWallet method. As a result, this error now appears every time, rather than just once when logging in (getFullWallet), which makes the issue more noticeable.

## Testing
* Step 1. Create Advanced payment
* Step 2. Check that there are no errors in the console
image

I would also really appreciate it if you could point out any other errors you spot, as this wallet/provider part is quite vital to our app. 🙏✨

Resolves #3113

### DIFF
--- a/src/hooks/useNetworkMotionStates.ts
+++ b/src/hooks/useNetworkMotionStates.ts
@@ -5,6 +5,8 @@ import { useAppContext } from '~context/AppContext/AppContext.ts';
 
 import useEnabledExtensions from './useEnabledExtensions.ts';
 
+import type { Provider } from '@ethersproject/providers';
+
 export type MotionStatesMap = Map<string, MotionState | null>;
 
 export type RefetchMotionStates = (motionIdsToRefetch?: string[]) => void;
@@ -23,7 +25,7 @@ const useNetworkMotionStates = (nativeMotionIds: string[], skip?: boolean) => {
   );
 
   useEffect(() => {
-    const { ethersProvider, address } = wallet || {};
+    const { ethersProvider } = wallet || {};
     if (
       skip ||
       !nativeMotionIds.length ||
@@ -44,11 +46,11 @@ const useNetworkMotionStates = (nativeMotionIds: string[], skip?: boolean) => {
     }
 
     // Properly initialize the signer with the current wallet address
-    const signer = ethersProvider.getSigner(address);
-
     const votingRepClient = VotingReputationFactory.connect(
       votingReputationAddress,
-      signer,
+      // We need to cast this to a Provider since colonyJS methods expect a "standard" provider,
+      // while ours is a custom one (which extends the base, standard provider)
+      ethersProvider as unknown as Provider,
     );
 
     const fetchMotionStates = async () => {
--- a/src/redux/sagas/utils/safeHelpers.ts
+++ b/src/redux/sagas/utils/safeHelpers.ts
@@ -10,6 +10,7 @@ import {
   type NetworkInfo,
   SUPPORTED_SAFE_NETWORKS,
 } from '~constants/index.ts';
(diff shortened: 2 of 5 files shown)


iamsamgibbs (changes requested):
Sorry to be the bearer of bad news, but this breaks the metamask wallet. Notice that when I try connecting with the metamask wallet it doesn't prompt me when connecting. It also doesn't prompt me when making the transactions for the advanced payment action.

https://github.com/user-attachments/assets/07621251-8369-4efa-8de4-cb1dba2633af

jakubcolony:
@Nortsova I think the issue is that RetryProvider's Extender assumes local dev always uses a dev wallet and not Metamask. The proper solution would be to choose the Extender based on the wallet type and not the environment, but as this will be refactored soon your idea with overriding `getSigner` without object spread would work well too

rdig:
@iamsamgibbs is correct in the issue he found with the Metamask wallet. It's correct in assessing that no prompt to sign message or transactions show up, but that's because the provider got confused and it used the ganache one _(this worked because he used the same wallets for both)_

@Nortsova you are correct in assessing that this logic should not be based on the environment, but on the wallet. This is is on me, since I introduced the faulty logic initally.

I have since pushed to fixes to this branch, one that refactors the `RetryProvider` a bit and makes this work based on the wallet type

Another, that reverts the fix made during #2871 since it's no longer needed

_(I'll provide the demo for the fixes in the review)_

rdig (approved):
As I have mentioned in the previous comment, I have now refactored the `RetryProvider` so this now works as expected.

Connecting / signing via Metamask:

https://github.com/user-attachments/assets/f2f4e7ea-a313-435b-a96c-da84cf593625

Connecting / signing via Local Dev Wallet: _(ganache/hardhat)_

https://github.com/user-attachments/assets/ee811530-fc43-49bb-8fa9-9d7711d0420b

Motion states correctly showing up after a page refresh:

https://github.com/user-attachments/assets/5300a8d6-2a44-4fe6-aacb-e70a34e07488

iamsamgibbs (approved):
🎉 Amazing, all working as expected.

Tested with metamask:

https://github.com/user-attachments/assets/89ee980b-cc5d-42e3-b4ce-1c573cd262a3

And with the dev wallet:

https://github.com/user-attachments/assets/0d7aebdd-b531-4f33-8efe-48ec6e45025a

Also tested creating motions with both just to be safe, and checked that the motion states are correct even after uninstalling and reinstalling the extension.

Screenshot 2024-09-20 at 09 43 10
Screenshot 2024-09-20 at 09 43 14

jakubcolony (approved):
Thank you @Nortsova and @rdig for your efforts, I can connect with Metamask and the block time hook works flawlessly 👍

PR #3149, opened 2024-09-23
feat: cancel streaming payment

joanna-pagepro:
## Description

Cancel streaming payments using the permissions decision method

## Testing

* Step 1. Create streaming payment action using different end conditions each time
* Step 2. Select the `Cancel payment` option from the kebab menu. A modal should appear where you should select the `Permissions` decision method and then cancel the payment. The status of the action will change to `Cancelled`

## Diffs

**New stuff** ✨

* Cancel modal

**Changes** 🏗

* 

**Deletions** ⚰️

* 


src/redux/sagas/utils/safeHelpers.ts
```diff
@@ -10,6 +10,7 @@ import {
   type NetworkInfo,
```
jakubcolony, line 10: So are these...

src/components/v5/common/CompletedAction/partials/StreamingPayment/StreamingPayment.tsx
```diff
@@ -119,13 +132,15 @@ const StreamingPayment: FC<StreamingPaymentProps> = ({
   const hasPermissions = addressHasRoles({
     address: user?.walletAddress || '',
     colony,
-    requiredRoles: [ColonyRole.Arbitration],
+    requiredRoles: [ColonyRole.Administration],
     requiredRolesDomains: [streamingPaymentData.nativeDomainId],
   });
 
   // @todo: update cancel-related logic in separate PR
```
jakubcolony, line 139: is that comment still relevant?

src/components/v5/common/CompletedAction/partials/StreamingPayment/partials/CancelModal/CancelModal.tsx
```diff
@@ -0,0 +1,140 @@
...
+      >
+        {({ watch, formState: { isSubmitting } }) => {
+          const method = watch('decisionMethod');
+          const { roleToShow } = getMemberPermissions(
+            colony,
+            user?.walletAddress || '',
+            streamingPayment.nativeDomainId,
+          );
+
+          const cancelDecisionMethodDecscriptions =
```
jakubcolony, line 94: typo

src/components/v5/common/CompletedAction/partials/StreamingPayment/partials/CancelModal/CancelModal.tsx
```diff
@@ -0,0 +1,140 @@
...
+        streamingPayment,
+        userAddress: user?.walletAddress ?? '',
+        shouldWaive,
+      };
+
+      await cancel(payload);
+
+      await refetchStreamingPayment({
+        streamingPaymentId: streamingPayment.id,
+      });
```
jakubcolony, line 55: That's a race condition as there's no guarantee the cancel action has been processed by the time it refetches. I can't tell exactly what depends on it but it either needs to use a concept similar to "expectedStepKey" in expenditures or "waitForDbAfterExtensionAction".

src/components/v5/common/CompletedAction/partials/StreamingPayment/partials/CancelModal/CancelModal.tsx
```diff
@@ -0,0 +1,140 @@
...
+          const method = watch('decisionMethod');
+          const { roleToShow } = getMemberPermissions(
+            colony,
+            user?.walletAddress || '',
+            streamingPayment.nativeDomainId,
+          );
+
+          const cancelDecisionMethodDecscriptions =
+            getCancelDecisionMethodDescriptions(
+              roleToShow?.name || formatText({ id: 'role.mod' }),
```
jakubcolony, line 96: This seems like a lot of leg work to show the current user role, which is not necessarily the 'minimum' role that is required. Would it be easier to hardcode the role, even if it means having different i18n messages for various cancel modals?

src/utils/streamingPayments.ts
```diff
@@ -81,7 +81,7 @@ export const getStreamingPaymentAmountsLeft = (
     amountAvailableToClaim = amountAvailableSinceStart.sub(amountClaimedToDate);
 
     amountAvailableToClaim = amountAvailableToClaim.lt(0)
-      ? BigNumber.from(0)
+      ? amountAvailableToClaim.mul(-1)
```
jakubcolony, line 84: If the amount is less than 0, showing 0 seems more correct than negating the amount. Are you handling a specific edge case here?

jakubcolony (changes requested):
Thanks for your work @joanna-pagepro, left a few comments for you to look at, it doesn't look far though 👍

src/components/v5/common/CompletedAction/partials/StreamingPayment/partials/CancelModal/CancelModal.tsx
joanna-pagepro: I've asked about it at the daily meeting and I have to show the current user role here

src/utils/streamingPayments.ts
joanna-pagepro: I have to check this because maybe this is something that I've added before your fixes

joanna-pagepro: I wasn't needed anymore, so I reverted this change

joanna-pagepro:
@jakubcolony can you review it again?

jakubcolony (changes requested):
Thanks for the changes @joanna-pagepro 

Unfortunately I ran into a couple of issues when cancelling a payment that hasn't yet started:
- After cancelling, the UI was stuck in a loading state despite the payment having updated in the DB
- After page refresh, the payment was still showing as "Not started" and I could try cancelling again (which fails on the contracts)



https://github.com/user-attachments/assets/b3633407-c55e-47a7-b0b1-0e4b14b96a9e

When cancelling an already active stream, the UI also got stuck, however after a page refresh the status updated correctly to Cancelled:

image

jakubcolony (approved):
I can't reproduce the loading issue anymore, and everything else looks good. Thanks for your work @joanna-pagepro! 👍 

cancel-stream-not-active
cancel-stream-active

arrenv (changes requested):
@joanna-pagepro Thank you for you patience on the review, it all functions as expected. I noticed a couple of things.

1. The loader button should use the pending version, as opposed to the spinner.

image

### Correct pending state of transactions
Except the font size should be the same in the standard state. Here is the correct style - https://www.figma.com/design/l1dOM5qiQYwF0ElvKDqqjg/Design-System---Colony-v3?node-id=1801-431&t=R9tKCzHbolLlEki7-4

image


2. The spinner is used here as well:

image


3. I see that after claiming a cancelled stream, you are left with a very small number. Which results in the pay funds always being present.

image

joanna-pagepro:
@arrenv buttons updated

arrenv (changes requested):
@joanna-pagepro Thank you for the updates. Looks like just aa couple more things noticed.

1. The text on the cancel button says "Yes, cancel and return stake", there is no stake, it should be "Confirm cancellation".
image

2. The dropdown of the Decision method shows blue options.
image

Here is the correct design:
image

joanna-pagepro:
@arrenv 
1. updated
2. You have the `permissions` option selected and that's why it's blue. If no option is selected it's the same as on a design. If there were 2 options available, then the other one wouldn't be blue in this case.

arrenv (changes requested):
@joanna-pagepro Thank you for the update and your patience in my review. 

Sorry, I have just tested it and noticed these issues.

1. The action gets stuck in a loading view, that only resolves after reloading the action.
image


2. The transaction list is missing a translation for the stream cancelattion transaction.
image

joanna-pagepro:

src/components/v5/common/CompletedAction/partials/StreamingPayment/partials/StreamingPaymentWidget/StreamingPaymentWidget.tsx
```diff
@@ -221,42 +226,48 @@ const StreamingPaymentWidget: FC<StreamingPaymentWidgetProps> = ({
...
+                    isLoading={
+                      loadingStreamingPayment || isStatusLoading || isLoading
+                    }
                     className="h-[1.125rem] w-14 rounded"
                   >
                     <AvailableToClaimCounter
                       hasEnoughFunds={hasEnoughFunds}
                       status={paymentStatus}
-                      onTimeEnd={fetchCurrentBlockTime}
+                      // onTimeEnd={fetchCurrentBlockTime}
```
iamsamgibbs, line 237: This can be removed now?

src/components/v5/common/CompletedAction/partials/StreamingPayment/partials/CancelModal/CancelModal.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+      const payload: CancelStreamingPaymentPayload = {
+        colonyAddress: colony.colonyAddress,
+        streamingPayment,
+        userAddress: user?.walletAddress ?? '',
+        shouldWaive,
+      };
+
+      await cancel(payload);
+      onSuccess();
```
iamsamgibbs, line 55: I think we should refetch the streaming payment here or as part of the onSuccess handler so that we have the latest claim amount.

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/CancelModal/consts.ts
```diff
@@ -12,9 +12,14 @@ export const cancelDecisionMethodItems: SelectBaseOption[] = [
...
 export const cancelDecisionMethodDescriptions = {
-  [DecisionMethod.Permissions]: formatText({
-    id: 'cancelModal.permissionsDescription',
-  }),
+  [DecisionMethod.Permissions]: formatText(
+    {
+      id: 'cancelModal.permissionsDescription',
+    },
+    {
+      role: 'Payer',
```
iamsamgibbs, line 20: Should Payer be hardcoded as the role here?

src/components/v5/common/CompletedAction/partials/StreamingPayment/StreamingPayment.tsx
```diff
@@ -290,6 +306,14 @@ const StreamingPayment: FC<StreamingPaymentProps> = ({
       {action.annotation?.message && (
         <DescriptionRow description={action.annotation.message} />
       )}
+      <CancelModal
+        isOpen={isCancelModalOpen}
+        streamingPayment={streamingPaymentData}
+        onClose={toggleCancelModalOff}
+        onSuccess={() =>
+          setExpectedPaymentStatus(StreamingPaymentStatus.Cancelled)
```
iamsamgibbs, line 314: Or could add refetchStreamingPayment here as part of this onSuccess handler.

iamsamgibbs (changes requested):
Nice work on this @joanna-pagepro ! It is mostly there I think. I am able to cancel the streaming payment, but there are some weird behaviours with claiming after cancelling.

I had this scenario where I had claimed some tokens whilst the payment was active, then when I cancel and try to claim the rest, the UI does not update correctly.

https://github.com/user-attachments/assets/5f4caa51-2850-44f0-854e-05e08f24f71f

Screenshot 2024-10-16 at 15 51 11

I tried adding `refetchStreamingPayment` to the `handleCancelStreamingPayment` function so that is has the latest claim amount when moving to the cancelled state which seemed to help - I've left a comment.

There is still an issue with the transaction string:

Screenshot 2024-10-16 at 16 11 26

The claim funds section is also just quite slow to update, the "Pay funds" button switches back from the pending state and is shows the stale values in the "Paid to date" and "Available to claim" fields for a full 2-3 seconds before updating to the correct values without any loading state. It's probably out of scope for this issue but will need addressing.

adam-strzelec:
@iamsamgibbs done 😉

arrenv (changes requested):
@adam-strzelec Thank you for the updates. I seem to still be having the issue when clicking on the cancel transactions in the Transactions tab in the Userhub linking to a "Not implemented yet" page. This should like to the relevant stream action.

image

arrenv (changes requested):
@adam-strzelec Thank you for the update, that link works now. Although, what I have noticed is that when you click on the link the Stream card is not visible.

You will notice that they are different transaction hashes, which makes what you have done here even more impressive, although, probably a little overkill.

Perhaps it is worth looking at this issue that has resolved a similar thing - https://github.com/JoinColony/colonyCDapp/pull/4007

#### Stream
image

### Clicking in the cancel transaction
image

arrenv:
> @adam-strzelec Thank you for the update, that link works now. Although, what I have noticed is that when you click on the link the Stream card is not visible.
> 
> You will notice that they are different transaction hashes, which makes what you have done here even more impressive, although, probably a little overkill.
> 
> Perhaps it is worth looking at this issue that has resolved a similar thing - #4007
> 
> #### Stream
> image
> 
> ### Clicking in the cancel transaction
> image

adam-strzelec:
@arrenv @iamsamgibbs done 😉

iamsamgibbs (changes requested):
There are still a few problems with this I'm afraid!

The changes from #4007 won't just work without additional changes. I don't think the sagas for cancelling or claiming streaming payments are on master, so they don't have the new `associatedActionId` prop which has been added elsewhere.

If you look at `cancelExpenditure` for example, it has this: `associatedActionId: getExpenditureCreatingActionId(expenditure),`

The `cancelStreamingPayment` saga will need `associatedActionId: getStreamingPaymentCreatingActionId(streamingPayment),` and a new `getStreamingPaymentCreatingActionId` utility function created for it. I'm happy to take a look at adding this in if you need a hand with it.

Claiming a streaming payment has the same issue.

Aside from that, I'm still having issues with the claim box. It always shows a remaining amount to be claimed even after attempting to claim multiple times, reloading the page, etc.

https://github.com/user-attachments/assets/faac7da9-66d5-48c8-bfd5-68be934a1231

I didn't look much beyond this so there may be other issues, but lets resolve these first.

iamsamgibbs:
@adam-strzelec I hope you don't mind, but I just pushed a fix for `getStreamingPaymentCreatingActionId`. Your previous commit got it most of the way there, but the changelog is empty until there is an edit so your implementation would have only worked after edits had been made. I've refactored it to use the transaction hash from the action that created the streaming payment.

iamsamgibbs (approved):
Thanks for the hard work on this! I think it's all good to go (after a rebase to resolve the conflict!)

I'm no longer experiencing an issue with the "available to claim" amount not updating correctly, whether I have previously made claims or not, it is working great now.

Screenshot 2025-01-27 at 10 00 13

I think this has been previously mentioned, but notifications aren't working for any of the streaming payments related actions, this should be it's own issue.

Lets finally get it merged!

src/components/common/ColonyActionsTable/partials/ActionDescription/ActionDescription.tsx
```diff
@@ -42,8 +41,9 @@ const ActionDescription: FC<ActionDescriptionProps> = ({
     true,
   );
 
-  const { expenditure, loadingExpenditure } =
-    useGetExpenditureData(expenditureId);
+  const { expenditure, loadingExpenditure } = useGetExpenditureData(
+    action.streamingPaymentId,
```
jakubcolony, line 45: This change breaks the description of all expenditure related actions (advanced, split, staged payment) in the activity feed.

jakubcolony (changes requested):
Thanks for picking this up @adam-strzelec, I had another go at reviewing as it's been some time since my approval.

One of the changes (I left a comment which) appears to break the action description for advanced payments:
Screenshot 2025-01-27 at 23 01 11

image


----


As I resize the window, action sidebar is overlapping the navigation sidebar. I cannot reproduce this on master or `feat/streaming-payments-ui`, so I suspect it was introduced somewhere in this PR:

https://github.com/user-attachments/assets/8bd89692-1068-46df-8837-457755bc25ef




----

Otherwise, the functionality works pretty well:

image

image

image

adam-strzelec:
@iamsamgibbs thank you for help 🙂 @jakubcolony sidebar width and description are already fixed

jakubcolony (approved):
Nice work here, the description looks correct and the sidebar issue is no longer reproducible: 

image

https://github.com/user-attachments/assets/4f27fe38-13a3-4ae3-9947-8b4ee4985229

arrenv (approved):
@adam-strzelec Everything looks to be working great, very nice! Thank you!

### Activity feed

image

### Cancel transaction

image

### Claiming

image

Issue #3146, opened 2024-09-23
[Dashboard] Reputation chart hover state


mmioana:
### Specification
- Hovering over a section of the chart:
    - It should show the hover state of the section.
    - It should highlight the team in the list.

### Implementation
In `ChartCustomTooltip.tsx` we'll need to add a `onActive` optional callback which will be set only for `src/components/v5/frame/ColonyHome/partials/ReputationChart/partials/ChartTooltip.tsx`

Probably the most straight forward change will be to add an `useEffect` in `ChartCustomTooltip.tsx`

```
useEffect(() => {
 if (isVisible && onActive) {
  onActive();
}
}, [isVisible, onActive]);
```

But feel free to choose the implementation that makes more sense 😄 

The design change for the legend item is related to changing the color
![Image](https://github.com/user-attachments/assets/5e15a38d-cbac-4159-a82a-6784ac356be1)

[Figma link](https://www.figma.com/design/gac6xVgGNCAge5Bia933dp/Dashboard?node-id=4878-89693&node-type=frame&t=Ulm1ehb6c6R4OttL-0)

PR #3152, opened 2024-09-23
[Dashboard] Add legend item active state on non-empty chart hover

mmioana:
Feat: Add legend item active state on non-empty chart hover

## Description

Hovering over a section of the chart:
- [x] It should show the hover state of the section. 
- [x] It should highlight the team in the list.

[Figma link](https://www.figma.com/design/gac6xVgGNCAge5Bia933dp/Dashboard?node-id=4878-89693&node-type=frame&t=3xGH3p1iB0FF8qa7-0)

## Testing

TODO: Check the hover state over a section of the chart matches Figma.

* Step 1. Make sure you have data to test against
* Step 2. Go to http://localhost:9091/planex
* Step 2. Either leave the `All teams` or `General` team
* Step 3. Hover over a section of the `Influence by team` chart
* Step 4. Check hover state for the legend item is matching Figma

Screenshot 2024-09-23 at 19 14 24

* Step 5. Select another team
* Step 3. Hover over a section of the `Top contributors` chart
* Step 4. Check hover state for the legend item is matching Figma

Screenshot 2024-09-23 at 19 14 14

## Diffs

**New stuff** ✨

* `ReputationChartContext` and provider
* `onVisible` optional property for the `ChartCustomTooltip`

Resolves [#3146](https://github.com/JoinColony/colonyCDapp/issues/3146)

### DIFF
--- a/src/components/v5/frame/ColonyHome/partials/ChartCustomTooltip/ChartCustomTooltip.tsx
+++ b/src/components/v5/frame/ColonyHome/partials/ChartCustomTooltip/ChartCustomTooltip.tsx
@@ -1,34 +1,13 @@
 import { useTooltipState } from '@nivo/tooltip';
 import { AnimatePresence, motion } from 'framer-motion';
-import React, { type PropsWithChildren, type FC, useEffect } from 'react';
+import React, { type PropsWithChildren, type FC } from 'react';
 
-interface ChartCustomTooltipProps extends PropsWithChildren {
-  onVisible?: (isVisible?: boolean) => void;
-}
+interface ChartCustomTooltipProps extends PropsWithChildren {}
 
 export const ChartCustomTooltip: FC<ChartCustomTooltipProps> = ({
   children,
-  onVisible,
 }) => {
   const { isVisible } = useTooltipState();
-
-  useEffect(() => {
-    return () => {
-      if (onVisible) {
-        onVisible(false);
-      }
-    };
-    // We are not including any deps here as we want to run react only to mounting/unmounting state
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
-
-  useEffect(() => {
-    if (isVisible && onVisible) {
-      onVisible(true);
-    }
-    // We are not including onVisible here as we really only want to react to changes of isVisible
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [isVisible]);
 
   return (
     <AnimatePresence>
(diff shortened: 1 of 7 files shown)


src/components/v5/frame/ColonyHome/partials/ChartCustomTooltip/ChartCustomTooltip.tsx
```diff
@@ -1,13 +1,35 @@
...
+  onVisible?: (isVisible?: boolean) => void;
+}
 
 export const ChartCustomTooltip: FC<ChartCustomTooltipProps> = ({
   children,
+  onVisible,
 }) => {
   const { isVisible } = useTooltipState();
+
+  useEffect(() => {
```
bassgeta, line 15: Would it work to use the `onActiveIdChange` for the pie chart? https://nivo.rocks/pie/ supports this and programatically setting the `activeId` which we could then also add to clicking/hovering the legend item.
I think we can keep the context as it is, but just use the API instead of the tooltip for setting the activeId into the context

bassgeta (changes requested):
Great work on this! It works nicely
image
I just left a comment if we could utilize Nivo's API for this, and control the hovered segment via a prop instead of using `CustomTooltip`

If that ends up a pain, ping me and I will approve :D

src/components/v5/frame/ColonyHome/partials/ChartCustomTooltip/ChartCustomTooltip.tsx
mmioana: Will convert it to draft until we explore this change 🙌

rdig (approved):
Tested and looks like it's rendering as expected. Nicely done.

Screenshot from 2024-09-25 13-03-45
Screenshot from 2024-09-25 13-04-51

rumzledz (approved):
Can confirm that the active style for the chart legends are correct ✅ 

active hover

src/context/ReputationChartContext/ReputationChartContext.ts
```diff
@@ -0,0 +1,20 @@
+import { createContext, useContext } from 'react';
+
+export const ReputationChartContext = createContext<
+  | {
+      setActiveLegendItem: (id: string) => void;
+      resetActiveLegendItem: () => void;
```
bassgeta, line 6: technically we don't need this one anymore, but it may be useful to keep it down the line

bassgeta (approved):
Nice job, this change was actually so simple, nice :sunglasses: 
Hover still works:
image
image

src/context/ReputationChartContext/ReputationChartContext.ts
mmioana: Yeah, I was thinking about removing it, but thought we might make use of resetting the active legend item at some point 😅

PR #3157, opened 2024-09-23
[Dashboard] Top contributors numbers should add up to 100

### LINKED ISSUE #3119: [Dashboard] Top contributors chart values don't add up to 100%
### Issue
We need to check from where this rounding issue is coming from, given the sum of all values shown in the chart don't add up to 100%.

Eg. 
Image
The sum of reputations will add up to 99.97%

### Implementation details
We need to check where the reputation is computed - probably in the `updateContributorsWithReputation` lambda

- [ ] The reputation percentages shown in the chart add up to 100%

chmanie:
## Description

This PR makes sure that percentages add up to 100 if there are fewer than 5 people with reputation in the colony.

Bonus: I also did that for the domains graph :1st_place_medal:

## Testing

Look at this graph within the dashboard

image

The percentages should add up to 100 if there are 4 or less contributors.

## Diffs

**New stuff** ✨

* Add a `adjustPercentagesTo100` function

**Changes** 🏗

* Apply that function to domains and top contributors

Resolves #3119

### DIFF
--- a/src/components/v5/frame/ColonyHome/partials/ReputationChart/utils.ts
+++ b/src/components/v5/frame/ColonyHome/partials/ReputationChart/utils.ts
@@ -24,10 +24,10 @@ export const isThereReputationInDomains = (colonyDomains: Domain[]) => {
   );
 };
 
-export const getNormalisedDomainReputationPercentage = (
-  domain: Domain,
+export const getNormalisedReputationPercentage = (
+  reputation: Domain['reputation'] | number,
   normalisedTotalReputation: number,
-) => (Number(domain.reputation || 0) * 100) / normalisedTotalReputation;
+) => (Number(reputation || 0) * 100) / normalisedTotalReputation;
 
 export const getTeamReputationChartData = (
   allTeams: Domain[],
@@ -63,37 +63,37 @@ export const getTeamReputationChartData = (
     0,
   );
 
-  let topTeams: ReputationChartDataItem[] = domainsWithoutRoot
+  const domainsWithoutRootWithReputationPercentage = domainsWithoutRoot
     // Filter out the domains without reputation in order to not display blank spaces in the chart
-    .filter((domain) => !!domain.reputationPercentage)
-    .slice(0, WIDGET_TEAM_LIMIT)
-    .map((domain) => {
-      return {
-        id: domain.id,
-        label: domain.metadata?.name || '',
-        value: getNormalisedDomainReputationPercentage(
-          domain,
-          normalisedTotalReputation,
-        ),
-        color: getTeamHexColor(domain.metadata?.color),
-      };
-    });
+    .filter((domain) => !!domain.reputationPercentage);
 
-  const topTeamsTotalReputation = topTeams.reduce(
-    (reputation, team) => reputation + team.value,
-    0,
-  );
+  let topTeams: ReputationChartDataItem[] =
+    domainsWithoutRootWithReputationPercentage
+      .slice(0, WIDGET_TEAM_LIMIT)
+      .map((domain) => {
+        return {
+          id: domain.id,
+          label: domain.metadata?.name || '',
+          value: getNormalisedReputationPercentage(
+            domain.reputation,
+            normalisedTotalReputation,
+          ),
+          color: getTeamHexColor(domain.metadata?.color),
+        };
+      });
 
-  /**
-   * To reach the 100% reputation percentage, we exclude the top teams reputation from 100%
-   */
-  const reputationInOtherTeams = 100 - topTeamsTotalReputation;
+  const reputationInOtherTeams = domainsWithoutRootWithReputationPercentage
+    .slice(WIDGET_TEAM_LIMIT)
+    .reduce((reputation, team) => reputation + Number(team.reputation || 0), 0);
 
   if (reputationInOtherTeams > 0) {
     topTeams.push({
       id: 'allOtherTeams',
       label: formatText(MSG.otherLabel),
-      value: reputationInOtherTeams,
+      value: getNormalisedReputationPercentage(
+        reputationInOtherTeams,
+        normalisedTotalReputation,
+      ),
       color: '--color-gray-400',
       shouldTruncateLegendLabel: false,
     });
--- a/src/components/v5/frame/ColonyHome/partials/ReputationChart/utils.ts
+++ b/src/components/v5/frame/ColonyHome/partials/ReputationChart/utils.ts
@@ -3,6 +3,7 @@ import { defineMessages } from 'react-intl';
 
 import { type Domain } from '~types/graphql.ts';
 import { formatText } from '~utils/intl.ts';
+import { adjustPercentagesTo100 } from '~utils/numbers.ts';
 import { getTeamHexColor } from '~utils/teams.ts';
 
 import { CONTRIBUTORS_COLORS_LIST } from './consts.ts';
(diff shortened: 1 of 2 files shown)


src/utils/numbers.ts
```diff
@@ -21,3 +21,34 @@ export const convertEnotationToNumber = (value: string) => {
...
+
+/**
+ * Adjusts an array of percentages so that they sum up to exactly 100% after rounding to the specified number of decimal places.
+ * It does this by calculating the rounding difference and adding it to the largest percentage value.
+ *
+ * @param percentages - An array of percentage values.
+ * @param decimals - The number of decimal places to round to.
+ * @returns A new array of percentages adjusted to sum up to 100%.
+ */
+export const adjustPercentagesTo100 = (
```
mmioana, line 33: Loving this helper 🙌

src/components/v5/frame/ColonyHome/partials/ReputationChart/utils.ts
```diff
@@ -126,7 +152,7 @@ export const getContributorReputationChartData = (
     return [
       ...topContributors,
       {
-        id: 'allOtherTeams',
+        id: 'allOtherUsers',
```
mmioana, line 155: 👍

src/components/v5/frame/ColonyHome/partials/ReputationChart/utils.ts
```diff
@@ -107,16 +120,29 @@ export const getTeamReputationChartData = (
...
+    const adjustedValues = adjustPercentagesTo100(
+      topContributors.map((contributor) => contributor.value),
+      // getNormalisedDomainReputationPercentage rounds to two decimals
+      2,
+    );
+    topContributors = topContributors.map((contributor, idx) => ({
+      ...contributor,
+      value: adjustedValues[idx],
+    }));
+  }
```
mmioana, line 145: The problem with this condition is that the `topContributors` always has a max length of `WIDGET_TEAM_LIMIT`, so we can end up normalising the values for the `topContributors`, but there is still reputation remaining in the remaining contributors, leading to this situation where the values add up to `104.33%`

Screenshot 2024-09-24 at 10 29 40

So, maybe it will make sense to apply `adjustPercentagesTo100` to the entire returned array

src/components/v5/frame/ColonyHome/partials/ReputationChart/utils.ts
```diff
@@ -78,6 +79,18 @@ export const getTeamReputationChartData = (
...
+      topTeams.map((team) => team.value),
+      // getNormalisedDomainReputationPercentage rounds to two decimals
+      2,
+    );
+    topTeams = topTeams.map((team, idx) => ({
+      ...team,
+      value: adjustedValues[idx],
+    }));
+  }
+
```
mmioana, line 93: For the teams, again a similar situation will happen. The max length of `topTeams` will be `WIDGET_TEAM_LIMIT`, so we'll end up normalising the reputation for them to add up to `100%`.
However for the teams, the way it's implemented right now, the remaining reputation is determined by adding up the `topTeams` reputation and subtracting it from `100`.
By calling `adjustPercentagesTo100`, the remaining reputation will always be `0`.

So, I believe the best solution will be to refactor the computation for the remaining reputation (only for consistency purposes)
```
const reputationInOtherTeams = domainsWithoutRoot
    .filter((domain) => !!domain.reputationPercentage)
    .slice(WIDGET_TEAM_LIMIT)
    .reduce((reputation, team) => reputation + Number(team.reputation), 0);
```
and then apply `adjustPercentagesTo100` to the entire returned value

src/components/v5/frame/ColonyHome/partials/ReputationChart/utils.ts
```diff
@@ -107,16 +120,29 @@ export const getTeamReputationChartData = (
...
+      // getNormalisedDomainReputationPercentage rounds to two decimals
+      2,
+    );
+    topContributors = topContributors.map((contributor, idx) => ({
+      ...contributor,
+      value: adjustedValues[idx],
+    }));
+  }
 
   const reputationInOtherContributors = contributorsList
```
mmioana, line 147: I know it wasn't done by you, but now that I have noticed, maybe we can rename the current value in the `reduce` function from `team` to `contributor`

mmioana (changes requested):
Hey @chmanie thank you for picking this issue up 🙌 

Really love the helper function you created, but I noticed a bug while testing, so I left some comments, if you could please have a look over them 🙏

src/components/v5/frame/ColonyHome/partials/ReputationChart/utils.ts
chmanie: Will do, I did it in other places but must have missed this one :)

src/components/v5/frame/ColonyHome/partials/ReputationChart/utils.ts
chmanie: Ah yeah, good point, I'll just apply that to the whole array then.

chmanie: Ah, actually, it was just a typo. I meant to use `contributorsList.length`

src/components/v5/frame/ColonyHome/partials/ReputationChart/utils.ts
```diff
@@ -89,50 +90,69 @@ export const getTeamReputationChartData = (
   const reputationInOtherTeams = 100 - topTeamsTotalReputation;
```
mmioana, line 90: I still believe this line should be refactored into
```
const reputationInOtherTeams  = domainsWithoutRoot
    .slice(WIDGET_TEAM_LIMIT)
    .reduce(
      (reputation, team) => reputation + Number(team.reputation || 0),
      0,
    );
```
Because with this implementation I artificially considered there might be some reputation in the other domains if the `topTeamsTotalReputation` isn't `100%`

And then your call to adjust the values inside `getTeamReputationChartData` will not really have an effect. 
What do you think?

mmioana (approved):
Great work on this one and thank you for applying my suggestions @chmanie 🙌 

Now the values add perfectly up to `100%`
Screenshot 2024-09-24 at 20 06 59
Screenshot 2024-09-24 at 20 07 04
Screenshot 2024-09-24 at 20 07 10

I did leave another comment, but feel free to tackle it as you see fit 👍 
I will approve it either way

src/components/v5/frame/ColonyHome/partials/ReputationChart/utils.ts
chmanie: Yes, let's do it, seems more proper

rdig (approved):
All good here. Nicely done Chris! 

Screenshot from 2024-09-25 13-12-26

Everything seems to add up to `100`: `90.93 + 5.45 + 1.81 + 1.81 = 100` 

Screenshot from 2024-09-25 13-13-12

src/utils/numbers.ts
```diff
@@ -21,3 +21,34 @@ export const convertEnotationToNumber = (value: string) => {
...
+  // Round each value to the nearest integer
+  const roundedValues = scaledValues.map((value) => Math.round(value));
+  // Sum the rounded values
+  const sumRounded = roundedValues.reduce((acc, val) => acc + val, 0);
+  // Calculate the difference to reach the total
+  const delta = total - sumRounded;
+  // Find the index of the largest value in the original array
+  const maxIndex = roundedValues.indexOf(Math.max(...roundedValues));
+  // Adjust the largest value by adding the delta
+  roundedValues[maxIndex] += delta;
```
Nortsova, line 51: wow. pretty interesting solution! thanks! 🤩

Nortsova (approved):
Thank you @chmanie for the fix ✨

image
image

It seems that the calculation of all items will total 100%. Great work 👏

src/utils/numbers.ts
chmanie: It was @mmioana's idea!

mmioana: 😈

PR #3163, opened 2024-09-25
build: update colony package versions

bassgeta:
## Description

Update versions of `colony-js`, `colony-events` and `colony-abis` to their latest releases which support streaming payments, multi-sig and many other goodies :candy: 

[Block ingestor PR](https://github.com/JoinColony/block-ingestor/pull/270/)

## Testing

You can just make sure everything works

## Diffs

**Changes** 🏗

* Updated package versions


rdig (approved):
All good here.

PR #2761, opened 2024-07-24
New KYC `pending` status

jakubcolony:
## Description

Please refer to issue for context.

This PR improves the UX around the KYC verification by adding an intermediate status between the user completing the KYC flow and Bridge receiving a response. It cleans up some types and todos and introduces a GQL enum for the statuses.

**Important**: You will need to add a new variable to `.env.secrets.local`: `PERSONA_ENVIRONMENT_ID`, and run `npm run prepare`. 

## Testing

When Persona notifies of the flow being complete, a flag is set on the user profile which then overrides the status in the lambda handler. 

To avoid having to complete the entire KYC flow, you can apply the following patch to mimic the behaviour when you cancel the KYC flow as well. Save it in a file called `diff` and run `git apply diff` in the CDapp directory:

```
diff --git a/src/components/frame/v5/pages/UserCryptoToFiatPage/partials/Verification/Verification.tsx b/src/components/frame/v5/pages/UserCryptoToFiatPage/partials/Verification/Verification.tsx
index 99cbfe44e..b013548fc 100644
--- a/src/components/frame/v5/pages/UserCryptoToFiatPage/partials/Verification/Verification.tsx
+++ b/src/components/frame/v5/pages/UserCryptoToFiatPage/partials/Verification/Verification.tsx
@@ -53,6 +53,18 @@ const Verification: FC<CryptoToFiatPageComponentProps> = ({
           },
         });
 
+        refetchStatus();
+      },
+      async onCancel() {
+        await updateProfile({
+          variables: {
+            input: {
+              id: user?.walletAddress ?? '',
+              hasCompletedKYCFlow: true,
+            },
+          },
+        });
+
         refetchStatus();
       },
     });

```

In `amplify/backend/function/bridgeXYZMutation/src/index.js` replace `apiKey` with sandbox API key (ping me if you don't have it) and start the KYC verification. When you get to the KYC modal close it at any point. The status should refresh. In sandbox, the KYC is automatically approved however in production the lambda handler would return `pending` until KYC is processed.

Resolves #2751

PR #3184, opened 2024-09-26
fix: delete old bank account only _after_ creating a new one

### LINKED ISSUE #2963: [C2F] Failed form submission clears out existing banking details & defaults the UI
## Description

Within the banking details modal, when I successfully submit my bank details, the details show in the UI and the button updates correctly.

However, if I select 'update details' and then submit the form with an error, all of my existing details in the UI and the form are cleared. The CTA button updates to the default 'Add state' and the bank details are cleared. 

Errors shouldn't override details. A failed form submission shouldn't clear out existing details. 

** Using EU details identified this issue but should not be present across all forms in C2F or global. 

image

## Steps to reproduce

1. Complete the bank details modal and submit.
2. The UI and CTA should populate and change labels once the skeleton has loaded the information.
3. Select the 'update details' button
4. Add the missing information in, but intentionally create errors in the inputs.
5. Select the CTA and the C2F UI will go back to the default state under the modal, and all information within the banking section will be cleared the CTA changed. 

## Expected behaviour

* Errors shouldn't override details. A failed form submission shouldn't clear out existing details.

chmanie:
## Description

This PR fixes the bug where bank accounts where deleted before a new one was created, resulting in deleting _all_ bank accounts when there was an error.

It's done by first getting _all_ bank accounts for that customer, then creating a new account, then deleting all of the old ones (most of the time that should be just a single one).

## Testing

Testing this is _hard_, so bring some time.

* First, follow the steps lined out in [this PR](https://github.com/JoinColony/colonyCDapp/pull/2761), to be able to do and bail out of the KYC flow
* Then, create a bank account using USD
* Then, update this bank account to use EUR instead of USD and fill all the forms (EUR will not work in the sandbox and produce an error similar to the one in the issue)
* Then see that the bank details are not deleted

## Diffs

**New stuff** ✨

* Add a `getExternalAccounts` helper method in the `bridgeXYZMutation`

**Changes** 🏗

* Create a new bank account first, then delete old ones
* Delete _all_ old accounts should there be more than one

Resolves #2963

### DIFF
--- a/amplify/backend/function/bridgeXYZMutation/src/handlers/updateExternalAccount.js
+++ b/amplify/backend/function/bridgeXYZMutation/src/handlers/updateExternalAccount.js
@@ -1,6 +1,6 @@
 const fetch = require('cross-fetch');
 const { graphqlRequest } = require('../utils');
-const { createExternalAccount, getLiquidationAddresses, getExternalAccounts } = require('./utils');
+const { createExternalAccount, deleteExternalAccount, getLiquidationAddresses, getExternalAccounts } = require('./utils');
 
 const { getUser } = require('../graphql');
 
@@ -41,23 +41,8 @@ const updateExternalAccountHandler = async (
     account,
   );
 
-  const deletePromises = externalAccounts.map(async ({ id }) => {
-    const deleteAccountRes = await fetch(
-      `${apiUrl}/v0/customers/${bridgeCustomerId}/external_accounts/${input.id}`,
-      {
-        headers: {
-          'Api-Key': apiKey,
-        },
-        method: 'DELETE',
-      },
-    );
-
-    if (deleteAccountRes.status !== 200) {
-      console.log(
-        `Error deleting external account: ${await deleteAccountRes.text()}`,
-      );
-    }
-  });
+  const deletePromises = externalAccounts.map(async ({ id }) =>
+    deleteExternalAccount(apiUrl, apiKey, bridgeCustomerId, id));
 
   await Promise.all(deletePromises);
 
--- a/amplify/backend/function/bridgeXYZMutation/src/handlers/updateExternalAccount.js
+++ b/amplify/backend/function/bridgeXYZMutation/src/handlers/updateExternalAccount.js
@@ -1,6 +1,6 @@
 const fetch = require('cross-fetch');
 const { graphqlRequest } = require('../utils');
-const { createExternalAccount, getLiquidationAddresses } = require('./utils');
+const { createExternalAccount, getLiquidationAddresses, getExternalAccounts } = require('./utils');
 
 const { getUser } = require('../graphql');
 
@@ -32,22 +32,7 @@ const updateExternalAccountHandler = async (
 
   const bridgeCustomerId = graphQlData?.getUser?.bridgeCustomerId;
 
-  const deleteAccountRes = await fetch(
-    `${apiUrl}/v0/customers/${bridgeCustomerId}/external_accounts/${input.id}`,
-    {
-      headers: {
-        'Api-Key': apiKey,
-      },
-      method: 'DELETE',
-    },
-  );
-
-  // Exit if deleting account fails to avoid creating multiple accounts
-  if (deleteAccountRes.status !== 200) {
-    throw Error(
-      `Error deleting external account: ${await deleteAccountRes.text()}`,
-    );
-  }
+  const externalAccounts = await getExternalAccounts(apiUrl, apiKey, bridgeCustomerId);
 
   const newAccount = await createExternalAccount(
     apiUrl,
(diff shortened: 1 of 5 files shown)


amplify/backend/function/bridgeXYZMutation/src/handlers/updateExternalAccount.js
```diff
@@ -56,6 +41,26 @@ const updateExternalAccountHandler = async (
...
+        method: 'DELETE',
+      },
+    );
+
+    if (deleteAccountRes.status !== 200) {
+      console.log(
+        `Error deleting external account: ${await deleteAccountRes.text()}`,
+      );
+    }
+  });
```
mmioana, line 60: Would it make sense to extract also this into an utility function `deleteExternalAccount`?

```
const deleteExternalAccount = async (apiUrl, apiKey, bridgeCustomerId, id) => {
  const deleteAccountRes = await fetch(
    `${apiUrl}/v0/customers/${bridgeCustomerId}/external_accounts/${id}`,
    {
      headers: {
        'Api-Key': apiKey,
      },
      method: 'DELETE',
    },
  );

  if (deleteAccountRes.status !== 200) {
    console.log(
      `Error deleting external account: ${await deleteAccountRes.text()}`,
    );
  }
}
```

And then use it here

```suggestion
  const deletePromises = externalAccounts.map(({ id }) => 
     deleteExternalAccount(apiUrl, apiKey, bridgeCustomerId, id)
  );
```

amplify/backend/function/bridgeXYZMutation/src/handlers/updateExternalAccount.js
```diff
@@ -56,6 +41,26 @@ const updateExternalAccountHandler = async (
     account,
   );
 
+  const deletePromises = externalAccounts.map(async ({ id }) => {
+    const deleteAccountRes = await fetch(
+      `${apiUrl}/v0/customers/${bridgeCustomerId}/external_accounts/${input.id}`,
```
mmioana, line 46: And btw, is there a reason why we don't actually use this `id` in the request URL but use the `input.id`?

mmioana (changes requested):
Nice and simple solution @chmanie 🥇 

Though I think for the testing steps you should add this git patch, otherwise applying the patch from the other PR will throw an error

```
diff --git a/src/components/frame/v5/pages/UserCryptoToFiatPage/partials/Verification/Verification.tsx b/src/components/frame/v5/pages/UserCryptoToFiatPage/partials/Verification/Verification.tsx
index 5bb55ece9..dd606a473 100644
--- a/src/components/frame/v5/pages/UserCryptoToFiatPage/partials/Verification/Verification.tsx
+++ b/src/components/frame/v5/pages/UserCryptoToFiatPage/partials/Verification/Verification.tsx
@@ -51,6 +51,18 @@ const Verification = () => {
           },
         });
 
+        refetchKycData();
+      },
+      async onCancel() {
+        await updateProfile({
+          variables: {
+            input: {
+              id: user?.walletAddress ?? '',
+              hasCompletedKYCFlow: true,
+            },
+          },
+        });
+
         refetchKycData();
       },
     });

```

Anyhow, can confirm having the bank details for the KYC and trying to add an `EUR` account will fail, but the previous bank details will not be removed 🙌 

Previous `USD` bank details
Screenshot 2024-09-26 at 17 54 58
Adding an `EUR` account
Screenshot 2024-09-26 at 17 55 23
Screenshot 2024-09-26 at 17 55 27
Previous bank details remain
Screenshot 2024-09-26 at 17 55 31
Updating the bank details for `USD` still works as expected
Screenshot 2024-09-26 at 18 08 34
Screenshot 2024-09-26 at 18 08 56

Left some minor refactoring comment, but awesome job 🙌

amplify/backend/function/bridgeXYZMutation/src/handlers/updateExternalAccount.js
chmanie: Oh wow good catch that's an oversight. I wonder why this was even working.

jakubcolony (approved):
> First, follow the steps lined out in https://github.com/JoinColony/colonyCDapp/pull/2761, to be able to do and bail out of the KYC flow

You can get away without doing this step if you refresh the page once you get to Persona popup.

----
Looks great @chmanie, thanks for picking it up! 💯 

Failed bank details update no longer clears the previous details:

https://github.com/user-attachments/assets/7cdfe8f2-df5c-48ce-ad0e-01c3e25b0a0f

Also checked updating to a different USD account which worked without issues.

amplify/backend/function/bridgeXYZMutation/src/handlers/updateExternalAccount.js
mmioana: I guess that as there was an external account already created, the `input.id` was holding the correct external account `id` 🤔

mmioana (approved):
Awesome work @chmanie thanks for picking up my comment 🙌 

Can confirm getting an error while filling up the `EUR` form, will not delete the old data 🎉 

Screenshot 2024-09-30 at 13 34 32
Screenshot 2024-09-30 at 13 34 59
Screenshot 2024-09-30 at 13 35 31
Screenshot 2024-09-30 at 13 35 34

Previous data still shows up
Screenshot 2024-09-30 at 13 35 41

And also can update the `USD` account and the changes are reflected in the UI
Screenshot 2024-09-30 at 13 36 45
Screenshot 2024-09-30 at 13 36 51

jakubcolony (approved):

PR #3185, opened 2024-09-26
Hotfix: Use uuid as idempotency key when creating liquidation address

jakubcolony:
## Description

Bridge lambda used the external account ID as the idempotency key when creating a linked liquidation address to prevent it being created twice (which happened in local dev, since each useEffect runs twice). 

However, as a result when the create address call fails, it cannot be retried due to the idempotency key having been used already.

This PR changes the idempotency key to a random uuid as in other calls to Bridge API.

## Testing

Sadly the instructions below do not work with Sandbox which doesn't complain about reusing idempotency keys. I ran those against production and it fixes the issue.

~~To test, we'll need to simulate a failed Bridge API call and then retry it.~~

~~1. Create an `.env` file inside `bridgeXYZMutation` directory and copy over the values from `.env.example`.~~
> [!IMPORTANT]
> Let me know if you need the API key for Bridge sandbox.

~~2. Complete KYC, but don't add any bank details yet.~~
> [!IMPORTANT]
> Try to use a "fresh" email address, otherwise you may already have bank details stored against your customer.

~~3. Apply the following patch (save as a file, then run `git apply <filename>`). It makes the payload of create liquidation address call invalid and it'll be rejected by Bridge.~~

```
diff --git a/amplify/backend/function/bridgeXYZMutation/src/handlers/checkKyc.js b/amplify/backend/function/bridgeXYZMutation/src/handlers/checkKyc.js
index e494ec419..f72ed0a20 100644
--- a/amplify/backend/function/bridgeXYZMutation/src/handlers/checkKyc.js
+++ b/amplify/backend/function/bridgeXYZMutation/src/handlers/checkKyc.js
@@ -167,7 +167,7 @@ const checkKYCHandler = async (
           },
           method: 'POST',
           body: JSON.stringify({
-            chain: 'arbitrum',
+            // chain: 'arbitrum',
             currency: 'usdc',
             external_account_id: firstAccount.id,
             destination_payment_rail:

``` 

~~4. Go back to C2F account area and add USD bank details (EUR is not supported in sandbox). You can use the following test details:~~
```
Account Number: 68470390
Routing Number: 211149530
```

~~Notice no liquidation address shows up.~~

~~5. Revert the changes applied with the patch and refresh the page.~~

~~The liquidation address should show up.~~

~~On `master`, the retried calls would fail due to using the same idempotency key, and the flow would be stuck.~~


amplify/backend/function/bridgeXYZMutation/.env.example
```diff
@@ -0,0 +1,2 @@
+BRIDGE_API_URL=https://api.sandbox.bridge.xyz
+BRIDGE_API_KEY=
```
jakubcolony, line 2: `.env` files inside lambdas indeed work out of the box, thanks for the heads up @davecreaser!

rdig (approved):
This can only be actually tested in production, while actually checking a live account.

This was already deployed successfully yesterday and tested, so all good here!

bassgeta (approved):
Since it's been deployed and it works, I'll do my part and send it off with a green checkmark :saluting_face:

iamsamgibbs (approved):
Changes look sensible and I believe everyone who has said it works.

Issue #3121, opened 2024-09-18
Make avatars on team cards clickable


arrenv:
Teamwork task: [Make avatars on team cards clickable](https://tw.auroracreation.com/#/tasks/16083206)

## Description

This is a small UX improvement that has come up a couple of times and seems like an easy fix to reduce frustration and increase expected behaviour of elements in the app.

On a few reported occasions it has been observed and commented on that the avatar icons should be clickable and work to filter out the list of members. Specifically on the Teams page, there is a list of member avatars that currently are not clickable at all.

![image](https://github.com/user-attachments/assets/9202fcca-53a7-4fdf-a51e-a43ecaa54f7e)

## Acceptance criteria
- Member avatar list in the teams cards should be clickable.
    - Clicking on the avatars should navigate the user to the Members > Contributor page with the relevant team filter applied.
    - The hover state of the icons should show a highlighted border on individual avatars using the primary blue/400 color token.

PR #3198, opened 2024-09-27
feat: make avatars clickable

adam-strzelec:
## Description
Make avatars on teams page clickable
Screenshot 2024-09-27 at 13 21 20

## Testing
- Go to the `/{colony_name}/teams` page
- Click the avatars on teams card
- After click, You should be redirected to contributors page with the relevant team filter applied

## Diffs

**New stuff** ✨

* New stuff goes here

Add link for avatars inside `TeamCard` component

Resolves [#31415](https://github.com/JoinColony/colonyCDapp/issues/3121)

### DIFF
--- a/src/components/v5/common/TeamCard/TeamCard.tsx
+++ b/src/components/v5/common/TeamCard/TeamCard.tsx
@@ -27,6 +27,7 @@ const TeamCard: FC<TeamCardProps> = ({
   meatBallMenuProps,
   links,
   balance,
+  searchParams,
   className,
 }) => {
   const isMobile = useMobile();
@@ -78,11 +79,11 @@ const TeamCard: FC<TeamCardProps> = ({
             <Link
               to={{
                 pathname: `/${colonyName}/${COLONY_CONTRIBUTORS_ROUTE}`,
-                search: `${links?.[0]?.to.search}`,
+                search: `${searchParams?.team}`,
               }}
             >
               {!!members?.length && (
-                <div className="ml-auto flex-shrink-0">
+                <div className="group ml-auto flex-shrink-0">
                   <UserAvatars
                     maxAvatarsToShow={4}
                     className="[&_.placeholder]:bg-gray-200 [&_.placeholder]:text-gray-900"
--- a/src/components/v5/common/TeamCard/TeamCard.tsx
+++ b/src/components/v5/common/TeamCard/TeamCard.tsx
@@ -2,7 +2,9 @@ import { Cardholder } from '@phosphor-icons/react';
 import clsx from 'clsx';
 import React, { type FC } from 'react';
 
+import { useColonyContext } from '~context/ColonyContext/ColonyContext.ts';
 import { useMobile } from '~hooks/index.ts';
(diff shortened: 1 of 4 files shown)


src/components/v5/common/TeamCard/TeamCard.tsx
```diff
@@ -70,7 +75,12 @@ const TeamCard: FC<TeamCardProps> = ({
               />
             </div>
           ) : (
-            <>
+            <Link
+              to={{
+                pathname: `/${colonyName}/${COLONY_CONTRIBUTORS_ROUTE}`,
+                search: `${links?.[0]?.to.search}`,
```
mmioana, line 81: I would suggest refactoring this line, as the `links` might update and is not a reliable source of truth for the parameter we want to use

So my suggestion would be to add an extra property to the `TeamCardProps` either `nativeId` or `searchParams`
and then update the `useTeams` hook to include either the `nativeId` or 
```
searchParams: {
  team: `${TEAM_SEARCH_PARAM}=${nativeId}`
}
```

and then use it here either as
`search: `?${TEAM_SEARCH_PARAM}=${nativeId}`,`
or
`search: `?${searchParams.team}`,`

mmioana (changes requested):
Nice start on this PR @adam-strzelec 🎉 

Everything works as expected, but I would appreciate if you can tackle my comment 🙏 

https://github.com/user-attachments/assets/f5f556f3-b934-4bf6-a3d4-8e7d68f29973

iamsamgibbs (changes requested):
Nice job on this @adam-strzelec ! Can confirm the avatars are linking to the correct page with the correct team filter. :)

https://github.com/user-attachments/assets/c0a9a9ff-bdd7-47b9-a8a3-e42adc2247ef

However, this part of the original issue has been missed:

> The hover state of the icons should show a highlighted border on individual avatars using the primary blue/400 color token.

When the component is hovered, the border on each of the icons should change to blue-400:

Screenshot 2024-09-30 at 10 10 49

Please also address Ioana's code suggestion.

src/components/v5/common/TeamCard/TeamCard.tsx
adam-strzelec: @mmioana done 🙂

iamsamgibbs (changes requested):
Thanks Adam! Can you also add the `transition-colors` class so that the border transitions in the same way as the other hover states on the site?

https://github.com/user-attachments/assets/f2b28c14-0244-476c-af26-3bfc2049abe6

https://github.com/user-attachments/assets/be94cc02-7f03-4931-be84-09b630b98200

iamsamgibbs (approved):
Nice one, this looks great now:

https://github.com/user-attachments/assets/1f57abd5-84bb-4b60-8bdb-a0c3a10a4bf2

mmioana (approved):
Thank you for your changes and continuous effort @adam-strzelec 🥇 

It all looks great now and behaves as expected 🎉 

https://github.com/user-attachments/assets/6a0c6724-1616-44c3-8f7a-f75f1d642e0c

Issue #3047, opened 2024-09-02
[New Navigation] PageHeader Component UI Adustments


rumzledz:
## Issues

* [ ] Issue 1
![Image](https://github.com/user-attachments/assets/48f1bf87-4856-4473-b3c5-43916a4b0bee)

* [ ] Issue 2
![Image](https://github.com/user-attachments/assets/6b3c9fc9-8068-427d-859b-b5374228a5e6)

PR #3200, opened 2024-09-27
Maintenance: Do not clear existing search params when adding a new one

### LINKED ISSUE #3176: [Dashboard] Newly created transaction remove previous query param
## Steps to reproduce
query-string-cleaned

1. Navigate to the dashboard and select the team (domain id = 2) 
2. Create any transaction
3. After a new transaction is created it has the "tx={hash}" query string
4. And this new query string removed the "team=2" query
5. This means after new transaction is created and the motion modal is closed you would be redirected to the general, not selected domain 

## Expected behaviour
- Selected team and query string for team should not disapear

## Actual behaviour
- New tx=hash rewrite previous team=2 param

## Suggestions (optional)
Instead of navigating and creating new URL we can just use URLSearchParams to add new param, and we can unify this solution for all actions:
image

rumzledz:
## Description

This PR removes the redundant navigation action whenever a saga completes successfully.

An example of this redundancy is:

```js
// In src/redux/sagas/actions/editDomain.ts

// This already has the logic needed to update the URL with the new search param before proceeding to navigate to it
setTxHash?.(txHash);

// Redundant and removed in this PR
navigate(`/${colonyName}?tx=${txHash}`, {
  state: { isRedirect: true },
});
```

The function `setTxHash` is passed in as a meta param via `setActionFormProps`, and you can find this in `useActionFormProps.ts`

```js
withMeta({
  setTxHash: (txHash: string) => {
    navigate(setQueryParamOnUrl(window.location.pathname, 'tx', txHash), {
      replace: true,
    });
  },
});
```

**AND OH YEAH BEFORE I FORGET**, the ACTUAL reason why this PR exists is to make sure that new search params added do not overwrite the existing ones:

persist params

## Testing

1. Visit the Colony Dashboard
2. Select Team Andromeda
3. Verify that the URL is now http://localhost:9091/planex?team=2
4. Create and submit a Simple Payment
5. Verify that the URL is http://localhost:9091/planex?team=2&tx={txHash}. Basically, the new search param should just be appended

## Diffs

**Changes** 🏗

- Now using `useSearchParams` when updating the search params after a successful transaction
- Slight tweaks to the `setQueryParamOnUrl` function so that it uses an "update as required" strategy instead of replace and it can now also accept multiple search params as a function argument

**Deletions** ⚰️

* Removed the redundant navigation step whenever a saga completes successfully

Resolves #3176

### DIFF
--- a/src/components/v5/common/ActionSidebar/hooks/useActionFormProps.ts
+++ b/src/components/v5/common/ActionSidebar/hooks/useActionFormProps.ts
@@ -1,11 +1,11 @@
 import { useCallback, useRef, useState } from 'react';
-import { useNavigate } from 'react-router-dom';
+import { useNavigate, useSearchParams } from 'react-router-dom';
 
 import { type Action } from '~constants/actions.ts';
 import { ActionTypes } from '~redux/index.ts';
+import { TX_SEARCH_PARAM } from '~routes';
 import { type ActionFormProps } from '~shared/Fields/Form/ActionForm.tsx';
 import { mapPayload, pipe, withMeta } from '~utils/actions.ts';
-import { setQueryParamOnUrl } from '~utils/urls.ts';
 
 import {
   ACTION_BASE_VALIDATION_SCHEMA,
@@ -26,6 +26,7 @@ const useActionFormProps = (
     mode: 'onChange',
     validationSchema: ACTION_BASE_VALIDATION_SCHEMA,
   });
+  const [searchParams, setSearchParams] = useSearchParams();
 
   const getFormOptions = useCallback<ActionFormBaseProps['getFormOptions']>(
     async (formOptions, form) => {
@@ -57,16 +58,8 @@ const useActionFormProps = (
                 })),
                 withMeta({
                   setTxHash: (txHash: string) => {
-                    navigate(
-                      setQueryParamOnUrl(
-                        window.location.pathname,
-                        'tx',
-                        txHash,
-                      ),
-                      {
-                        replace: true,
-                      },
-                    );
+                    searchParams.set(TX_SEARCH_PARAM, txHash);
+                    setSearchParams(searchParams);
                   },
                 }),
               ),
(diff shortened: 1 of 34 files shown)


src/components/v5/common/ActionSidebar/hooks/useActionFormProps.ts
mmioana, line 1: Cool cool 💯

mmioana (approved):
Thank you @rumzledz for fixing this issue! 10/10 🥇 

Can confirm the previous navigation to `${colonyName}?tx=${txHash}` was replaced
Screenshot 2024-09-27 at 18 29 53
with actually setting the `tx` search param
Screenshot 2024-09-27 at 18 30 34

Screenshot 2024-09-27 at 18 31 03
Screenshot 2024-09-27 at 18 31 42
Screenshot 2024-09-27 at 18 31 50

https://github.com/user-attachments/assets/5834db77-722b-4af9-a310-b35f38cf1f15


Let's merge this 💯

iamsamgibbs (approved):
Great job tidying this up, it works so much more nicely now.

https://github.com/user-attachments/assets/74ab0de0-e697-4ea9-a4c6-9f10bf411974

https://github.com/user-attachments/assets/f243bf74-47fb-4309-8457-79bd90359ae4

Nice one removing all the unused navigate functions.

rdig (approved):
All good. Nice way of dealing with the ever more complicated URL composition.

A side note, that this will most likely have to change to accommodate proxy colonies, but we'll deal with that when we get to it

Nicely done! :100: 

Screenshot from 2024-09-30 14-24-08
Screenshot from 2024-09-30 14-24-14
Screenshot from 2024-09-30 14-25-18
Screenshot from 2024-09-30 14-25-05
Screenshot from 2024-09-30 14-25-13

A thing that I noticed, don't know it if was raised before, but the element for the hamburger menu is no longer round, it's now a rounded square(ish)

Screenshot from 2024-09-30 14-23-59

Nortsova:
Thank you, @rumzledz, for this refactoring! Amazing work 🙌

Nortsova (approved):
Sep-30-2024 13-47-12

Thank you @rumzledz for your work! Now it works as expected and code is much cleaner 🙌

rumzledz:
@rdig 

> A side note, that this will most likely have to change to accommodate proxy colonies, but we'll deal with that when we get to it

Agreed! 👌 

> A thing that I noticed, don't know it if was raised before, but the element for the hamburger menu is no longer round, it's now a rounded square(ish)

This was actually a required change that was done to fix issue no. 2 of #3047 
image

PR #3202, opened 2024-09-30
[Notifications] feat: do not show unread counters and the notification tab if user has notifications disabled

### LINKED ISSUE #3168: Notifications: Handle notifications disabled
### Story

Example: As a *user who has notifications disabled globally*, I want to *not see the notifications tab in the user hub*, so that *I am not confused*

### Description

When a user disabled notifications, we should not attempt any connection with Magicbell inside NotificationContextProvider, and we should not show the notifications tab inside the user hub.

- [ ] No calls to Magicbell are made if the user has notifications turned off globally
- [ ] The notifications tab doesn't show for any users who have notifications turned off globally

bassgeta:
## Description

This PR achieves what the title says by splitting up the `NotificationsContext` into two parts: `NotificationsUserContext` for user's configured data regarding notifications such as muted colonies and if notification are even enabled and `NotificationsDataContext` for reading data from MagicBell and interacting with it.
For ease of use I've created a component called `NotificationsEnabledWrapper` that renders any component conditionally if the user has notifications enabled.

## Testing

Should be relatively straight forward!
In case you haven't tested any other notification PRs add the following to your `.env.local.secrets` and run `npm run prepare`

```
MAGICBELL_API_KEY=<ask somebody for this>
MAGICBELL_API_SECRET=<ask somebody for this>
MAGICBELL_DEV_KEY=
```

1. Login as `leela`, create some simple payment actions paying `leela`
image
2. Notice that you get a notification dot on the user hub button (needs a few seconds)
image
3. Verify that when you open the user hub you get an unread notification count of 2 and you can open the tab
image
4. Go to your settings under the advanced tab (`http://localhost:9091/account/advanced`) and disable your notifications
image
5. Go back to `/planex` and verify that you can't see the notification dot anymore
image
6. Open the user hub and verify that you don't see the notifications tab 
image

## Diffs

**New stuff** ✨

* `NotificationsEnabledWrapper` which conditionally renders a component if notifications are enabled

**Changes** 🏗

* Split `NotificationsContext` into `NotificationsUserContext` for user's config data and `NotificationsDataContext` for interacting with notifications
* Split the notification dot into its own component

Resolves  #3168

### DIFF
--- a/src/components/common/Extensions/UserHub/UserHub.tsx
+++ b/src/components/common/Extensions/UserHub/UserHub.tsx
@@ -3,20 +3,21 @@ import React, { type FC, useState, useContext } from 'react';
 import { defineMessages } from 'react-intl';
 
 import { FeatureFlagsContext } from '~context/FeatureFlagsContext/FeatureFlagsContext.ts';
-import { useNotificationsContext } from '~context/NotificationsContext/NotificationsContext.ts';
+import { useNotificationsUserContext } from '~context/Notifications/NotificationsUserContext/NotificationsUserContext.ts';
 import CryptoToFiatContextProvider from '~frame/v5/pages/UserCryptoToFiatPage/context/CryptoToFiatContextProvider.tsx';
 import { useMobile } from '~hooks/index.ts';
 import { formatText } from '~utils/intl.ts';
 import Select from '~v5/common/Fields/Select/index.ts';
+import NotificationsEnabledWrapper from '~v5/common/NotificationsEnabledWrapper/NotificationsEnabledWrapper.tsx';
 import TitleLabel from '~v5/shared/TitleLabel/index.ts';
 
 import { tabList } from './consts.ts';
-import CountBadge from './partials/CountBadge.tsx';
 import CryptoToFiatTab from './partials/CryptoToFiatTab/CryptoToFiatTab.tsx';
 import NotificationsTab from './partials/NotificationsTab/NotificationsTab.tsx';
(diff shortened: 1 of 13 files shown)


davecreaser (changes requested):
Very nice @bassgeta. I think the solution of splitting the context is no only the best way for this ticket, but it also actually makes a lot of sense and makes them both super clean and easy to read. Also NotificationsEnabledWrapper is really nice, and good idea adding the comment explainer in there. 

I tested it out and it works as expected. I also tried sending some simple payments as Leela while notifications were disabled, including one to Amy who had notifications enabled - that also worked out, Amy still received the notifications and Leela did not. 🎉 

Screenshot 2024-10-01 at 11 00 03
Screenshot 2024-10-01 at 11 15 51
Screenshot 2024-10-01 at 11 16 07

Just one potential change: could you move the MagicbellProvider from NotificationsUserContextProvider into NotificationsDataContextProvider? It should only be needed in the data context really, that way we _really_ keep Magicbell separate from the user context which will always be rendered.

davecreaser (approved):
Thanks for the changes + comment!

rdig (approved):
Working as expected. Nicely done!

Screenshot from 2024-10-07 14-53-11
Screenshot from 2024-10-07 14-53-16
Screenshot from 2024-10-07 14-53-30
Screenshot from 2024-10-07 14-53-47
Screenshot from 2024-10-07 14-54-03

Issue #3210, opened 2024-09-30
[Dashboard] Team reputation widget meatball menu UI fixes


mmioana:
### Issue

The meatball menu should change when a parent team is selected, since the same menu items are not relevant to contributors/ members. It should show 'View members' that opens up the members page with the parent team filter already selected applied. 

Currently:
<img width="470" alt="image" src="https://github.com/user-attachments/assets/27437881-31d9-481e-b19a-8879fc0c871a">

Figma:
<img width="502" alt="image" src="https://github.com/user-attachments/assets/298bd87f-74c5-4f68-ae84-efc5502ada83">

Link in Figma: https://www.figma.com/design/gac6xVgGNCAge5Bia933dp/Dashboard?node-id=6830-36359&t=cXj2gpagsBjz8b9Q-4

rdig:
Closed by #3228

PR #3228, opened 2024-09-30
[Dashboard] Top contributors meatball menu

mmioana:
## Description

The meatball menu should change when a parent team is selected, since the same menu items are not relevant to contributors/ members. It should show 'View members' that opens up the members page with the parent team filter already selected applied. 

Currently:
image

Figma:
image

Link in Figma: https://www.figma.com/design/gac6xVgGNCAge5Bia933dp/Dashboard?node-id=6830-36359&t=cXj2gpagsBjz8b9Q-4

## Testing

* Step 1. Check for `All teams` and `General` the actions in the `Influence by team` widget remained unchanged (you can also check each item still behaves as before)
Screenshot 2024-09-30 at 21 38 42
* Step 2. Select another team/domain
* Step 3. Check only `View members` dropdown item is shown
image
* Step 4. Clicking `View members` should navigate to http://localhost:9091/planex/members

## Diffs

**New stuff** ✨

* `useDropdown` hook
* `DropdownMenu`, `DropdownItem` and `ContributorActionsMenu` components

**Changes** 🏗

* `ReputationChart` renders either `TeamActionsMenu` or `ContributorActionsMenu`

Resolves [#3210](https://github.com/JoinColony/colonyCDapp/issues/3210)


src/components/v5/frame/ColonyHome/partials/ReputationChart/partials/DropdownMenu.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+interface DropdownMenuProps extends PropsWithChildren {
+  isDisabled?: boolean;
+  setTriggerRef: (node: HTMLElement | null) => void;
+  setTooltipRef: (node: HTMLElement | null) => void;
+  getTooltipProps: () => Record<string, any>;
+  visible: boolean;
+}
+
+const DropdownMenu: FC<DropdownMenuProps> = ({
```
bassgeta, line 13: should we maybe already extract this to a new shared component? maybe we can reuse this for the team filter too :eyes:

src/components/v5/frame/ColonyHome/partials/ReputationChart/partials/DropdownItem.tsx
```diff
@@ -0,0 +1,24 @@
+import { type IconProps } from '@phosphor-icons/react';
+import React, { type ComponentType, type FC } from 'react';
+
+import { tw } from '~utils/css/index.ts';
+
+// eslint-disable-next-line max-len
+const dropdownItemClassName = tw`flex w-full items-center gap-2 rounded-s px-3 py-2 text-md text-gray-900 hover:font-medium md:hover:bg-gray-50`;
```
bassgeta, line 7: maybe this can be used directly instead of from a variable

src/components/v5/frame/ColonyHome/partials/ReputationChart/partials/DropdownItem.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+interface DropdownItemProps {
+  onClick?: () => void;
+  icon: ComponentType<IconProps>;
```
bassgeta, line 11: should we make this optional so we can reuse this for the teams dropdown? 
Or maybe we can refactor that down the line and have two components like
`DropdownItem` and
`DropdownIconItem` :shrug:

bassgeta (approved):
Very nice and simple, I like the new dropdown component :100: 
We probably would need to hide `popper` implementation details from that component, but let's not scope creep this and let's do that down the line :D 

The teams dropdown:
image

COntributors dropdown:
image

bassgeta (approved):
Still looking good after the refactor, nice going :sunglasses: 
image
image

Nortsova (changes requested):
Looks great! 🙌 And works perfect 😍 Thank you @mmioana ✨

image
image

However it would be nice to redirect to selected team members. This is the only reason why I am rejecting now ❤️

iamsamgibbs (changes requested):
Great changes here! Love the new components.

Screenshot 2024-10-03 at 12 19 50

Screenshot 2024-10-03 at 12 20 45

Only rejecting for the same reason as @Nortsova that the "View members" link should keep the selected team.

Also tested out how the new components handle not having an icon set and looks great!

Screenshot 2024-10-03 at 12 23 47

mmioana:
Thank you guys for your review!
I have fixed the filter when `View members` is clicked and also removing the filter when clicking on `View all teams`, as it didn't make sense to keep it.

Also, noticed there was a problem with the back navigation with the new `ColonyFiltersContext` and pushed a fix for it.
However, please keep in mind the initial route is first rendered, but if there is a domain present in the context, the route is re-rendered with the search param set. If you go back, you'll see this empty state when the initial route is rendered.
It seemed to take way too much time to also fix this now, but I'll open a new issue for it 🙌

iamsamgibbs (approved):
Nice job! This is a nice fix. I see what you mean with the back issue, but agree this can be addressed as a separate issue.

https://github.com/user-attachments/assets/117364ca-6afd-4b8c-8fe7-9b2d0180a740

mmioana:
@iamsamgibbs I opened https://github.com/JoinColony/colonyCDapp/issues/3263

Nortsova (approved):
redirect-on-view-member

Perfect! Thank you for the fast fix and opened issue ✨

PR #3229, opened 2024-09-30
Hotfix: Handle Bridge drains history when no receipt is available

jakubcolony:
## Description

Thanks to @Nortsova for alerting me of an issue she ran across when off-ramping her crypto. Once the transfer was picked up by Bridge, the drains history disappeared in the UI as the `bridgeGetDrainsHistory` query started returning an error. After a while, the issue went away and drains history was visible again.

I looked through the logs and realised it was trying to read the drain receipt's URL, but it was not available:
```
bridgeXYZMutation handler getDrainsHistoryHandler failed with error: TypeError: Cannot read properties of undefined (reading 'url')
```

Most likely, when a new transfer is created, the receipt is not available immediately but only after a few minutes or so.

This PR changes the `getDrainsHistory` handler to gracefully handle a case where there's no receipt as well as introduces an extra state in the UI. 

## Testing

I could not test it even when connected to prod, it's an intermittent issue that only reproduces under the right circumstances.

I attached all the investigation details above in a hope this justifies my fixes.

Also, here's a screenshot from local dev connected to Bridge prod, just to prove the drains history still works:
image


iamsamgibbs (dismissed):
These changes all look sensible, nicely done!

The only thing I would suggest, is that if we know we will definitely get the receipt at some point, maybe the wording could be "Receipt not yet available", or if that is too long maybe just "Pending"?

jakubcolony:
Hey @iamsamgibbs, I've checked in with Mel and updated the copy to "Generating...":

image

iamsamgibbs (approved):

rdig (approved):
Approving this with and offer of trying of reproducing this in production :)

bassgeta (approved):
The changes all look like they **should** solve the issue, so :crossed_fingers: they do.
Nice job going into the trenches and solving the issue :100:

PR #3231, opened 2024-10-01
[Dashboard] Team switcher dropdown fixes

### LINKED ISSUE #3215: [Dashboard] Team/domain filter dropdown UI fixes
### Issues

- [ ] The menu items in the meatball menu dropdown need additional padding on the left and right and top and bottom of each menu item to match Figma.

Currently:
image

Figma:
image

---

- [ ]  A very minor issue, but can we avoid the jump on hover in the meatball menu if possible?

Export-1727695641373

**For this last point, we might need to make use of `absolute` positioning**
Or please discuss with @melyndav what other solutions we could have, given the width of the copy text changes with the weight of the used font

iamsamgibbs:
## Description

This PR addresses an issue where hovering over the team switcher dropdown items would cause the text font weight to change to bold, this would cause the container width to increase. This PR resolves this by adding an after pseudo element with the same text content set to bold to "pre-reserve" the necessary space for the bold text when hovered. It is hidden so shouldn't cause any accessibility concerns, but please raise in the comments if you see any issues with this approach.

## Testing

Shrink the screen width enough so that teams drop into the dropdown filter (or add new teams). Hover each of the items, ensure that the dropdown width does not increase when hovering the longest item.

https://github.com/user-attachments/assets/56c8a191-f28b-4653-a4ad-62fc147b7f13

## Diffs

**Changes** 🏗

* After pseudo element added to dropdown items

## TODO

The original issue also mentioned padding and gap issues, but comparing against figma it seems to line up perfectly already.

Screenshot 2024-09-30 at 17 25 30

Figma:

Screenshot 2024-10-01 at 09 50 50

Screenshot 2024-10-01 at 09 51 00

The font size and line heights are also correct. Let me know if I am missing something here.

Resolves #3215

### DIFF
--- a/src/components/v5/shared/TeamFilter/partials/TeamsDropdown/TeamsDropdown.tsx
+++ b/src/components/v5/shared/TeamFilter/partials/TeamsDropdown/TeamsDropdown.tsx
@@ -57,7 +57,7 @@ const TeamsDropdown: FC<TeamsDropdownProps> = ({ domains }) => {
               key={`TeamsDropdown.${domain.nativeId}`}
               onClick={() => handleClick(domain)}
               aria-label={domain.metadata?.name} // Used to set the after content, which is used to pre-reserve the space required for the bold hover state
-              className="w-full text-start text-sm font-medium text-gray-700 after:invisible after:block after:h-0 after:overflow-hidden after:font-semibold after:content-[attr(aria-label)] hover:font-semibold hover:text-gray-900"
+              className="w-full text-start text-sm font-medium text-gray-700 bold-on-hover hover:text-gray-900"
             >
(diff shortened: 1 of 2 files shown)


mmioana (approved):
A veeery nice solution @iamsamgibbs 👏 

https://github.com/user-attachments/assets/0b69fc94-8ab1-4120-a5d7-5b5ddde00e2c

bassgeta (approved):
nicely done and nice utility class :sunglasses: 
image
image
Firefox also looks legit:
image
image

Nortsova (approved):
Thank you @iamsamgibbs for this perfect solution 🙌✨ It looks really good:
bold-on-hover

PR #3236, opened 2024-10-01
fix: update UI to match the design

### LINKED ISSUE #3213: [Dashboard] Funds cards row UI/UX fixes
### Issues

- [x] The styling of the create team box such as the font size and padding between the icon should match Figma. 

Currently: 
image

Figma:
image

---

- [x]  When individual teams are selected, they should go to the 'Balances' page with the selected team applied. Currently they just have a hover state and don't link anywhere.

image

---

- [x]  The 'Create team' card shouldn't apply when a parent team is selected. It should only show where there are no parent teams in the colony.

Currently showing:

image

As per Figma:

image

Nortsova:
## Description
- [x]  The styling of the create team box such as the font size and padding between the icon should match Figma.
- [x]  When individual teams are selected, they should go to the 'Balances' page with the selected team applied. Currently they just have a hover state and don't link anywhere.
- [x]  The 'Create team' card shouldn't apply when a parent team is selected. It should only show where there are no parent teams in the colony.

## Testing
1. Run `node scripts/create-colony-url.js` and create new team
2. See that "Create team" item is shown under "All teams" and "General" filter 
Design requirements:
image
Result:
image
image

3. Check that all items match the design (link: https://www.figma.com/design/gac6xVgGNCAge5Bia933dp/Dashboard?node-id=4867-40639&node-type=frame&t=yyk6XVtMJFk4l0TJ-0 ):
Pixel perfect example: 
image

4. Create a team and check that the "Create team" item is no longer shown
image
image

5. Go to planex colony and check that items matches the design (link: https://www.figma.com/design/gac6xVgGNCAge5Bia933dp/Dashboard?node-id=4859-36367&node-type=frame&t=pXOwAsU0iYnga3iH-0 ):

Pixel perfect example:
image

6. Click on any team is selected, it should redirect to balance page with selected team:
click-on-team-item

## Diffs
**Changes** 🏗

* Updated NumeralCurrency formating in case value is 0, now instead of 0.00 it will show only 0

Resolves  #3213

### DIFF
--- a/src/components/shared/Numeral/Numeral.tsx
+++ b/src/components/shared/Numeral/Numeral.tsx
@@ -29,7 +29,7 @@ const Numeral = ({
 
   return (
     <NumeralBase
-      prefix={prefix}
+      prefix={prefix && `${prefix} `}
       suffix={suffix}
       className={className}
       appearance={appearance}
--- a/src/components/v5/frame/ColonyHome/partials/FundsCards/hooks.ts
+++ b/src/components/v5/frame/ColonyHome/partials/FundsCards/hooks.ts
@@ -10,6 +10,7 @@ import {
 } from '~gql';
 import { useCurrencyHistoricalConversionRate } from '~hooks/useCurrencyHistoricalConversionRate.ts';
 import useGetSelectedDomainFilter from '~hooks/useGetSelectedDomainFilter.tsx';
+import { useSubDomains } from '~hooks/useSubDomains.ts';
 import { convertFromTokenToCurrency } from '~utils/currency/convertFromTokenToCurrency.ts';
 import { type CoinGeckoSupportedCurrencies } from '~utils/currency/index.ts';
 
@@ -134,4 +135,17 @@ export const usePreviousTotalData = () => {
      */
     previousTotal: convertAmount(previousTotal),
   };
+};
+
+export const useIsAddNewTeamVisible = () => {
+  const subTeams = useSubDomains();
+
+  const selectedDomain = useGetSelectedDomainFilter();
+
+  // In case "All teams" selected we have "General" team created by default
+  // @TODO: probably with nested teams this functionality will change
+  if (!selectedDomain && (!subTeams || subTeams.length <= 1)) {
+    return true;
+  }
+  return selectedDomain?.isRoot && (!subTeams || subTeams.length < 1);
 };
--- a/src/components/shared/Numeral/NumeralBase.tsx
+++ b/src/components/shared/Numeral/NumeralBase.tsx
@@ -21,7 +21,7 @@ export const NumeralBase: FC<Props> = ({
       className={classNames(getMainClasses(appearance, styles), className)}
       {...rest}
     >
-      {prefix && `${prefix} `}
+      {prefix}
       {children}
       {suffix}
     </span>
--- a/src/components/shared/Numeral/NumeralCurrency.tsx
+++ b/src/components/shared/Numeral/NumeralCurrency.tsx
@@ -23,10 +23,9 @@ const NumeralCurrency = ({
   appearance,
   ...rest
 }: NumeralProps) => {
-  const fixedValue = value;
-  const convertedValue = convertToDecimal(fixedValue, decimals || 0);
+  const convertedValue = convertToDecimal(value, decimals || 0);
 
-  const formattedValue = getFormattedCurrencyValue(convertedValue, fixedValue);
+  const formattedValue = getFormattedCurrencyValue(convertedValue, value);
 
   return (
     <NumeralBase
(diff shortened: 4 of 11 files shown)


src/components/v5/frame/ColonyHome/partials/FundsCards/hooks.ts
```diff
@@ -142,3 +143,19 @@ export const useTotalFunds = () => {
...
+
+  // In case "All teams" selected we have "General" team created by default
+  // @TODO: probably with nested teams this functionality will change
+  if (!selectedDomain && (!subTeams || subTeams.length <= 1)) {
+    return true;
+  }
+  const isAddNewTeamVisible =
+    selectedDomain?.isRoot && (!subTeams || subTeams.length < 1);
+
+  return isAddNewTeamVisible;
```
mmioana, line 160: Would it make sense to directly return `selectedDomain?.isRoot && (!subTeams || subTeams.length < 1)` without creating the `isAddNewTeamVisible`?

mmioana (approved):
Nice and pixel-perfect fixes @Nortsova ✨❤️

 Inside `planex`

https://github.com/user-attachments/assets/f8d0c4ca-1198-4720-a489-98204894e4e2

Inside a new colony
Screenshot 2024-10-01 at 18 25 57
Screenshot 2024-10-01 at 18 26 21

Creating a new domain in the brand new colony
Screenshot 2024-10-01 at 18 26 54
Screenshot 2024-10-01 at 18 26 58
Screenshot 2024-10-01 at 18 27 01

Thanks so much for your effort 🌟 let's get it mergeeeed!

src/components/v5/frame/ColonyHome/partials/FundsCards/hooks.ts
Nortsova: good spot, thank you!

src/components/shared/Numeral/Numeral.tsx
```diff
@@ -29,7 +29,7 @@ const Numeral = ({
 
   return (
     <NumeralBase
-      prefix={prefix}
+      prefix={prefix && `${prefix} `}
```
bassgeta, line 32: should we maybe just call the prefix with a space in front of it when we call this component instead? I know we sometimes don't want the prefix to have a space when displaying currency data and such

bassgeta (approved):
Another one smashed out of the park :sunglasses: 
The create new team shows up on a new colony :heavy_check_mark: 
image
Also shows up when `General` is selected :heavy_check_mark: 
image
New team makes it disappear:
image
You get redirected with a query param :heavy_check_mark: 

https://github.com/user-attachments/assets/85bcf6e4-8281-405a-a37a-9972905396bb

src/components/shared/Numeral/Numeral.tsx
Nortsova: For currency, it works as you mentioned. However, for the Numeral component, I decided to leave the space as is because it is used in many places, and we would need to check all of them in this case.

iamsamgibbs (changes requested):
Nice work on this @Nortsova ! The UI matches figma perfectly.

Screenshot 2024-10-04 at 12 41 24

Screenshot 2024-10-04 at 12 41 28

Screenshot 2024-10-04 at 12 42 06

Screenshot 2024-10-04 at 12 42 12

This works great:

https://github.com/user-attachments/assets/57105e47-661f-4267-84e5-0e2cef324f6b

However, I'm rejecting as I think the card here should still link to the balance page:

https://github.com/user-attachments/assets/381a47d6-64f2-46ed-998d-596bb4e89224

There is also an issue with the team selection not getting properly set. In this video, I click a funds card, it takes me to the balances page with the team selected as expected. However, when I then navigate to the dashboard, my team selection has been lost. This only happens when you are navigated via the funds card. If I select a team on the dashboard and navigate between a bunch of pages it is not an issue.

https://github.com/user-attachments/assets/23dc3d25-eaaa-4cef-b03e-402ef1e990b5

Good work getting it this far though, very almost done! (Equally if you would prefer to handle these in a new issue, let me know and I will switch this to an approval)

mmioana:
@iamsamgibbs @Nortsova I think with the newly introduced changes to the `feat/enhanced-dashboard` we can use 
`{ updateTeamFilter } = useColonyFiltersContext();` in the `onTeamClick` function from the `FundCardsItem` component, to persist the team filter 🙌

Nortsova:
Thank you @iamsamgibbs and @mmioana 
Now total is also clickable:
total-clickable

iamsamgibbs (approved):
Perfect! Working flawlessly now 🎉 

https://github.com/user-attachments/assets/5a7d0f32-efa0-4d9b-a497-8fa895de218a

Issue #3226, opened 2024-09-30
[Dashboard] Properly convert existing token to CLNY


mmioana:
### Issue
Currently we try to convert a token by it's id to the selected currency which can be `CLNY`

However, `CoinGecko` doesn't provide this information in the `${COINGECKO_API_URL}/api/v3/coins/${tokenId}/history` endpoint response.

### Implementation
We might need to come up with a robust solution, but an easy one would be to get all balance data - `useGetDomainBalanceQuery` - in `USD` if `SelectedCurrency` is `CLNY`.
Then call `getCLNYPriceInUSD`, followed by `convertTokenToCLNY(tokenPriceinUSD, clnyInUSD)`

The same will be available for the `useGetCachedDomainBalanceQuery`. However the data stored for it is in `USDC`, so once we get the value back, we should convert it to `USD` if `SelectedCurrency` is `CLNY`.
Then call `getCLNYPriceInUSD`, followed by `convertTokenToCLNY(tokenPriceinUSD, clnyInUSD)`
 
### Update
After a long and insightful discussion with @bassgeta, given `CoinGecko` doesn't provide token to token conversion, our best approach, at least for the moment, is the one mentioned above, even though there are going to be some exchange loss due to the number of conversions needed.

PR #3237, opened 2024-10-01
[Dashboard] Token balance conversion to CLNY token

mmioana:
## Description

Currently we try to convert a token by it's id to the selected currency which can be `CLNY`. However, `CoinGecko` doesn't provide this information in the `${COINGECKO_API_URL}/api/v3/coins/${tokenId}/history` endpoint response.
So the easy fix is to get all balance data in `USD` and then do the conversion from `USD` to `CLNY`

This PR addresses this issue, plus a bigger refactoring which introduces 3 contexts in order to avoid querying the same data multiple times.

## Testing

TODO: Please test the data displayed in the `Total` cards and the `Total in and out` card are updating properly based on the selected currency.

Screenshot 2024-10-01 at 17 04 41
Screenshot 2024-10-01 at 17 06 06

## Diffs

**New stuff** ✨

* `BalanceCurrencyContext`, `TotalInOutBalanceContext` and `TotalInOutBalanceChartContext` contexts
* `convertFromTokenToClny`

**Deletions** ⚰️

* `usePreviousLast30DaysData`
* `useLast30DaysData`
* `useData`
* `useChartYSteps`
* Unused `useTotalFunds` and `calculateTotalFunds`

Resolves [#3226](https://github.com/JoinColony/colonyCDapp/issues/3226)


bassgeta (approved):
fur such large change blocks, this is very readable and it makes a lot of sense when you go through it, so great job overall!
Loving the fact that we use a context to dedupe calls for this hefty lambda, optimization :chart_with_upwards_trend: 

Fingers crossed this is gonna be one of the last things you need to do on this giga chonker :sweat_smile: 

All teams:
image
Switching to ETH shows up loaders:
image
Gets updated (ofc with dummy values locally)
image
CLNY also works:
image

:ship:

src/context/BalanceCurrencyContext/BalanceCurrencyContextProvider.tsx
```diff
@@ -0,0 +1,74 @@
...
+const BalanceCurrencyContextProvider = ({
+  children,
+}: {
+  children: ReactNode;
+}) => {
+  const { currency } = useCurrencyContext();
+  const queryCurrency =
+    currency === SupportedCurrencies.Clny ? SupportedCurrencies.Usd : currency;
+
+  const defaultClnyInUSD = isDev ? 1 : 0;
```
davecreaser, line 26: Very handy, thank you 😆

src/components/v5/frame/ColonyHome/partials/FundsCards/hooks.ts
```diff
@@ -1,62 +1,115 @@
...
+  }, [selectedDomain?.id, colonyAddress]);
+
+  const { data, loading } = useGetCachedDomainBalanceQuery(
+    memoizedQueryVariables.queryOptions,
+  );
+
+  useEffect(() => {
+    return () => {
+      if (loading) {
+        memoizedQueryVariables.abortController.abort();
```
davecreaser, line 109: 👏 Really nice fix for this domain problem!

davecreaser (approved):
Wow! A lot of work here 😅 Really good stuff, I think the contexts are a neat idea, and it certainly made that hook file smaller!

I tried my best to understand everything that was written since this is a large, complex PR, but I think you've done a really good job with it from what I can tell. 🎉 

Also tested it, and it worked as expected:
Screenshot 2024-10-03 at 16 41 32
Screenshot 2024-10-03 at 16 41 46

Issue #3208, opened 2024-09-30
[Dashboard] Team reputation widget inconsistent state


mmioana:
### Issue

As mentioned in Discord, when I first load into a Colony or switched to a new colony, the team reputation shows three states.

1. Skeleton loader state.
2. The empty state with no teams or no reputation (ideally this shouldn't show) with vertical height issues.
3. The active state with teams and pic charts.

### Implementation
Given the `updateContributorsWithReputation` finishes before `db` domain reputation data is updated and subscribing to the `OnUpdateColony` will retrieve `lastUpdatedContributorsWithReputation` before the domain reputation data is available, our best approach will be to add an extra `isReputationUpdating` flag to the `ColonyContext` and reset it after a `2s` timeout. However, in the reputation chart in order to prevent the loading skeletons to show when there actually is domain data available, we can use the following piece of code
```
const isNoReputationAvailable = !isThereReputationInDomains(allTeams);
const isDataLoading = isContributorsLoading || (isReputationUpdating && isNoReputationAvailable);
```

mmioana:
Closed by #3247

PR #3247, opened 2024-10-02
[Dashboard] Reputation widget state

mmioana:
## Description

As mentioned in Discord, when I first load into a Colony or switched to a new colony, the team reputation shows three states.

1. Skeleton loader state.
2. The empty state with no teams or no reputation (ideally this shouldn't show) with vertical height issues.
3. The active state with teams and pic charts.

### Implementation
Given the `updateContributorsWithReputation` finishes before `db` domain reputation data is updated and subscribing to the `OnUpdateColony` will retrieve `lastUpdatedContributorsWithReputation` before the domain reputation data is available, our best approach will be to add an extra `isReputationUpdating` flag to the `ColonyContext` and reset it after a `2s` timeout. However, in the reputation chart in order to prevent the loading skeletons to show when there actually is domain data available, we can use the following piece of code
```
const isNoReputationAvailable = !isThereReputationInDomains(allTeams);
const isDataLoading = isContributorsLoading || (isReputationUpdating && isNoReputationAvailable);
```

## Testing

TODO: Make sure the Reputation chart empty state is not visible (or at least for a small amount of time given we can't really determine when there is reputation data available) if reputation data exists. 

* Step 1. Make sure to start with a clean environment and run `create-data` script first
* Step 2. Visit `planex`
* Step 3. Check reputation chart state
* Step 4. Visit `wayne`
* Step 5. Check reputation chart state

## Diffs

**New stuff** ✨

* `isReputationUpdating` in `ColonyContext`

**Changes** 🏗

* `ReputationChart` loading state

Resolves [#3208](https://github.com/JoinColony/colonyCDapp/issues/3208)

### DIFF
--- a/src/components/v5/frame/ColonyHome/partials/ReputationChart/ReputationChart.tsx
+++ b/src/components/v5/frame/ColonyHome/partials/ReputationChart/ReputationChart.tsx
@@ -36,13 +36,11 @@ const MSG = defineMessages({
 const ReputationChart = () => {
   const {
     colony: { domains },
+    isReputationUpdating,
   } = useColonyContext();
 
   const { contributorsList, loading: isContributorsLoading } =
     useContributorsByDomain();
-
-  // @TODO here we should also check if the colony data is loading
-  const isDataLoading = isContributorsLoading;
 
   const selectedDomain = useGetSelectedDomainFilter();
 
@@ -52,7 +50,11 @@ const ReputationChart = () => {
       (a, b) => Number(b.reputationPercentage) - Number(a.reputationPercentage),
     );
 
-  const chartDataTeams = !isThereReputationInDomains(allTeams)
+  const isNoReputationAvailable = !isThereReputationInDomains(allTeams);
+  const isDataLoading =
+    isContributorsLoading || (isReputationUpdating && isNoReputationAvailable);
+
+  const chartDataTeams = isNoReputationAvailable
     ? []
     : getTeamReputationChartData(allTeams);
 
--- a/src/context/ColonyContext/ColonyContext.ts
+++ b/src/context/ColonyContext/ColonyContext.ts
@@ -22,6 +22,7 @@ export interface ColonyContextValue {
   refetchColony: RefetchColonyFn;
   startPollingColonyData: (pollInterval: number) => void;
   stopPollingColonyData: () => void;
+  isReputationUpdating: boolean;
   isSupportedColonyVersion: boolean;
   colonySubscription: {
     canWatch: boolean;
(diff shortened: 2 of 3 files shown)


bassgeta (changes requested):
The cleanest solution I can think of for a problem this weird :melting_face: 
Good job resolving this, it was probably quite a painful journey to come to the conclusion that we need to wait for 2 seconds until we dare show the teams reputation chart :')

However, I've just opened `Wayne` and I got the `No reputation yet` jumping in :(

https://github.com/user-attachments/assets/7695123f-a790-4608-ab74-86bcff21f5d4

However, given that this only happens the first time a colony with teams and no reputation mined (so basically only the dev env), should we even solve this? It's most likely a non issue on prod and QA?

mmioana:
> The cleanest solution I can think of for a problem this weird 🫠 Good job resolving this, it was probably quite a painful journey to come to the conclusion that we need to wait for 2 seconds until we dare show the teams reputation chart :')
> 
> However, I've just opened `Wayne` and I got the `No reputation yet` jumping in :(
> 
>  reputation-loading-2024-10-03_14.19.01.mp4 
> However, given that this only happens the first time a colony with teams and no reputation mined (so basically only the dev env), should we even solve this? It's most likely a non issue on prod and QA?

I can totally agree that the `useTimeout` is a hacky solution, that can't really cover the situation of slow DB operations for updating the teams/domains reputation. However, please check if in QA or prod this can actually happen and if not, let's remove the `useTimeout` - I would argue though to keep the `isReputationUpdating` state as it is the only way to know when reputation updates are performed

mmioana:
The timeout solution was dropped, but we'll refetch the colony data after the colony reputation gets updated

rdig (approved):
Seems to be working as expected, nicely done!


https://github.com/user-attachments/assets/4314b6f4-7573-4a67-b002-c35245193482

iamsamgibbs (approved):
Nice job, this is working very nicely now!

https://github.com/user-attachments/assets/9d65e8f9-f9e9-496c-848a-c5841aee262d

Creating a new colony still looks as expected too:

https://github.com/user-attachments/assets/0e17f104-ca76-4f61-bc2e-de7a089d2dbd

bassgeta (approved):
Niiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiice :sunglasses: 
Vacation brainpower is real, really nice and smooth solution with the refetching of a colony after we recalculate the data!

https://github.com/user-attachments/assets/63515eb9-f80e-4d15-a15e-9284cd6f2989

Issue #3239, opened 2024-10-01
Create agreement flashing missing extension error banner


arrenv:
Teamwork task: [Create agreement flashing missing extension error banner](https://tw.auroracreation.com/#/tasks/16097785)

## Steps to reproduce

1. Open up a new Create agreement action.
2. Note the error banner that appears.

## Expected behaviour

* There should not be a banner that flashes in on opening up the Create agreement action.

## Actual behaviour

* There is a banner that flashes in when creating a new Create agreement action.

[agreement-banner.webm](https://github.com/user-attachments/assets/3c13cebb-02ed-46a4-bcf5-da22cff70ae3)

PR #3249, opened 2024-10-02
fix: prevent display error banner when extension data is loading

adam-strzelec:
## Description

After open Action Sidebar to create agreement, during fetching extensions data, the no extension error banner is visible even though all required extensions are installed.

Screenshot 2024-10-02 at 14 11 39

## Testing
- Set network connection to eg. 3G
- Click `Create agreement` on agreements page
- Error banner should not be visible when data is fetching

## Diffs

**Changes** 🏗

add checking if data is not loading inside `useIsFieldDisabled` hook

Resolves https://github.com/JoinColony/colonyCDapp/issues/3239

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/hooks.ts
+++ b/src/components/v5/common/ActionSidebar/partials/hooks.ts
@@ -74,14 +74,20 @@ export const useSubmitButtonDisabled = () => {
 };
 
 export const useIsFieldDisabled = () => {
-  const { extensionData } = useExtensionData(Extension.VotingReputation);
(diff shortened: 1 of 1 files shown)


iamsamgibbs (approved):
Nice job! This is working as expected.

Screenshot 2024-10-02 at 13 42 48

Tested it both with and without the extension installed.

It does have the reverse effect than the original issue where if you are a user without permissions, the banner will take a split second before it shows up - but I think this is the lesser of two evils and the 'Create agreement' button is still disabled so it is not as if the user could try to create the agreement.

Issue #3218, opened 2024-09-30
[Dashboard] Mobile navigation menu opened by default


mmioana:
### Issue

When I minimise the browser, the mobile nav is automatically triggered open. This shouldn't be triggered and take over the UI, instead the closed default mobile header should be displayed. 

![Export-1727697383382](https://github.com/user-attachments/assets/7c4be856-177d-4b6e-b740-3f1c6b97a77f)

PR #3254, opened 2024-10-02
[Dashboard] Mobile navigation menu opened by default

mmioana:
## Description

When I minimise the browser, the mobile nav is automatically triggered open. This shouldn't be triggered and take over the UI, instead the closed default mobile header should be displayed. 

Export-1727697383382

This PR addresses this issue by introducing a `resetSidebars` function that resets the navigation or colony switcher upon window resizing if on a device > `table`

## Testing

TODO: Please test opening either the navigation or colony switcher and then resize to a viewport wider than > `1024px`. Then switch again to a viewport less wide than < `1024px`. None of these sidebars should be opened by default.

https://github.com/user-attachments/assets/7ba3beb6-7d68-44fd-b9b1-a62c0c85ccf7

## Diffs

**New stuff** ✨

* `resetSidebars` callback used for `useResize` in `PageLayoutContextProvider`

Resolves [#3218](https://github.com/JoinColony/colonyCDapp/issues/3218)

### DIFF
--- a/src/context/PageLayoutContext/PageLayoutContextProvider.tsx
+++ b/src/context/PageLayoutContext/PageLayoutContextProvider.tsx
@@ -7,6 +7,8 @@ import React, {
 } from 'react';
 
 import { type UserHubTab } from '~common/Extensions/UserHub/types.ts';
+import { useTablet } from '~hooks';
+import { useResize } from '~hooks/useResize.ts';
 
 import { PageLayoutContext } from './PageLayoutContext.ts';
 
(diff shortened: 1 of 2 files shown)


iamsamgibbs (approved):
Nice job fixing this issue! All working great now.

https://github.com/user-attachments/assets/37490516-7a6d-4839-8f80-59ea0999d165

I wonder if this fix would be better applied directly to the navigation branch as it doesn't specifically touch any dashboard functionality.

I also think it might be a good idea to add a throttle function to the useResizeHook like this?

```
import { throttle } from 'lodash';
import { useEffect } from 'react';

export const useResize = (callback: () => void, wait: number = 200) => {
  useEffect(() => {
    const throttledCallback = throttle(callback, wait);

    throttledCallback();

    window.addEventListener('resize', throttledCallback);

    return () => {
      window.removeEventListener('resize', throttledCallback);
      throttledCallback.cancel();
    };
  }, [callback, wait]);
};
```

davecreaser (approved):
Nice fix @mmioana, it's working great 🚀 

https://github.com/user-attachments/assets/d35fb992-0bc2-4062-8101-58cfdef9b6a2

I agree with Sam too, it'd be nice to have some throttling on this function if you feel like implementing it. Won't stop me from approving though since that's more of an "improvement point" for the hook you've used, not tied to this ticket.

Issue #3260, opened 2024-10-03
[Streaming Payments] Streaming Payments Page


AdrianDudko:
Teamwork task:

## Related PR's

- [ ] pr#1
- [ ] pr#2
- [ ] pr#3

## Vision

You are able to easily find and view all active and past streaming payments in a colony.

## **Overview**

A dedicated streaming page will help to highlight the feature and provide transparency to members of the colony. It should be a view of all of the streaming payments a colony has made, including active and past streaming payments. It should include live updating values of total current streams and statistics. It should provide the ability to search, filter, and sort streaming payments to easily find the streaming payment you are looking for. Streaming payments should be easy to access with one click to open them in the action panel. 

## **User stories**

- As a USER, I want to be able to see a list of the streaming payments in a colony, both past streams and active.
- As a USER, I want to be able to see some values of important figures, such as the total current streaming value and total funds streamed to date.
- As a USER, I want to be able search and filter streams in a colony to be able to easily find a stream I am looking for.
- As a USER, I want to see streams grouped by recipient, to be able to see all streams for a specific recipient in one place and the total amount being streamed for the recipient.
- As a USER, I want to be able to open up and manage streams from the list of streams.
- As a USER, I want to be able to be able to create a new stream.
- AS a USER, I want to see the history of all funds that have been paid from active and past Streams.

## **Priority**

- Critical - Without it, users will easily lose track of active payment streams.

## **User Flows**

- View page
    - User clicks on the “Streaming” menu item to navigate to the Streaming page.
    - User is able to browse the list of active streams ordered by highest value first.
- Filtering & Searching Streams
    - User is able to click on the “Team” filter to filter streams by team.
    - User is able to click on the “Filter” button to open the filter menu.
    - User can select the filters or enter the search they want to apply and it will update the list.
    - Active filters will be visible at the top of the table.
    - Clicking the “x” on active filters and searches will reset the table to the default.
- Create new stream
    - User clicks on “create new stream” button.
    - Opens the Action panel with the “Create new stream” action type pre-selected.
- View stream
    - User clicks on an available stream.
    - Opens up the Streaming payment in the action panel.

## Design

**Streaming page**

Figma link: https://www.figma.com/file/5V8pr7iMwXsT9L3VAZsmUt/Colony-v5?type=design&node-id=20551-221564&mode=design&t=WWlxemq0eqFdRO5I-4

[...]

PR #3264, opened 2024-10-04
feat: add streaming payments subpage

adam-strzelec:
## Description

Create subpage for streaming payments
Screenshot 2024-10-09 at 10 00 50

## Testing

-Open `Finances` in navigation and click `Streaming payments`

## Diffs

**New stuff** ✨

Add new Route for `Streaming subpages`

TO DO
Implement rest of the features inside the subpage

Main issue - https://github.com/JoinColony/colonyCDapp/issues/3260

### DIFF
--- a/src/components/frame/Extensions/layouts/consts.ts
+++ b/src/components/frame/Extensions/layouts/consts.ts
@@ -15,7 +15,7 @@ import {
   // SealCheck,
   // ChartPieSlice,
   // ListChecks,
-  // ChartLine,
+  ChartLine,
   // Intersect,
   // Briefcase,
   // BookOpenText,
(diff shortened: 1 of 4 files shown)


jakubcolony (approved):
Thanks for your PR @adam-strzelec, all looks good 👍 

The page is accessible both via navigation and direct link:

image

image

Just noting Czarek used `/streaming` path instead of `/streaming-payments` in his PR. @arrenv could you help with which one should be used?

iamsamgibbs (approved):
Looks good to me! Nice job. (Although noting that this will need checking when these branches are rebased onto the new nav once that has been merged)

Screenshot 2024-10-10 at 11 46 32

Screenshot 2024-10-10 at 11 46 36

PR #3271, opened 2024-10-04
fix: replace div with span in SkeletonLoading component for html cons…

### LINKED ISSUE #3268: [Dashboard]: Console div inside p error
## Steps to reproduce
1. Open the dashboard page
2. See error in console
image

## Suggestions 
Probably would be a good idea to replace div with span, or update nesting in FundsCardsSubTitle

Nortsova:
## Description

Before:
image

## Testing

Step 1. Check that there is no errors in the console
image

Step 2. Check that all Skeletons still shown correctly:
image

Resolves #3268

### DIFF
--- a/src/components/common/LoadingSkeleton/LoadingSkeleton.tsx
+++ b/src/components/common/LoadingSkeleton/LoadingSkeleton.tsx
@@ -11,7 +11,7 @@ const LoadingSkeleton: React.FC<LoadingSkeletonProps> = ({
   children,
 }) =>
   isLoading ? (
-    <div className={clsx('overflow-hidden skeleton', className)} />
+    <span className={clsx('block overflow-hidden skeleton', className)} />
   ) : (
     <>{children}</>
   );


iamsamgibbs:
This looks like a sensible change! Can we do it on master rather than enhanced-dashboard? :)

rdig (approved):
All good here. I approved it, but as @iamsamgibbs said, please change the base branch and do this change on `master`, not on the dashboard feature branch

iamsamgibbs (dismissed):
All looks good! Let's get it merged into master.

Screenshot 2024-10-11 at 09 33 31

rumzledz (dismissed):
I'm not entirely sure if I agree with this fix since the root issue is that `FundsCardsSubTitle` is written like this:

```ts
<p className="flex items-center gap-2">
  <LoadingSkeleton className="h-[27px] w-[90px] rounded" isLoading={isLoading}>
    <span>{value}</span>
  </LoadingSkeleton>

  <LoadingSkeleton isLoading={isLoading} className="h-5 w-[30px] rounded">
    <span className="uppercase text-1">{currency}</span>
  </LoadingSkeleton>
</p>
```

We could just simply replace `p` with `div`, keep the LoadingSkeleton as a `div` and tadaaa

no errors

But at the same time, I don't feel strongly about a component that only momentarily appears and this code update gets rid of the error as well anyway so I'm approving this. Great stuff @Nortsova 💯

Nortsova:
Thank you all for the review! 🙌 I updated the base branch to `master`. ✨

Thank you @rumzledz for your comment! I really appreciate it 🙌 I considered that fix, but it really depends on our goal. I decided to implement a universal fix for the future, as I believe there could be other cases where we might want to place a Skeleton inside a `<p>` tag. While I think this will be a rare scenario, it's still possible.

iamsamgibbs (approved):
Thanks for the change! Still looking good.

Screenshot 2024-10-14 at 12 10 09

rdig (approved):
re-approving

rumzledz (approved):

PR #3276, opened 2024-10-06
Fix: set current wallet address as the default getSigner address

### LINKED ISSUE #3275: Local dev authenticates as the first wallet even if a different one is chosen
I spotted this bug when testing #3252 as I tried to check the notifications for a user other than leela. The lambda call would be blocked by auth proxy however, as it thought I was authenticated as the first dev wallet. 

## Steps to reproduce

1. Disconnect your wallet and clear your browser data, including cookies.
2. Connect any dev wallet other than the first one.
3. Inspect the `/auth` request in the network tab.

## Expected behaviour

You have been authenticated using your currently connected wallet.

## Actual behaviour

No matter which wallet you choose, you will be authenticated as the first dev wallet:
image

jakubcolony:
## Description

Please see linked issue #3275 for more details.

This PR ensures that `getSigner` will return the signer for the current wallet unless a different address is specified. Apparently not providing an address makes it return the very first address it gets from `eth_accounts`. 

## Testing

I noticed this issue in the notifications branch, but it also affects other areas, for example when updating a user profile. The auth proxy performs a similar check, comparing the GraphQL mutation's address variable with the authenticated wallet address. And because it always thinks it's the first dev wallet address, updating any other profile will fail.

1. Disconnect your wallet and clear your browser data, including cookies.
2. Connect any dev wallet other than the first one.
3. Navigate to User account and update any of the information on that page. The mutation should be allowed by auth proxy.

You can also inspect the network tab to see the `/auth` request contains the correct wallet address.

Resolves #3275

### DIFF
--- a/src/redux/sagas/wallet/RetryProvider.ts
+++ b/src/redux/sagas/wallet/RetryProvider.ts
@@ -3,6 +3,7 @@ import { providers, utils } from 'ethers';
 import { backOff } from 'exponential-backoff';
 
 import { GANACHE_LOCAL_RPC_URL } from '~constants/index.ts';
+import { type Address } from '~types/index.ts';
 import {
   RetryProviderMethod,
   IColonyContractMethodSignature,
@@ -13,7 +14,10 @@ type RetryProviderOptions = {
   delay?: number; // in milliseconds
 };
 
-const classFactory = (walletType: 'MetaMask' | string = '') => {
+const classFactory = (
+  walletType: 'MetaMask' | string = '',
+  walletAddress?: Address,
+) => {
   const devWallet = walletType !== 'MetaMask';
 
   const Extender = devWallet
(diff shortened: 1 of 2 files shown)


iamsamgibbs (approved):
Nice job picking up on this issue and resolving it Jakub. All tested and working as expected now:
https://github.com/user-attachments/assets/0b9bde22-1778-4877-937d-eb80474a142c

Compared to master:
https://github.com/user-attachments/assets/bdf20ef7-0f57-4199-81a4-4fda3a88d392

rdig (approved):
Nice catch. This was obviously an oversight on my part.

:100:

bassgeta (approved):
It's nice that this was actually easy to solve, on paper it sounded quite more severe :sweat_smile: Nice solution and quality of life improved :chart_with_upwards_trend: 
image

PR #3292, opened 2024-10-10
[Notifications] Feat: Notification should take you to relevant colony

### LINKED ISSUE #3281: Notifications: Clicking on a notification should take you to the relevant colony
## Steps to reproduce

1. Open a colony and make a simple payment. A notification should come through.
2. Go to a different colony and open the userhub.
3. Click on the notification. It opens the sidebar in the current colony, with a transaction from the other colony.

## Expected behaviour

* Example: The user should be navigated to the correct colony.

## Actual behaviour

* Example: The user stays in the current colony.

iamsamgibbs:
## Description

This PR ensures that when you click a notification that it opens actions in the colony in which it was created. (Note: Similar fixes will be required for some of the other open notifications PRs, it's a very small change being added here so it can be easily copied across regardless of which order the PRs are merged.)

## Testing

* Step 1 - Create a simple payment action in planex, the notification indicator should appear, don't click on the notification yet.
* Step 2 - Switch to the wayne colony.
* Step 3 - Open the UserHub and click on the notification. You should be redirected to the planex colony and the action sidebar should open with the relevant action open. Note also that the notification should be marked as read.
* Step 4 - Navigate to another page on planex, balances for example. Click the notification again. It should open in the sidebar without redirecting you away from the balances page.
* Step 5 - Switch to the wayne colony and create an advanced payment and advance it to the funding stage without checking any of the notifications.
* Step 6 - Switch to the planex colony. Click on any of the advanced payments notifications, you should be redirected to the wayne colony.

## Diffs

**New stuff** ✨

* handleNotificationClicked function now redirects if the notification was not created in the currently open colony

Resolves #3281

### DIFF
--- a/src/components/common/Extensions/UserHub/partials/NotificationsTab/partials/Notification/Expenditure/ExpenditureNotification.tsx
+++ b/src/components/common/Extensions/UserHub/partials/NotificationsTab/partials/Notification/Expenditure/ExpenditureNotification.tsx
@@ -19,12 +19,14 @@ const displayName =
 
 interface NotificationProps {
   colony: NotificationColonyFragment | null | undefined;
+  isCurrentColony: boolean;
   loadingColony: boolean;
   notification: NotificationInterface;
 }
 
 const ExpenditureNotification: FC<NotificationProps> = ({
   colony,
+  isCurrentColony,
   loadingColony,
   notification,
 }) => {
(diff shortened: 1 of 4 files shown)


src/components/common/Extensions/UserHub/partials/NotificationsTab/partials/Notification/Expenditure/ExpenditureNotification.tsx
```diff
@@ -62,14 +64,17 @@ const ExpenditureNotification: FC<NotificationProps> = ({
...
-        `${window.location.pathname}?${TX_SEARCH_PARAM}=${transactionHash}`,
-        {
-          replace: true,
-        },
-      );
+    if (!transactionHash) {
+      return;
     }
+
+    const path = isCurrentColony
```
bassgeta, line 71: This is actually quite crafty, so if we're on the current colony, we just open up the sidebar, if we're not, we navigate to that colony and open up the sidebar :100:

bassgeta (approved):
Nice job fixing this, I can confirm it works like a charm :sunglasses: 
I was thinking if there's a nicer way of solving this with react router, but I couldn't think of one except passing in a base path to the components, which is the same solution but done one level higher :shrug:

rdig (approved):
This works lovely. Very nice work! 

https://github.com/user-attachments/assets/20239d48-d91b-46f6-9dfa-ea83d6ae12a9

rumzledz (approved):
Great work on this @iamsamgibbs 🎉 

I can confirm that on a transaction notification takes me to the Colony where that transaction was made:

Wayne to Planex

balance wayne to planex

Planex to Wayne

PR #3316, opened 2024-10-14
fix: split payment table ui

CzarekDryl:
## Description

- Equal distribution method now uses fractions to calculate amount instead of percentages
- Changed decimals to dots everywhere in payment table
- Total amounts is not higher then the originally intended value
image

## Testing

* Create a split payment and play with it

Closed PR with comments - https://github.com/JoinColony/colonyCDapp/pull/3042

Resolves https://github.com/JoinColony/colonyCDapp/issues/3012

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/forms/SplitPaymentForm/partials/SplitPaymentRecipientsField/hooks.tsx
+++ b/src/components/v5/common/ActionSidebar/partials/forms/SplitPaymentForm/partials/SplitPaymentRecipientsField/hooks.tsx
@@ -1,5 +1,6 @@
 import { createColumnHelper, type ColumnDef } from '@tanstack/react-table';
 import Decimal from 'decimal.js';
+import moveDecimal from 'move-decimal-point';
 import React, { useMemo, useCallback, useEffect } from 'react';
 import { type FieldValues, type UseFieldArrayReturn } from 'react-hook-form';
 
@@ -31,6 +32,7 @@ export const useRecipientsFieldTableColumns = ({
   amount,
   fieldArrayMethods: { update },
   disabled,
+  distributionMethod,
 }: {
   name: string;
   token: Token;
@@ -38,6 +40,7 @@ export const useRecipientsFieldTableColumns = ({
   amount: string | undefined;
   fieldArrayMethods: UseFieldArrayReturn<FieldValues, string, 'id'>;
   disabled?: boolean;
+  distributionMethod?: SplitPaymentDistributionType;
 }): ColumnDef<SplitPaymentRecipientsTableModel, string>[] => {
   const isTablet = useTablet();
   const columnHelper = useMemo(
@@ -106,6 +109,10 @@ export const useRecipientsFieldTableColumns = ({
               <SplitPaymentPayoutsTotal
                 data={dataRef.current || []}
                 token={token}
+                value={
+                  distributionMethod === SplitPaymentDistributionType.Equal &&
+                  (amount ? moveDecimal(amount, token.decimals) : 0)
+                }
                 convertToWEI
               />
               {isTablet && (
@@ -113,7 +120,8 @@ export const useRecipientsFieldTableColumns = ({
                   {parseFloat(
                     (
                       dataRef.current?.reduce(
-                        (acc, _, index) => acc + getPercentValue(index),
+                        (acc, _, index) =>
+                          Number(acc) + Number(getPercentValue(index)),
                         0,
                       ) || 0
                     ).toFixed(4),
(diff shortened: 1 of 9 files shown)


jakubcolony (changes requested):
Dots instead of commas ✅ 
image

image

Redoing action ✅ 
redo-staked-test

I came across a slightly different issue which I'm hoping could be fixed here as well, when manually changing the percentage value (while Equal or Reputation split is selected), entering the first digit causes the field to lose focus. You can't really see that in the recording but I kept pressing 5 and had to refocus the field again:

https://github.com/user-attachments/assets/0ad6d453-caea-4bb8-b4d0-b9e3edb94530

AdrianDudko:
hi @jakubcolony

> I came across a slightly different issue which I'm hoping could be fixed here as well, when manually changing the percentage value (while Equal or Reputation split is selected), entering the first digit causes the field to lose focus. You can't really see that in the recording but I kept pressing 5 and had to refocus the field again:
> 
>  Screen.Recording.2024-10-21.at.22.38.09.mov

This PR (along with the closed, original one) is nearly 2 months old and is becoming unwieldy. @CzarekDryl  has several large open PRs that would be worth closing. I suggest addressing this in a separate issue. fyi @arrenv

jakubcolony:
Hey @AdrianDudko, I've just checked and the issue is not reproducible on master. Once rebased, I'll approve the PR if it is resolved on this branch too.

CzarekDryl:
@jakubcolony I've rebased to master but fix was needed 

https://github.com/user-attachments/assets/bc265faf-4709-40af-918b-3ee748caabdf

jakubcolony (dismissed):
LGTM, nice work @CzarekDryl 👍 


https://github.com/user-attachments/assets/9a5e96a8-a0d1-4217-93c0-908017c23863

iamsamgibbs (changes requested):
Thanks for your continued work on this form, appreciate it is a bit of a pain!

I've found a few issues unfortunately.

If you set the form to 'Equal' then manually change one of the percentages, the total percentage does not change from `100%`.

Screenshot 2024-11-06 at 08 48 09

Adding new rows doesn't change this:

Screenshot 2024-11-06 at 08 48 23

Completed view looks good, but if you redo the action, the total amount is incorrect:

Completed view:
Screenshot 2024-11-06 at 08 49 17

Redo action:
Screenshot 2024-11-06 at 08 49 29

The table rows are also missing a digit when setting to a value like `40.40` it will show `40.4`
 
Screenshot 2024-11-06 at 08 52 12

Let me know if any of these are out of scope for this issue!

CzarekDryl:
@iamsamgibbs `The table rows are also missing a digit when setting to a value like 40.40 it will show 40.4` this is happening also in other actions, so I think it will be good to make other issue for this.
image

src/components/v5/common/ActionSidebar/partials/forms/SplitPaymentForm/partials/SplitPaymentRecipientsField/hooks.tsx
```diff
@@ -106,14 +110,20 @@ export const useRecipientsFieldTableColumns = ({
               <SplitPaymentPayoutsTotal
                 data={dataRef.current || []}
                 token={token}
+                value={
+                  distributionMethodRef?.current ===
+                    SplitPaymentDistributionType.Equal &&
+                  (amount ? moveDecimal(amount, token.decimals) : 0)
```
bassgeta, line 116: can we safely pass `0` here? i would guess that 0 doesn't show up if other values are present, maybe we could pass `undefined` juuust to make sure
image

src/components/v5/common/Fields/InputBase/FormInputBase.tsx
```diff
@@ -47,7 +48,14 @@ const FormInputBase: FC<FormInputBaseProps> = ({
         if (type === 'number') {
           onChange(Number.isNaN(valueAsNumber) ? 0 : valueAsNumber);
         } else {
-          onChange(inputValue);
+          onChange(
```
bassgeta, line 51: i don't really understand this, we have an input that is not of type `number` but it can allow only numbers?
I feel like we should be using type="number" and maybe combine it with `cleave`?

bassgeta (approved):
Nice job with solving this, I am going to approve it, because I think you've probably thought about these edge cases already, but dropping 2 comments nonetheless if they may be useful :D 
I think we could get rid of that `shouldAllowOnlyNumbers` prop and try to use cleave if the issue was with formatting?

It splits nicely when using numbers that don't divide fully :heavy_check_mark: 
image
image
Errors are there and get removed when clicking out of the table :heavy_check_mark: 
image
image

iamsamgibbs (approved):
Nice job! On reflection, ignore my comment about the decimals, I was thinking in terms of GBP / EUR and not tokens.

All the issues I raised in my previous comment are resolved.

Total percentage updates correctly:

Screenshot 2024-11-19 at 14 30 42

Redoing the completed action has the correct amount.

https://github.com/user-attachments/assets/cd457084-e652-49dc-beff-790ca98a2d75

Issue #3286, opened 2024-10-08
Incorrect Recipient validation on search


arrenv:
Teamwork task: [Incorrect Recipient validation on search](https://tw.auroracreation.com/#/tasks/16111632)

## Steps to reproduce

1. Create a new Simple payment action.
2. In the Recipient field, click to open the user dropdown.
3. Enter a value in the search box.
4. Note, the validation for the recipient field is triggered incorrectly.

## Expected behaviour

* The recipient field validation should only trigger if no selection is made.
* The placeholder for the search input should be "Search or enter an address".

## Actual behaviour

* Validation incorrectly shows on the recipient field, when simply using the search input.
* The input is only showing the text "Search".

[validation-on-search.webm](https://github.com/user-attachments/assets/e2f22fb7-d9d9-43f7-b6a9-03655c907a6c)

PR #3317, opened 2024-10-14
fix: simple payment recipent validation on search

adam-strzelec:
## Description

Fix incorrect recipient validation on search

## Testing

- Create simple payment action
- Click `Select member`
- Enter text in input
- Validation is not triggered anymore during entering the text

https://github.com/user-attachments/assets/8b6232ec-8396-48b2-9903-f4e93acb11e8

**Deletions** ⚰️

Delete onSearch callback

Resolves https://github.com/JoinColony/colonyCDapp/issues/3286


iamsamgibbs (changes requested):
Nice job, this is working so much nicer now.

https://github.com/user-attachments/assets/30e1445e-7a7f-4e03-b2a4-4265a19a255d

Compared to master:

https://github.com/user-attachments/assets/b7bb3bf0-0d65-4b6d-a3bd-f858814d8444

The only thing missing from the original issue is the search input placeholder should be `Search or enter an address`

Screenshot 2024-10-14 at 09 18 58

iamsamgibbs (changes requested):
Thanks for the change, this looks right here now:
 
Screenshot 2024-10-14 at 11 03 33

However, it has also changed the team select which is now incorrect:

Screenshot 2024-10-14 at 11 04 03

I've had a quick check through the other action types and it doesn't seem like any other selects are affected, but it could be affecting other searches elsewhere - it might be better to add a specific one here rather than adjusting the search one.

iamsamgibbs (approved):
Nice job, thank you for this fix, looks great now.

Screenshot 2024-10-14 at 15 38 55

Screenshot 2024-10-14 at 15 38 59

jakubcolony (approved):
The fix looks good, not seeing the error as I'm typing anymore, nice work 👍 

recipient-fix-test

rdig (approved):
Tested and seems to be working as expected. Nicely done!


https://github.com/user-attachments/assets/957c6a02-79bc-45ad-a603-2ee0eee2ffcb

Issue #2256, opened 2024-04-18
[ADV.PAYMENTS][PAYMENT BUILDER][REPUTATION METHOD] As a Colony member, I can use REPUTATION in the PAYMENT BUILDER action to make decisions


AdrianDudko:
Github issue: 

## Description
Provides the ability to make decisions in a Payment Builder action, using Reputation Decision Method. 

## Testing
 - Reputation Extention has to be installed
 - 
 
## In the Scope of that story

- In align with Figma design
- Responsive design
- Lock Action
- Funding Step
- Release Step
- Payment step
- Cancel 
- Edit 

## Out of Scope of that story
-

PR #3322, opened 2024-10-14
Feat/15886093 reputation cancel

CzarekDryl:
## Description

This PR introduces options for penalizing or not penalizing the cancellation of staked payments. It also implements a reputation method for canceling after the locked state and the handling of multiple cancellation requests.

## Testing

* Install Reputation Weighted extension
* Create advanced payment using both permissions and staking method
* Lock it and then cancel the payment

Main issue - https://github.com/JoinColony/colonyCDapp/issues/2256


jakubcolony:
@CzarekDryl looks like it needs another rebase on `master`

arrenv (changes requested):
@CzarekDryl Thank you for the work on this, and your patience. I have noticed a couple of things.

1. This translation error appeared when the Advanced payment was created.
image
image


2. Two cancelled pills appeared, there should be only one pill in the timeline.
image
image

3. The staking decision method does not trigger the transaction to create the payment.

4. Cancelling with Reputation after the Funding step resulted in the following, you should still be able to cancel after funding and the cancel pill should appear after the funding step.
image

iamsamgibbs (changes requested):
Nice start of this, I'm able to get through the entire flow.

Screenshot 2024-11-14 at 11 22 45

There is an issue that if you reject the first motion and start another one, when the cancel step reappears it loads into the first motion rather than the staking stage of the second one, so it looks like it is in a broken state.
 
Screenshot 2024-11-14 at 11 21 23

And I'm experiencing the same issue as Arren when attempting to cancel after the funding stage.

Screenshot 2024-11-14 at 11 23 59

It also seemed odd that the cancel motion step completely disappears if you reject the motion, I think you should still be able to access the failed motions and the step should remain visible, especially if there are unclaimed stakes / rewards.

arrenv (changes requested):
@CzarekDryl Thank you for your hard work on this. I noticed a few things while testing:

1. Canceling with Permissions after the Review stage resulted in the first incomplete action timeline, and if after funding, the second screenshot.
image
    image


2. Can these have a hover state when there are multiple cancellation requests:
image

3. In this case I created a motion to cancel the action, then while voting I used Permissions to cancel it right away. It cancelled correctly but stayed in this view. Is it possible to remove all the following steps in the timeline and show a message in this case. Something like this [Figma link](https://www.figma.com/design/l1dOM5qiQYwF0ElvKDqqjg/Design-System---Colony-v3?node-id=6395-12258&t=MGgIIfGaNxjF5qji-4).
image

4. I saw this error in console when revealing a vote:
image

5. It seems if there are any still running motions then the extra steps in the timeline persists, despite a motion already passing and being finalized and the action being cancelled. It should remove all steps if the action has been cancelled.
image

6. Claiming staked tokens shows this broken pending button, however, is it possible to auto return this stakes when cancelling?
image

7. Cancelling with permissions with no penalty showed this timeline, it should have the step label be "Cancel", as the stakes have already be returned, the text should be "The action was canceled and cannot be executed."
image

[Figma link showing expected designs](https://www.figma.com/design/l1dOM5qiQYwF0ElvKDqqjg/Design-System---Colony-v3?node-id=1782-148404&t=MGgIIfGaNxjF5qji-4)

image


image

8. While testing, I noticed that we a missing the indicator as to whether the cancellation motion is penalising or not, can we add a pill to a motion item like this design? [Figma link](https://www.figma.com/design/l1dOM5qiQYwF0ElvKDqqjg/Design-System---Colony-v3?node-id=6396-12341&t=MGgIIfGaNxjF5qji-4).
image

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/CancelModal/CancelModal.tsx
```diff
@@ -1,69 +1,164 @@
...
+
+      if (
+        decisionMethod &&
+        decisionMethod.value === DecisionMethod.Reputation
+      ) {
+        if (penalise) {
+          await cancelStakedExpenditureViaMotion(stakedMotionPayload);
+        } else {
+          await cancelExpenditureViaMotion(motionPayload);
+        }
```
jakubcolony, line 146: Which saga to use should not depend on `penalise` but rather on the expenditure type - if staked, `cancelStakedExpenditureViaMotion` should always be used. You can pass the `penalise` as `shouldPunish` param to the saga.

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/CancelModal/CancelModal.tsx
```diff
@@ -1,69 +1,164 @@
...
+      } else if (penalise) {
+        await cancelStakedExpenditure(stakedPayload);
+        await reclaimExpenditureStake(reclaimPayload);
+      } else {
```
jakubcolony, line 150: Similarly here, `cancelStakedExpenditure` should always be used for staked expenditure regardless of `penalise` value.

Also, if creator was penalised, they lost their stake, so there is nothing to claim.

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/CancelRequests/CancelRequestItem.tsx
```diff
@@ -0,0 +1,80 @@
+import clsx from 'clsx';
```
jakubcolony, line 1: Is there a way to reuse code from similar components? Composition via props could be used for elements specific to this type of action, e.g. the penalise badges.

jakubcolony (changes requested):
Thanks for your work on this PR @CzarekDryl, generally most of the functionality is there, but I spotted a few issues.

There appears to be regression around how cancel in review step is displayed:

image

`master` for comparison:
image

----

The note in the Cancel modal should display the actual role held by the user instead of hardcoded Payer:
image
image

----

I was able to get into a broken state by quickly funding or releasing a payment while the funding motion was loading. The UI should be disabled during that period:


https://github.com/user-attachments/assets/ffe5afe0-9c2e-43c7-84fd-91f534808016

----

When cancelling a staked expenditure with penalise option, the modal gets stuck in Pending. There are errors in the console which are most likely due to the fact the UI is trying to reclaim stake (which has been forfeited) - I commented on where I think it happens in code.
image


image

The same thing happens with cancelling via motions, on motion finalization.

----

Penalise badge has a border that is not in the designs:
- Figma
image
- UI
image

----

## Other flows tested which mostly work well:

Cancelling via permissions, non-staked expenditure:

- At Funding:
image

- At Release:
image

----

Cancelling via reputation, non-staked expenditure:

image
image
image

----

Cancelling via permissions, staked:

image

----

Cancelling via reputation, staked expenditure:

image

image

image

jakubcolony:
Just a note this PR will most likely have a conflict with #3922 as the designs have been updated (as per #3776).

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/CancelRequests/CancelRequestItem.tsx
CzarekDryl: @jakubcolony Can we do this in other issue and refactor all request items?

CzarekDryl:
@jakubcolony Everything should be fixed, please check it :)

arrenv (changes requested):
Nice one @CzarekDryl, it is looking and working really well.

I did still run into a couple of things, not sure if they should be tackled in this PR.

1. If created using staking, and when finalizing a cancelation or using permissions to cancel I received this transaction error in the console. 
image

2. If cancelling with Permissions and penalising, or not, it does not show which option was selected in the card, where it has been added when you use Reputation.
image

3. I received this error in the console when trying to stake, despite having more then enough tokens to stake.
image
image

CzarekDryl:
@arrenv I've added badge while using permissions and it's in finalize pill
image

And I'm not able to replicate console errors:
https://github.com/user-attachments/assets/dc02f19f-90ee-4eff-a8a9-ab72fb9f831b

https://github.com/user-attachments/assets/cae79533-fa27-4bf8-8ab5-ab4a672733a2

arrenv (dismissed):
Thank you @CzarekDryl, looking great, tested all scenarios I could think of and all works as expected.

An active motion with a cancelled motion
image


Cancelling working correctly
image

Reputation and cancel, penalise and not
image

arrenv:
I have just seen this @CzarekDryl, is this related to this PR?

image

CzarekDryl:
@arrenv I've checked and all descriptions work fine for me 
image

arrenv (dismissed):
Re-approving after a rebase

jakubcolony (changes requested):
Thanks for another round of changes @CzarekDryl, mostly looking good but I'm still seeing some issues with staked expenditure.

When cancelling using permissions, the cards are different depending on whether the creator was punished or not.

If punished, no label is displayed informing of this:
image
image

If not punished, an extra Finalize step is added to the stepper: 

image

----

It appears that the UI is still trying to reclaim the stake even though the payment creator was penalised, leading to some errors in the console:
image

----

I am still able to click around the UI while the cancel modal is in pending state. That's how I got the "Paid" label as well as a cancellation request in progress:

image

----

Otherwise, things are working well:

image

image

image

image

image

CzarekDryl:
@jakubcolony All fixed

arrenv (changes requested):
@CzarekDryl Thank you for the updates. Going through Jakub's issued raised, mostly look resolved from what I could tell, except for the extra Finalize step, with some additional issues.

1. I created an Advanced payment using Staking, I confirmed the details, then I cancelled and did not penalize using Permissions. The result was both a "Cancel" and a "Finalize" stepper. Which both had different information, one saying "Penalised", and the other saying "No penalty".

    There should not be an extra Finalize step, the pill should only be "No penalty", the stake amount should be included in the cancel step along with the label "Cancelled" as opposed to "Created".

    image

    image

CzarekDryl:
@arrenv Now reclaim step is showing only when user is penalised: 

image

And here is version for not penalised:

image

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/ActionWithPermissionsInfo/ActionWithPermissionsInfo.tsx
```diff
@@ -124,14 +79,19 @@ const ActionWithPermissionsInfo: FC<ActionWithPermissionsInfoProps> = ({
                   </div>
                 </>
               )}
+              {additionlInfo && additionlInfo}
```
jakubcolony, line 82: typo

jakubcolony (changes requested):
Doesn't seem far off, thanks for your continued work on this.

The note should not be shown when cancelling a payment made using permissions (non-staked):


image

image

The spacing between labels in the action info box is inconsistent:

image

image

image

I _think_ the finalize step should not be shown in both cases (penalised and not penalised) as all the details are already in the cancel step. But I'll let @arrenv confirm that.

Cancelling staked payment while in review should automatically reclaim the stake (this is implemented on the contracts level). However it seems like the wrong method is called by the cancel saga, because it wasn't provided with `stakedExpenditureAddress`:

image

arrenv:
@CzarekDryl & @jakubcolony, yes, when an action has been cancelled, then there should be no finalize step. The person who originally staked the advanced payment should receive their stake back when the action has been cancelled.

The members who staked on the cancel motion, should still be able to claim their stakes back in the prior step within the cancel step, rather then there being an additional Finalize step.

CzarekDryl:
@jakubcolony Spacing between heading are different because of items height, between items they are the same.
image

Now there is only cancel step:
image

jakubcolony (changes requested):
All the previous issues have been resolved, there's just one problem I run into - when cancelling a staked payment with penalising creator, all the steps collapse and it's not possible to select any:

https://github.com/user-attachments/assets/ed036cd0-b683-4976-9b9d-c94e303cfe5e

----

Please have a look at the failing Github check:
image

----

The note is no longer displayed when cancelling a non-staked payment ✅ 
image

Stake is reclaimed when cancelling a staked payment in review stage ✅ 

https://github.com/user-attachments/assets/1b57f05e-dfd8-412f-9a5d-ac047e3b5873

There is no extra step when cancelling a staked payment ✅ 
image

Nice work on this so far, let's get it over the line!

jakubcolony (dismissed):
Cancelling a staked payment with penalising works well now 👍 Nice work!

image
image

iamsamgibbs (changes requested):
This is a huge amount of work, so great job so far!

There are unfortunately a few issues though, some may be out of scope, but I'll list them here regardless.

All cancellation motions have a "No Penalty" pill, even when the "Penalty" option selected.

https://github.com/user-attachments/assets/ebeb1bab-4f67-4c22-b70d-02dcd3627587

If you have an active funding motion when cancelling, the widget becomes unusable.

https://github.com/user-attachments/assets/698169df-6358-49fc-8551-e3fd391aeffd

If you have an active funding motion, then attempt to cancel via reputation, the widget becomes unusable.

https://github.com/user-attachments/assets/844ed8dc-312e-4204-b59d-296f3be9d8ce

Similar issues with an active release motion.

https://github.com/user-attachments/assets/8e9a7e95-8b9a-4235-b86c-794a8e198b84

Otherwise things are mostly working great, again there is a lot of good work here!

I've also left a few comments where translation strings have been missed.

adam-strzelec:
@iamsamgibbs I can't get info wether staker should be penalised because `willPunishExpenditureStaker` from graphql is null.

PR #3326, opened 2024-10-14
Feat: Add HMAC authentication to notifications

### LINKED ISSUE #3051: Notifications: Turn on HMAC Authentication
### Story

Example: As a *notifications user*, I want to *ensure that my notifications are kept private*, so that *I know my colony's data is safe*

### Description

We need to enable HMAC authentication for Magicbell. This can be done here https://app.magicbell.com/projects/8931/settings/security by following these docs https://www.magicbell.com/docs/channels/notification-inbox/hmac-authentication.

This will need a lambda since the HMAC cannot be calculated on the client.

- [ ] We have a new lambda for calculating the HMAC
- [ ] HMAC is turned on in our test Magicbell project (it will also be turned on in the real project)

davecreaser:
## Description

- Add lambda for generating the user notification HMAC. When HMAC authentication is turned on in Magicbell (it hasn't been up until now, but it will be in production), this code must be passed to Magicbell when requesting the notifications to provide an extra layer of security, so that it is not enough to just know the wallet address and our Magicbell API Key to receive a list of a user's notifications. The code is generated using our Magicbell API Secret, and the user wallet address. This is why it must be calculated inside a lambda, and not client-side.

## Testing

First ensure that you have all the correct notifications env vars in your `.env` file. By now, I'm assuming most of you have, but if you haven't worked on any notifications things before then let me know and I can tell you what you need to add.

⚠️ IMPORTANT: You need to also create a `.env` file inside the folder of the new lambda which is located at `amplify/backend/function/getUserNotificationsHMAC` and add the same MAGICBELL_API_SECRET here as you did in the `.env` file for `amplify/backend/function/createUserNotificationsData`. You will then need to restart you environment so that this environment variable is picked up in the lambda.

Now, to test:

* Just go to `http://localhost:9091/planex` with the network tab open.
* Search in your network tab for a call to magicbell to fetch notifications:
Screenshot 2024-10-14 at 17 50 19

* Check the request headers of this request. There should be a header called `x-magicbell-user-hmac`
Screenshot 2024-10-14 at 17 51 16

If you really want to, there's another thing you can do to ensure that this is working, but it's important that you please clean up afterwards.

* Go to `https://app.magicbell.com/projects/8931/settings/security` and toggle the HMAC authentication ON.
* Comment out line 69 of `src/context/Notifications/NotificationsDataContext/NotificationsDataContextProvider.tsx` so that the `userKey` prop is not passed to the `MagicbellProvider` component.
* Refresh the page. The network calls to magicbell should now be failing.
Screenshot 2024-10-14 at 17 53 59

* Go back to `https://app.magicbell.com/projects/8931/settings/security` and toggle the HMAC authentication OFF (this way it does not break Magicbell for everybody else using it to develop). I will toggle this on full-time when this PR is merged and everyone has rebased onto it.

## Diffs

**New stuff** ✨

* New lambda `getUserNotificationsHMAC`

**Changes** 🏗

* Added userKey prop to the MagicbellProvider
* Added `magicbellUserKey` to `notificationsData` model (queried using lambda)

Resolves #3051


davecreaser:
Yeah good catch @rdig - I blame Monday afternoon developing for not seeing that 😬 It was even worse, that actually when querying for all users they would just return all the HMACs with it 🤦 

I've now made it a query which gets the wallet address directly from the header in the lambda, so it will only ever return the HMAC of the user who is currently logged in. Should be much more secure this way.

amplify/backend/function/getUserNotificationsHMAC/src/index.js
```diff
@@ -0,0 +1,29 @@
...
+};
+
+exports.handler = async (event) => {
+  try {
+    await setEnvVariables();
+  } catch (err) {
+    throw new Error('Unable to set environment variables. Reason:', err);
+  }
+
+  const checksummedWalletAddress = event.request.headers['x-wallet-address'];
```
davecreaser, line 22: I took this pattern from the bridge lambdas. Is it safe to assume that the wallet address from the headers has already been checksummed, or do we need to pass it through ethers.utils.getAddress?

rdig: The issue here is that I don't really trust where the address is coming from. It might be checksumed or it might be not

So I'd run it through `getAddress` just to be sure

rdig, line 22: So I still don't think I trust this, as I can just pass the header onto the lambda itself and just get a valid HMAC back without any validation.

If I remember correctly, the `x-wallet-address` header was put in as a util to have access to the address, not necessarily as something secure, because in the end, it's just a string.

I think we might need to involve both @jakubcolony (he set up the header in the first place), @bogdan-1337 (so we know the limits of how we can call the lambdas) as well as any other interested parties and just have a call to figure this out.

But as it stands, I don't think it's secure enough.

davecreaser: So as far as I understand it (and after a brief chat with Bogdan to confirm), we cannot call the lambda from anywhere outside of the AWS console / api, and only with our Colony credentials. When we call the graphql query which resolves using the lambda, we call it with an auth token, not passing the wallet address directly, so even if somebody tried to hack that graphql call they wouldn't be able to just add a different wallet address.

Let's indeed have a call tomorrow when Bogdan is available though, we can figure it out together for sure 💪

rdig (approved):
Yeah, we've established this is not an attach vector, as I initially though.

This was established during a call w/ me, @davecreaser, @jakubcolony and @bogdan-1337 

This is all good now!

jakubcolony (approved):
Nice work, it was good to go over this in a call 👍 

I can confirm the header is being set:

image

I just couldn't test the scenario without the header as I don't have access to Magicbell

iamsamgibbs (approved):
Nice job figuring this out Dave!

hmac header appears:
Screenshot 2024-10-23 at 12 01 15

With hmac enabled and the API key removed:
Screenshot 2024-10-23 at 12 18 33

With hmac enabled and the API key provided:
Screenshot 2024-10-23 at 12 19 15

🎉

iamsamgibbs:
Is it worth adding an example.env to `getUserNotificationsHMAC` and `createUserNotificationsData` ?

PR #3338, opened 2024-10-15
[Notifications-ish] Action Sidebar should automatically close when tx params is removed from the URL

### LINKED ISSUE #3311: [Notifications-ish] Action Sidebar should automatically close when tx params is removed from the URL
## Steps to reproduce

I don't think this has been possible before introducing notifications as usually clicking outside of the action sidebar will force it to close, but notifications introduce components to the userhub which link to other pages without transactions. As a result this is only possible to recreate on `feat/in-app-notifications`.

* Step 1 - Create a mint tokens action
* Step 2 - Click the "mint tokens" notification to have a transaction open in the sidebar.
* Step 3 - Click the "incoming funds" notification which will navigate you to the balances page.

## Expected behaviour

You should be navigated to the balances page and the action sidebar should close.

## Actual behaviour

You are navigated to the balances page, however the action sidebar stays open but losses it's tx hash so the action sidebar changes to "create new action" state.

https://github.com/user-attachments/assets/646354e7-58a0-4c46-9313-defdbe50b8b2

## Suggestions (optional)

`setActionSidebarOpen` is automatically set if there is a tx param in the URL. We should store the previous tx value and if it changes to undefined, forcibly close the action sidebar. Note, this should not affect the ability to open the action sidebar and create a new action.

If this is not easily achievable, the notifications which navigate to a new page (at the time of writing, this applies to incoming funds, and all extension related notifications) should ensure the tx hash is preserved in the URL so if the action sidebar is open it won't switch to the "create new action" state and will instead keep the action open.

mmioana:
## Description

- This PR fixes the issue of having the `Create new action` sidebar replacing the completed action sidebar when navigating to another route by closing it if previously the `tx` parameter was present or the action sidebar is opened.

## Testing

TODO: Please check the action sidebar is closed when navigating to another route which removes the `tx` parameter. Easiest way to test this is via the notifications

* Step 1. Create a mint tokens action

Screenshot 2024-10-15 at 16 22 15

* Step 2. Click the "mint tokens" notification to have a transaction open in the sidebar.
* Step 3. Click the "incoming funds" notification which will navigate you to the balances page.

https://github.com/user-attachments/assets/c3ac460e-abea-4f26-bff5-ac8669adac2f

## Diffs

**New stuff** ✨

* `useLocationChange` hook

Resolves #3311

### DIFF
--- a/src/components/frame/Extensions/layouts/ColonyLayout.tsx
+++ b/src/components/frame/Extensions/layouts/ColonyLayout.tsx
@@ -95,7 +95,7 @@ const ColonyLayout: FC<PropsWithChildren> = ({ children }) => {
   const previousTransactionId = usePrevious(transactionId);
 
   useLocationChange(() => {
-    if ((!!previousTransactionId && !transactionId) || isActionSidebarOpen) {
+    if (!!previousTransactionId && !transactionId && isActionSidebarOpen) {
       toggleActionSidebarOff();
     }
   });
--- a/src/components/frame/Extensions/layouts/ColonyLayout.tsx
+++ b/src/components/frame/Extensions/layouts/ColonyLayout.tsx
@@ -21,6 +21,8 @@ import { useColonyCreatedModalContext } from '~context/ColonyCreateModalContext/
 import { useMemberModalContext } from '~context/MemberModalContext/MemberModalContext.ts';
 import { usePageHeadingContext } from '~context/PageHeadingContext/PageHeadingContext.ts';
 import { useTablet } from '~hooks';
+import useLocationChange from '~hooks/useLocationChange.ts';
+import usePrevious from '~hooks/usePrevious.ts';
 import { TX_SEARCH_PARAM } from '~routes/index.ts';
 import ActionSidebar from '~v5/common/ActionSidebar/index.ts';
 import ColonyCreatedModal from '~v5/common/Modals/ColonyCreatedModal/index.ts';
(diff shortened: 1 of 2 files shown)


rdig (changes requested):
This isn't working as (I) expected. As soon as an action is completed the action sidebar goes away. 

I'm not sure if you intended it to work like this, but this is not a valid fix here


https://github.com/user-attachments/assets/1043a8ef-86f5-4328-ad68-6ccc40a33c45

iamsamgibbs (changes requested):
Thanks for picking up this issue and investigating it!

I'm experiencing the same issue as Raul where the action sidebar disappears when creating an action.

I've also noticed this issue where the action sidebar disappears if you already have a transaction open and click a notification to open a new transaction (the txHash in the URL gets updated and is then forcibly removed by toggleActionSideBarOff)

https://github.com/user-attachments/assets/b487a3b5-fdc0-46c7-9824-fa93e443fa7b

I think you're on the right track though as this approach makes sense, just need to update the if statement inside of useLocationChange. I think `|| isActionSidebarOpen` is just making it close every time there is a location change if it is already open, in fact it seems like changing it to `&& isActionSidebarOpen` might resolve both of these issues.

mmioana:
@rdig @iamsamgibbs thank you for reviewing this. Totally not intended to have this behaviour 🫠 

@iamsamgibbs changing the condition to `&& isActionSidebarOpen` seems to do the trick, thank you so much 🫶 
I am wondering if there are other test scenarios (beside the ones you tried) that I should consider 🤔

iamsamgibbs (approved):
Amazing, thanks for the fix @mmioana - I can't think of any test scenarios that would break this, so looks good to me!

https://github.com/user-attachments/assets/de81cb20-fb47-4e23-be1c-d33857c99fa7

https://github.com/user-attachments/assets/bc043dc9-79b0-47e2-9d25-5798d1d95326

davecreaser (approved):
Nice fix @mmioana, looks to be working great! Tested a few things and I couldn't break it :)

https://github.com/user-attachments/assets/e7bd0e3a-3f27-4e02-ac02-e8deb9a03a24

rdig (approved):
Works as expected now. Nice turn around


https://github.com/user-attachments/assets/579dc62f-a517-4ae0-bf4d-af9200643c93

Issue #3299, opened 2024-10-11
Loading spinner on action timeline for Adv. payments


arrenv:
Teamwork task: [Loading spinner on action timeline for Adv. payments](https://tw.auroracreation.com/#/tasks/16115955)

## Steps to reproduce

1. Create an Advanced payment action.
2. Note, the loading spinner in the right hand side action timeline.

## Expected behaviour

* The skeleton loader should remain until the content loads as opposed to both the skeleton and the loading spinner.

## Actual behaviour

* The loading spinner appears in the action timeline, which should be timeline.
* You see both the skeleton loader and loading spinner on motions.

[loading-wheel.webm](https://github.com/user-attachments/assets/6524de6a-ca86-4353-837e-416ef2673496)

![image](https://github.com/user-attachments/assets/9e4ab673-3f20-4ba7-a8e9-0066835edd46)

jakubcolony:
Hey @arrenv, the screen recording shows a simple payment motion so could this be a motion-specific issue and not advanced payments? I think I'm seeing the correct behaviour on QA:

![adv-payment-loading](https://github.com/user-attachments/assets/d7bc6184-cd66-455b-b6c8-1981eeeba75b)

arrenv:
@jakubcolony I knew there was something funky there, when verifying, I seemed to only be able to replicate it on Adv. payments, but I seemingly ignored my own recording.

It seems fairly inconsistent, I just went and tested it again and am now getting it in the content area.

It seems unrelated to this PR, so, I will raise with Pagepro.

![image](https://github.com/user-attachments/assets/64318344-b63f-4e61-ae06-96fc710a024d)

adam-strzelec:
@arrenv @jakubcolony I noticed that the problem occurs when you create action with reputation method. This might not be so easy to fix, since the content on the right only starts loading once the content on the left has been rendered. I'll have to think about how to fix that.

PR #3347, opened 2024-10-16
fix: loading skeleton for loading actions with reputation

adam-strzelec:
## Description

Fix loading skeleton when open sidebar with reputation action.

## Testing
- Create action with reputation eg. Simple payment
- Open action sidebar with reputation
- Loader spinner on the right is no longer appears

## Diffs

**New stuff** ✨

Add information about motion loading to context

**Changes** 🏗

Before:

[loading-wheel.webm](https://github.com/user-attachments/assets/6524de6a-ca86-4353-837e-416ef2673496)

After:

https://github.com/user-attachments/assets/907dc754-f741-4301-914a-4961de988952

Resolves https://github.com/JoinColony/colonyCDapp/issues/3299

### DIFF
--- a/src/components/v5/common/ActionSidebar/ActionSidebar.tsx
+++ b/src/components/v5/common/ActionSidebar/ActionSidebar.tsx
@@ -107,35 +107,13 @@ const ActionSidebar: FC<PropsWithChildren<ActionSidebarProps>> = ({
   const isLoading =
     transactionId !== undefined && (loadingAction || loadingExpenditure);
 
+  const isContentLoading = isLoading || motionsLoading;
+
   const actionNotFound = transactionId && !action;
 
   const getSidebarContent = () => {
-    if (isLoading && !action) {
-      return <ActionSidebarLoadingSkeleton />;
-    }
-
     if (action) {
-      return (
-        <div className="relative left-0 top-0 flex flex-grow">
-          <div
-            className={clsx('flex flex-grow', {
-              hidden: isLoading || motionsLoading,
-            })}
-          >
-            <CompletedAction action={action} />
-          </div>
-          <div
-            className={clsx(
-              'absolute left-0 top-0 flex h-full w-full flex-grow',
-              {
-                hidden: !motionsLoading && !isLoading,
-              },
-            )}
-          >
-            <ActionSidebarLoadingSkeleton />
-          </div>
-        </div>
-      );
+      return <CompletedAction action={action} />;
     }
 
     if (actionNotFound) {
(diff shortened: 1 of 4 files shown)


rdig (dismissed):
The testing steps are missing the actual extension installation

Screenshot from 2024-10-16 13-14-05

> Create action with reputation 

This is called a "motion"

Other then the above, everything works as expected, the loading spinner has been replaced with a skeleton loader


https://github.com/user-attachments/assets/b2540871-1aac-41d5-89b7-2d4401952422

src/components/v5/common/ActionSidebar/ActionSidebar.tsx
```diff
@@ -109,12 +110,32 @@ const ActionSidebar: FC<PropsWithChildren<ActionSidebarProps>> = ({
...
 
   const getSidebarContent = () => {
-    if (isLoading) {
+    if (isLoading && !action) {
       return <ActionSidebarLoadingSkeleton />;
     }
 
     if (action) {
-      return <CompletedAction action={action} />;
+      return (
```
bassgeta, line 118: could we just return the skeleton `if (isLoading || motionsLoading)` otherwise return the `CompletedAction`?
It's a bit funky to render both at the same time just hide them with CSS

bassgeta (changes requested):
Stuff works nicely, so good job solving this!
I may be missing some context here, what's stopping us from just returning the loading skeleton from the `Motions` component? Feels a bit unnecessary to add yet another property to that large context.

rdig (dismissed):
Looking like you've fixed all of Kerry's issues, so this is still good from my side

bassgeta (changes requested):
Hey! Sorry for pestering your with this PR, but I think that we could probably solve it in the `Motions` component instead of the context.
I've pasted a solution in a comment, could you check it out and see if there's something that it doesn't work for? :D

adam-strzelec:
@bassgeta I have just realised that we don't need context as `isLoading` contains `loadingAction` that is true when motion is fetching. However this was not worked in old approach as motion had been rendered after `isLoading` had been set to false. Check how it works now for you, because it works perfect for me.

bassgeta (approved):
Very nice, really nice catch that we didn't even need to use that and it just works from this component :100: 
image

jakubcolony (approved):
Working as expected, nice work @adam-strzelec 💯 

motion-loader-test

iamsamgibbs (approved):
Nice job, looks good to me! Everything is still loading correctly, looks like a nice simple fix.

Screenshot 2024-11-04 at 11 01 21

PR #3379, opened 2024-10-17
Feat/16099336 table filters

CzarekDryl:
## Description

- Added buttons for filtering and searching streaming payments table
- Added pagination to table
- Added empty state
- Sort by active was removed by Arren
image

## Testing

* Create multiple streaming payments
* Navigate to `/{colony}/streaming-payments` page

Main issue - https://github.com/JoinColony/colonyCDapp/issues/3260


CzarekDryl:
@arrenv Currently filtering by token, status, end condition and sorting by status and streamed amount is not available on backend.

arrenv (changes requested):
@CzarekDryl Thank you for the PR, I noticed a few things while testing.

1. The filters or data in the tables do not seem accurate. Only one in the list should not be started, but all 4 are showing under the active filter, despite there being 3 showing as Not started.
image

2. The End condition also does not seem to be accurate. Only one of this is a Fixed date, for example.
image

3. The End condition filter does not seem to allow you to remove the filter when clicking to remove it.
image

4. Filtering by token type and selecting ETH then CREDS crashed the app:
image

5. Filtering by date does not seem easy to test in local, but, I am getting this error in the console:
image

6. I had trouble get the date range to work, as it would not accept clicking dates, without clicking apply, but the apply button always hidden under the screen. Can this auto fill the date fields when clicking on them and the modal closes.
[date-filter-stream.webm](https://github.com/user-attachments/assets/b05a2130-52f5-4698-bed6-d36f3c18fbe0)

CzarekDryl:
@arrenv Everything should work fine now, please take a look at this again :)

CzarekDryl:
@jakubcolony ETH symbol is back and app shouldn't crash while filtering
image

jakubcolony (changes requested):
Thank you @CzarekDryl, the ETH symbol is showing now but the app still crashes on token/date filter:


https://github.com/user-attachments/assets/fe0d090e-3b3d-4140-beec-5a2e1a43a70c


https://github.com/user-attachments/assets/d8aa5091-54b2-4fa3-aef8-ae5b6bf8af3d

CzarekDryl:
@jakubcolony This should be fixed now, please check it again

https://github.com/user-attachments/assets/e620a16c-2705-4192-9972-9ec50cde58c6

arrenv (changes requested):
@CzarekDryl Thank you for the updates. Please see comments below:

NOTE: none of the streams actually started after a few restarts, so it was hard to test completely. i.e. Stream status, Created date, Total streamed where not able to be tested.

1. Can we please change the empty state description to: "There are no streaming payments to show." So that it is still relevant with filters active.

2. The list should update live when new streams are created, so you don't get the empty state, or missing
image

3. The sorting arrows should only be visible for columns with an active sort, otherwise, they should show on hover.
image

4. Can we prevent the text from moving on expand and collapse.

    [moving-expand.webm](https://github.com/user-attachments/assets/6dea3b38-0c09-400b-b470-0f9821dc958a)

5. There is no tooltip on the amount value - [Figma link](https://www.figma.com/design/v2y9n3CksZFG8dytUcnZWx/Finances?node-id=30402-30459&t=LGEsKtWZ69bVTnBc-4)
image


6. There should be a hover state on the arrow and on that whole row as well - [Figma link](https://www.figma.com/design/v2y9n3CksZFG8dytUcnZWx/Finances?node-id=30402-30495&t=LGEsKtWZ69bVTnBc-4):
image


7. The pagination order changes on load more. Can they stay the same order.

[pagination-order.webm](https://github.com/user-attachments/assets/8998e9da-4e91-4b03-9764-69f221ac9d5a)

8. The search does not seem to filter completely. In this search, I would expect only one payment for leela and fry to show, not all the results for both:
image


9. When filtering by token, I get this error in console.
image

src/components/frame/v5/pages/StreamingPaymentsPage/partials/StreamingPaymentsTable/partials/StreamingPaymentFilters/partials/TokenFilters/TokenFilters.tsx
```diff
@@ -0,0 +1,58 @@
...
+    name: token?.tokenAddress || '',
+  }));
+
+  return (
+    <div>
+      <h5 className="hidden px-3.5 pb-2 uppercase text-gray-400 text-4 sm:block">
+        {formatText({ id: 'balancePage.filter.approvedTokenTypes' })}
+      </h5>
+      <ul>
+        {tokenItems.map(({ label, name }) => {
```
bassgeta, line 37: I don't think it's necessary to map over all tokens and then map over them again, let's just map over `tokenTypesFilters` here and render it without storing it in a variable first

src/components/v5/common/ActionSidebar/partials/ActionSidebarContent/ActionSidebarContent.tsx
```diff
@@ -306,6 +306,11 @@ const ActionSidebarContent: FC<ActionSidebarContentProps> = ({
                 include: [SearchActionsDocument],
               });
             }
+            if (isQueryActive('SearchStreamingPayments')) {
```
bassgeta, line 309: would it be possible to do this invalidation somewhere else more streaming payment related? like its form

CzarekDryl: Moved to completed action component

jakubcolony:
@CzarekDryl Could you rebase when you get a chance please?

CzarekDryl:
@jakubcolony rebase done

jakubcolony (changes requested):
Thanks for another round of changes, I can confirm the app is not crashing on filter selection anymore.

A couple of things I noted:

I have a single stream that ended. The filtering works fine that it shows leela when no filters or "Ended" status filter is selected, but I cannot expand the group:

https://github.com/user-attachments/assets/2ecbc699-bc10-43d0-8d0c-7dcb3916a9a0

This seems to be the case regardless of what streams each group contains:

https://github.com/user-attachments/assets/fb933b56-705d-48cb-9f90-095c41ef4b34

----

There appears to be a possible regression when creating a new stream with Start immediately option. My colony has 4M+ of CLNY so plenty to cover the claimable amount yet I get the insufficient funds warning:

https://github.com/user-attachments/assets/8e389664-a5da-4c21-ab33-89f2a7d14c60

It doesn't happen on `feat/streaming-payments-ui`:

https://github.com/user-attachments/assets/185ef855-26fd-46f6-9648-8591d293d7dc

----

When changing values of an already applied custom date filter, they don't seem to update:

https://github.com/user-attachments/assets/b88a3c80-9f15-4761-8e31-833906074f08

----

Selecting a preset date filter when custom one is not reflected in the filter pills:


https://github.com/user-attachments/assets/56085139-06e2-41f2-babc-22f4311e58ba

This raises a question whether the date filter should be a multi-select. A counterargument would be that selecting a higher order option (e.g. last month) already includes all the lower options (last hour, last day). 

@arrenv Could you confirm the expected behaviour for the above?

----

I believe search should be non-case sensitive. I have several streams called `Active stream`, yet the term `active` doesn't bring any results:
image

image

arrenv:
@jakubcolony I don't it needs to be multi-select. Which, is also the case with the activity feed.

However, it could provide a little tiny big of extra flexibility if you want to show the past hour as well as transactions between a specific date, small use case, but I felt it was easy to address this with filtering at the time I created this issue https://github.com/JoinColony/colonyCDapp/issues/2792 then to change the fields. So, open to other thoughts on this, but for now, we should at least be consistent with the activity feed.

jakubcolony (changes requested):
Thanks for your continuous work on this PR @CzarekDryl.

> I have no idea what could be wrong with the action, as no file or function related to it was changed in this PR — unless something happened during the rebase.

I managed to reproduce it on `feat/streaming-payments-ui` and opened #3901 to track. Not related to this PR.

----

> @jakubcolony I fixed the table opening issue — in the Table, I commented out the code added by Mmioana, which will need to be fixed later (I left a TODO).

I posted a video in one of the comments showing the issue the original code block was solving which unfortunately is reproducing again.

----

I tested the date filtering by manually updating streaming payments createdAt field in the DB (to match their titles). While selecting individual values works fine:
image

image

image
image

Multiple selection behaves as an AND filter instead of an OR. I would expect 3 streams displayed in both cases below:
image
image

----

Previously tested and checked again:

Collapsed view:
image

----

Expanded view:
image

----

No results view:
image

----

Filtering by status:
image
image
image

----

Filtering by end condition:
image

arrenv (changes requested):
@CzarekDryl It is working great, nice work on the updates, it will be great to get this wrapped up so we can release Streaming payments.

I did notice a couple of minor things:

1. This one still needs to be addressed:
    > Can we please change the empty state description to: "There are no streaming payments to show." So that it is still relevant with filters active.

    image

2. The text and icons are still moving when collapsing and expanding rows.

https://github.com/user-attachments/assets/74f203a0-4f81-489a-a10a-bf5dd599ee57

davecreaser (changes requested):
Hey @CzarekDryl, nice work here! Looks like a lot of effort has gone into this and it's looking pretty nice. There's nothing that stands out to me code-wise, so all good there!

I did some testing too and it looks to me like the filters are almost all working. I just ran into two issues that should be looked into (you can see them in my screen recording). These are:
- When I try to filter by both ETH and CREDS at the same time, I get no results.
- When I try to use the sort by in the filters, it doesn't do anything (sorting on the table itself does work however)

https://github.com/user-attachments/assets/56a341d1-2a9b-4c9a-933d-5d295cbfd2eb

Another point - it would be helpful if your PR description and testing instructions can be a little more detailed next time. It's not super easy to come into this with little context (especially a PR this size) and have to spend some time understanding everything by just reading the code and comments. Thanks!

Overall nice job so far though! 👍

CzarekDryl:
@arrenv @davecreaser Fixed and ready to check again

@davecreaser Sorting does work but you need two or more recipients
https://github.com/user-attachments/assets/296c1eee-3865-45a0-a290-500e7e2f6a39

davecreaser (approved):
Thanks for fixing the issues @CzarekDryl. Looking good now! I can filter by both ETH and CREDS and the streams show up as expected.

Screenshot 2025-01-07 at 13 27 40
Screenshot 2025-01-07 at 13 27 57
Screenshot 2025-01-07 at 13 29 01


And you're right, the sorting by total streams works to sort the larger dataset (users) rather than the individual streams. All good there.

arrenv (approved):
@CzarekDryl Thank you for the updates, I am still seeing a couple of minor layout issues, however, I will approve and create a separate issue to avoid this PR continuing on any longer.

Issue #3293, opened 2024-10-10
Skeleton loading states on dark mode


melyndav:
Teamwork task: [Skeleton loading states on dark mode](https://tw.auroracreation.com/#/tasks/16115953)

## Description

Currently, the skeleton loaders when shown in dark mode match the light UI only and don't align with the dark mode colour tokens. 

Having a supported dark state for the skeleton loaders would make the UI in dark mode more cohesive. The scope of this issue is to add support for dark mode using the specified values in [Figma](https://www.figma.com/design/Jn8wu7cmMceG0z7uBjU51s/Dark-mode?node-id=26721-8076&t=CLYLv4HTQQWPTnR4-4). 

The current implementation for light mode should be applied for dark mode across all responsive sizes. 

This is a low priority issue. 

<img width="414" alt="image" src="https://github.com/user-attachments/assets/1402fb60-b064-4c2a-b9a6-a72012cdd9e7">

### Figma link: https://www.figma.com/design/Jn8wu7cmMceG0z7uBjU51s/Dark-mode?node-id=26721-8076&t=CLYLv4HTQQWPTnR4-4

rdig:
Shouldn't this go Pagepro ?

adam-strzelec:
@melyndav I checked in the app and loading skeleton displays with exactly the same colors as in Figma. Both in dark and light mode. I don't quite understand what you mean. Did you mean gradient color? 🙂

melyndav:
Hey @adam-strzelec, when dark mode is active, I want the skeleton loader to match the dark tokens above. Currently, it shows the lighter token and doesn't match with the dark theme. So dark mode needs a custom colour when active.

<img width="312" alt="image" src="https://github.com/user-attachments/assets/7fa4802d-0144-4f93-b51c-98bd466fd1bb">

<img width="362" alt="image" src="https://github.com/user-attachments/assets/920085ad-6222-44b9-a499-93e3b2a277ed">

adam-strzelec:
@melyndav Something like that?
Before:

![Screenshot 2024-10-17 at 15 07 28](https://github.com/user-attachments/assets/8d88401d-8ced-44b2-85be-c3facd69cdec)

After:

![Screenshot 2024-10-17 at 15 08 09](https://github.com/user-attachments/assets/4d8b45fa-33cc-4bc9-85b0-1e19b788850b)

PR #3380, opened 2024-10-17
fix: change loading skeleton gradient for dark mode

adam-strzelec:
## Description

Change gradient color for loading skeleton in dark mode

## Testing

- Set app to dark mode
- Open any loading skeleton while fetching data

**Changes** 🏗

Before:

Screenshot 2024-10-17 at 15 07 28

After:

Screenshot 2024-10-17 at 15 08 09

Resolves [#31415](https://github.com/JoinColony/colonyCDapp/issues/3293)

### DIFF
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -257,6 +257,12 @@ module.exports = {
             animation: 'shine 1.125s ease infinite',
           },
         },
+        '.skeleton-dark': {
+          '.skeleton&::before': {
+            'background-image':
+              'linear-gradient(90deg,rgba(146, 149, 152, 0),rgba(146, 149, 152, 0.5),rgba(146, 149, 152, 0))',
+          },
+        },
         '.break-word': {
           'word-break': 'break-word',
         },
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -107,6 +107,9 @@ module.exports = {
         100: 'var(--color-teams-grey-100)',
         500: 'var(--color-teams-grey-500)',
       },
+      gradient: {
+        base: 'var(--color-skeleton-gradient)',
+      },
     },
     fontFamily: {
       inter: ['Inter', 'sans-serif'],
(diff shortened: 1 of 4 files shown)


rdig (approved):
This seems to be all good, at least as much as I've tested.

However I want to ask you to be more specific with your testing instructions.

> Open loading skeleton

This is very generic, and I can't really tell _which_ loading skeleton I need to test

Screenshot from 2024-10-22 16-19-38
Screenshot from 2024-10-22 16-19-57

iamsamgibbs (changes requested):
I'm afraid this might be a much bigger bit of work than initially anticipated.

This fix only applies to places where the LoadingSkeleton component is used.

So it works great here:
Screenshot 2024-10-22 at 17 08 25

But the fix isn't applied anywhere that just uses the `skeleton` class and not the LoadingSkeleton, such as here:

Screenshot 2024-10-22 at 17 09 30

There are at least 47 places where it is used just as a class name and having to use the `usePageThemeContext` hook everywhere just to apply this change is going to be a real pain. Is there definitely no way to make this change in the tailwind config to use this background-image for the `skeleton` class when in dark mode so we can avoid adding the `skeleton-dark` class everywhere?

arrenv:
47 file changes does not seem worth the effort at all for a slight color change, so, if this cannot be unified as a part of this PR then I would suggest scrapping this issue.

rdig (changes requested):
In light of recent changes I will rescind my approval

adam-strzelec:
@arrenv @rdig @mmioana @iamsamgibbs You're right. I didn't notice that skeleton can be called in other ways in the application than just through LoadingSkeleton. I've fixed that and it should work in the whole application now.

iamsamgibbs (approved):
Nice solution, this seems to work for me.

Screenshot 2024-10-24 at 08 34 56

Screenshot 2024-10-24 at 08 35 07

Screenshot 2024-10-24 at 08 35 13

Light mode is unaffected:

Screenshot 2024-10-24 at 08 35 57

Screenshot 2024-10-24 at 08 36 04

mmioana (approved):
Definitely a global and more clean approach to solving this issue 💯 

Nice work @adam-strzelec 🙌 

Dark mode
Screenshot 2024-10-24 at 10 26 15
Screenshot 2024-10-24 at 10 27 07

Light mode
Screenshot 2024-10-24 at 10 26 25
Screenshot 2024-10-24 at 10 26 50

tailwind.config.js
```diff
@@ -250,8 +253,7 @@ module.exports = {
             'background-color': theme('colors.gray.100'),
             'background-repeat': 'no-repeat',
             'min-height': '1em',
-            'background-image':
-              'linear-gradient(90deg,rgba(255, 255, 255, 0),rgba(255, 255, 255, 0.5),rgba(255, 255, 255, 0))',
+            'background-image': `linear-gradient(90deg,rgba(255, 255, 255, 0),${theme('colors.gradient.base')},rgba(255, 255, 255, 0))`,
```
rumzledz, line 256: Yessssss yes yes yes 💯

rumzledz (approved):
Clean and wonderful fix @adam-strzelec !! ✨ 

dark mode loader

adam-strzelec:
@rdig I think I need your approval to merge it as now the merging is blocked.

rdig (approved):

rdig:
@adam-strzelec done! :100:

PR #3417, opened 2024-10-21
Add colony sdk custom abi contracts

chmanie:
## Description

This is an attempt to allow for custom contracts, based on AbiType and Colony SDK. This will set the foundation of moving to Colony SDK entirely (from colony-js) and for a more rapid dev cycle, without the need to always create new ABIs and a new ColonyJS version.
The API will be the same as in Colony SDK so it is typed and (hopefully) well documented.

## Testing

There is an example here: https://github.com/JoinColony/colonyCDapp/pull/3417/files#diff-928abc3bd5c819369d41139150531cc7243d674015496e3fd983b7e298f355a8R268-R270.

This is the way a custom contract could be initialized and used (using the ABI directly from the network docker container)

## Diffs

**New stuff** ✨

* Add a docker mount to network-files to access the ABIs directly
* Update ColonyJS, Colony SDK, `@colony/events`, to allow for custom contracts
* Add a `getCustomContract` function to the ColonyManager 

**Changes** 🏗

* THIS DISABLES THE VERSION CHECK OF COLONY JS! WITH GREAT POWER COMES GREAT RESPONSIBILITY. IF SOMETHING DOES NOT WORK, CHECK IF YOUR CONTRACT VERSION IS CORRECT!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,7 @@
         "@colony/colony-js": "^8.0.0-next.0",
         "@colony/events": "^4.0.0-next.0",
         "@colony/redux-promise-listener": "^1.2.0",
-        "@colony/sdk": "^3.0.0-next.1",
+        "@colony/sdk": "^3.0.0-next.2",
         "@colony/unicode-confusables-noascii": "^0.1.2",
         "@hookform/resolvers": "^2.9.10",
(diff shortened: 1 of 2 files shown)


chmanie:
Already incorporated in a bunch of other PRs

Issue #3392, opened 2024-10-18
[Landing Page] Landing Page updates


AdrianDudko:
## Related PR's

- [ ] 1
- [ ] 2
- [ ] 3
- [ ] 4

## Vision

Create an engaging and informative landing page for the Colony app, where users can quickly understand what the platform offers, connect their wallet, and navigate to where they need to go. The landing page should serve as the gateway to the rest of the app, with a strong call to action for new users while providing returning users with an easy way to continue where they left off.

## Scope
Notion link:
https://www.notion.so/colony/Landing-page-8b8e63c5b0d444b09b41229d5c139415

## **Overview**

The **Landing Page** will act as the main entry point for Colony users. It will communicate the value proposition of Colony, highlight key  features (e.g., decentralized finance for teams), and provide users with a call to action (e.g., connect their wallet, join the beta).

The landing page will also inform users of benefits of the platform, thereby also serving as a conversion or lead generator.

## **User stories**

- **As a first-time user**, I want to understand the key features of Colony and quickly connect my wallet so I can start exploring the platform or request access.
- **As a returning user**, I want to re-connect my wallet quickly to get back into my existing colonies and continue working on my decentralized finance tasks.
- **As a potential user**, I want to easily take action to register my interest or understand where to find more information.

## **User Flows**

- **New User Flow**:
    - The user arrives at the landing page.
    - The user reads about the key features of Colony in the “Decentralized finance for teams” section.
    - The user sees a prominent “Connect Wallet” button.
    - The user sees a strong call to action to request access.
- **New Invited User Flow**:
    - The user arrives at the landing page.
    - The user reads about the key features of Colony in the “Decentralized finance for teams” section.
    - The user sees a prominent “Connect Wallet” button.
    - User sees the option to join the colony they were invited to.
- **Returning User Flow**:
    - The user arrives at the landing page.
    - The user clicks on the “Connect Wallet” button.
    - After connecting their wallet, the user sees a list of their joined colonies to return to.
    - Clicking on a colony takes them to that colony’s dashboard.

## Design

- Figma design - Landing page
    - https://www.figma.com/design/j0VEfFryOkmjrVVq2PItso/Beta-Launch?node-id=16532-6757&t=tIyBQePetzAI1muS-4
        
[...]

PR #3425, opened 2024-10-22
feat: create colony cards for landing page

adam-strzelec:
## Description

Create colony cards in all variants for landing page

## Testing

- Run storybook
- Open `Colony Card`

## Diffs

**New stuff** ✨

Screenshot 2024-10-22 at 11 52 25

Main issue - https://github.com/JoinColony/colonyCDapp/issues/3392


src/components/frame/LandingPage/ColonyCard.tsx
```diff
@@ -0,0 +1,107 @@
+import { Plus } from '@phosphor-icons/react';
+import clsx from 'clsx';
+import React from 'react';
+
+import LoadingSkeleton from '~common/LoadingSkeleton/LoadingSkeleton.tsx';
+import { formatText } from '~utils/intl.ts';
+import Button from '~v5/shared/Button/Button.tsx';
+
+export interface ColonyCardProps {
+  colonyName?: string;
```
bassgeta, line 10: I think all of the conditional rendering we're doing with skeletons and handling the create new colony function is putting way too many things into this component.
I would recommend the following steps to make this more sustainable:
1. Create 2 new components, a `ColonyCard` and `CreateNewColonyCard`
2. Instead of having all the props optional because the app may be loading, I would recommend creating a file structure something like this:
- `ColonyCardHandler` is a component which gets the colony and the loading state, if the app is loading it renders `ColonyCardSkeleton`, if we have all the necessary data, we render `ColonyCard` with all the props sent in defined and not nullable
- then the same for `CreateColonyCard` (can this one even be impacted by anything loading?)

bassgeta (changes requested):
Nice start on this, design wise stuff is looking quite on point :ok_hand: 
However, I think we could benefit from some more splitting of components according to what they do, I've left a comment regarding that, feel free to reach out if I wasn't clear enough!
But essentially, I think we shouldn't have all of our new components receive a loading flag and render everything wrapped in `LoadingSkeleton` components.
It would be much more concise and readable to handle loading in a different component

mmioana (changes requested):
Nice work so far on this one @adam-strzelec 👏 

All 3 scenarios for the cards look good, but please check @bassgeta and my comments for refactoring the `ColonyCard` component as it will make it easier to maintain and add more types of cards in the future

Screenshot 2024-10-23 at 09 58 38
Screenshot 2024-10-23 at 09 58 45
Screenshot 2024-10-23 at 09 58 54

src/components/frame/LandingPage/ColonyCard.tsx
adam-strzelec: @bassgeta Ok, I will add `ColonyCardHandler` when I start creating logic for Landing page

bassgeta (changes requested):
Lightspeed turnaround here, much appreciated :ok_hand: 
All the components look good, and my oh my this composition is truly sweet to look at, really great job!
The only reason I'm requesting changes is because I would like to see us pulling all the i18n strings into the components that use them, for [example](https://github.com/JoinColony/ColonyFrontEndLivingStandard/blob/master/codeQuality/index.md#i18n)

After this is resolved, I will happily approve this :D

bassgeta (approved):
:fire: looking good, thanks for the changes and bringing the landing page one step closer towards completion!

mmioana (approved):
Way more composable implementation of these cards 🙌 thanks so much for your fast changes @adam-strzelec 💯

 
Screenshot 2024-10-24 at 10 35 54
Screenshot 2024-10-24 at 10 36 02
Screenshot 2024-10-24 at 10 36 09

Everything looks nice and will approve your PR 👍 
However a question is popping in my mind if it wouldn't be better to structure all these variants under a single storybook `Story` as in the end they should only represent a variant of a component?

arrenv (changes requested):
@adam-strzelec Thank you for the PR, I have noticed a couple of things.

### Colony card
1. Long Colony names are not handled with an ellipsis, breaking the layout.
image

2. Can we have the case of "1 members", be "1 member". I think it is worth doing as it will be seen by people when first creating a colony.
image

3. Can we have `cursor: pointer` when hovering over the card.

### Create New Colony Card
1. Can we have `cursor: pointer` when hovering over the whole card, as well as the button, as the whole should be clickable.

adam-strzelec:
@mmioana I added it to one story in my last commit 🙂

src/components/frame/LandingPage/ColonyCards/CreateNewColonyCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+export interface ColonyCreateCardProps {
+  invitationsRemaining: number;
+  onCreate: () => void;
+}
+
+const MSG = defineMessages({
+  remaining: {
+    id: `${displayName}.remaining`,
+    defaultMessage: `{remaining} INVITATIONS REMANING`,
```
rumzledz, line 19: I think we should set a proper pluralisation rule here:

```js
defaultMessage: `{remaining} {remaining, plural, one {INVITATION} other {INVITATIONS}} REMANING`,
```

src/components/frame/LandingPage/ColonyCards/CreateNewColonyCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  return (
+    <BaseColonyCard
+      onClick={() => onCreate()}
+      onMouseOver={() => setIsHovered(true)}
+      onMouseLeave={() => setIsHovered(false)}
+      isClickable
+      avatarPlaceholder={
+        <div className="flex h-8 w-8 items-center justify-center rounded-full bg-gray-100">
+          <Plus size={18} className="text-gray-900 " />
```
rumzledz, line 45: **[Nitpick alert]:** Extra whitespace at the end 👀

src/components/frame/LandingPage/ColonyCards/CreateNewColonyCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+        </div>
+      }
+    >
+      <div className="flex flex-1 flex-col gap-1">
+        <div className="flex">
+          <span className="rounded bg-blue-100 px-[.1875rem] py-[.1563rem] text-2xs font-extrabold text-blue-400">
```
rumzledz, line 51: Font weight should be 700: `font-bold`

Screenshot 2024-10-26 at 00 15 57

src/components/frame/LandingPage/ColonyCards/ColonyCard.tsx
```diff
@@ -0,0 +1,51 @@
...
+        src={colonyAvatar}
+        alt="Colony avatar"
+        className="w-8 rounded-full"
+      />
+    }
+  >
+    <p className="flex-1 truncate text-ellipsis text-md font-medium">
+      {colonyName}
+    </p>
+    <p className="max-w-[150px] justify-self-end truncate text-ellipsis text-xs font-normal">
```
rumzledz, line 39: This should also use the `text-gray-600` class as per Figma designs:

Screenshot 2024-10-26 at 00 20 27

src/components/frame/LandingPage/ColonyCards/ColonyCard.tsx
```diff
@@ -0,0 +1,51 @@
...
+      <FormattedMessage
+        {...MSG.members}
+        values={{ members: membersCount.toLocaleString('en-US') }}
+        {...(membersCount === 1 ? { defaultMessage: '1 Member' } : {})}
```
rumzledz, line 43: You can also use i18n's built in pluralisation rule:

Remove: `{...(membersCount === 1 ? { defaultMessage: '1 Member' } : {})}`

And update your string to:

```js
defaultMessage: `{members} {members, plural, one {Member} other {Members}}`,
```

rumzledz (changes requested):
Thanks for working on this feature @adam-strzelec 👊 

## Loading state

Please correct me if I'm wrong but this is what I found on Figma:

https://www.figma.com/design/j0VEfFryOkmjrVVq2PItso/CDapp-Landing-page-%2B-Beta-Launch?node-id=16596-34185&t=cHgkD8OJ2DkdRRkM-4

image

And it seems that the Loading Skeleton is a slightly different from your version:

| Your version | Figma version |
| ---- | ---- |
| image | image |

Maybe I'm looking at the wrong place or something?

## Create new colony hover state

It seems like on Figma, the SVG isn't needed when hovering over it:

image

But please confirm this with product, it might just be a slight mistake

## Tailwind linting

I noticed that your code hasn't automatically been formatted by your linter. Can you please double check your local dev environment?

image

adam-strzelec:
@rumzledz Thans for review. It seems like designs for skeleton changed recently. I already added improvements.

src/components/frame/LandingPage/ColonyCards/CreateNewColonyCard.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+export interface ColonyCreateCardProps {
+  invitationsRemaining: number;
+  onCreate: () => void;
+}
+
+const MSG = defineMessages({
+  remaining: {
+    id: `${displayName}.remaining`,
+    defaultMessage: `{remaining} {remaining, plural, one {INVITATION} other {INVITATIONS}} REMANING`,
```
iamsamgibbs, line 19: The wording here should be "colonies" remaining, not "invitations".

Screenshot 2024-10-28 at 10 38 15

tailwind.config.js
```diff
@@ -112,6 +112,7 @@ module.exports = {
       inter: ['Inter', 'sans-serif'],
     },
     fontSize: {
+      '2xs': ['.5rem', 1.5],
```
iamsamgibbs, line 115: If we are adding this we should update the other uses of `text-[0.5rem]` and `text-[.5rem]` to use this. (Maybe at a later point in a separate PR). It is also used in frequently enough that it would be fine to just use this as an arbitrary value here instead of adding it to the config.

iamsamgibbs (changes requested):
Nice work on this so far! The component structure is very nice.

I've left one comment about the wording here:

Screenshot 2024-10-28 at 10 48 14

Not matching the figma design:

Screenshot 2024-10-28 at 10 38 15

The loading state is missing the bar on the right:

Screenshot 2024-10-28 at 10 32 07

Figma:
Screenshot 2024-10-28 at 10 32 18

Large numbers over 100 million should use abbreviations (100M members):

Screenshot 2024-10-28 at 10 53 44

Which should remove the need to truncate the member count like this (I was going to point out that the word members should remain visible when the count is truncated, but with the abbreviation change, it shouldn't occur):

Screenshot 2024-10-28 at 10 40 06

arrenv (changes requested):
@adam-strzelec Thank you for the updates.

1. I am still getting the text wrapping issue with large numbers and names. It seems the cause of this is that the number abbreviation stops after a million. While I don't think this numbers are realistic to ever be hit, could still be worth fixing.
image

2. Yes, the component was auto updated, the "+" should still be visible on button hover. Also, can the hover for the button and the card be independent, i.e. the button hover state only triggers if specifically hovering over the button.
image

iamsamgibbs:
@adam-strzelec Might be worth checking the Numeral component and seeing how numbro is used there for abbreviations.

arrenv (approved):
Very nice @adam-strzelec, thank you for the updates! All looks good to me now!

image

image

PR #3428, opened 2024-10-22
fix: validation mode for checking if colony url exists form

### LINKED ISSUE #3309: [Create Colony Flow]: Incorrect message is shown for a taken URL in the "Custom Colony URL" field
## Steps to reproduce

1. Go to the "Create a colony" page
2. Enter an already existing URL in the "Custom Colony URL" field
3. Observe the message displayed

## Expected behaviour

* The message "This colony URL is already taken" should be displayed when entering an existing URL

## Actual behaviour

* The message "URL available" is incorrectly shown even though the URL is already taken

image

bassgeta:
## Description

I solved this by comparing it to the colony name creation, figuring out that one works and this one doesn't, what's the difference?
Well basically it wasn't a race condition, but it instead "revalidated" when the field changed focus :melting_face: 

## Testing

1. Run the create data script
2. Run `node ./scripts/create-colony-url.js` and open the URL
image
3. Into the URL field, enter `planex` and wait for this to appear
image
4. Enter something else and URL should be available
image

Notice how you don't see the URL available flashing into URL already taken behaviour described in the original issue

## Diffs

**Changes** 🏗

* `onChange` revalidation mode to the colony name step

Resolves #3309

### DIFF
--- a/src/components/common/Onboarding/wizardSteps/CreateColony/StepColonyName.tsx
+++ b/src/components/common/Onboarding/wizardSteps/CreateColony/StepColonyName.tsx
@@ -45,7 +45,11 @@ const StepColonyName = ({
     user?.profile?.displayName ?? splitWalletAddress(user?.walletAddress ?? '');
 
   return (
-    <Form<Step1> onSubmit={nextStep} validationSchema={validationSchema}>
+    <Form<Step1>
+      onSubmit={nextStep}
+      validationSchema={validationSchema}
+      mode="onChange"
+    >
       <HeaderRow
         heading={MSG.heading}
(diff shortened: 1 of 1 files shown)


iamsamgibbs (approved):
What a lovely simple fix. 10 / 10

https://github.com/user-attachments/assets/a0954381-19fc-418c-9542-04abf0328130

rumzledz (approved):
Clean fix! ✨ 

fixxx

rdig (approved):
Very nice!


https://github.com/user-attachments/assets/93f4f180-c976-4dea-be06-319f42bf4268

PR #3432, opened 2024-10-22
Feat: Allow overriding user's liquidation address

jakubcolony:
## Description

In preparation for this month's payroll, we needed a way to override some of the liquidation addresses returned from Bridge and replace them with ones created under Metacolony. This PR adds the ability to override liquidation address for a given wallet address using env variable (locally) and SSM parameter (in QA and prod).

## Testing

Create or update the .env file of the bridgeXYZMutation lambda (`amplify/backend/function/bridgeXYZMutation/.env`). It should have the following 3 values in it:
```
BRIDGE_API_URL=https://api.sandbox.bridge.xyz
BRIDGE_API_KEY=<sandbox key, ping me if you don't know it>
LIQUIDATION_ADDRESS_OVERRIDES="{"0xb77D57F4959eAfA0339424b83FcFaf9c15407461": "test address override"}"
```

The last variable is a JSON object mapping wallet addresses to their respective overrides. In the example, it will return the test string for leela's address. 

**Important:** Restart your dev env for the changes to take effect.

Run the following query:
```gql
query {
  bridgeGetUserLiquidationAddress(userAddress:"0xb77D57F4959eAfA0339424b83FcFaf9c15407461")
}
```

Confirm the value you get back matches the override.

Note: This will not impact the address shown in the C2F account tab, only the address that is retrieved when making payments.

### DIFF
--- a/amplify/backend/function/bridgeXYZMutation/src/handlers/getUserLiquidationAddress.js
+++ b/amplify/backend/function/bridgeXYZMutation/src/handlers/getUserLiquidationAddress.js
@@ -1,4 +1,3 @@
-const fetch = require('cross-fetch');
 const { graphqlRequest } = require('../utils');
 const { getLiquidationAddresses, getExternalAccounts } = require('./utils');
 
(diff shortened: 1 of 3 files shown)


iamsamgibbs (approved):
Nice job Jakub, seems to be working as expected!

Screenshot 2024-10-23 at 12 35 02

rdig (approved):
All good here. Nicely done! :100: 

Screenshot from 2024-10-23 16-24-07

bassgeta (approved):
Looks legit and works legit :sunglasses:

PR #3437, opened 2024-10-23
fix: validate that token name doesn't contain only spaces

### LINKED ISSUE #3298: [Create Colony Flow]: It is possible to enter only spaces in the "Token name" field
## Steps to reproduce

1. Open Create your new colony page
2. Proceed to the your colony native token step
3. Select create a new token radio button
4. Enter spaces in the "Token name" field

## Expected behaviour

* It is impossible to enter only spaces in the "Token name" field

## Actual behaviour

* It is possible to enter only spaces in the "Token name" field

image

bassgeta:
## Description

Added a validation test to check if the string isn't just whitespaces. I however allow for leading and trailing whitespaces, but they get trimmed out when creating the token? :shrug: 

## Testing

1. Start up your dev env and run `node ./scripts/create-colony-url.js`
2. Select "Create a new token"
image
3. Now try to break the token name validation, entering only spaces should fail
image
But spaces at the start/end should work if you have any content
image

## Diffs

**Changes** 🏗

* `createTokenValidationSchema` now includes whitespace only validation for `tokenName` 

Resolves #3298

### DIFF
--- a/src/components/common/Onboarding/wizardSteps/CreateColony/validation.ts
+++ b/src/components/common/Onboarding/wizardSteps/CreateColony/validation.ts
@@ -42,6 +42,7 @@ const { formatMessage } = intl({
   'error.invalidToken': 'Invalid address.',
   'error.tokenNotFound':
     'Token data not found. Please check the token contract address.',
+  'error.tokenNameInvalid': 'Invalid token name',
   'error.tokenNameRequired': 'Enter a token name to continue',
   'error.tokenSymbolRequired': 'Enter a token symbol to continue',
   'error.tokenNameLength': 'Token name should be 255 characters or fewer',
@@ -116,6 +117,11 @@ export const createTokenValidationSchema = object({
       then: (schema) =>
         schema
           .max(MAX_TOKEN_NAME, '')
+          .test(
+            'isNotOnlySpaces',
+            formatMessage({ id: 'error.tokenNameInvalid' }),
+            (value) => value.trim().length > 0,
+          )
           .required(formatMessage({ id: 'error.tokenNameRequired' })),
       otherwise: (schema) => schema.notRequired(),
     }),


iamsamgibbs (approved):
Nice job this is working as expected!
 
Screenshot 2024-10-24 at 09 13 43

I guess we need a product decision on whether leading / trailing spaces?

Screenshot 2024-10-24 at 09 14 12

Either way it looks like we'll need to address how the colony name is displayed with spaces in elsewhere in the app as it doesn't deal well with multiple spaces:
 
Screenshot 2024-10-24 at 09 15 00

Screenshot 2024-10-24 at 09 14 18

It does retain the spaces in the db though, so this is purely a UI issue:

Screenshot 2024-10-24 at 09 15 20

None of that prevents this from getting merged though!

rdig (approved):
All good here, it now validates against empty spaces

Screenshot from 2024-10-24 20-33-47
Screenshot from 2024-10-24 20-34-00
Screenshot from 2024-10-24 20-34-03

However I can still bypass it with Unicode spaces _(in this case `U+0020`)_

Screenshot from 2024-10-24 20-35-06
Screenshot from 2024-10-24 20-35-14
Screenshot from 2024-10-24 20-35-17

This should not make the focus of this PR, and arguably we shouldn't bother with Unicode validates as most nobody does _(just test this on any major service's forms, and you won't get validated against -- this is actually my go to when I don't want to provide a name :stuck_out_tongue:)_.

However one could also argue that since this is a Token and it's going to be deployed "forever" with that name, it might become confusing, so we might want to prevent against that. _(maybe run this validation through our "confusable characters" lib that we use for usenames?)_

PS: for testing unicodes I use this website https://www.editpad.org/tool/invisible-character

bassgeta:
> All good here, it now validates against empty spaces
> 
> Screenshot from 2024-10-24 20-33-47 Screenshot from 2024-10-24 20-34-00 Screenshot from 2024-10-24 20-34-03
> 
> However I can still bypass it with Unicode spaces _(in this case `U+0020`)_
> 
> Screenshot from 2024-10-24 20-35-06 Screenshot from 2024-10-24 20-35-14 Screenshot from 2024-10-24 20-35-17
> 
> This should not make the focus of this PR, and arguably we shouldn't bother with Unicode validates as most nobody does _(just test this on any major service's forms, and you won't get validated against -- this is actually my go to when I don't want to provide a name 😛)_.
> 
> However one could also argue that since this is a Token and it's going to be deployed "forever" with that name, it might become confusing, so we might want to prevent against that. _(maybe run this validation through our "confusable characters" lib that we use for usenames?)_
> 
> PS: for testing unicodes I use this website https://www.editpad.org/tool/invisible-character

I agree that we could run it through our username validator, on it! Good old Alt + 255 spaces <3

iamsamgibbs (commented):
This change does indeed now prevent against the invisible character, however it also prevents emoji from being used in a token name which was previously allowed:
 
Screenshot 2024-10-24 at 09 25 44

Screenshot 2024-10-25 at 11 49 31

(The invisible character issue also affects colony names on the first step of the process, and the error validation there is a bit different than it is here. It would be nice to unify the experience on both fields, but I don't want to increase the scope here if you don't want to tackle it now).

rdig:
Unicode chars in general are dabateable, but emoji definitely needs to be allowed

rumzledz:
Hey @bassgeta ! This isn't a change request but more of a question. I noticed that the field now seems to be validating on blur 🤔 Is this intentional?

blur-validation

bassgeta:
Apologies for ignoring this one for so long :sweat_smile:  I've reset it to the first commit and we can deal with unicode shennanigans later on I guess? :')

iamsamgibbs (approved):
All good!

Just spaces are invalid

Screenshot 2024-11-29 at 10 29 40

With spaces at start / middle / end

Screenshot 2024-11-29 at 10 30 20

Emoji are allowed

Screenshot 2024-11-29 at 10 30 35

rdig (approved):
In that case, all good to go! :100:

mmioana (approved):
All good on this one @bassgeta 💯 

Entering only spaces triggers the error message to be shown
Screenshot 2024-12-03 at 09 50 35

While using leading and trailing spaces doesn't
Screenshot 2024-12-03 at 09 51 00

Nice job! 🥇

src/components/common/Onboarding/wizardSteps/CreateColony/validation.ts
```diff
@@ -116,6 +117,11 @@ export const createTokenValidationSchema = object({
       then: (schema) =>
         schema
           .max(MAX_TOKEN_NAME, '')
+          .test(
+            'isNotOnlySpaces',
+            formatMessage({ id: 'error.tokenNameInvalid' }),
+            (value) => value.trim().length > 0,
+          )
           .required(formatMessage({ id: 'error.tokenNameRequired' })),
```
mmioana, line 125: @bassgeta however in order to not trigger the validation upon blur when there is no value, I believe the 
```
         .test(
            'isNotOnlySpaces',
            formatMessage({ id: 'error.tokenNameInvalid' }),
            (value) => value.trim().length > 0,
          )
```
has to be after `.required(formatMessage({ id: 'error.tokenNameRequired' }))`

PR #3400, opened 2024-10-18
Feat: Add payout claimed notification

davecreaser:
## Description

Add a new notification for when a payout is claimed from an expenditure, and change the expenditure finalized notification message to instead say "Payment released".

Block ingestor changes here: https://github.com/JoinColony/block-ingestor/pull/284

## Testing

* Create an advanced payment with a few different claim delays.
Screenshot 2024-10-18 at 15 46 37

* Go through all the steps for the expenditure, until you get to the final stage (when it has been finalized). You should receive a notification about "Payment released". If you had any payments with a zero claim delay, you should also receive a notification about "Payment made".
Screenshot 2024-10-18 at 15 47 36

* Advance time to unlock some more payouts for claiming by running `npm run forward-time -- 1` in a terminal. The final number in this script is the number of hours to advance by. In my example, I advanced time by 1 hour to unlock two more payouts.
Screenshot 2024-10-18 at 15 47 55
Screenshot 2024-10-18 at 15 48 14

* In the action sidebar, make the payments that are available and check that you received a notification for each claim.
Screenshot 2024-10-18 at 15 48 30
Screenshot 2024-10-18 at 15 48 38

* Repeat the above two steps until all payouts are claimed.
Screenshot 2024-10-18 at 15 48 51
Screenshot 2024-10-18 at 15 49 07
Screenshot 2024-10-18 at 15 49 17
Screenshot 2024-10-18 at 15 49 43

## Diffs

**New stuff** ✨

* New notification type `EXPENDITURE_PAYOUT_CLAIMED`

**Changes** 🏗

* Changes in the block ingestor https://github.com/JoinColony/block-ingestor/pull/284
* Changed the message for expenditure finalized notifications to "Payment released:"

Resolves #3389

PR #3438, opened 2024-10-23
Feat: In app notifications

davecreaser:
In app notifications full feature.

Auth proxy: https://github.com/JoinColony/colonyCDappAuthProxy/pull/31
Block Ingestor: https://github.com/JoinColony/block-ingestor/pull/287


jakubcolony (approved):
Very nice, well done @davecreaser and everyone else who contributed! 🎉 

Had a quick look through the PRs and all looks good. When I try to run the branch locally, I'm still having the issue I noted in #3400:
> I had a bit of a trouble getting the notifications dev key to generate. After I switched branches and ran npm run prepare, there was no entry for MAGICBELL_DEV_KEY in .env so the script could not replace it properly. I had to add that var with empty value to .env.local and run prepare again for it to work.

What are your thoughts on the above? Should the dev key entry be included in the `.env.local` as well? Or maybe the script can append it to the file if it's not there already?

davecreaser:
> Very nice, well done @davecreaser and everyone else who contributed! 🎉
> 
> Had a quick look through the PRs and all looks good. When I try to run the branch locally, I'm still having the issue I noted in #3400:
> 
> > I had a bit of a trouble getting the notifications dev key to generate. After I switched branches and ran npm run prepare, there was no entry for MAGICBELL_DEV_KEY in .env so the script could not replace it properly. I had to add that var with empty value to .env.local and run prepare again for it to work.
> 
> What are your thoughts on the above? Should the dev key entry be included in the `.env.local` as well? Or maybe the script can append it to the file if it's not there already?

Yeah thanks for raising this again Jakub - I've added all the notifications vars to .env.local 😄

PR #3446, opened 2024-10-23
fix: update UserHub transactions to have correct styles for 720-768px screen

### LINKED ISSUE #3056: UserHub Transactions Errors Styles Break on Medium Viewport Width
When viewing the UserHub transactions list at a resolution above `720px` and below `768px`, the error state for the transaction gets wrapped up almost on a single vertical line _(see screenshot)_.

This is most likely the fault of some `flexbox` configuration that acts upon this resolution range and does not shrink accordingly

Screenshot from 2024-09-03 12-48-18

I do want to note however that anything below `720` and above `768` works as expected

Screenshot from 2024-09-03 12-48-10
Screenshot from 2024-09-03 12-48-45

Nortsova:
## Description
For 720-768px error message was broken: 
image

I really hope that my fix will not break anything else, but all I tested now works 😄  

## Testing

* Step 1. Login with MetaMask
* Step 2. Create Simple payment motion
* Step 3. Reject transaction
* Step 4. Open UserHub -> Transactions and ensure that all screens looks good

image
image
image
image

Resolves  #3056

### DIFF
--- a/src/components/v5/shared/NotificationBanner/NotificationBanner.tsx
+++ b/src/components/v5/shared/NotificationBanner/NotificationBanner.tsx
@@ -14,6 +14,7 @@ const NotificationBanner: FC<NotificationBannerProps> = ({
   callToAction,
   descriptionClassName,
   callToActionClassName,
+  contentClassName,
 }) => {
   return (
     <div
@@ -40,7 +41,12 @@ const NotificationBanner: FC<NotificationBannerProps> = ({
           })}
         />
       ) : null}
-      <div className="flex flex-1 flex-col items-start gap-2 @[600px]/notificationBanner:flex-row @[600px]/notificationBanner:items-center">
+      <div
+        className={clsx(
+          'flex flex-1 flex-col items-start gap-2 @[600px]/notificationBanner:flex-row @[600px]/notificationBanner:items-center',
+          contentClassName,
+        )}
+      >
         <div className="flex flex-1 flex-col items-start gap-2 text-md break-word">
           {children}
           {description ? (
--- a/src/components/common/Extensions/UserHub/partials/TransactionsTab/partials/GroupedTransactionContent.tsx
+++ b/src/components/common/Extensions/UserHub/partials/TransactionsTab/partials/GroupedTransactionContent.tsx
@@ -99,6 +99,7 @@ const GroupedTransactionContent: FC<GroupedTransactionContentProps> = ({
           <NotificationBanner
             status="error"
             callToActionClassName="w-full no-underline"
+            contentClassName="@[600px]/notificationBanner:flex-col"
             callToAction={
               <div className="flex justify-between">
                 <button
--- a/src/components/v5/shared/NotificationBanner/types.ts
+++ b/src/components/v5/shared/NotificationBanner/types.ts
@@ -10,4 +10,5 @@ export type NotificationBannerProps = PropsWithChildren<{
   callToAction?: ReactNode;
   descriptionClassName?: string;
   callToActionClassName?: string;
+  contentClassName?: string;
 }>;


iamsamgibbs (changes requested):
This fixes transactions with long error messages:

Screenshot 2024-10-24 at 09 47 29

But short ones are aligned centre:

Screenshot 2024-10-24 at 09 46 32

And it breaks notification banners in a bunch of places unfortunately:

Screenshot 2024-10-24 at 09 44 09

Screenshot 2024-10-24 at 09 47 45

The notification banner is used in a bunch of different places so you'll have to be careful with the changes here:

Screenshot 2024-10-24 at 09 37 31

I found the original PR where this change was made #1381 so the reasoning for it is a bit clearer.

src/components/v5/shared/NotificationBanner/NotificationBanner.tsx
```diff
@@ -40,7 +40,7 @@ const NotificationBanner: FC<NotificationBannerProps> = ({
           })}
         />
       ) : null}
-      <div className="flex flex-1 flex-col items-start gap-2 @[600px]/notificationBanner:flex-row @[600px]/notificationBanner:items-center">
+      <div className="flex flex-1 flex-col items-start gap-2 @[600px]/notificationBanner:items-center">
```
rumzledz, line 43: This is just a humble suggestion @Nortsova -- to save you the hassle since this seemingly affects other areas where we use this NotificationBanner component, maybe you can add a prop for this component and adjust these styles specifically for the GroupedTransactionContent, which appears to be the only area that's being affected

And because we know your style changes are definitely fixing the issue, we can try:

```js
<div
  className={clsx(
    'flex flex-1 flex-col items-start gap-2 @[600px]/notificationBanner:flex-row @[600px]/notificationBanner:items-center',
    contentClassName,
  )}
>
```

Then in GroupedTransactionContent:

```js
<NotificationBanner
  ...existingProps
  contentClassName="@[600px]/notificationBanner:flex-col" // <<-- Add me 😄 
>
```

Nortsova: Thank you @rumzledz , I decided to go with this approach. 🙌 

After some investigation, I realized that this issue arises because:
1. There's a rule for a container width of 600px (only for 720-768px does the container width become larger).
2. In UserHub, we have a `callToAction` prop containing two buttons that attempt to fill the full width.

An alternative solution I considered is to remove the `callToAction` from the component entirely.
And instead of this:
```
<NotificationBanner
 ...existingProps
  callToAction={<Buttons />}
>
```

Do something like:
```
<NotificationBanner
 ...existingProps
>
   {....content}
   <Buttons>
</NotificationBanner>
```

but this case requires me to fix a lot of styles.

Nortsova:
Thank you, @iamsamgibbs and @rumzledz, for reviewing! It’s crucial to update this component without impacting other areas where it’s used. 🙌

rumzledz (approved):
Retested and all good on my side @Nortsova ✨ 
 
all good anastasiia

iamsamgibbs (approved):
Nice job fixing this up, all looks good now!

Screenshot 2024-10-29 at 10 19 30
Screenshot 2024-10-29 at 10 20 20
Screenshot 2024-10-29 at 10 20 48
Screenshot 2024-10-29 at 10 22 54

I agree with you that this should probably be refactored at some point so the content / call to action can just be composed in which would give us more flexibility, but appreciate that would be a much bigger change and this works well for now.

rdig (approved):
All good now.

Screenshot from 2024-10-29 18-25-10

https://github.com/user-attachments/assets/a18f3b6a-1b0a-4b35-8409-73e0d0dd8bad

Issue #3305, opened 2024-10-11
Staged payment metadata and activity feed items not showing correctly


arrenv:
Teamwork task: [Staged payment metadata and activity feed items not showing correctly](https://tw.auroracreation.com/#/tasks/16115956)

## Steps to reproduce

1. Create a Staged payment action.
2. Navigate to the activity feed page and view the newly created Staged payment.
3. Note that the activity feed metadata shows incorrectly.

## Expected behaviour

* The metadata line in the action and activity feed should following the spec in the format:
 `Staged Payment of {token_amount} {token_symbol} to {recipient_username} with {number_of_milestones} milestones by {creator_username}`
* For multiple token types:
 `Staged Payment with {number_of_token_types} tokens to {recipient_username} with {number_of_milestones} milestones by {creator_username}`

## Actual behaviour

* Staged payment metadata and activity feed items not showing correctly, in the action it just shows as the initial version, in the activity feed it displays as if it is a standard Advanced payment.

![image](https://github.com/user-attachments/assets/a3ce9143-79fd-4051-be02-88ae2264edce)

PR #3512, opened 2024-10-29
feat: add staged payment description

CzarekDryl:
## Description

- Added description to staged payment
- I had to modify a few components to retrieve the recipient's data

## Testing

* Install staged payment extension
* Create action with one and many token addresses to check both descriptions

Resolves https://github.com/JoinColony/colonyCDapp/issues/3305


CzarekDryl:
@jakubcolony Redo action is working fine now, I've also refactored code that you mentioned.

https://github.com/user-attachments/assets/d26d3023-00e0-438f-8374-604e02c364ef

src/components/v5/common/CompletedAction/partials/CompletedExpenditureContent/CompletedExpenditureContent.tsx
```diff
@@ -64,7 +70,42 @@ const CompletedExpenditureContent: FC<CompletedExpenditureContentProps> = ({
...
+    };
+  });
+
+  const summedAmount = stages.reduce((acc, { amount, tokenAddress }) => {
+    const token = allTokens.find(
+      ({ token: currentToken }) => currentToken.tokenAddress === tokenAddress,
+    );
+    const formattedAmount = moveDecimal(
+      amount || '0',
+      getTokenDecimalsWithFallback(token?.token.decimals),
```
jakubcolony, line 98: It seems there is no need for this conversion from and then back to wei.

jakubcolony (dismissed):
Yep, redoing payment works fine now. Nice work @CzarekDryl 👍 

image
image

src/components/v5/common/CompletedAction/partials/CompletedExpenditureContent/CompletedExpenditureContent.tsx
CzarekDryl: Conversions removed

bassgeta (approved):
Seems legit, really nice catch with our current action description not following through if we used the liquidation address! Sensible changes all around, the new hook instead of the helper is also very simple and easy to use :100: 
1 token payment:
image
2 tokens payment: 
image

Just a thought here (not sure if it should be tackled here), but in the recent activity table it says `Payment to 2 recipients` instead of using the correct format. Probably it's enough work for another issue, just pointing it out in case we forgot to add it in here!

jakubcolony (approved):

rdig (approved):

PR #3520, opened 2024-10-29
Fix: "Upgrade" button on Calmity Banner should be enabled for all users

### LINKED ISSUE #3340: Colony version upgrades: Only display sitewide banner to those with ROOT permissions
## Description

Currently in Production, I am shown the sitewide calamity banner within MetaColony, notifying me of an available network upgrade. However, I do not have the necessary permissions to perform the upgrade, resulting in a banner with a disabled button.

For everyday users who experience the same issue, this could lead to a frustrating experience, as they may not understand why the button is disabled or why they cannot perform the upgrade.

This sitewide notification banner for new network versions should only be displayed to users with the appropriate ROOT permissions across all responsive sizes. 

image

iamsamgibbs:
## Description

This PR ensures the "Upgrade" button on the Calamity Banner is always enabled for all users when a new colony version is available.

## Testing

To save you the hassle of deploying a colony on an older version, we must first update the `canColonyBeUpgraded` function in `src/utils/checks/canColonyBeUpgraded.ts` to return true.

* Step 1 - Connect to a colony as a user with root permissions. Confirm the Calamity Banner is visible and the upgrade button is enabled.

Screenshot 2024-10-29 at 15 13 47

* Step 2 - Connect to a colony as a user without root permissions. Confirm the Calamity Banner is visible and the upgrade button is enabled.

## Diffs

**Deletions** ⚰️

* Removed `canUpgradeColony` and related uses from `useCalamityBannerInfo`

Resolves #3340

### DIFF
--- a/src/components/frame/Extensions/layouts/hooks.tsx
+++ b/src/components/frame/Extensions/layouts/hooks.tsx
@@ -13,17 +13,13 @@ import { useLocation } from 'react-router-dom';
 
 import { Action } from '~constants/actions.ts';
 import { useActionSidebarContext } from '~context/ActionSidebarContext/ActionSidebarContext.ts';
-import { useAppContext } from '~context/AppContext/AppContext.ts';
 import { useColonyContext } from '~context/ColonyContext/ColonyContext.ts';
 import useColonyContractVersion from '~hooks/useColonyContractVersion.ts';
-import useTransformer from '~hooks/useTransformer.ts';
 import {
   COLONY_FOLLOWERS_ROUTE,
   COLONY_MULTISIG_ROUTE,
 } from '~routes/routeConstants.ts';
-import { getAllUserRoles } from '~transformers/index.ts';
-import { canColonyBeUpgraded, hasRoot } from '~utils/checks/index.ts';
-import { extractColonyRoles } from '~utils/colonyRoles.ts';
+import { canColonyBeUpgraded } from '~utils/checks/index.ts';
 import { formatText } from '~utils/intl.ts';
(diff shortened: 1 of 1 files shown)


arrenv:
@iamsamgibbs Thank you for picking this up, sorry for the confusion on this issue as you have done the work already. @melyndav happened to not have Reputation enabled when running into this originally.

However, hiding the banner completely for everyone but root is not the desired behavior. As it could pose a security risk, and everyone should be informed of a new version.

We already handle permissions in the Upgrade colony version action, and we also allow a colony to be upgraded using Reputation.

So, the correct behavior here should be that the upgrade button is never disabled, and should always open the "Upgrade colony version" action.

arrenv (changes requested):
As above.

iamsamgibbs:
No problem Arren! Updated to match the new specs.

rumzledz (approved):
Smashed it Sam! 💪 

## 1. Updated the canColonyBeUpgraded function to return true

Screenshot 2024-10-31 at 18 28 46

## 2. Connected Leela's wallet (has root perms)

Screenshot 2024-10-31 at 18 29 24

## 3. Connected Amy's wallet (has no root perms)

Screenshot 2024-10-31 at 18 29 42

rdig (approved):
Nicely fixed! :100: 

As `leela` (root)

Screenshot from 2024-10-31 17-20-15

As `fry`

Screenshot from 2024-10-31 17-20-39

arrenv (approved):
@iamsamgibbs Thank you, and sorry again for the mess around. Looks good to me :)

PR #3560, opened 2024-10-30
[Dashboard] Fix: Scrollbar shifts main content

### LINKED ISSUE #3513: QA Dashboard: Switching team shifts the main content container
## Description

When selecting a team from the dashboard in the new team switcher component, the main content blocks shifts on load. 

Export-1730205124136

And for inside pages:

Export-1730205251665

This appears to be only on desktop. 

## Steps to reproduce

1. Log in with a MetaMask wallet
2. On the Dashboard, select a different team using the new team switcher component.
3. As the content loads, watch the main content container move. 
4, navigate to the teams page.
5. Select the general team.
6. Watch the content move on load.

## Expected behaviour

* All content globally in the main content container should not shift when a different team is selected using the new team switcher component.

iamsamgibbs:
## Description

This PR adds scrollbar-gutter: stable to the main content area. It reserves space for the scrollbar so it will prevent content from jumping around when the scrollbar appears / disappears. It is not fully supported, so we should be aware that this does not resolve this issue for Safari / iOS (https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter).

## Testing

Set the screen size so that there is no scrollbar. Resize so that the scrollbar appears. The content should not jump around.

This branch:
https://github.com/user-attachments/assets/e3b42737-69cd-46a0-92d5-8d3848d0b209

Master:
https://github.com/user-attachments/assets/5d7e8ce7-ad9c-4678-857e-5453722e358e

## Diffs

**New stuff** ✨

* Added `scrollbar-gutter-stable` utility class
* Added new class to main content area

Resolves #3513

### DIFF
--- a/src/components/v5/frame/PageLayout/PageLayout.tsx
+++ b/src/components/v5/frame/PageLayout/PageLayout.tsx
@@ -66,7 +66,7 @@ const PageLayout: FC<PropsWithChildren<PageLayoutProps>> = ({
             </section>
             <section
               className={clsx(
-                'modal-blur h-full w-full overflow-auto px-6 md:p-8 md:pb-0 md:pt-2',
+                'modal-blur h-full w-full overflow-auto px-6 scrollbar-gutter-stable md:p-8 md:pb-0 md:pt-2',
                 {
                   'md:!pt-[1.125rem]': !pageTitle,
                 },
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -270,10 +270,13 @@ module.exports = {
         '.no-scrollbar::-webkit-scrollbar': {
           display: 'none',
         },
+        '.scrollbar-gutter-stable': {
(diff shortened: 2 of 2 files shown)


mmioana (approved):
Nice and lovely fix @iamsamgibbs thanks for picking this issue up 🙌 

The content no longer jumps when the scrollbar is shown/hidden 💯 

https://github.com/user-attachments/assets/2f3d6af3-fc4c-479e-aeab-c7dbadf6a486

rumzledz (approved):
Gotta love these CSS trickeries 🪄 🎩 ⭐ 

no shift

Nortsova (approved):
Thanks @iamsamgibbs for this fix, it seems that not only dashboard was fixed but other pages as well 🙌 

no-scroll-jump
no-jump-2

PR #3582, opened 2024-10-31
Fix: Missing permissions banner should not show when reenabling

### LINKED ISSUE #3337: Multi-sig permissions extension installation issues
## Steps to reproduce

1. Install the Multi-sig permissions extension.
2. Note that during the installation, the transaction will fail.
3. Reload the app.
4. Note that the extension will be installed, and enabled but the permissions banner will appear.
5. Note that the status will say Installed and not Enabled.
6. Deprecate the extension.
7. Reenabled the extension again.
8. Note that the permission banner will appear again.

## Expected behaviour

* When installing the Multi-sig permissions extension, it should be enabled and assign the correct permissions in one flow.
* The transaction should not fail during installation.
* The status should be accurate. Raised here https://github.com/JoinColony/colonyCDapp/issues/3266

## Actual behaviour

* Installing the Multi-sig permissions extension fails.
* You should not have to manually assign permissions to the extension.

[multi-sig-permissions.webm](https://github.com/user-attachments/assets/4d9a618f-edc7-4820-822c-ae9926c8c736)

### LINKED ISSUE #3430: [Bug] Missing permissions banner incorrectly shows when reenabling an extension
Very low priority as it is pretty convoluted to reproduce, is not breaking and is easily resolved with a page refresh. It also only occurs if you deprecate an extension immediately after giving it permissions via the missing permissions banner then immediately re-enable the extension without refreshing at any step in between - so incredibly unlikely anyone will run into this without deliberately trying to break it.

## Steps to reproduce

* Step 1 - Switch to the metamask wallet so you can reject specific transactions
* Step 2 - Install the Staged Payments extension. Accept the `install` transaction but reject the `setUserRoles` transaction.
* Step 3 - Refresh so that the missing permissions banner appears (it shouldn't be necessary to refresh, there is a PR fixing this: #3349)
* Step 4 - Click "Enable permissions" in the missing permissions banner and accept the `setUserRoles` transaction.
* Step 5 - Deprecate the extension.
* Step 6 - Re-enable the extension.

## Expected behaviour

The "missing permissions banner" should not be visible as the extension already has the required permissions.

## Actual behaviour

The "missing permissions banner" will be visible.

https://github.com/user-attachments/assets/b5d6af2c-f9e6-4c69-b9dc-cdfec3c7503c

## Suggestions (optional)

This only occurs when assigning permissions via the "missing permissions banner", if done as part of the install transactions group it won't occur. This is because the extension is refetched after being installed, but is not after assigning permissions via the banner. I've left a TODO in #3349 for where this could be added, but it will break the current implementation of the success step which will require refactoring. (I would suggest either waiting for #3349 to be merged or branching off of that to tackle this issue).

iamsamgibbs:
## Description

This PR fixes an issue where the "Missing permissions" banner would show when re-enabling an extension, even if the extension had previously been assigned permissions.

## Testing

* Step 1 - Switch to the metamask wallet so you can reject specific transactions.
* Step 2 - Navigate to the Multi-Sig extension page.
* Step 3 - Click Install. Accept the "Installation" transaction but reject the "setUserRoles" transaction.
* Step 4 - The "Missing permissions" banner should be visible. Do not interact with it yet.
* Step 5 - Deprecate the extension. The "Missing permissions" banner should disappear.
* Step 6 - Re-enable the extension.
* Step 7 - The "Missing permissions" banner should reappear.
* Step 8 - Click "Enable permissions" on the banner and accept the transaction.
* Step 9 - The "Missing permissions" banner should switch to a green success state.
* Step 10 - Deprecate the extension. The success banner should remain.
* Step 11 - Re-enable the extension. The success banner should still remain.
* Step 12 - Refresh. There should be no permissions banner visible.

https://github.com/user-attachments/assets/9ab1bf6a-46a9-4ce7-8122-139d1102deb6

Further testing:
- Repeat the above steps, refreshing each step of the way to make sure the correct banner shows at the expected times.

## Diffs

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

* Refactored the PermissionsNeededBanner to show the success state correctly

Resolves #3430
Resolves #3337

### DIFF
--- a/src/components/frame/Extensions/pages/ExtensionDetailsPage/partials/ExtensionDetailsHeader/ExtensionDetailsHeader.tsx
+++ b/src/components/frame/Extensions/pages/ExtensionDetailsPage/partials/ExtensionDetailsHeader/ExtensionDetailsHeader.tsx
@@ -7,7 +7,7 @@ import { isInstalledExtensionData } from '~utils/extensions.ts';
 import { formatText } from '~utils/intl.ts';
 import ExtensionStatusBadge from '~v5/common/Pills/ExtensionStatusBadge/index.ts';
 
-import PermissionsNeededBanner from '../PermissionsNeededBanner.tsx';
+import ExtensionPermissionsBanner from '../ExtensionPermissionsBanner/ExtensionPermissionsBanner.tsx';
 
 import ActiveInstalls from './ActiveInstalls.tsx';
 import { extensionsBadgeModeMap, extensionsBadgeTextMap } from './consts.ts';
@@ -53,9 +53,10 @@ const ExtensionDetailsHeader: FC = () => {
 
   return (
     <>
-      {isPermissionsBannerVisible && (
-        <PermissionsNeededBanner extensionData={extensionData} />
-      )}
+      <ExtensionPermissionsBanner
+        extensionData={extensionData}
+        isPermissionsBannerVisible={isPermissionsBannerVisible}
+      />
       <div className="flex min-h-10 flex-col flex-wrap justify-between sm:flex-row sm:items-center sm:gap-6">
         <div className="flex w-full flex-col flex-wrap gap-4 sm:flex-row sm:flex-nowrap sm:items-center sm:gap-6">
           <div className="flex flex-col sm:grow sm:flex-row sm:items-center sm:gap-2">
--- a/src/components/frame/Extensions/pages/ExtensionDetailsPage/partials/ExtensionPermissionsBanner/ExtensionPermissionsBanner.tsx
+++ b/src/components/frame/Extensions/pages/ExtensionDetailsPage/partials/ExtensionPermissionsBanner/ExtensionPermissionsBanner.tsx
@@ -1,0 +1,40 @@
+import React, { useState } from 'react';
+
+import { type AnyExtensionData } from '~types/extensions.ts';
+
+import PermissionsEnabledBanner from './PermissionsEnabledBanner.tsx';
+import PermissionsNeededBanner from './PermissionsNeededBanner.tsx';
+
+const displayName =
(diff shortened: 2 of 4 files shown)


rdig (approved):
Tested this and works as expected. Very nice work here! :100: 

Screenshot from 2024-11-01 17-45-41
Screenshot from 2024-11-01 17-45-45
Screenshot from 2024-11-01 17-46-06
Screenshot from 2024-11-01 17-46-19
Screenshot from 2024-11-01 17-46-35
Screenshot from 2024-11-01 17-46-50
Screenshot from 2024-11-01 17-47-11
Screenshot from 2024-11-01 17-47-23

bassgeta (approved):
Really nice solution with the handler component and breaking up the banners into 2 specific components, a bit less black magic is always appreciated :D
Rejected set user roles, banner for missing permissions shows up :heavy_check_mark: 
image
Disappears when deprecated :heavy_check_mark: 
image
It's back after enabling :heavy_check_mark: 
image
Green banner shows up :heavy_check_mark: 
image
Stays there when deprecating it :heavy_check_mark: 
image
And it disappears on refresh :heavy_check_mark: 
image

rumzledz (approved):
All good on my side Sam!

## Installed but rejected the `setUserRoles` transaction:

Screenshot 2024-11-06 at 04 12 00

## Yellow banner notifying you that you're missing permissions

Screenshot 2024-11-06 at 04 12 04

## Deprecated it, banner not present

Screenshot 2024-11-06 at 04 12 22

## Re-enabled it, yellow banner present

Screenshot 2024-11-06 at 04 12 38

## Accepted the `setUserRoles` transaction
 
Screenshot 2024-11-06 at 04 12 58

## Reload the page, banner gone!

Screenshot 2024-11-06 at 04 13 39

PR #3586, opened 2024-10-31
Fix: Notifications do not load when switching wallets

### LINKED ISSUE #3508: Notifications don't load when switching wallets
## Steps to reproduce

1. Ensure you have two users with notifications.
2. While connected to the app with one user, go into Metamask and switch wallets to the other user.
3. The app will prompt you to connect with the second account.
4. After connecting it will direct you back to the colony.
5. Check your notifications tab in the Userhub.
6. Note if the expected notifications load.

## Expected behaviour

* Switching connect wallets should still allow you to see relevant notifications for the user.

## Actual behaviour

* Notifications don't load at all if you switch your connected wallets.
* In order to get notifications to work, you need to reload the page or fully disconnect your wallet.

[notifications-switch-wallet.webm](https://github.com/user-attachments/assets/a9a1e0e0-7394-4500-a95d-47c0ac940d6e)

iamsamgibbs:
## Description

[https://github.com/JoinColony/colonyCDappAuthProxy/pull/33](The actual fix is in this auth proxy PR: https://github.com/JoinColony/colonyCDappAuthProxy/pull/33)

This PR aims to address an issue where notifications do not load when switching wallets.

This was being caused by the `useGetUserNotificationsHmacQuery` in the `NotificationsDataContextProvider` being sent before the auth cookie had been set. Which resulted in the `getUserNotificationsHMAC` lambda returning undefined as the x-wallet-address header is not set.

My proposed solution here is to change how the auth proxy handles un-auth queries. By default the auth proxy allows all un-auth queries and only blocks mutations. This change will still allow un-auth queries by default, but will restrict certain queries (currently just `getUserNotificationsHMAC`) from executing when not auth.

This causes the `useGetUserNotificationsHmacQuery` to be reattempted when the wallet connection has properly finished and notifications will load correctly.

Note: I have tried to avoid making changes to the auth proxy by preventing the `useGetUserNotificationsHmacQuery` from being triggered until the auth headers are set, but was unable to get this working. If this is the preferred solution and anyone has any ideas on how to resolve this, I'm happy to re-investigate this.

## Testing

* Step 1 - Spin up the dev environment from scratch so that it uses the new auth proxy commit (or stop your auth proxy docker and run the branch separately).

We first want to make sure we have two users with notifications enabled and setup.

* Step 2 - Connect with dev wallet 1 and go to planex. Refresh to be extra sure the `createUserNotificationsData` mutation has ran.
* Step 3 - Disconnect your wallet.

Screenshot 2024-10-31 at 15 17 18

* Step 4 - Connect with dev wallet 2, you should automatically load into planex. Refresh to be extra sure the `createUserNotificationsData` mutation has ran.

Now that we have two users with notifications enabled we can test properly.

* Step 5 - Disconnect your wallet again and connect with dev wallet 1.
* Step 6 - Create a simple payment action and send tokens to Amy.
* Step 7 - Check you receive a notification.
* Step 8 - Disconnect your wallet. Without refreshing, connect with dev wallet 2, you should automatically load into planex.
* Step 9 - Check you have notifications without refreshing.

## Diffs

**New stuff** ✨

* Auth proxy can now block non-auth queries as necessary

Resolves #3508


mmioana (approved):
Awesome work on this one @iamsamgibbs 👏 🥇 💯 

After connecting the wallets, the magic bell user id was created for both `leela` and `amy`
Screenshot 2024-11-01 at 13 00 54
Screenshot 2024-11-01 at 13 01 14

Created a simple payment from `leela` to `amy`
Screenshot 2024-11-01 at 13 02 01

`leela` received notifications
Screenshot 2024-11-01 at 13 02 06

And switched to `amy`'s wallet without a page refresh and the notifications were there
Screenshot 2024-11-01 at 13 02 23

Also the check in `auth-proxy` for the `getUserNotificationsHMAC` query has passed ✅ 
Screenshot 2024-11-01 at 13 05 57
Screenshot 2024-11-01 at 13 06 07

rdig (approved):
All good here, tested and working as you're describing, the notifications show up properly for the second account.

Screenshot from 2024-11-04 15-04-20
Screenshot from 2024-11-04 15-04-34
Screenshot from 2024-11-04 15-04-51

PR #3590, opened 2024-10-31
fix: hide multi-sig error if there are still loadings

### LINKED ISSUE #3242: Not enough multi-sig permissions error flashes
## Steps to reproduce

Note: This does not happen all the time, but did happen a number of times.

1. Create a New team action.
2. Select the "Multi-sig" decision method.
3. Note the error banner that flashes in then disappears.

## Expected behaviour

* No error banner should appear unless it actually needs to, it should also not flash in and out at all.

## Actual behaviour

* The not enough multi-sig permissions error banner flashes when selecting the "Multi-sig" decision method.

[error-flashing-multi.webm](https://github.com/user-attachments/assets/d5615d2e-5119-410f-ad4d-b4f47fdcc742)

Nortsova:
## Description
The not enough multi-sig permissions error banner flashes when selecting the "Multi-sig" decision method.
[372548146-d5615d2e-5119-410f-ad4d-b4f47fdcc742.webm](https://github.com/user-attachments/assets/b4058c9d-3e49-4e49-98de-0b33993adc34)

## Testing
Step 1. Enable Multi-sig extension 

> [!TIP]
> Give leela Multi-sig permission in General to have Authority "via Multi-sig" available 

Step 2. Create a Simple payment motion
image

Step 3. Verify that Multi-Sig error is NOT flashed

no-multi-sig-flash

### I will give you a small explanation on this solution: 

image

Because `isLoading` was `false` by default and set to `true` using the async function, it was the very first render with `"isLoading": false`. This led to our component understanding it as "okay, we have all data rendered and it's empty—let's render an error then."

So I changed it for the very first rendering `isLoading` will be `true`, and at line 60 we call the **async** function, and after the function is executed `isLoading` state will be changed to `false`. 🙌 

Resolves  #3242

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/ActionSidebarContent/partials/MultiSigMembersError/useHasEnoughMembersWithPermissions.ts
+++ b/src/components/v5/common/ActionSidebar/partials/ActionSidebarContent/partials/MultiSigMembersError/useHasEnoughMembersWithPermissions.ts
@@ -51,7 +51,7 @@ export const useHasEnoughMembersWithPermissions = ({
 
   const isLoading = isDomainThresholdLoading || areEligibleSigneesLoading;
 
-  if (!thresholdPerRole || !requiredRoles) {
+  if (!thresholdPerRole || !requiredRoles || isLoading) {
     return {
       hasEnoughMembersWithPermissions: true,
       isLoading,
--- a/src/hooks/multiSig/useEligibleSignees.ts
+++ b/src/hooks/multiSig/useEligibleSignees.ts
@@ -24,7 +24,7 @@ export const useEligibleSignees = ({
   const {
     colony: { colonyAddress },
   } = useColonyContext();
-  const [isLoading, setIsLoading] = useState(false);
+  const [isLoading, setIsLoading] = useState(true);
   const [isError, setIsError] = useState(false);
   const [signeesResult, setSigneesResult] = useState<GetEligibleSigneesResult>({
     countPerRole: {},


mmioana (approved):
Nice fix @Nortsova 🫶 the error is not flashing anymore

https://github.com/user-attachments/assets/0e9a72d7-99c8-490a-815f-8c4159362c22

Though to make the `Multi-sig` option available as a decision method, we need to assign at least one user the `Owner` permissions

Screenshot 2024-11-01 at 10 59 06

However I did notice a strange bug that is not related to your changes. If you enter a fixed threshold, for eg. `2` and you don't refresh the page, the not enough permissions error banner is not showing up at all

https://github.com/user-attachments/assets/c6a607dd-43c9-4cb9-8839-3536a3ee0621

However, after refreshing the page, the error banner is right in place

Screenshot 2024-11-01 at 11 02 17

Either way, we can merge this and open up a new issue for the error I mentioned above 🙌 

Nice job!!! 🥇

iamsamgibbs (approved):
Nice job! I was expecting this to be a much bigger fix considering the issues encountered when first setting this up!

I set the threshold to 10 and turned on 3G throttling to see this working more accurately. Banner only appears when the query has resolved:

https://github.com/user-attachments/assets/b4d20491-c8bb-4c9e-8864-02798cb450be

Then set the threshold to 1, and banner does not appear at all (as expected):

https://github.com/user-attachments/assets/a9f0e02b-d9a7-4f65-81a3-e76efe01b1d6

bassgeta (approved):
Nice and easy :sunglasses: 

https://github.com/user-attachments/assets/01d4d1b1-ed61-4e94-b112-9e2901670026

PR #3602, opened 2024-11-01
Fix: CLNY conversion breaks the app

### LINKED ISSUE #3557: QA - Switching currency to CLNY breaks the app
## Steps to reproduce

1. Using the hamburger icon in the top right-hand corner of the user header, select CLNY as the currency. 
2. The app breaks and a blank page is shown.
3. Refresh the browser.
4. The app remains broken.
5. Using the address bar, return to the main Colony landing page.
6. Using the Colony switcher, navigate back to the Colony previously used to switch tokens to CLNY.
5. The app remains broken. 

_You can unbrick this by:_

1. Returning to the main landing page
2. Switching to a different currency from the hamburger menu while on the landing page (not CLNY)
3. Using the Colony switcher, navigate back to the Colony previously used to switch tokens to CLNY.
4. The Colony will load, and the user is not bricked.

Switching straight back to CLNY bricks, the app again. 

## Expected behaviour

Switching tokens should not brick the app. All other currencies work correctly.

## Actual behaviour

image

mmioana:
## Description

- This PR ensures selecting `CLNY` as currency no longer breaks the app. 
- The issue was caused by getting `0` as an exchange rate from CLNY in USD, causing the reverse to be `Infinity`
- Also, updated the `getCLNYPriceInUSD` to use the proper chain id and token address

## Testing

TODO: Select `CLNY` as preferred currency and make sure the app no longer breaks.

Screenshot 2024-11-05 at 00 52 16

Currently the values for the total cards remain to `0` because of a bug that is resolved in #3614

Resolves #3557

### DIFF
--- a/src/utils/currency/convertFromTokenToCurrency.ts
+++ b/src/utils/currency/convertFromTokenToCurrency.ts
@@ -5,9 +5,10 @@ export const convertFromTokenToCurrency = (
   conversionRate?: number | null,
 ) => {
   const precision = conversionRate ? 1000 : 1;
-  const formattedConversionRate = conversionRate
-    ? Math.floor(conversionRate * precision)
-    : 0;
+  const formattedConversionRate =
+    conversionRate && Number.isFinite(conversionRate)
+      ? Math.floor(conversionRate * precision)
+      : 0;
 
   return BigNumber.from(value ?? 0)
     .mul(formattedConversionRate)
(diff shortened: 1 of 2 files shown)


bassgeta (approved):
Better to have 0 CLNY than 0 app my mom always used to say
Nice find with the chainId shennanigans, a pretty nice solution!
image

rumzledz (approved):
Henceforth, thou shalt be known as Ioana, Master of Conversions, ruler of numbers and forger of metrics!

mmioana:
You guys @bassgeta and @rumzledz gave me a good laughing for the day 😂

rdig (approved):
Very nice! All good here!


https://github.com/user-attachments/assets/d22d8918-4bd0-403d-a9b2-bc4f4be2fa80

Issue #3555, opened 2024-10-30
[QA Dashboard] Total funds bug


mmioana:
## Steps to reproduce

1. Login on QA to the colony with address `0x997B915CaC293D69f14a8a85e660EE87f1523FcD` - `Jakub Colony`
2. On dashboard you'll see the balance has a small negative value
![Screenshot 2024-10-30 at 12 37 53](https://github.com/user-attachments/assets/771417a4-c5d4-4eda-9b7c-80785d5ea9b2)
3. However, if you switch to the finances pages, on the balance you'll notice the total ETH is positive
![Screenshot 2024-10-30 at 12 38 04](https://github.com/user-attachments/assets/0b6b7318-2b8a-43e0-8c25-3a58d94642e1)
But there is no incoming funds - in the past once an incoming funds was claimed, the entry was deleted
![Screenshot 2024-10-30 at 12 38 17](https://github.com/user-attachments/assets/cd0d2f74-4ef6-42fd-bcc1-8ea9e30ef1b7)

4. Also, another bug related to the funds, was found while scraping through the `getDomainBalance` query: if for the `timeframePeriod` you enter a number of days > than the number of days in a month (considering you selected `TimeframeType.DAILY`), as well as if you enter a number of months > than the number of months in a year (considering you selected `TimeframeType.MONTHLY`), you will not have an entry for each expected key.
So the solution would be to use the entire date `yyyy-mm-dd` as key and then adjust the displayed key in the UI.

https://github.com/user-attachments/assets/17fd48aa-0963-42f4-b2b3-e702a187562c

## Implementation details

Given the incoming funds entries are not available for prior to May 2024 - previously when an incoming funds action was claimed, the entry got deleted, we need to change a bit our approach to the balances.

So, in order to get our data right, at least for the overall amounts shown in the Total cards we need to make use of the actual network balance at colony level or domain level.

For this please make use of the following code put in place by @Nortsova in the UI

```
const calculateTotalFunds = async (
  balances: ColonyBalances,
  currency: SupportedCurrencies,
  selectedDomainNativeId: number | undefined,
): Promise<Decimal | null> => {
  let isError = false;
  const funds = balances.items
    ?.filter(notNull)
    .filter(({ domain }) =>
      !selectedDomainNativeId
        ? domain === null
        : domain?.nativeId === selectedDomainNativeId,
    )
    .reduce(
      async (total, { balance, token: { tokenAddress, decimals } }) => {
[...]

PR #3614, opened 2024-11-04
Fix: [QA Dashboard] Total funds bug

mmioana:
## Description

- Given the incoming funds entries are not available for prior to May 2024 - previously when an incoming funds action was claimed, the entry got deleted, we need to change a bit our approach to the balances. So, in order to get our data right, at least for the overall amounts shown in the Total cards we need to make use of the actual network balance at colony level or domain level. This PR addresses that ✨ 

## Testing

TODO: Let's test the changes in this PR retrieve the correct total balance for a domain, for `All teams` and also the caching.

* Step 1. First things first, run your dev environment
* Step 2. Run the `create-data` script
* Step 3. Visit http://localhost:9091/planex check the `All teams` or `General` `Total` card correctly reflects the `Income` - `Payments`
Screenshot 2024-11-04 at 17 08 30

* Step 4. Now select `Andromeda` and check again the difference is correct.
Screenshot 2024-11-04 at 17 08 34

* Step 5. Please replace
`const closestBlock = await provider.getBlock(closestBlockNumber);`
with 
`const closestBlock = await provider.getBlock(closestBlockNumber || currentBlockNumber);`
in `amplify/backend/function/fetchDomainBalance/src/services/networkBalance.js`

> [!IMPORTANT]
> This is needed because when we start our local dev env, the first block will have the current timestamp, so if we try to cache data from the past for the `Total` timeframe type, there will be no block having that timestamp and it will fallback to `0`

* Step 6. For ease of testing let's already cache the balance values
```
query MyQuery {
  cacheAllDomainBalance {
    success
  }
}
```
* Step 7. You can run the following query to check the `Total` cache has been properly updated
```
query MyQuery {
  listCacheTotalBalances(
    filter: {
      timeframeType: {eq: TOTAL}, 
      domainId: {eq: "<DOMAIN_ID>"} # or use <COLONY_ADDRESS>_2
    }
  ) {
    items {
      colonyAddress
      timeframeType
      timeframePeriod
      totalUSDC
      domainId
      totalUSDCIn
      totalUSDCOut
      date
    }
  }
}
```
* Step 8. Make sure to refresh the app as the new cached values have not yet been fetched
* Step 9. Now let's create a `Simple payment` with `Permissions`
Screenshot 2024-11-04 at 17 09 48

* Step 10. Check the `Total`, the trend for `Total`, `Income` and `Payments` properly update
Screenshot 2024-11-04 at 17 10 18

## Important
Locally it's a bit tricky to properly test the `findClosestPastBlock` for a given date, and even cache the `Total` funds trends only because with the latest network hash change, the first block should be at the current date.

However, let's do some magic ✨ 

Now run the following query
```
query MyQuery {
  getDomainBalance(
    input: {
      colonyAddress: "<COLONY_ADDRESS>", 
      timeframePeriod: 1, 
      timeframeType: TOTAL, 
      timeframePeriodEndDate: "<DATE_HERE>" # put a date here with at least one extra day compared to the current date
      domainId: "<DOMAIN_ID>" # you can either remove this or use <COLONY_ADDRESS>_<TEAM_NUMBER>
  }) {
    total
    totalIn
    totalOut
  }
}

```
And see if in the logs you get 
`Closest block to target date: 1025` - the number should update based on the used date and the existing block in the network

Last, but not least - and forever grateful you made it so far - please don't let these testing steps hinder your creativity. 

Also, if you want to test the cache values for the `Income`/`Payments` sections you can check [this PR](https://github.com/JoinColony/colonyCDapp/pull/3080)

## Diffs

**Changes** 🏗

* Make use of `colonyClient.getFundingPotBalance` for `TimeframeType.TOTAL`

## TODO

- [ ] We don't get all relevant timeframe keys if we were to change our timeframePeriod query input - this will be solved in another PR as it doesn't impact the data that we currently use in the UI

Contributes to [#3555](https://github.com/JoinColony/colonyCDapp/issues/3555)

### DIFF
--- a/amplify/backend/function/fetchDomainBalance/src/services/networkBalance.js
+++ b/amplify/backend/function/fetchDomainBalance/src/services/networkBalance.js
@@ -13,49 +13,109 @@ const { getDomains, getAllColonyTokens } = require('../api/graphql/operations');
 const ExchangeRatesService = require('./exchangeRates');
 const { getTotalFiatAmountFor } = require('./tokens');
 const { getTokensDatesMap } = require('./actions');
+const NetworkConfig = require('../config/networkConfig');
 
-const getAverageBlockTime = async ({
-  provider,
-  currentBlockNumber,
-  sampleSize,
-}) => {
-  const currentBlock = await provider.getBlock(currentBlockNumber);
-  const previousBlock = await provider.getBlock(
-    currentBlockNumber - sampleSize,
-  );
-  const timeDifference = currentBlock.timestamp - previousBlock.timestamp;
-  return timeDifference / sampleSize;
+// Helper function to fetch block timestamp in milliseconds
+const getBlockTimestamp = async (provider, blockNumber) => {
+  const block = await provider.getBlock(blockNumber);
+  return block ? block.timestamp * 1000 : null; // convert to milliseconds
 };
 
-const getEstimatedBlockNumber = async ({
-  timeframeEndDate,
-  currentBlockNumber,
-  averageBlockTime,
+// Final check to compare left and right bounds for closest timestamp
+const checkFinalBounds = async ({
+  provider,
+  left,
+  right,
+  targetTimestamp,
+  closestBlock,
+  minTimestampDiff,
 }) => {
-  const currentBlockTimestamp = new Date(
-    currentBlockNumber * 1000,
-  ).toISOString();
-  const numberOfSecondsAgo = getDifferenceInSeconds(
-    currentBlockTimestamp,
-    timeframeEndDate,
-  );
-  const estimatedBlocksDifference = Math.floor(
-    numberOfSecondsAgo / averageBlockTime,
-  );
+  const leftTimestamp = await getBlockTimestamp(provider, left);
+  const rightTimestamp = await getBlockTimestamp(provider, right);
 
-  /**
-   * If estimatedBlocksDifference is negative (due to numberOfSecondsAgo being negative)
-   * It means timeframePeriodEndDate is ahead of current network block time
-   */
-  if (estimatedBlocksDifference < 0) {
-    return currentBlockNumber;
+  if (
+    leftTimestamp !== null &&
+    Math.abs(leftTimestamp - targetTimestamp) < minTimestampDiff
+  ) {
+    closestBlock = left;
   }
-  // @TODO If estimatedBlocksDifference is higher than the existing number of blocks, we need to get the data for the (current or first)? block
-  if (estimatedBlocksDifference > currentBlockNumber) {
-    return currentBlockNumber;
+  if (
+    rightTimestamp !== null &&
+    Math.abs(rightTimestamp - targetTimestamp) < minTimestampDiff
+  ) {
+    closestBlock = right;
   }
 
-  return currentBlockNumber - estimatedBlocksDifference;
+  return closestBlock;
+};
+
+// Finds the closest past block to a target date using binary search
+const findClosestPastBlock = async ({
+  targetDate,
+  averageBlockTime,
+  currentBlockNumber,
+  provider,
+}) => {
+  if (!targetDate) {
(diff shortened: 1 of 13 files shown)


rumzledz (approved):
Another one smashed @mmioana !

## Did the caching process

Screenshot 2024-11-05 at 19 31 03

Screenshot 2024-11-05 at 19 31 49

## Pre-payment

Screenshot 2024-11-05 at 20 08 42

## Post-payment (12 CREDS)

Screenshot 2024-11-05 at 20 08 53

rdig (approved):
Wow, such a nice fix and quick turnaround! :100: 

Everything seems to be working as expected! Very nice!

Screenshot from 2024-11-05 16-53-18
Screenshot from 2024-11-05 16-53-25
Screenshot from 2024-11-05 16-54-05
Screenshot from 2024-11-05 16-54-25
Screenshot from 2024-11-05 16-54-55
Screenshot from 2024-11-05 16-55-42
Screenshot from 2024-11-05 16-55-56
Screenshot from 2024-11-05 16-57-00

iamsamgibbs (approved):
Great job! This seems to all be working nicely! Thank you for the great testing steps.

Screenshot 2024-11-05 at 17 01 38
Screenshot 2024-11-05 at 17 01 46
Screenshot 2024-11-05 at 17 02 41
Screenshot 2024-11-05 at 17 03 35

The only thing I wasn't sure about, is that the total and totalIn appear to be the same here:

Screenshot 2024-11-05 at 17 05 42

Screenshot 2024-11-05 at 17 06 18

mmioana:
@iamsamgibbs thanks for the review (also @rdig and @rumzledz)!
Unfortunately for the `totalIn` and `totalOut` when querying the `Total` timeframeType might not be correct given these values rely only on the incoming funds, expenditures and the remaining financial actions. 
However the fix on this PR is focused only on the `total` field to get the value from the network as on a real environment there might be some incoming funds not stored in the db

amplify/backend/function/fetchDomainBalance/src/services/networkBalance.js
```diff
@@ -0,0 +1,233 @@
...
+  // Step 1: Estimate initial block guess near target timestamp
+  const currentBlockTimestamp = await getBlockTimestamp(
+    provider,
+    currentBlockNumber,
+  );
+  const timeDiffSeconds = (currentBlockTimestamp - targetTimestamp) / 1000;
+  let estimatedBlock =
+    currentBlockNumber - Math.floor(timeDiffSeconds / averageBlockTime);
+
+  // Step 2: Set up binary search bounds
```
jakubcolony, line 70: This is a nice solution, but is it actually worth the extra computational effort? Your initial estimation might be a few seconds off, which is unlikely to matter when fetching past balance (it would only matter if the balance changed exactly within that few-second window). On the other hand, doing binary search will slow down the lambda at each execution.

Consider fetching a balance from 30 days ago. The estimated block number will be approximately n - 10,368,000 (where n is the current block number), so your initial bounds will be [n - 10,369,000, n]. The target block is much closer to the left bound than to the right one, and in the worst case scenario it will take roughly 24 calls (log₂ of the search range) to `getBlock()` to find the most precise block.

Unless I'm missing something else, this seems like a high price to pay for getting a block that is only a few seconds "more correct", if that makes sense.

jakubcolony (commented):
Very nice work @mmioana 🎉 

As mentioned I left a comment on the binary search bit, would love to hear your thoughts on that.

amplify/backend/function/fetchDomainBalance/src/services/networkBalance.js
mmioana: As I mentioned in our chat, your comment is very valid and worth considering, so I'll proceed creating another issue for this improvement 🙌 thanks so much for your review!!

PR #3617, opened 2024-11-04
Fix: Prevent shifting of other team filter copies

### LINKED ISSUE #3564: QA Dashboard - Switching teams causes team labels to shift
When switching teams, there is a minor jump of the content inside the individual team tabs.

## Steps to reproduce

1. On the dashboard, select the 'General' team tab.
2. Continually select and deselect this 'General' tab.
3. Notice the team tabs next to it. You'll see the team label shifting. 

## Expected behaviour

* Selecting a team tab should not cause the other tabs to move. 

## Actual behaviour

* When selecting an individual team tab, the content of the others move horizontally. 

https://github.com/user-attachments/assets/46e1bac7-4399-4119-858d-2d47783a5b73

rumzledz:
## Description

I guess the best way to explain this fix is we're currently faking the semibold effect via the `after` pseudo-element to prevent the font weight change from affecting the rest of the layout. The problem is that `font-medium` is only being applied for team items that are not selected. So when a team item is selected, this `font-medium` doesn't get applied and it's causing the rest of the layout to sort of "shift".

So I'm just applying the `font-medium` style as the base style for these team items regardless if they're selected or not which should keep the layout in tact. When a team item does get selected, our "fake" semibold effect gets applied which doesn't affect our layout.

prevent-shift

## Testing

1. Visit a Colony Dashboard and make sure there's at least 2 subdomains in there
2. Keep on clicking the first subdomain on the Team Filter
3. Verify that the layout of the Team Filter component doesn't "shift"
4. Verify this on mobile view as well please

Resolves #3564

### DIFF
--- a/src/components/v5/shared/TeamFilter/partials/TeamItem.tsx
+++ b/src/components/v5/shared/TeamFilter/partials/TeamItem.tsx
@@ -53,10 +53,10 @@ const TeamItem: FC<TeamItemProps> = ({
       type="button"
       aria-label={label}
       className={clsx(
-        'box-border inline-flex h-full w-full items-center justify-center border-y border-solid bg-base-white px-4 py-2 text-sm text-transparent bold-on-hover',
+        'box-border inline-flex h-full w-full items-center justify-center border-y border-solid bg-base-white px-4 py-2 text-sm font-medium text-transparent bold-on-hover',
         selected ? teamColor : null,
         {
-          'border-gray-200 font-medium hover:bg-gray-50': !selected,
+          'border-gray-200 hover:bg-gray-50': !selected,
           'border-l border-transparent after:font-semibold after:text-base-white hover:after:text-base-white':
             selected,
           'border-r': hasDelimiter,


iamsamgibbs (approved):
Nice job, team switcher is no longer shifting around when switching teams:

https://github.com/user-attachments/assets/fb89cbc1-8bd7-4b1f-8f4e-fd425d76114d

Drop down menu is unaffected:

https://github.com/user-attachments/assets/af9f672a-368c-4d1a-bce1-c207f52eb5d1

mmioana (approved):
And with this PR @rumzledz the domains no longer shift 🥇 

https://github.com/user-attachments/assets/26432f98-19a8-4680-8488-8f3ac13b23ad

While on master

https://github.com/user-attachments/assets/d68e55db-1365-425b-a863-be6038f5f6ae

rdig (approved):
All good here. No more layout shifting

https://github.com/user-attachments/assets/b98f08b0-6f6e-415c-bc88-c89ed548408b

PR #3621, opened 2024-11-04
Fix: Add action sidebar overflow

### LINKED ISSUE #3620: Content overflows the action sidebar and it is not scrollable
## Steps to reproduce

* Open the action sidebar and select the "Edit colony" action - the content overflows and it is not scrollable

Screenshot 2024-11-04 at 16 30 47

This also affects other actions which have enough content to overflow the page:

Screenshot 2024-11-04 at 16 30 28

## Expected behaviour

* The action sidebar should not overflow and be scrollable

rumzledz:
## Description

This adds a vertical overflow for the action sidebar content

sidebar overflow

## Testing

1. Go to a Colony
2. On the left sidebar, click Manage Colony
3. Verify that:
  - the action sidebar content isn't overflowing
  - the action sidebar content's hidden components can be viewed via scrolling
4. Click Edit Colony Details
5. Make sure that the content is scrollable if the Colony Description text area causes the rest of the form contents to go out of view 

Resolves #3620

### DIFF
--- a/src/components/v5/common/ActionSidebar/ActionSidebar.tsx
+++ b/src/components/v5/common/ActionSidebar/ActionSidebar.tsx
@@ -314,7 +314,7 @@ const ActionSidebar: FC<PropsWithChildren<ActionSidebarProps>> = ({
       </div>
       {isLoading && <ActionSidebarLoadingSkeleton />}
       <div
-        className={clsx('flex flex-grow', {
+        className={clsx('flex flex-grow overflow-y-auto', {
           hidden: isLoading,
         })}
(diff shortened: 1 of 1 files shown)


mmioana (approved):
Thanks for fixing this issue @rumzledz 🥇 

I actually noticed this bug while reviewing another PR, but with your fix, it all works great now and we can scroll the action sidebar if the content is overflowing 💯 


https://github.com/user-attachments/assets/6e6e509d-d9ab-4f62-be7a-1bc51d41f4b5

PR #3642, opened 2024-11-05
Fix: Locked token validation should properly refresh

### LINKED ISSUE #3403: [Payment FLow] Unable to make payments using native tokens from other colony
## Steps to reproduce

1. Navigate to the "Wayne Enterprises" page on the platform
2. Make a simple payment and send 100GGG to the "Planet Express" colony
3. Select Manage colony ->  Unloke token
4. Make Unlock actions
5. Navigate to the "Planet Express" colony -> Finances -> Incoming
6. Press the "Accept all the available funds" button
7. Try to make simple payment with GGG tokens in the "Planet Express" colony

## Expected behaviour

* Users should be able to make payments using native tokens without encountering any restrictions once the token is unlocked

## Actual behaviour

* The platform fails to recognize that the token is available for use, blocking further actions

image

image

https://github.com/user-attachments/assets/209dc35c-be04-4e69-9661-1ec7a85fc563

iamsamgibbs:
## Description

This PR removes useCallback from `fetchTokenStates` and instead calls it on initial render to ensure the latest token lock state is used.

## Explanation if you're interested

Bit of an edge case here. The `fetchTokenStates` function previously used `useCallback` and is only called when either the `getColonyTokenLockedState` or `tokens` passed from the colony context change.

When a token is unlocked, the `tokens` array passed from the colony context doesn't actually change as the locked status is not stored on the token model. This means the `fetchTokenStates` `useCallback` is not re-run as there are no changes. As a result, if a user has the app opened, and a token's locked status is changed by another user, this change will not be reflected in the token select component / validation until the user refreshes the entire app.

This PR removes the `useCallback` and changes the `useEffect` in `useTokenLockStates` to call `fetchTokenStates` when it is first rendered to fetch the fresh token locked state.

## Testing

For testing, open two tabs, one in the Planex colony and one in the Wayne colony both as the Leela wallet. Do not refresh at any point during these steps.

* Step 1 - Planex tab: Go to the incoming funds page and claim the incoming GGG tokens.
* Step 2 - Planex tab: Create a Simple Payment action. Switch the token to GGG. Note the validation error:

Screenshot 2024-11-05 at 15 53 48

* Step 3 - Planex tab: Close the action sidebar
* Step 4 - Wayne tab: Create an unlock tokens action.

Screenshot 2024-11-05 at 15 54 06

* Step 5 - Planex tab: Create a Simple Payment action. Switch the token to GGG. There should be no validation error:

Screenshot 2024-11-05 at 15 54 20

## Diffs

**Changes** 🏗

* `fetchTokenStates` is called on initial render

Resolves #3403

### DIFF
--- a/src/hooks/useTokenLockStates.ts
+++ b/src/hooks/useTokenLockStates.ts
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useState } from 'react';
+import { useEffect, useState } from 'react';
 
 import { useColonyContext } from '~context/ColonyContext/ColonyContext.ts';
 import { useGetColonyTokenLockedStateLazyQuery } from '~gql';
@@ -10,7 +10,7 @@ const useTokenLockStates = (): Record<string, boolean> => {
   const [getColonyTokenLockedState] = useGetColonyTokenLockedStateLazyQuery();
   const [tokenLockStatesMap, setTokenLockStatesMap] = useState({});
 
-  const fetchTokenStates = useCallback(async () => {
+  const fetchTokenStates = async () => {
     if (!tokens) {
       return;
     }
@@ -39,11 +39,11 @@ const useTokenLockStates = (): Record<string, boolean> => {
     });
 
     setTokenLockStatesMap(newTokenLockStatesMap);
-  }, [getColonyTokenLockedState, tokens]);
+  };
 
   useEffect(() => {
     fetchTokenStates();
-  }, [fetchTokenStates]);
+  }, []);
 
   return tokenLockStatesMap;
 };


rumzledz (approved):
Neattttt fix! ✨ 

## Locked token, form validation is throwing an error

Screenshot 2024-11-06 at 04 24 35
Screenshot 2024-11-06 at 04 24 52

## Unlocked token, no more form validations and successful payment

Screenshot 2024-11-06 at 04 25 44
Screenshot 2024-11-06 at 04 26 04

mmioana (approved):
This is a really nice fix @iamsamgibbs 👏 

Accepting incoming funds 
Screenshot 2024-11-06 at 11 19 45
Screenshot 2024-11-06 at 11 19 50
Trying to create a simple payment using `GGG` shows up error
Screenshot 2024-11-06 at 11 20 08
Unlocking token
Screenshot 2024-11-06 at 11 20 42
And the error no longer shows up when creating a simple payment
Screenshot 2024-11-06 at 11 21 02

Thanks so much for fixing this issue 💯 🥇

PR #3657, opened 2024-11-06
Fix: Dismiss lingering tooltips when the page is scrolled

### LINKED ISSUE #3486: [Dashboard] Tooltips display incorrect behavior on Dashboard screen on mobile
## Steps to reproduce

1. Open Dashboart
2. Tap on the "Copy contact address"  and "Copy this colony URL" icons
3. Scroll page up and down several times

## Expected behaviour

* Tooltips should disappear after scrolling up or down

## Actual behaviour

* Tooltips remain visible even after scrolling

https://github.com/user-attachments/assets/df17ba38-158d-45ed-97cd-2cf6de484f29

rumzledz:
## Description

TLDR: The `trigger` and `isOpen` props were clashing and led to situations whereby the tooltip still lingers even if we don't want it to. As a workaround, I had to manually detect hover states for the Tooltip's children and conditionally update the `isOpen` prop ✌️ 

| Production | Current Master | This Branch |
| ---- | ---- | --- |
| Production | master | my branch |

This is how it behaves on mobile now, on this branch:

mobile-tooltip

## Testing

1. Hover on the Copy Contract Address button
2. Click it
3. Scroll the page
4. Verify that its tooltip is not visible
5. Do the same for the Share button please

## Diffs

**New stuff** ✨

A new hook called `useOnElementScroll` which executes a function when the page is scrolled:

```js
useOnElementScroll({
  scrollableElementId: 'scrollable-element',
  callback: () => console.log('Scrolled!'),
});
```

I also had to slightly tweak the `useCopyToClipboard` hook.

Resolves #3486

### DIFF
--- a/src/hooks/useOnElementScroll.tsx
+++ b/src/hooks/useOnElementScroll.tsx
@@ -1,0 +1,47 @@
+import { noop } from 'lodash';
+import { useEffect } from 'react';
+
+/**
+ * Hook to run a callback function when a specific element is scrolled.
+ *
+ * @param {string} scrollableElementId - The ID of the element to listen for scroll events.
+ * @param {() => void} callback - The function to run on each scroll.
+ * @param {boolean} [shouldExecute=true] - Whether to set up the scroll listener.
+ *
+ * @example
+ * useOnElementScroll({
+ *   scrollableElementId: 'content',
+ *   callback: () => console.log('Scrolled!'),
+ *   shouldExecute: true,
+ * });
+ */
+export const useOnElementScroll = ({
+  scrollableElementId,
+  callback,
+  shouldExecute = true,
+}: {
+  scrollableElementId: string;
+  callback: () => void;
+  shouldExecute?: boolean;
+}) => {
+  useEffect(() => {
+    if (!shouldExecute) {
+      return noop;
+    }
+
+    const scrollElement = document.getElementById(scrollableElementId);
+
+    if (!scrollElement) {
+      console.error(`Element with ID '${scrollableElementId}' not found`);
+      return noop;
+    }
+
+    const handleScroll = () => callback();
+
+    scrollElement.addEventListener('scroll', handleScroll, { passive: true });
+
+    return () => {
+      scrollElement?.removeEventListener('scroll', handleScroll);
+    };
+  }, [callback, scrollableElementId, shouldExecute]);
+};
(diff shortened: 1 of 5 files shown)


rdig (approved):
Tooltip is no longer rendered when scrolling. Very nice fix here!


https://github.com/user-attachments/assets/9caf3c38-dfc1-46b8-a93a-77275f76fb2c


https://github.com/user-attachments/assets/28cad51a-41c4-48a1-b3e3-535b5afb9895

src/hooks/useOnElementScroll.tsx
```diff
@@ -0,0 +1,47 @@
...
+    const scrollElement = document.getElementById(scrollableElementId);
+
+    if (!scrollElement) {
+      console.error(`Element with ID '${scrollableElementId}' not found`);
+      return noop;
+    }
+
+    const handleScroll = () => callback();
+
+    scrollElement.addEventListener('scroll', handleScroll, { passive: true });
```
iamsamgibbs, line 41: Can we throttle handleScroll here?

iamsamgibbs (changes requested):
Nice job, this is working as expected. Disappears on scroll. Still disappears after expected amount of time.

On desktop:
https://github.com/user-attachments/assets/ab9044c2-8ea2-47fd-83fe-8b02da79a5f9

On mobile:
https://github.com/user-attachments/assets/f3a42757-c329-403e-af7f-81891939337b

I'm just rejecting as this tooltip has the wrong text when copied (I've checked master and it isn't an issue there, so is something introduced by this PR):

Screenshot 2024-11-07 at 13 41 29

Also left a comment about throttling the scroll event listener.

src/hooks/useOnElementScroll.tsx
rumzledz: Can do! And nice spot with the copy, that's my bad!

rumzledz: All right added a `scrollThrottle` prop set to 100ms by default and updated the copy! 🚀

iamsamgibbs (approved):
Thanks for fixing the tooltip and implementing the throttle!

Screenshot 2024-11-07 at 16 09 05

Nortsova (approved):
Very nice solution! 🙌 

I investigated tooltip documentation and it should be something with a boundary, but unfortunately no luck to implement that "boundary" thing. So I am happy that you found custom solution 🥳  

tooltip-disapear

rdig (approved):

PR #3671, opened 2024-11-07
Fix: Prevent the loading view to render more than once on page load

### LINKED ISSUE #3662: [QA] App content flashes before showing the loader again
I assume this is related to the recent changes that were deployed to QA, since it is not reproducible in production.

## Steps to reproduce

Refresh the app.

## Expected behaviour

Loading spinner shows until the app has loaded, after which it should disappear. 

## Actual behaviour

Loading spinner disappears briefly flashing the loaded app content, then the spinner shows once again:

https://github.com/user-attachments/assets/d449532a-7065-4e2e-bfdc-06e0f4488073

rumzledz:
## Description

> [!NOTE]
> I appreciate that this may be more of a band-aid solution.
> Justification is that, this hook sits on a top-level component and it only ever makes a network request again when the `colonyName` changes i.e. when you switch Colonies and `colonyName` is plucked directly from the URL via `useParams()` so it should remain constant whilst you're browsing the same Colony.

The issue is that the following options are causing the hook to make the request twice:

```js
fetchPolicy: 'network-only',
nextFetchPolicy: 'cache-only', // <<-- This one
```

Which then causes this component to render twice:

```js
if (
  walletConnecting ||
  isColonyLoading ||
  userLoading ||
  isContributorLoading
) {
  return <LoadingTemplate loadingText={MSG.loadingText} />;
}
```

## Testing

1. Directly visit a colony: http://localhost:9091/planex/
2. Verify that the following loader only shows up once:

image

Resolves #3662

### DIFF
--- a/src/routes/ColonyRoute.tsx
+++ b/src/routes/ColonyRoute.tsx
@@ -50,7 +50,6 @@ const ColonyRoute = () => {
       name: colonyName,
     },
     fetchPolicy: 'network-only',
-    nextFetchPolicy: 'cache-first',
   });
 
   const { user, userLoading, walletConnecting } = useAppContext();


bassgeta (approved):
Legit solution, and I agree with your reasoning :+1: 

https://github.com/user-attachments/assets/9f539bf5-a8b1-45e2-b645-aefde8fc264c

Nortsova (approved):
Nice solution, thank for for fix 👍  I agree, we have quite complex app so sometimes we have a problems with reloading 😢 

| Before | Now |
| ----- | ----- |
| before-loading | after-loading |

iamsamgibbs (approved):
Works for me! Only one loader.

https://github.com/user-attachments/assets/094630da-3dcb-4ce7-8d1d-d8eef0aed4a9

Issue #3302, opened 2024-10-11
Empty action created on Metatransactions "Too many requests" error


arrenv:
## Steps to reproduce

I am not really sure how to replicate this as all I did was create a new Staged payment action, which has worked after this one, I received the error in the console "Too many requests", which, I assume is something to do with Metatransactions as I could not replicate without it, or even again while testing now. Although, I feel like I have seen it before, just not with the empty action created.

So, I suppose, what would be good is to work out if we can get out of this empty action state, or if we can prevent it from being created.

This is the transaction in question - https://arbitrum-sepolia.colony.io/cleanenergy?tx=0x3856aec9555a857520d51af25b38c162f773c95f4f8c62f0953b0030122f9c91

## Expected behaviour

* Metatransactions "Too many requests" error should not create empty actions.

## Actual behaviour

* In running into the Metatransactions "Too many requests" error, an empty action was created.

![image](https://github.com/user-attachments/assets/e699bebc-0aa7-4a53-9c33-eeaa5fec7892)

![image](https://github.com/user-attachments/assets/4b15d19c-526e-48a1-bcd5-d6ac49ca89c3)

jakubcolony:
I'm afraid there's not much we can do if the second transaction (setting expenditure recipients and payouts) fails like so, since the expenditure is already created on contracts level. We can't multicall both transactions as the second one depends on the results of the first.

Users would be able to get out of this state once editing expenditure is ready.

rdig:
The "too many requests" errors comes from the provider and not the broadcaster. 

Also this is triggering the RetryRPC provider meaning the transaction somehow fails and it's retrying it

arrenv:
@jakubcolony What are the options to better handle this failure point?

Such as, if the second transaction fails, can we keep the user on the creation page, show an error in the UI, and allow them to create again without losing all the details they added.

Can we then also exclude 0 recipient, 0 payout expenditure from the activity feed? Making sure that it is included again if it is edited outside of the UI and recipients and payouts are added. Otherwise, it can safely be ignored.

arrenv:
Just experienced this again, however, the second and following transactions went through.

![image](https://github.com/user-attachments/assets/64277e6e-f82c-4c36-a474-7405c7b9177c)

arrenv:
Not entirely sure if it is the same issue, but a user just reported the second creation transaction failing for them.

[...]

PR #3678, opened 2024-11-08
Fix: Advanced payments crashing due to unsafe array access

### LINKED ISSUE #3677: Advanced payments crashing intermittently
There is a race condition involved in this issue (CDapp has to refetch expenditure specifically in between 2 contract events), so multiple attempts may be necessary to reproduce.

## Steps to reproduce

1. Create an advanced payment.

## Expected behaviour

* Completed action view is shown.

## Actual behaviour

* Occasionally, if CDapp refetches the expenditure precisely when it exists, but does not have any slots yet, the entire app will crash due to unsafe array access.

jakubcolony:
## Description

This small PR fixes unsafe array access when generating action description.

I came across a case where a newly created expenditure gets returned from the DB, but it does not yet have any slots (block-ingestor has not processed them yet). In such a scenario, the entire app would crash due to it trying to read the recipient of the first slot.

It would also be possible to run into this issue when an empty expenditure is created, such as in #3302 or through direct contract interaction (it is possible to create an expenditure with 0 slots).

## Testing

Create a couple of advanced payments and confirm the app does not crash.

Resolves #3677

### DIFF
--- a/src/components/common/ColonyActions/helpers/mapItemToMessageFormat.tsx
+++ b/src/components/common/ColonyActions/helpers/mapItemToMessageFormat.tsx
@@ -79,7 +79,7 @@ const useGetRecipientData = (
   } = actionData || {};
   const safeRecipient = safeTransaction?.transactions?.items[0]?.recipient;
   const stagedPaymentRecipientAddress =
-    expenditure?.slots[0].recipientAddress || '';
+    expenditure?.slots[0]?.recipientAddress || '';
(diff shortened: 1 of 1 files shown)


Nortsova (approved):
Nice fix ✨  No crashes now

| before: | now: |
| ------ | ------ |
| image | image |

iamsamgibbs (approved):
Nice job Jakub, sensible change and advanced payments are working:

Screenshot 2024-11-11 at 17 49 41

Issue #3565, opened 2024-10-30
[Create Colony] Input field validation fails when typing quickly


olgapod:
## Steps to reproduce

1. Navigate to the Create colony form.
2. Type quickly into the Custom colony URL field word "account" or any other invalid value.
3. Observe whether the validation triggers correctly.

## Expected behaviour

*  Validation should work consistently, regardless of the typing speed, and should provide feedback on invalid inputs.
* The 'CheckColonyNameExists' network request payload should include the full “account” string as entered in the input field
* 
![photo_2024-10-31 08 56 53](https://github.com/user-attachments/assets/0853fd1d-528b-403b-819d-b7315f96b534)

## Actual behaviour

* Validation allows invalid inputs being accepted.

<img width="1259" alt="image" src="https://github.com/user-attachments/assets/b29363aa-8577-49ab-be0e-62647282f345">

rumzledz:
Hi @olgapod 👋 Can you please confirm if this is still an issue? I cannot seem to replicate it:

![Image](https://github.com/user-attachments/assets/62b97e55-c622-4c6b-9fd4-861e049f4a42)

olgapod:
@rumzledz thanks, I also cannot reproduce anymore

PR #3682, opened 2024-11-08
est: add e2e tests for final step of create colony flow 

olgapod:
## Description

- added e2e tests for the Create Colony flow
- enabled parallelization for tests of a single file
- use `fill` instead of `pressSequentially` to speed up the tests a bit

## Testing

**Prerequisites**: The test script simulates user interactions with the local instance of the web app, which means that the BE is expected to be set up and running. For now, backend services should be started manually and the initial test data should be seeded . Therefore, ensure that the development environment is properly set up by following the instructions in the README file

- Start your local dev environment  `npm run dev` 
- Seed the mock data `node scripts/create-data.js` 
- If it's the first time you run e2e tests on your machine - run `npm e2e:install` 

- run `npm run e2e` or `npm run e2e:ui`
- Ensure all tests pass (the 'should reject invalid colony name' test may intermittently fail due to the input validation issue described in #3565)

## Diffs

**New stuff** ✨

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -51,6 +51,7 @@ module.exports = {
       extends: 'plugin:playwright/recommended',
       rules: {
         'no-await-in-loop': 'off',
+        'playwright/no-conditional-expect': 'off',
       },
     },
   ],
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -8,6 +8,7 @@ dotenv.config({ path: path.resolve(__dirname, '.env') });
  * See https://playwright.dev/docs/test-configuration.
  */
 export default defineConfig({
+  fullyParallel: true,
   testDir: './playwright/e2e',
   /* Fail the build on CI if you accidentally left test.only in the source code. */
   forbidOnly: !!process.env.CI,
--- a/playwright/e2e/create-colony.spec.ts
+++ b/playwright/e2e/create-colony.spec.ts
@@ -1,8 +1,8 @@
 import { expect, test } from '@playwright/test';
 
 import {
-  acceptCookieConsentBanner,
   fillInputByLabelWithDelay,
+  setCookieConsent,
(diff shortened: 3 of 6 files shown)


Nortsova (changes requested):
Hey there! ✨ 

Nice work on this 👍 

However there is one test that failed:

image

I tested 10 times and 4 of them this test failed. 

Here is more info: 
image

image

I also checked our code base why it could happen and found this: 
image

It’s likely that the form was updated too quickly. If we can replicate it manually, then it’s a bug. :)

olgapod:
> Hey there! ✨
> 
> Nice work on this 👍
> 
> However there is one test that failed:
> 
> image
> I tested 10 times and 4 of them this test failed.
> 
> Here is more info: image
> 
> image
> I also checked our code base why it could happen and found this: image
> 
> It’s likely that the form was updated too quickly. If we can replicate it manually, then it’s a bug. :)

it is actually expected the test to fail, as it's a bug that we have an issue for and that can be also reproduced manually https://github.com/JoinColony/colonyCDapp/issues/3565

iamsamgibbs (changes requested):
Nice job! I am also having the same issue as @Nortsova but otherwise all testings are passing! (Am I remembering correctly that there was an issue raised recently with this? So maybe there is a legit issue here?)

Screenshot 2024-11-11 at 17 43 31

iamsamgibbs:
Ah, literally as I was writing it up! I guess it is fine to approve this for now. It would be helpful to adjust the testing steps and mention if we are expecting a specific test to fail.

Also when this is merged, could you add a comment to that issue saying that the test should pass in order for that issue to be resolved.

Nortsova (approved):
Oh, okay! Great! Thank you for raising issue 🙌  I am approving in this case 🎉 

Great work 🦸‍♀️

PR #3709, opened 2024-11-14
Fix: Add disabled styling to multi-sig setting selects

### LINKED ISSUE #3227: MultiSig: threshold dropdowns are missing a disabled state
## Steps to reproduce

1. Provided you don't have root permissions in Betacolony, navigate to https://app.colony.io/beta/extensions/MultisigPermissions.
2. Select Extension settings tab.

## Expected behaviour

The form is disabled and cannot be interacted with.

## Actual behaviour

The form is indeed disabled, however the domain specific threshold dropdowns are missing a disabled state, making them look interactive (even though they are not):

image

## Suggestions (optional)

This issue is also present in #3114, so the fix should be merge after/into that branch.

iamsamgibbs:
## Description

This PR adds disabled styling to the selects on the multi-sig extension settings page. (The inputs were already disabled, just missing the styling.)

## Testing

* Step 1 - Login as a user with permissions to install extensions (Leela)
* Step 2 - Install the multi-sig extension and check the inputs are enabled on the extension settings page

Screenshot 2024-11-14 at 10 09 55

* Step 3 - Switch to a user who does not have permissions to update extensions (Amy / Fry)
* Step 4 - Navigate to the extension settings page and check the inputs are disabled and the select styling is as expected

Screenshot 2024-11-14 at 10 10 06

## Diffs

**New stuff** ✨

* Added disabled styling to select inputs

Resolves #3227

### DIFF
--- a/src/components/v5/common/Fields/Select/Select.module.css
+++ b/src/components/v5/common/Fields/Select/Select.module.css
@@ -14,6 +14,14 @@
   @apply bg-transparent text-gray-300;
 }
 
+.wrapper :global(.select__control--is-disabled .select__single-value) {
+  @apply text-gray-300;
+}
+
+.wrapper :global(.select__control--is-disabled .select__indicator) {
+  @apply text-gray-300;
+}
+
 .wrapper :global(.select__placeholder) {
   @apply font-normal text-gray-400;
 }


Nortsova (approved):
Nice and tiny fix, works great ✨ 

image

bassgeta (approved):
I would've never guessed the solution was a CSS one :D 
Nice catch, simple and it works, what more can you ask for?
image

PR #3722, opened 2024-11-15
RESERVED_ROUTES check for create new colony is only done on the front end

### LINKED ISSUE #3406: The RESERVED_ROUTES check for create new colony is only done on the front end
When creating a new colony you must select a URL. When creating a colony via the UI we check if the URL matches any RESERVED_ROUTES and reject it if so. However, if this front end check was bypassed, it would be possible to create a colony using one of these reserved routes as there is no restriction on the network side or block-ingestor.

This might require some further discussion on what this affects and changes we may want to make. I also suspect it would be quite difficult to exploit this anyway due to the way we handle colony creation and colony ethereal metadata, so maybe nothing is required here / not urgently required here.

Some suggestions:

* The majority of the RESERVED_ROUTES are not currently in use, if a colony was created using one of these the colony would appear when navigating directly to one of these pages (`/staff` for example). For extra safety we could ensure these RESERVED_ROUTES always navigate to the 404 page even if a colony exists with the same slug.

* We could add a check on the block-ingestor and if the colony name matches a reserved route, we ~~could either~~ *should* not finish creating the colony, ~~or forcibly change the name~~.

iamsamgibbs:
## Description

See block-ingestor counterpart

This PR adds a RESERVED_ROUTES check to the block-ingestor and will prevent a colony from being created if the name matches a reserved route.

It also adds an extra safety fallback that if any unused reserved route will now redirect to a 404 before checking for any colony pages, so even if a colony is somehow created using a reserved route, it will fail to load.

## Testing

* Step 1 - This has a new block-ingestor hash, so boot up from scratch I'm afraid.
* Step 2 - Change the `isValidName` function in: `src/components/common/Onboarding/wizardSteps/CreateColony/validation.ts` to this:

```
function isValidName(name: string) {
  return name ? new RegExp(COLONY_NAME_REGEX).test(name) : true;
}
```

This will remove the reserved routes check from the frontend and allow us to test these extra checks.

* Step 3 - Run the `node scripts/create-colony-url.js` script.
* Step 4 - Try creating a colony using one of the names from the reserved routes list: `src/routes/routeConstants.ts`

The app should hang at the final stage of the setup as the block-ingestor rejects it. We don't need to handle this nicely as the validation will prevent the form getting to this state.

Screenshot 2024-11-15 at 13 48 42

* Step 5 - Refresh to start the colony creation flow over again.
* Step 6 - Kill the block-ingestor docker container, and run the block-ingestor on master (this will remove the block-ingestor check that has been added in this PR).
* Step 7 - Try creating a colony using one of the names from the reserved routes list: `src/routes/routeConstants.ts`

This time at the end of the colony creation flow, you should be redirected to a 404. The colony will still have been accepted by the block-ingestor and created in the database, but now the route check will redirect it to a 404 even though the colony exists. The colony will also show under the colony switcher, if you try to select it, you will be redirected to the 404.

## Diffs

**Changes** 🏗

* Unused reserved routes will now redirect to 404.
* Block-ingestor now prevents creating colonies with a reserved route as the name.

Resolves #3406

### DIFF
--- a/docker/colony-cdapp-dev-env-block-ingestor
+++ b/docker/colony-cdapp-dev-env-block-ingestor
@@ -1,6 +1,6 @@
 FROM colony-cdapp-dev-env/base:latest
 
-ENV BLOCK_INGESTOR_HASH=d2c9d9ee531640abe8e561cc7a06cfd662de18b3
+ENV BLOCK_INGESTOR_HASH=7d91f791e1e7aefe0c8c0e3495f4d012fa9bb54f
 
 # Declare volumes to set up metadata
 VOLUME [ "/colonyCDapp/amplify/mock-data" ]
--- a/src/routes/Routes.tsx
+++ b/src/routes/Routes.tsx
@@ -64,6 +64,7 @@ import {
   COLONY_MULTISIG_ROUTE,
   COLONY_AGREEMENTS_ROUTE,
   USER_CRYPTO_TO_FIAT_ROUTE,
+  RESERVED_ROUTES,
   // ACTIONS_PAGE_ROUTE,
   // UNWRAP_TOKEN_ROUTE,
   // CLAIM_TOKEN_ROUTE,
@@ -118,6 +119,11 @@ const Routes = () => {
           path={CREATE_COLONY_ROUTE}
           element={<OnboardingPage flow={Flow.Colony} />}
         />
+
+        {/* If a reserved route has not been used by this point, redirect to NotFoundRoute */}
+        {[...RESERVED_ROUTES].map((route) => (
+          <Route path={route} element={<NotFoundRoute />} />
+        ))}
 
         {/* Colony routes */}
         <Route path={COLONY_HOME_ROUTE} element={<ColonyRoute />}>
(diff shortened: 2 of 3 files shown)


bassgeta (approved):
Nice one!
Created a colony called `general`:
image
And it "successfully" hung at the last step :heavy_check_mark: 
image

After restarting the block ingestor without those changes i get to the 404 page :heavy_check_mark: 
image

rdig (approved):
Very nice here, having the validation for colony names where it actually matters, namely on the ingestor.


https://github.com/user-attachments/assets/320449c2-3220-4f35-b310-a369b4e2eff1

As well as the ingestor spewing out the relevant error

Screenshot from 2024-11-20 17-26-31

A further improvement here, but one that it's for the future, is to move all the reserved routes to the database, so that this list could be updated and maintained from a more central place

rumzledz (approved):
Great work on this one @iamsamgibbs !

## 1. Updated the isValidName function

image

## 2. Created a colony named colonystaff

The colony creation process hangs after step 4 has been marked as completed:

Screenshot 2024-11-21 at 23 11 18

## 3. Switched to the master branch for the block ingestor

I am redirected to a 404 page after creating a colony named "colonystaff"

Screenshot 2024-11-21 at 23 12 55

## 4. Clicking the colonystaff colony from the Colony Switcher

I am redirected to a 404 page

colonystaff

PR #3723, opened 2024-11-15
Fix: Previous steps in colony creation should not be clickable

iamsamgibbs:
## Description

The previous steps in the colony creation stepper are to show progress only, they should not be clickable. This PR adds a Stepper prop to disable previous steps.

## Testing

* Step 1 - Run: `node scripts/create-colony-url.js`
* Step 2 - Progress the create colony flow to the confirmation step

## Diffs

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

* Changes go here

**Deletions** ⚰️

* Deletions go here

## TODO

- [ ] Add TODOs

### DIFF
--- a/src/components/common/Onboarding/Wizard/MobileWizardHeader.tsx
+++ b/src/components/common/Onboarding/Wizard/MobileWizardHeader.tsx
@@ -52,6 +52,7 @@ const MobileWizardHeader = ({ currentStep, wizardSteps }: WizardProps) => {
       <Stepper
         activeStepKey={activeStepKey}
         items={steps.filter((step) => !!step)}
+        disablePreviousSteps
       />
     </div>
   );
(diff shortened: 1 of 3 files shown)


src/components/common/Onboarding/Wizard/MobileWizardHeader.tsx
```diff
@@ -52,6 +52,7 @@ const MobileWizardHeader = ({ currentStep, wizardSteps }: WizardProps) => {
       <Stepper
         activeStepKey={activeStepKey}
         items={steps.filter((step) => !!step)}
+        disablePreviousSteps
```
rumzledz, line 55: ✨

rumzledz (approved):
I can confirm that previous steps are no longer clickable 🚀 

prev-step-not-clickable

Nortsova (approved):
Nice and tiny fix ✨ 

Previous steps cannot be clicked now 👍 
create-colony-prev-steps

PR #3725, opened 2024-11-18
Fix: Optimise token sums calculation for multi-row advanced payments

### LINKED ISSUE #3673: Advanced Payments Table is incredibly slow when there is a large number of rows
## Steps to reproduce

* Create an advanced payment
* Add a ton of rows (via CSV is quicker)

## Expected behaviour

* The form should load smoothly

## Actual behaviour

* The form becomes incredibly laggy and takes a long time to load in via CSV

https://github.com/user-attachments/assets/59e005b7-9019-48ca-8840-90dd4a406379

Screenshot 2024-11-07 at 14 42 39

Screenshot 2024-11-07 at 14 47 55

Screenshot 2024-11-07 at 14 48 05

Payments also seem to be coming in in the incorrect order on the completed action screen too (should be starting from 1 not 37), I'm hoping it is somehow related to this laggy behaviour and will work properly once it is resolved rather than being a deeper issue:

Screenshot 2024-11-07 at 14 55 57

rumzledz:
## Description

> [!NOTE]
>
> **💡 What this PR aims to do**
> Prevent an app crash when uploading a CSV file with at least 200 rows
>
> **🚫 What this PR doesn't attempt to fix:**
> This PR is not aiming to address the issue with the Completed Action component's payments incorrect order

Currently, there is a palpable performance hit whenever we're calculating the sums for each token present on the Advanced Payment form. This is because we're performing this calculation for each and every single payment row present. This means that when you upload a CSV file with like 200 rows, you are pretty much performing the calculation 200 times which is causing the huge lag and could oftentimes cause our app to crash.

This PR aims to semi-resolve this lag issue by building a map of token sums outside of the validation functions. So whenever the Payment Builder form needs needs to know the token sum for a payment row's token, it just needs to perform an object search, using the token as the key i.e. `tokenSums[tokenAddress]`. At this point, `tokenSums` is something that has already been built before the validation check, which means that the validator doesn't need to calculate the token sum again.

This solution is not perfect because you can still see the table lagging when you're scrolling it up and down. But I guess this is a small win since it prevents the app from crashing/being non-responsive 😅 

advanced payment

## Testing

1. Bring up the Advanced Payments form
2. Upload this [CSV file](https://drive.google.com/file/d/1iMFNpRtMwyc3teh2mVPeBlRPEyZBRwFQ/view?usp=sharing), it should be 400 rows
3. Verify that the app does not crash

Resolves #3673

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/forms/PaymentBuilderForm/hooks.ts
+++ b/src/components/v5/common/ActionSidebar/partials/forms/PaymentBuilderForm/hooks.ts
@@ -26,7 +26,7 @@ import {
   getPaymentBuilderPayload,
 } from './utils.ts';
 
-export const useValidationSchema = (networkInverseFee: string | undefined) => {
+export const useValidationSchema = () => {
   const { colony } = useColonyContext();
   const colonyTokens = useMemo(
     () =>
@@ -41,6 +41,11 @@ export const useValidationSchema = (networkInverseFee: string | undefined) => {
     () =>
       object()
         .shape({
+          /**
+           * Stores a map of the sums for each token present on the form
+           * @internal
+           */
+          _tokenSums: object(),
           from: number().required(
             formatText({ id: 'errors.fundFrom.required' }),
           ),
@@ -90,7 +95,6 @@ export const useValidationSchema = (networkInverseFee: string | undefined) => {
                         value,
                         context,
                         colony,
-                        networkInverseFee,
                       }),
                     ),
                   tokenAddress: string()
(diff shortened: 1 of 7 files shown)


bassgeta (approved):
I actually really like this solution of internal form values which help for faster lookup!
It's much nicer and much more proprietary than smashing everything in yet another context provider :100: 
I guess `react-virtualized` is the next step :D ? 
image

Nortsova (approved):
It works much better now than it was before 👏 , great solution 🎉 

Nov-19-2024 16-02-05

As mentioned earlier, the next step will be to completely eliminate this lag 😄

Issue #3334, opened 2024-10-15
Incorrect description on release step


arrenv:
Teamwork task: [Incorrect description on release step](https://tw.auroracreation.com/#/tasks/16120160)

## Steps to reproduce

1. Create an Advanced payment action.
2. Go through the whole process as the Payment creator.
3. Release the payment using Permissions.
4. Return to the Release step and review the history card.
5. Note that despite releasing using Permissions, it still says it was released by the Payment creator.

## Expected behaviour

* The history of decisions should accurately show what happened and how the decision was made.

## Actual behaviour

* History of release shows "Payment creator released the payment." however, Permissions were used.

![image](https://github.com/user-attachments/assets/aa504ca7-c5d3-4012-905a-cdc67837c887)

PR #3726, opened 2024-11-18
fix: description in release step

adam-strzelec:
## Description

Add `Permissions` for `Payer` and `Custom - arbitration` role and `Payment creator` for `Owner`

## Testing

- Create advanced payment
- Go thru entire flow
- Go back to release step

## Diffs

**Changes** 🏗

Screenshot 2024-12-19 at 11 31 56

Resolves [3334](https://github.com/JoinColony/colonyCDapp/issues/3334)


src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/PaymentBuilderWidget/PaymentBuilderWidget.tsx
```diff
@@ -455,18 +453,9 @@ const PaymentBuilderWidget: FC<PaymentBuilderWidgetProps> = ({ action }) => {
...
-                    <FinalizeByPaymentCreatorInfo
-                      userAdddress={expenditure?.ownerAddress}
-                    />
-                  ) : (
-                    <ActionWithPermissionsInfo
-                      action={finalizingActions?.items[0]}
-                    />
-                  )}
-                </>
+                <ActionWithPermissionsInfo
```
bassgeta, line 456: Hmmm? :D Alright so we always display that the payment was released via permissions, if that's what we always need to do (even if the user used their temporary payment creator permissions), we can safely remove `FinalizeByPaymentCreatorInfo`

bassgeta (changes requested):
Well it does in fact do the job, but if we always want to display this, I think we should remove `FinalizeByPaymentCreator` info so we get rid of some extra lines :sunglasses:

mmioana (changes requested):
Hey @adam-strzelec thanks for picking this issue up 🙌 

However, I believe this is not working as intended.

If I release the advanced payment using `Permissions` the right information is displayed
Screenshot 2024-11-21 at 17 32 10

However, releasing the advanced payment using `Payment creator` the same details as for the previously used decision method (`Permissions`) is shown, which is not correct
Screenshot 2024-11-21 at 17 33 11

jakubcolony:
The permissions vs payment creator selection never worked properly - the saga always uses payment creator if user is such:

image

While we can tell if permissions or reputation was used, we lack details to tell if permissions were specifically chosen over payment creator.

@arrenv How should we proceed with this one?
- We could add a separate payment creator finalization action (but this might be overkill just for history tracking).
- Simply prevent payment creators from selecting permissions (assuming there's no real need for them to use permissions). This lets us keep using the existing expenditure owner address check.

arrenv:
@adam-strzelec @jakubcolony 

Thank you for the input on this. I think it is useful to be able to communicate the difference of how the payment was released. So, out of the two options provided, I would opt for the second one.

> Simply prevent payment creators from selecting permissions (assuming there's no real need for them to use permissions). This lets us keep using the existing expenditure owner address check.

To add to that, I feel it would also be good to have the select field to auto select the option if there is only one. I do think a new issue should be created for this update.

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/PaymentBuilderWidget/PaymentBuilderWidget.tsx
```diff
@@ -455,18 +453,10 @@ const PaymentBuilderWidget: FC<PaymentBuilderWidgetProps> = ({ action }) => {
...
+                  action={finalizingActions?.items[0]}
+                  statusText="action.executed.creator.description"
```
jakubcolony, line 458: This needs to show payment creator/permissions depending on the decision method used

jakubcolony (changes requested):
It looks like there was a copy change that was not specified in the issue. I think this should still read "Payment creator released this payment."

image

----

The info is also incorrectly showing payment creator decision method when the payment was finalized by a different user via permissions.
image

mmioana (commented):
Thank you for your continuous effort on this one @adam-strzelec 🙌 

Now the copy is the correct one for the `Release` step as the `Payment creator`
Screenshot 2024-12-20 at 11 38 36

Created another `Advanced payment`
Screenshot 2024-12-20 at 11 40 51

And tried to finalise it as `amy`, but it seems I can no longer do that. Is this as intended?
Screenshot 2024-12-20 at 11 41 06

adam-strzelec:
@mmioana I have no idea. @arrenv If I am a person who did not create the action, should I receive such information?

arrenv:
> @mmioana I have no idea. @arrenv If I am a person who did not create the action, should I receive such information?

Yes, this is intended, the has to be an available decision method for a non-creator of the Payment to be able to release it.

So, amy would need either the right permissions ("Payer" bundled or Arbitration), or, the Lazy consensus extension would need to be installed and there would need to be reputation in the domain.

mmioana (changes requested):
Hey @adam-strzelec @arrenv thanks for the clarification 🙌 

The description correctly updates when releasing as a payment creator
Screenshot 2025-01-13 at 09 18 33
Screenshot 2025-01-13 at 09 18 43

Created another `Advanced payment`
Screenshot 2025-01-13 at 09 19 28

Then after installing the `Reputation` extension
Screenshot 2025-01-13 at 09 21 53

And tried to release it as `amy`, the `Reputation` decision method was available in the dropdown
Screenshot 2025-01-13 at 09 21 49

However @arrenv @adam-strzelec, if I assign `amy` the `Payer` bundled permissions, the permissions is not available to `Release` the advanced payment ❌ 
Screenshot 2025-01-13 at 09 16 20
Screenshot 2025-01-13 at 09 20 56

adam-strzelec:
@mmioana done 🙂

jakubcolony (changes requested):
Thanks for another round of changes @adam-strzelec

As leela (who created the payment), I can correctly see Payment creator as the only available decision method:
image

As a user without arbitration permission, I get an error as expected:
image

However, after being assigned Payer role, the newly available method reads "Payment creator" instead of "Permissions":
image

src/components/v5/common/CompletedAction/partials/PaymentBuilder/partials/PaymentBuilderWidget/hooks.tsx
```diff
@@ -145,7 +145,9 @@ export const useGetFinalizeStep = ({
             )}
 
             {selectedFinalizeAction && !selectedFinalizeMotion && (
-              <ActionWithPermissionsInfo action={selectedFinalizeAction} />
+              <FinalizeByPaymentCreatorInfo
+                userAdddress={selectedFinalizeAction.initiatorAddress}
+              />
             )}
```
mmioana, line 151: Here we should perform a similar check for `userIsCreator` to decide which description to actually show

mmioana (changes requested):
Thanks for your continuous efforts on this issue @adam-strzelec 🙌 

Released a payment as `Payment creator` ✅ 
Screenshot 2025-01-16 at 08 34 15

Gave `amy` payer permissions
Screenshot 2025-01-16 at 08 35 00

Released payment using `Payer` permissions ✅ 
Screenshot 2025-01-16 at 08 36 47
However the widget from the `Release` step showed the payment was released as `Payment creator`, which is not correct ❌ 
Screenshot 2025-01-16 at 08 37 09

Gave `fry` `Custom-Arbitration` permissions
Screenshot 2025-01-16 at 08 38 55
Released payment using `Custom-Arbitration` permissions ✅ 
Screenshot 2025-01-16 at 08 40 39
However, also here the same problem with the widget showing the payment was released as `Payment creator`, which is not correct ❌ 
Screenshot 2025-01-16 at 08 40 49

adam-strzelec:
@mmioana Do we have any designs for how it should look like when the user uses permissions? The main issue only says very generally that `The history of decisions should accurately show what happened and how the decision was made.` Should it be `Member used permissions to create this action.`?

jakubcolony:
@adam-strzelec I believe that's the copy that should be used in case of permissions, it would be consistent with the cancel step:
image

mmioana (approved):
Thanks a lot for all the changes @adam-strzelec it all works as expected now 🥇 

Released the payment as `Payment creator`
Screenshot 2025-01-28 at 20 04 14
Screenshot 2025-01-28 at 20 04 30

Gave `amy` `Payer` permissions and released the payment using `Permissions`
Screenshot 2025-01-28 at 20 06 13
Screenshot 2025-01-28 at 20 07 34

Now gave `fry` `Custom-Arbitration` permissions
Screenshot 2025-01-28 at 20 07 41
And released the payment using `Permissions`
Screenshot 2025-01-28 at 20 08 36
Screenshot 2025-01-28 at 20 08 46

Removed `fry`'s permissions and installed the `Reputation` extension
Screenshot 2025-01-28 at 20 11 51
Created another payment as `fry` and used the `Reputation` decision method for releasing the payment
Screenshot 2025-01-28 at 20 11 57
Screenshot 2025-01-28 at 20 14 01

@adam-strzelec Regarding the wording I think @arrenv is best to provide an input to when releasing a payment using `Permissions` we should use `Member used permissions to create this action.`. Personally, I believe `Member used permissions to release this payment.` would be a better fit. Either way, I'm happy to approve your PR 💯 

Really nice work and thanks again for all the changes! 🎉

bassgeta (approved):
Thanks for all the hard work on this, looking real good right now :muscle: 
As payment creator :heavy_check_mark: 
image
As member with permissions :heavy_check_mark: 
image
With reputation :heavy_check_mark: 
image

arrenv:
@adam-strzelec Regarding wording as @mmioana has raised, this is definitely the desired behavior, we want these descriptions to be as accurate as possible so, if it can be `Member used permissions to release this payment.` in this case, that would be preferred.

adam-strzelec:
@arrenv @mmioana I already changed the copy 🙂

jakubcolony (approved):
Nice work on this PR, looks good to be merged. Thank you @adam-strzelec 👍 

Releasing as payment creator:
image
image

Releasing as another user with permissions:
image
image

Trying to release as a user without permissions:
image

Releasing using reputation:
image

AdrianDudko:
@bassgeta Can you please take a look at this PR? We have 2 approvals already :)

iamsamgibbs (approved):
Great job getting to the bottom of this issue. All working nicely now.

Releasing as creator:
Screenshot 2025-02-05 at 09 32 46

Releasing with permissions:
Screenshot 2025-02-05 at 09 36 51

Releasing with reputation:
Screenshot 2025-02-05 at 09 34 33

🎉

PR #3728, opened 2024-11-18
Fix: Not enough funds validation message

### LINKED ISSUE #3336: [Payment Flow] Clarify fee validation logic in payment process
### Story

Update the payment validation logic to include a comment or explanation about the inclusion of the network fee in the validation process

### Description

Example: During the colony creation wizard, there is an activity indicator for pending transactions, but there is no indicator of other async operations (e.g. for the DDB) in between these transactions. It's important that the indicator remains visible until all async activity is finished, because otherwise, the user has no feedback that they're waiting for something to finish; they might leave the wizard or refresh the page, leading to frustration.

image

iamsamgibbs:
## Description

The validation message when there are not enough tokens lacked context about the network fee. This PR updates the error message.

## Testing

* Step 1 - Switch to a sub-domain and confirm the available funds in the domain.
* Step 2 - Create a new simple payment action in the sub-domain.
* Step 3 - Enter an amount just below the maximum available funds, but where the network fee (1%) would take it over the amount available. Check the error validation message shows as expected.

Screenshot 2024-11-18 at 14 27 59

(Note, the same validation message is used even when the amount excluding the network fee is over the maximum available funds).

## Diffs

**Changes** 🏗

* Updated `notEnoughTokens` message.

**Deletions** ⚰️

* Removed unused `notEnoughTokensIn` message.

Resolves #3336


rumzledz (approved):
Can confirm that the copy update applies 🚀 

Screenshot 2024-11-19 at 20 37 11

Nortsova (approved):
Nice fix ✨ 
Just wondering, can we calculate the network fee and show the user the maximum amount they can input in this field?

image

iamsamgibbs:
> Just wondering, can we calculate the network fee and show the user the maximum amount they can input in this field?

Yes we could! There is a rework planned which will show a proper break down of the tokens and network fee, designs are here if you're interested! :) https://www.figma.com/design/5V8pr7iMwXsT9L3VAZsmUt/Actions-%26-Motions?node-id=26076-279777

rdig (approved):
Working as expected.

Total balance in subdomain was `423`, and as expected, anything above `418` will throw up the network fee validation error.

Nicely done

Screenshot from 2024-11-20 17-01-48
Screenshot from 2024-11-20 17-02-02
Screenshot from 2024-11-20 17-02-09
Screenshot from 2024-11-20 17-02-20

PR #3741, opened 2024-11-20
Fix: Do not prematurely end the GetExpenditure query polling

### LINKED ISSUE #3719: [Payments] Payment process occasionally stalls at "Review" stage
## Steps to reproduce

1. Create a staged payment with two milestones. (or advanced , split payments)
2. Observe the payment flow, noting that some payments get stuck in the "Pending" state.

## Expected behaviour

* The payment process should consistently move through each stage without stalling at "Pending."

## Actual behaviour

* In roughly half of the cases, the payment process stalls at "Pending," with console errors indicating a possible URL construction issue.

image

rumzledz:
## Description

So the main issue is that we're prematurely ending the GetExpenditure query polling on the PaymentBuilderWidget. Which means that we won't ever get newer expenditure data information.

I refactored the `useGetExpenditureData` hook be removing a redundant setTimeout because we don't actually need it. Basically, this hooks just needs to be very simple -- start or end the polling based on `shouldPoll`, and that's pretty much it. 

This hook is used in other areas of the app. I'm not entirely sure what the use case is in those other areas, but I am certain that as far as the Payment Builder Widget goes, we need  to poll for information whilst the Widget is mounted. So I ended up adding a `pollUntilUnmount` option which pretty much forces the poll to continue in the background regardless of original condition.

payment-widget-poll

## Testing

1. Create an Advanced Payment
2. On the Review step, click Confirm details
3. Verify that you can eventually reach the Funding step, without you having to reload the app/close and reopen the completed action component

Resolves #3719


bassgeta (approved):
Nice catch! It was quite a painful bug, so nice job sorting it out nicely :sunglasses: 
We don't really need to do that here, but we should think about using a subscription query for this instead of polling.
I've tried it out multiple times, and the advanced payment always goes through, nice one!

rumzledz:
Agreed that a subscription would be great! But I feel like if I did that here, I would be forced to refactor all places that could benefit from it 👀 Maybe we can treat it as a "something to do for the future" sort of thing!

Also apologies your approval got wiped because I had to push a small change 😅

bassgeta (approved):
Still working and looking good, reapproving :sunglasses:

iamsamgibbs (approved):
Nice job, this seems to have helped make sure the review step progresses properly. I created a few expenditures without issue.

https://github.com/user-attachments/assets/a1c0887a-b64b-45cb-8d53-c06fe4de41f0

Definitely agree we should update this (and motions) to use subscriptions but that will definitely be a bigger rework.

rdig (approved):
Seamless transition all the way to releasing the payment. Nicely done


https://github.com/user-attachments/assets/72a43bdd-8ea7-4a67-bfb4-ad418ebbb7d0

PR #3750, opened 2024-11-21
fix: button outline in dark mode

### LINKED ISSUE #3610: Team Switcher focus state on Dark Mode
## Steps to reproduce

Select a team then tab through the different teams.

Screenshot 2024-11-04 at 11 33 23

Screenshot 2024-11-04 at 11 33 28

Screenshot 2024-11-04 at 11 33 31

The focus states have extra white around the borders which should be removed.

Nortsova:
## Description
This was a pretty tricky bug where the outline, for some reason, added a second white outline 😵‍💫
image
image

## Testing
Step 1. Switch to the dark mode
Step 2. Focus on any button.

## Explaining my solution: 
**Default Focus Outline:**
Here’s how the focus looked with the default outline:
image

**Focus with Solid Outline Style:**
Applying a solid outline style caused this issue:
image

**Second Problem – Outline-Solid with Parent `overflow: hidden`**
When the parent container had `overflow: hidden`, this problem arose:
image

For some reason, the default outline with a white line worked fine in this case 🤯:
image

**Fix: Using offset: -1px 🎉**
To resolve this issue, I adjusted the outline offset by -1px:
image

**Final Touch: Removing Unnecessary Outlines**
For non-focused buttons, I removed unnecessary outlines by applying `outline-color: transparent`:
image

Resolves #3610


iamsamgibbs (commented):
Nice job deciphering this strange issue! It does indeed resolve the issues with the team switcher:

Screenshot 2024-11-22 at 09 56 45

Screenshot 2024-11-22 at 09 56 52

Screenshot 2024-11-22 at 09 56 57

I'm not sure whether the second white outline is an intentional design decision or not, but I think removing it does look better in most cases. Keeping it in and dealing with all the overflow hidden issues sounds like it won't be possible to make it look good. If we remove it, we should aim to remove it everywhere though as currently it is quite inconsistent:

https://github.com/user-attachments/assets/96ee32d4-e0fd-4eaf-880c-05a87555bbfc

And results in inconsistent focus states here:

https://github.com/user-attachments/assets/ff360384-83c2-4121-aa95-976828b62baa

The changes have created a bit of a weird look on these buttons though:
 
Screenshot 2024-11-22 at 09 58 58

Compared to master:

Screenshot 2024-11-22 at 09 59 42

It also changes the look of the focus states here:

https://github.com/user-attachments/assets/9b3ee324-fe7f-4f7b-8fdf-3a9359960807

Compared to master:

https://github.com/user-attachments/assets/a41fc41f-b096-4e75-93f7-6effa3c0ce07

We might need a definite design confirmation on how focus states should look across the app, then an investigation into how achievable that is, then applying the changes consistently across the app. Which sounds like a much much bigger job! I wonder if you could apply the changes here specifically to the team switcher (and leave all other buttons untouched for now) as those focus states looked particularly bad before this fix?

Nortsova:
Oh, thank you, @iamsamgibbs, for your review! 🙌 I initially thought the problem was caused by the white color of the outline. In this case, you are right - we can fix only the specific buttons with this issue. (The issue occurs due to the parent container's `overflow: hidden` property.)

bassgeta (approved):
This sure was a weird error with an actual okay solution, glad you were able to find a solution!
image
When i use tab to go to the button:
image

Other teams:
image
And when I tab onto the button
image

Nortsova:
Thank you, @iamsamgibbs 🙌  

I tried another approach now. And now it should be good: 
outline-tab


I fixed #3753 in my last commit as well (that bug also affected outlines)

iamsamgibbs (approved):
10/10 this looks lovely! Thanks for sticking with this pain of an issue!

Screenshot 2024-11-25 at 14 19 22

Screenshot 2024-11-25 at 14 19 29

Screenshot 2024-11-25 at 14 19 35

rumzledz (approved):
Outline's looking prettier than ever! ✨ 

pretty-outline

PR #3553, opened 2024-10-30
[Multi-sig] feat: implement multi-sig support for funding an advanced payment

bassgeta:
## Description

This PR adds Multi-sig support for the funding step.
Note: for some reason, the payment builder widget can get stuck in a pending state, both in the confirm details and funding step... This was happening to me on `master` too, so we should definitely investigate why it happens, but I don't think we should fix it here...

[Block ingestor PR](https://github.com/JoinColony/block-ingestor/pull/290)
[Figma link](https://www.figma.com/design/SaO0Zj6e7yGqLzjEsxqVhy/Permissions?node-id=22101-46905&node-type=frame&t=88D3UBSl2i7HmUac-0)

## Testing

The testing itself isn't going to be a lot of work, but if you spot anything being broken, give me a shout.
Test it with `amy`, `fry` and `leela` with 3 browsers so we can cover the flow properly.

1. Install the Multi-sig extension, set the global threshold to 2
image
2. Give `leela` Multi-sig `Owner` permissions in `General` , `amy` the Multi-sig `Payer` in `General` and `fry` Multi-sig Payer in `Andromeda`
image
image
image
3. Create an advanced payment action sending out tokens to `amy` and `leela` from `General`
image
4. Get to the `Funding` step, choose Multi-sig as the decision method. Also verify that you get a notification
image
image
image
5. Refresh the page, verify that the widget loading isn't janky (just a skeleton should appear)
6. Let's try out rejection first, Remove your vote as `leela` and then Reject the multisig. Verify that the status pill changed to `Rejected` and that you get the funding button back. Verify that you get a rejected multisig notification.
image
image
image
7. Now create a multi sig motion for funding again. Verify that you can switch between the two.
image
8. Open up the payment action as `amy` and verify that she can sign it and you can switch between actions. The first one should be open by default
image
9. Open the action up as `fry` and verify that you can't vote on it, but you can peruse at your own leisure
image
10. Right back at it, as `amy` approve the funding step and Finalize the motion. Verify that you get notifications for approval and finalization
image
11. Now release as `leela`, wrap it up and verify that `amy` and `leela` got the funds
image
image
Right additionally let's see if `fry` can sign from `Andromeda`.
1. As `leela` create an advanced payment sending some CREDS to 1 recipient (we test an edge case where an advanced payment to 1 recipient is parsed as just `moveFunds` in the block-ingestor.
image
2. Get to the funding step and as `fry` choose Multi-sig as the decision method (just so we verify not only the payment creator can do that)
image
image
[...]

PR #3761, opened 2024-11-23
Fix: Prevent TX button from being stuck in a pending state

### LINKED ISSUE #3760: The Transaction Button can oftentimes be stuck in the Pending state
## Steps to reproduce

1. Install the Multi-Sig extension
2. Set the threshold to 2
3. Give Leela and Amy permissions to take actions via Multi-Sig
4. Create an Advanced Payment
5. Progress until you reach the Funding stage
6. Click the "Fund" button:

image

7. Select the Multi-Sig decision method:

image

8. Submit the modal
9. Remove your vote by clicking the Remove button:

image

10. Reject the motion
11. Repeat steps 4 - 10, at some point, the app's Transaction Button will be stuck in the Pending state:

perpetual-pending

12. Install the Reputation extension

## Expected behaviour

The app's Transaction Button should correctly reflect the correct global transaction state.

## Actual behaviour

The app's Transaction Button is stuck on the Pending state.

pending-after-reputation-installation

rumzledz:
## Description

I found that some grouped transactions do legitimately have transactions with a zero address hash so I removed the part inside `getGroupStatus` where it determines if a transaction group is considered to have succeeded if there are no transactions with an address zero hash.

perpetual-pending-fix

## Testing

> [!NOTE]
> If you do end up seeing the TX Button in a "stuck" pending state, verify if it is indeed pending by checking the same transaction inside the User Hub. If the transaction is marked as Pending inside the User Hub then, it really is in a "Pending" state. Maybe it's an environment issue or something 🤔 

1. Connect Leela's wallet
2. Install the Reputation extension
3. Uninstall it
4. Repeat steps 1-2 a couple of times and verify that the TX Button is not stuck in a Pending state
5. Install the Multi-Sig extension
6. Give Amy and Leela permissions to take actions via Multi-Sig
7. Create an advanced payment
8. Fund it via Multi-Sig
9. Remove your default Yay vote
10. Reject the Multi-Sig motion
11. Repeat steps 8-10
12. Verify that the TX Button is not stuck in a perpetual loading state

Resolves #3760

### DIFF
--- a/src/components/common/Extensions/UserHub/partials/TransactionsTab/partials/GroupedTransaction.tsx
+++ b/src/components/common/Extensions/UserHub/partials/TransactionsTab/partials/GroupedTransaction.tsx
@@ -15,6 +15,7 @@ import {
   getGroupValues,
   getActiveTransactionIdx,
   getGroupId,
+  getHasZeroAddressTransaction,
 } from '~state/transactionState.ts';
 import { TRANSACTION_METHODS } from '~types/transactions.ts';
 import { arrayToObject } from '~utils/arrays/index.ts';
@@ -55,6 +56,8 @@ const GroupedTransaction: FC<GroupedTransactionProps> = ({
   const groupId = getGroupId(transactionGroup);
   const status = getGroupStatus(transactionGroup);
   const values = getGroupValues<TransactionType>(transactionGroup);
+  const hasZeroAddressTransaction =
+    getHasZeroAddressTransaction(transactionGroup);
 
   const groupMsgId = `transaction.group`;
 
@@ -95,7 +98,8 @@ const GroupedTransaction: FC<GroupedTransactionProps> = ({
     !GROUP_KEYS_WHICH_CANNOT_LINK.includes(
       values.group.key as TRANSACTION_METHODS,
     ) &&
-    status === TransactionStatus.Succeeded;
+    status === TransactionStatus.Succeeded &&
+    !hasZeroAddressTransaction;
 
   const createdAt =
     transactionGroup?.[0].createdAt &&
@@ -127,7 +131,7 @@ const GroupedTransaction: FC<GroupedTransactionProps> = ({
               onClick={handleNavigateToAction}
               disabled={!canLinkToAction || hideSummary}
               className={clsx(
-                'flex w-full flex-col items-start gap-1  sm:px-6',
+                'flex w-full flex-col items-start gap-1 sm:px-6',
                 {
                   'cursor-default': !canLinkToAction,
                 },
--- a/src/state/transactionState.ts
+++ b/src/state/transactionState.ts
@@ -115,17 +115,14 @@ export const getGroupStatus = (txGroup: TransactionType[]) => {
   if (txGroup.some((tx) => tx.status === TransactionStatus.Failed)) {
     return TransactionStatus.Failed;
   }
-  if (
-    txGroup.every(
-      (tx) =>
-        tx.status === TransactionStatus.Succeeded &&
-        tx.hash !== DEFAULT_TX_HASH,
-    )
-  ) {
+  if (txGroup.every((tx) => tx.status === TransactionStatus.Succeeded)) {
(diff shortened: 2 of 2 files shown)


src/components/common/Extensions/UserHub/partials/TransactionsTab/partials/GroupedTransaction.tsx
```diff
@@ -55,6 +56,8 @@ const GroupedTransaction: FC<GroupedTransactionProps> = ({
   const groupId = getGroupId(transactionGroup);
   const status = getGroupStatus(transactionGroup);
   const values = getGroupValues<TransactionType>(transactionGroup);
+  const hasZeroAddressTransaction =
+    getHasZeroAddressTransaction(transactionGroup);
```
rumzledz, line 60: Still needed to preserve the fix done on https://github.com/JoinColony/colonyCDapp/pull/2945

iamsamgibbs (approved):
Nice job! I would have expected this to be a much more complex issue, so I'm glad it was a relatively small fix!

Tested installing and uninstalling the extension a bunch of times and the pending button no longer gets stuck.

I tried following the rest of your testing steps to create an advanced payment funding expenditure, but it is not possible on master (I'm guessing you originally worked on this on Kerry's branch?) So I instead did a couple of funding reputation motions instead without issue.

I tried clicking a few transactions from the transactions tab to make sure they were still clickable and all seems to still be working.

I believe @chmanie worked on this pending button quite a lot, so if possible might be good to get a sanity check from him that this change won't have any intended consequences!

mmioana (changes requested):
Nice start on this issue @rumzledz 🙌 

Installing and un-installing the `Reputation Weighted` extension

https://github.com/user-attachments/assets/3866601c-7cba-4cac-b585-66ac88ad21a3


As @iamsamgibbs I couldn't test the `Multi-sig` funding step for an `Advanced payment` as I guess the base branch should've been [feat/3497-funding-expenditure-multisig](https://github.com/JoinColony/colonyCDapp/pull/3553)


Even though with your fix I couldn't reproduce any longer getting the `TX button` in a never-ending `Pending` state, I believe there is an issue if you try to open a transaction related to the extension installation. 
However @rumzledz  if you believe it makes more sense to treat it in a separate PR let me know and I'll approve 🙌 


https://github.com/user-attachments/assets/f7b3da5a-8b5b-4ead-83d2-a2bc80c45f69

rumzledz:
Hey @iamsamgibbs ! That's a good shout! I'll speak to @chmanie just in case 🤙 

The issue that @mmioana mentioned about the User Hub is something that's already on `master`(I didn't introduce it on this PR) and I believe it's been there for quite a while. It's also a very different issue from the stuck Pending button. So I've requested that we create a different issue for it so we can deal with it separately 🙌

chmanie:
I believe I was summoned?

rumzledz:
@iamsamgibbs I have just received a 🦭 of approval from @chmanie 🚀

rdig (changes requested):
I manage to get it stuck again while voting/revealing a motion

Granted, on all the other instances I've tried (installing, uninstalling extensions) the bending state went again after the transaction was finished, so I think it's almost there


https://github.com/user-attachments/assets/dee6e62a-34d5-492c-a650-aa83400549c7

rdig (approved):
Never mind, I can no longer replicate it.

Most likely you were correct in your assessment Romeo, that it could have been that the actual tx wasn't finished, hence why the pending state was still displaying

Here's another go at it, and after the transactions finish, the pending state goes away.

Nicely done!


https://github.com/user-attachments/assets/b8110d07-a1b0-4515-8800-c1bc16a72f95

rumzledz:
@rdig @iamsamgibbs oh right I see! I saw a similar thing in some other PR I reviewed recently as well where there was a tiny window of opportunity for me to click the "Change vote" button before the form properly progresses to the next step. In any case, I'll stay vigilant about this pending state 👌

PR #3780, opened 2024-11-26
Fix: Multi-sig extension settings issues

### LINKED ISSUE #3034: Multi-sig extension customize thresholds should be expanded if different
## Steps to reproduce

1. Ensure you have teams in the colony
2. Install and enabled the Multi-sig extension.
3. Navigate to the extensions settings screen.
4. Set custom thresholds for a team, different to colony-wide.
5. Save settings.
6. Reload the page and return to the settings screen.
7. Note that the configuration is hidden within a collapsed draw.

Note: I feel this was raised previously, so, credit to whomever raised it in the past.

## Expected behaviour

* You should be able to see custom team threshold settings by default on page load.

## Actual behaviour

* Extension settings are hidden within a collapsed draw.

image

image

### LINKED ISSUE #3598: Multi-Sig threshold settings change is not applied until after refresh
This affects the settings on the extension settings page not updating correctly, but also the insufficient members banner not being shown and the incorrect number of required approvals being shown when creating a multi-sig action.

## Steps to reproduce

* Step 1 - Install multi-sig extension.
* Step 2 - Change the threshold setting to fixed.
* Step 3 - Navigate to the dashboard.
* Step 4 - Navigate back to the multi-sig extension settings page.

## Expected behaviour

The newly saved settings should be shown without refreshing.

## Actual behaviour

The previous settings will appear, requiring a refresh before the new saved settings appear.

https://github.com/user-attachments/assets/15e84d1a-4741-4345-b692-06d0909d489d

## Steps to reproduce

* Step 1 - Install multi-sig extension
* Step 2 - Give Leela Owner Multi-Sig permissions.
* Step 3 - Change the threshold setting to fixed 1 and save changes.
* Step 4 - Refresh the page.
* Step 5 - Set the threshold to fixed 10 and save changes. Do not refresh.
* Step 6 - Create a simple payment action and select the multi-sig decision method. Notice how you are still able to create the action despite there only being 1 member and the threshold being changed to 10.
* Step 7 - Notice how it goes straight to the finalize step and if you check the approval step it will show enough signatures. (Note if you attempt to finalize here the transaction will fail as the contract has correctly updated to the new threshold of 10).
* Step 8 - Refresh. Notice the number of approvals has now increased to match the threshold.

## Expected behaviour

The threshold should correctly update when saving extension setting changes. The "insufficient members with permissions" banner should show preventing you from being able to create a multi-sig action with insufficient members.

## Actual behaviour

https://github.com/user-attachments/assets/7af704cf-5871-4169-a13c-6b7972521d8b

iamsamgibbs:
## Description

This PR adds an improvement and fixes an issue with the extensions settings tabs.

The extension data is now refetched when saving changes to an extension and properly waits for these updates to be reflected in the database.

The multi-sig custom settings accordion is now open by default if any threshold is different than the colony threshold.

## Testing

* Step 1 - Install the Multi-Sig extension
* Step 2 - Change the threshold setting to "Fixed threshold" with a value of `2`.

Screenshot 2024-11-27 at 10 06 02

* Step 3 - Without refreshing, navigate to the dashboard, then back to the multi-sig extension page. Check the threshold is still set to 2. (If you really want to be sure the threshold has updated correctly, then go through the extra steps of assigning multi-sig permissions and checking the threshold is as expected)

Screenshot 2024-11-27 at 10 06 25

* Step 4 - Open the "Customize thresholds per team" accordion and set one of the domains to "Majority approval", then click "Save changes"

Screenshot 2024-11-27 at 10 09 29

* Step 5 - Without refreshing, navigate to the dashboard, then back to the multi-sig extension page. Check the "Customize thresholds per team" accordion is already open.

## Diffs

**New stuff** ✨

* Added new `waitForCondition` utility function
* Added new `waitForDbAfterExtensionSettingsChange` function which is called on save changes with configurable extensions to ensure extension data is refetched after the change in settings is applied

**Changes** 🏗

* Added new useEffect to MultiSigSettings component to compare domain thresholds against the global threshold and toggle the accordion open if any domain thresholds do not match

Resolves #3598
Resolves #3034

### DIFF
--- a/src/components/frame/Extensions/pages/ExtensionDetailsPage/partials/ExtensionSettings/MultiSigSettings/MultiSigSettings.tsx
+++ b/src/components/frame/Extensions/pages/ExtensionDetailsPage/partials/ExtensionSettings/MultiSigSettings/MultiSigSettings.tsx
@@ -1,11 +1,12 @@
 import { CaretDown } from '@phosphor-icons/react';
-import React, { type FC } from 'react';
+import React, { useEffect, type FC } from 'react';
 import { defineMessages } from 'react-intl';
 
 import { InputGroup } from '~common/Extensions/Fields/InputGroup/InputGroup.tsx';
 import RadioBase from '~common/Extensions/Fields/RadioList/RadioBase.tsx';
 import { useExtensionDetailsPageContext } from '~frame/Extensions/pages/ExtensionDetailsPage/context/ExtensionDetailsPageContext.ts';
 import useToggle from '~hooks/useToggle/index.ts';
+import { isInstalledExtensionData } from '~utils/extensions.ts';
 import { formatText } from '~utils/intl.ts';
 import Select from '~v5/common/Fields/Select/Select.tsx';
 import AccordionItem from '~v5/shared/Accordion/partials/AccordionItem/index.ts';
@@ -105,8 +106,31 @@ const MultiSigSettings: FC = () => {
     handleGlobalThresholdTypeChange,
   } = useThresholdData(extensionData);
 
-  const [isCustomSettingsVisible, { toggle: toggleCustomSettings }] =
+  const [isCustomSettingsVisible, { toggle: toggleCustomSettings, toggleOn }] =
     useToggle();
+
+  useEffect(() => {
+    const multiSigConfig = isInstalledExtensionData(extensionData)
+      ? extensionData.params?.multiSig || null
+      : null;
+
+    const { colonyThreshold, domainThresholds } = multiSigConfig || {};
+
+    if (!domainThresholds) {
+      return;
+    }
+
+    // If any custom settings have been adjusted, show the custom thresholds accordion on load
+    const shouldShowCustomSettings = domainThresholds.some((domain) => {
+      const { domainThreshold } = domain || {};
+
+      return domainThreshold !== colonyThreshold;
+    });
+
+    if (shouldShowCustomSettings) {
+      toggleOn();
+    }
+  }, [extensionData, toggleOn]);
 
   const handleOnKeyDown = (e) => {
     if (['.', ','].includes(e.key)) {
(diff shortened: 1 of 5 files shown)


mmioana (commented):
Hey @iamsamgibbs thanks for your super nice fix 🥇 👏 

Tested updating the `global` threshold and upon returning to the extension page, the value is still displayed

https://github.com/user-attachments/assets/e7a278b2-3531-4059-8db7-fe2b63ad6d35

Also if we update the threshold per team, the values get saved and upon return to the extension page, this section is toggled on

https://github.com/user-attachments/assets/1163439e-8091-4afd-836d-7a24ed3a5f4b


Just one thing that I have noticed: if you try to save the same values, after the timeout expires, an error toast is shown

https://github.com/user-attachments/assets/2b00fcfa-9a31-4eca-978a-2b110ca162d7

So I'm wondering if there is anything we can do about it (maybe not show the `Save settings` button if the values are identical with the already saved ones) or even if it makes sense to solve this issue in this PR. What do you think @iamsamgibbs?

iamsamgibbs:
> Just one thing that I have noticed: if you try to save the same values, after the timeout expires, an error toast is shown

I actually tried to implement hiding the save button as you suggested, but it turns out to be a bit of a pain. There is already an `isDirty` check to show / hide the button, but it doesn't reset itself if you re-select the same values. So yes to implement properly will require comparing all the form values against the currently saved values which is a bit more involved. I'll open a new issue for it. :)

mmioana (approved):
Thanks @iamsamgibbs! Then I'll approve this PR 💪

rdig (approved):
Tested this all without refreshing and all values show up as expected. Nicely fixed!


https://github.com/user-attachments/assets/91b40aa0-e539-43d4-9e85-ee09a2073008

bassgeta (changes requested):
`waitForCondition` is such a great helper, kudos, we'll be able to use this to solve a lot of cases, (looking at you `PaymentBuilderWidget` :eyes: )
The threshold is correctly updated :heavy_check_mark:  And I can also confirm that I could immediately start craeting multisig actions when I had enough memebrs
https://github.com/user-attachments/assets/84edac2b-2ded-471b-985f-aeb3e7c05ef9

I have however found a bug which.... I hope won't be too much of a pain to solve :')
If you set the global threshold to something else, it will update it, but not the domain values, so it opens the accordion.
I couldn't reproduce it 100% of the time, but basically
1. set the threshold to 2, refresh the page
2. set the threshold to 1 and see that the accordion opens and the previous values are used
Attaching a video where you can see how this failed to happen the first time, but it happened the second time :shrug: 

https://github.com/user-attachments/assets/71db1e54-e2fe-4505-a84b-497e75fac07c

bassgeta (approved):
As discussed in Discord, it's actually quite a lot of work fixing this bug, so let's do it as a separate issue :ship:

PR #3782, opened 2024-11-27
Fix: Handle decision method validation properly for Split and Staged payments

### LINKED ISSUE #3774: [Permission] User doesn't have ability to make the payments
## Steps to reproduce

1. Assign the Payer role to the user in Andromeda.
2. Disable "Staking Advanced Payments" and "Reputation Weighted" extentions.
3. User opens staged and split payments.

## Expected behaviour

* The user with the required permissions should be able to create a payment without encountering the permission error in Andromeda team.

## Actual behaviour

* The system displays an error message, preventing the user from creating a payment despite having necessary permissions.

image

rumzledz:
## Description

We have a hook called `getPermissionsDomainIdForAction` which checks if there is an available decision method for a user on any given form. It has a section where it checks if the user has permissions to do the current action based on the currently selected team on the form itself.

The problem was that, we weren't handling the Staged and Split payment types. As a result, the form's "team" value is being incorrectly set to the root domain ID even before the user gets the chance to select the team on the form. So what ends up happening is that the hook incorrectly asserts that the user doesn't have permissions to do the current action type since the user doesn't have any permissions on the root domain. This is wrong because the user hasn't even selected the team at this point.

This PR adds cases for the Split & Staged payment types so that the `getPermissionsDomainIdForAction` hook correctly waits and correctly processes the selected team before it does its magic.

split-staged-fixes

## Testing

1. Disable the Reputation and Staking extensions
2. Enable the Staged payments extension
3. Assign Payer permissions for Fry in team Andromeda
4. Connect Fry's wallet and bring up the Split Payments form
5. Verify that the form is editable
6. Select team General
8. Verify that you can see this error banner:

image

9. Verify that the form's submit button is disabled
10. Set the team field to Andromeda
11. Verify that the banner goes away
12. Verify that the form's submit button is enabled
13. Set the action type to Split Payments
14. Repeat steps 5 - 12

Resolves #3774

### DIFF
--- a/src/components/v5/common/ActionSidebar/hooks/permissions/helpers.ts
+++ b/src/components/v5/common/ActionSidebar/hooks/permissions/helpers.ts
@@ -124,6 +124,7 @@ export const getPermissionsDomainIdForAction = (
   switch (actionType) {
     case Action.SimplePayment:
     case Action.PaymentBuilder:
+    case Action.StagedPayment:
       if (!isMotion) {
         return from;
       }
@@ -133,6 +134,7 @@ export const getPermissionsDomainIdForAction = (
       return undefined;
     case Action.ManageReputation:
     case Action.EditExistingTeam:
+    case Action.SplitPayment:
       if (!isMotion) {
         return team;
       }


rdig (approved):
Working as expected. Nicely done!

Screenshot from 2024-11-28 12-10-37
Screenshot from 2024-11-28 12-11-24
Screenshot from 2024-11-28 12-12-37
Screenshot from 2024-11-28 12-13-01

bassgeta (approved):
Ez fix, ez game, let's go :ship: 
As fry the form is editable :heavy_check_mark: 
image
Error banner when I choose permissions :heavy_check_mark: 
image
Can create it for Andromeda :heavy_check_mark: 
image
And the same is true for staged payments :heavy_check_mark: 
image
image
image

iamsamgibbs (approved):
Thanks for this fix, all working as expected!

Screenshot 2024-12-02 at 10 43 22

Screenshot 2024-12-02 at 10 43 28

Screenshot 2024-12-02 at 10 43 48

Screenshot 2024-12-02 at 10 43 56

PR #3788, opened 2024-11-27
fix: use correct domain permission proofs for move funds motion

### LINKED ISSUE #3704: bug: you can't create a motion to transfer funds from a  subdomain to the parent domain
## Steps to reproduce

1. Install and enable the voting reputation extension
2. Try to create a `Transfer funds` action moving some CREDS from `Andromeda` to `General`
image
3. The transaction estimation fails with reason `voting-rep-invalid-domain-id` 

## Suggestions (optional)

This happens because our `moveFunds` action creation doesn't set the third parameter correctly. We should have an action domain which we check permissions for, and then use that. Instead it always uses the `fromDomain`, but the "action domain" depends on where are we moving funds from and to.
image

While you're at it, change the third parameter of the multisig motion call too, to use the correct "action domain" instead.

bassgeta:
## Description

This PR just uses the new helpers for getting move funds permission proofs for motions and thus fixes the bug :100: 

## Testing

Best way is to just test **everything** :D 
1. Install the voting reputation extension
2. Install the Multi-sig extension, set the global threshold to 1, give `leela` the `Owner`/`Payer` multi-sig role in `General`
3. Create a move funds voting reputation motion moving 1 CREDS from `Andromeda` to `General`. Fully support it, `npm run forward-time 1` and finalize it.
image
4. Create a move funds voting reputation motion, moving 1 CREDS from `Andromeda` to `Rocinante`. Fully support it, `npm run forward-time 1` and finalize it.
image
5. Create a move funds voting reputation motion moving 1 CREDS from `General` to `Rocinante`. Fully support it, `npm run forward-time 1` and finalize it.
image
6 7 and 8 are the exact same but for multi-sig :D 
image
image
image

## Diffs

**Changes** 🏗

* the move funds motion saga now uses new helpers to get the action domain and skill indices

Resolves #3704

### DIFF
--- a/src/redux/sagas/motions/moveFundsMotion.ts
+++ b/src/redux/sagas/motions/moveFundsMotion.ts
@@ -22,7 +22,8 @@ import {
   initiateTransaction,
   createActionMetadataInDB,
   getPermissionProofsLocal,
-  getChildIndexLocal,
+  getMoveFundsActionDomain,
+  getMoveFundsPermissionProofs,
 } from '../utils/index.ts';
 
 function* moveFundsMotion({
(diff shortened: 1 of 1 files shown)


rdig (approved):
All good here, now transfer from subdomains work properly.

Screenshot from 2024-11-28 12-32-21
Screenshot from 2024-11-28 12-32-58
Screenshot from 2024-11-28 12-35-17
Screenshot from 2024-11-28 12-37-01
Screenshot from 2024-11-28 12-39-10

iamsamgibbs (approved):
Nice job fixing this up! All working as expected with both reputation motions and multi-sig.

Before:
 
Screenshot 2024-11-29 at 10 35 48

After:

Screenshot 2024-11-29 at 10 36 21

I only took screenshots of the amounts, so won't upload them all here, but amounts all correctly updated after each motion.

mmioana (approved):
Nice job on this one @bassgeta 🎉 

Installed the extensions
Screenshot 2024-12-03 at 09 37 09
Screenshot 2024-12-03 at 09 37 34

Gave `leela` multi-sig permissions
Screenshot 2024-12-03 at 09 38 16

Transferred funds using `Reputation`
Screenshot 2024-12-03 at 09 39 38
Screenshot 2024-12-03 at 09 40 45
Screenshot 2024-12-03 at 09 42 02

Transferred funds using `Multi-sig`
Screenshot 2024-12-03 at 09 42 40
Screenshot 2024-12-03 at 09 43 00
Screenshot 2024-12-03 at 09 43 21

Good to be merged 💯

PR #3794, opened 2024-11-27
Fix: Adjust the input allowed for the Claim Delay input field

### LINKED ISSUE #3479: Entering a letter or space in the claim delay field breaks input
## Steps to reproduce

1. Open up an advanced payment action.
2. Enter a letter into the Claim delay field.
3. Note the appearance of the input.

## Expected behaviour

* Entering a letter or space into a number field should have no impact.

## Actual behaviour

* The alignment breaks and the placeholder is cut off upon entering a letter or space as the first character in the claim delay field.

image

rumzledz:
## Description

I'm just simply doing a check to see if the most recent version of the input conforms with our cleave rule. If not, then don't allow the field value to be updated.

## Testing

1. Bring up the Advanced Payment form
2. Enter a space or a letter inside the Claim delay field
3. Verify that the field value doesn't get updated

Resolves #3479

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/forms/PaymentBuilderForm/partials/ClaimDelayField/ClaimDelayField.tsx
+++ b/src/components/v5/common/ActionSidebar/partials/forms/PaymentBuilderForm/partials/ClaimDelayField/ClaimDelayField.tsx
@@ -67,6 +67,7 @@ const ClaimDelayField: FC<ClaimDelayFieldProps> = ({
     >
       <input
         {...field}
+        value={field.value ?? ''}
         ref={inputRef}
         name={name}
         className={clsx(
@@ -85,6 +86,13 @@ const ClaimDelayField: FC<ClaimDelayFieldProps> = ({
             event.target.value,
             formattingOptions,
           );
+
+          if (
+            (!field.value && !formattedValue) ||
+            formattedValue === field.value
+          ) {
+            return;
+          }
 
           if (
             wrapperRef.current &&


src/components/v5/common/ActionSidebar/partials/forms/PaymentBuilderForm/partials/ClaimDelayField/ClaimDelayField.tsx
```diff
@@ -86,6 +86,10 @@ const ClaimDelayField: FC<ClaimDelayFieldProps> = ({
             formattingOptions,
           );
 
+          if (formattedValue === field.value) {
+            return;
+          }
+
```
mmioana, line 92: From my testing, in order to avoid the issue upon initially entering a space, I believe we have to update this condition to 

```
          if (
            (!field.value && !formattedValue) ||
            formattedValue === field.value 
          ) {
            return;
          }
```

mmioana (changes requested):
Nice start on this issue @rumzledz 🙌 

Unfortunately I can still reproduce the issue upon initially entering a space

https://github.com/user-attachments/assets/a8a86512-a311-4713-aa03-4387bfebf205

Also noticed this error in the `console`
Screenshot 2024-11-28 at 10 07 28

So probably we need to add this prop to the input field `value={field.value ?? ''}`

Also left a comment for how we should update the `if` condition for the issue to not be present anymore on this field. However, please feel free to be ✨ creative ✨

mmioana (approved):
Smashed 💯 

https://github.com/user-attachments/assets/64a67a06-2f1a-4e86-90f2-10e66425efbf

And no error regarding the `uncontrolled input` is showing up in the console

Screenshot 2024-11-28 at 13 12 10

Nortsova (approved):
Niice ✨  I tested different whitespace characters and everything works great 👏 

claim-delay-enter-space

PR #3795, opened 2024-11-27
Fix: Properly calculate the reputation percentage

### LINKED ISSUE #3687: [Manage awards - QA] Incorrect percentage representation (~0%) when adding reputation awards.
## Steps to reproduce

1. Go to the "Manage reputation" action.
2. Add award.

## Expected behaviour

* The percentage change should accurately reflect the added reputation points.

## Actual behaviour

* The UI displays the percentage change as ~0%

image

rumzledz:
## Description

The calculation wasn't happening because this boolean check fails:

```js
typeof newPercentageReputation === 'number' && typeof percentageReputation === 'number'
```

It was failing because I guess that the real intention was to check if the values are numerically parsable. So I just tweaked it to set it free and reach its full potential 🪽

This is an issue on the Completed Action component as well so I fixed it there too.

BUT WAIT ✋🏻There's more!

I also updated the Completed Action form to handle the Manage Reputation Multi-Sig action. The Reputation motion's logic is that it shows the `<ManageReputationTableInMotion />` table when the motion hasn't finished yet. So I just did the equivalent of that for the Multi-Sig motion.

rep-calc

## Testing

1. Bring up the Manage Reputation form
2. Set it to Award reputation
3. Set the Member to ollie-orbit
4. Update the Change field
5. Verify that the percentage value shown for the Change field is (FINAL PERCENTAGE - ORIGINAL PERCENTAGE)
6. Keep track of the Change field percentage value
8. Submit the form
9. On the Completed Action component, verify that the percentage is correct based on the percentage value from the Change field

Resolves #3687

### DIFF
--- a/src/components/v5/common/CompletedAction/partials/ManageReputation/ManageReputation.tsx
+++ b/src/components/v5/common/CompletedAction/partials/ManageReputation/ManageReputation.tsx
@@ -57,6 +57,7 @@ const ManageReputation: FC<ManageReputationProps> = ({ action }) => {
   const { colony } = useColonyContext();
   const { nativeToken } = colony;
   const {
+    isMultiSig,
     isMotion,
     transactionHash,
     metadata,
@@ -66,12 +67,16 @@ const ManageReputation: FC<ManageReputationProps> = ({ action }) => {
     amount,
     fromDomain,
     annotation,
+    multiSigData,
   } = action;
+
   const { networkMotionState } = useGetActionData(transactionHash);
   const motionFinished =
     networkMotionState === NetworkMotionState.Finalizable ||
     networkMotionState === NetworkMotionState.Finalized ||
     networkMotionState === NetworkMotionState.Failed;
+
+  const multiSigFinished = multiSigData?.isRejected || multiSigData?.isExecuted;
 
   const isSmite = [
     ColonyActionType.EmitDomainReputationPenalty,
@@ -190,7 +195,7 @@ const ManageReputation: FC<ManageReputationProps> = ({ action }) => {
       )}
       {positiveAmountValue && (
         <>
-          {isMotion && !motionFinished ? (
+          {(isMotion || isMultiSig) && !motionFinished && !multiSigFinished ? (
             <ManageReputationTableInMotion
               isSmite={isSmite}
               amount={positiveAmountValue}
--- a/src/utils/reputation.ts
+++ b/src/utils/reputation.ts
@@ -6,6 +6,8 @@ import {
   getFormattedNumeralValue,
 } from '~shared/Numeral/index.ts';
 
+import { isNumeric } from './numbers.ts';
+
 export enum ZeroValue {
   Zero = '0',
   NearZero = '~0',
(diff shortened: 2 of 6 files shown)


mmioana (approved):
Nice nice nice @rumzledz the reputation percentage seems to properly update 🥇 

Screenshot 2024-11-28 at 09 28 46
Screenshot 2024-11-28 at 09 28 55
Screenshot 2024-11-28 at 09 29 30
Screenshot 2024-11-28 at 09 30 21
Screenshot 2024-11-28 at 09 30 28

@rumzledz I left a tiny refactoring comment so we don't duplicate the logic for computing the reputation percentage, but I'll approve this PR nonetheless 🙌

mmioana (approved):
Lovely, thanks for the supaaaa-fast refactoring @rumzledz 🥇 🤩 

Screenshot 2024-11-28 at 10 18 46
Screenshot 2024-11-28 at 10 19 03

src/components/v5/common/CompletedAction/partials/ManageReputation/ManageReputation.tsx
```diff
@@ -190,7 +195,7 @@ const ManageReputation: FC<ManageReputationProps> = ({ action }) => {
       )}
       {positiveAmountValue && (
         <>
-          {isMotion && !motionFinished ? (
+          {(isMotion || isMultiSig) && !motionFinished && !multiSigFinished ? (
```
davecreaser, line 198: This logic check works, but seems a little bit prone to false negatives if certain things change in the future. I think it would be neater as:

```ts
(isMotion && !motionFinished) || (isMultiSig && !isMotionFinished)
```

But, that's just personal preference so feel free to ignore me!

src/utils/reputation.ts
```diff
@@ -50,3 +52,11 @@ export const formatReputationChange = (
   const value = adjustConvertedValue(absoluteChange, decimals);
   return getFormattedNumeralValue(value, absoluteChange);
 };
+
+export const getReputationDifference = (
```
davecreaser, line 56: Nice to have this as a util!

A little question though - should we handle a case where both percentages are `ZeroValue.Zero`, in which case return `ZeroValue.Zero` instead of `ZeroValue.NearZero`?

davecreaser (approved):
Nice fix @rumzledz! Very neat. Just left a couple of very minor comments, but I'll leave those in your hands and approve.

Works great:

Screenshot 2024-11-29 at 11 45 13
Screenshot 2024-11-29 at 11 45 37

src/utils/reputation.ts
rumzledz: Hey @davecreaser ! If both values are `ZeroValue.Zero` / `"0"`, it will be parsed as numeric type `0` so in the end, the result will still be `0 - 0 = 0` 👌

davecreaser: Ahh of course of course, nice explanation! I completely ignored that of course `isNumeric` will return true for `ZeroValue.Zero`

rdig (approved):
The reputation percentage value shows correctly now. Nicely done.

Screenshot from 2024-12-03 20-08-03
Screenshot from 2024-12-03 20-08-15

Issue #3802, opened 2024-11-28
[Payments] A new validation error appears when the "From" field is left empty


olgapod:
## Steps to reproduce

1. Open any payment form.
2. Leave the "From" field empty.
3. Submit the form.
4. Observe the error message displayed in the validation section.

## Expected behaviour

* The error message should clearly state:
"The 'From' field is required. Please select a team."

## Actual behaviour

* The error message reads:
"From must be a 'number' type, but the final value was: 'NaN' (cast from value '')".
* This message is confusing and not actionable for users.

<img width="699" alt="image" src="https://github.com/user-attachments/assets/5f3bf76d-06c9-4648-b0d3-8a5dd666b56d">

PR #3804, opened 2024-11-29
test: simple payment e2e tests

olgapod:
## Description

 Added tests for Simple Payment action.
Test cases included in the PR:
- Should successfully create a simple payment
- Should display validation errors for required fields and invalid inputs
- Should display tooltips and user info on hover and click actions
- Should display the confirmation modal when the user tries to close the form
- Should not allow to create a simple payment with no permissions

## Testing

Step 1. Spin up the dev env and seed the test data 

`npm run dev`  followed by `node scripts/create-data.js`

Step 2. Remove your local playwright trace files

 `rm -rf playwright-report/` (from the root of the project )

Step 3. Build and start the frontend in preview mode

`npm run preview` 

Step 4. Run the tests in a separate terminal 

`npm run e2e simple-payment`

Step 5. The tests should pass ( if any failures - please send me .zip files from /playwright-report/data directory)

### DIFF
--- a/playwright/utils/simple-payment.ts
+++ b/playwright/utils/simple-payment.ts
@@ -10,7 +10,7 @@ export const validationMessages = {
   },
   allRequiredFields: [
     'decisionMethod must be defined',
-    // TODO: Uncomment once issue #38102 is fixed
+    // TODO: Uncomment once issue #3802 is fixed
     // 'from is a required field',
     'recipient is a required field',
     'Amount must be greater than zero',
--- a/playwright/utils/simple-payment.ts
+++ b/playwright/utils/simple-payment.ts
@@ -1,0 +1,51 @@
+/* eslint-disable no-warning-comments */
+import { type Page } from '@playwright/test';
+
+export const validationMessages = {
+  title: {
+    maxLengthExceeded: 'Title must not exceed 60 characters',
+    isRequired: 'Title is required',
+  },
+  allRequiredFields: [
+    'decisionMethod must be defined',
+    // TODO: Uncomment once issue #38102 is fixed
+    // 'from is a required field',
+    'recipient is a required field',
+    'Amount must be greater than zero',
+    'Title is required',
+  ],
+  amount: {
+    isRequired: 'Amount is required',
+    mustBeGreaterThanZero: 'Amount must be greater than zero',
+    maxExceeded: 'Not enough funds to cover the payment and network fees',
+  },
+  permissions:
+    "You don't have the right permissions to create this action. Try another action type.",
+};
+
+export const tooltipMessages = {
+  from: /Team source of the funds/i,
+  recipient: /Member or address designated to receive the funds/i,
+  amount: /Quantity and type of funds of the payment/i,
+  decisionMethod: /Mechanism behind the decision-making process/i,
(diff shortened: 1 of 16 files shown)


rdig (approved):
Tested this and everything works, test run nice and fairly fast.

Nicely done

Screenshot from 2024-12-03 20-51-49

playwright/utils/simple-payment.ts
```diff
@@ -0,0 +1,83 @@
...
+import { selectWallet } from './common.ts';
+
+export const validationMessages = {
+  title: {
+    maxLengthExceeded: 'Title must not exceed 60 characters',
+    isRequired: 'Title is required',
+  },
+  allRequiredFields: [
+    'decisionMethod must be defined',
+    // TODO: Uncomment once issue #38102 is fixed
```
iamsamgibbs, line 13: The issue number is actually #3802 - can you leave a note on the issue to re-enable this as part of the issue?

src/components/v5/common/ActionSidebar/partials/ActionSidebarContent/ActionSidebarContent.tsx
```diff
@@ -193,14 +196,19 @@ const ActionSidebarFormContent: FC<ActionSidebarFormContentProps> = ({
...
+              testId="action-sidebar-error"
+            >
               {customError.message?.toString()}
             </NotificationBanner>
           </div>
         )}
         {flatFormErrors.length ? (
           <div className="mt-7">
             <NotificationBanner
+              testId="action-sidebar-error"
```
iamsamgibbs, line 211: I think it might be possible for this to show at the same time as the customError above, so maybe these need different testIds?

iamsamgibbs (dismissed):
Nice start on adding these simple payment tests! I think the scope of the tests here is a bit limited and there is a lot we can add (shouldn't be able to set the amount higher than the available funds, should check if the network fee would bump it over the available funds, shouldn't be able to make a payment with a locked token outside of it's native colony, etc) but these are good tests of the basic form's functionality and we should get this merged and add the more advanced tests later on.

Screenshot 2024-12-05 at 11 48 02

iamsamgibbs:
This is a personal opinion thing, but it would be nice if these PRs included a list of what tests are included in the PR description just so it is easy to see without having to dig through the test files, just something like: 

* Should display validation errors for required fields and invalid inputs
* Should display tooltips and user info on hover and click actions
* Should display the confirmation modal when the user tries to close the form

Only if it's not a ton of extra hassle to add though, not sure if others would find this useful?

olgapod:
> shouldn't be able to set the amount higher than the available funds

Thank you for the review @iamsamgibbs. One note: the `shouldn't be able to set the amount higher than the available funds` is already there within the test for validation of the form. I will add the additional 2 you mentioned in the next PR. I will reach out to you in chat to ask about other possible ideas you have  for "advanced" tests. for this action

iamsamgibbs (approved):
Tests are all still passing

Screenshot 2024-12-09 at 11 20 23

Sounds good! I will have a think of what other tests we could add.

rdig (approved):
Re-approving

PR #3867, opened 2024-12-06
Fix: Equal split payment amount issue

### LINKED ISSUE #3751: [Split payment] App crashes when entering value in the amount field of the Split Payment form
## Steps to reproduce

1. Navigate to the Split Payment form.
2. Select Equal distribution.
3. Focus on the Amount field.
4. Enter 0.000000000000001 value into the field.
5. Observe that the application crashes immediately.

## Expected behaviour

* The app should accept numeric input in the Amount field without crashing.

## Actual behaviour

* The app crashes as soon as a numeric value is entered in the Amount field of the Split Payment form.

image
image

https://github.com/user-attachments/assets/1957c1d6-5ee1-42cd-84ff-5fccf1301de4

iamsamgibbs:
## Description

This PR fixes an issue where small amounts using the equal split type were defaulting to scientific notation which was breaking the amount fields in the table.

Whilst working on this I also found a weird edge case issue where entering `1M` (as in 1 million) would get adjusted by the cleave-zen `unformatNumeral` function to `1.` and if you added another `M` (`1.M`) the amount field would appear unchanged, but the action subtitle would show `1M00000000`.

## Testing

* Step 1 - Create a split payment action
* Step 2 - Select the "Equal" split type
* Step 3 - Enter the amount `0.00000001` - check that the app does not crash

Screenshot 2024-12-06 at 16 31 29

* Step 4 - Clear the amount field and enter `1M` check that the `M` is not entered and that the amount field just shows `1`, enter another `M` and check the amount field still just shows `1`.

Screenshot 2024-12-06 at 16 31 39

Compared to master:

Screenshot 2024-12-06 at 16 32 09

## Diffs

**Changes** 🏗

* unformatNumeral fix
* Equal amount now uses `toFixed`

Resolves #3751

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/AmountField/AmountField.tsx
+++ b/src/components/v5/common/ActionSidebar/partials/AmountField/AmountField.tsx
@@ -79,7 +79,8 @@ const AmountField: FC<AmountFieldProps> = ({
   const handleFieldChange = (e: ChangeEvent<HTMLInputElement>) => {
     const formattedValue = formatNumeral(e.target.value, formattingOptions);
 
-    const unformattedValue = unformatNumeral(e.target.value);
+    // Strip 'M' from the input as the character 'M' is used as a placeholder in the unformatNumeral function and can result in weird values
+    const unformattedValue = unformatNumeral(e.target.value.replace('M', ''));
 
     if (
       wrapperRef.current &&
@@ -95,7 +96,7 @@ const AmountField: FC<AmountFieldProps> = ({
     onChange?.();
 
     field.onChange(unformattedValue.replace('-', ''));
-    setValue(formatNumeral(e.target.value, formattingOptions));
+    setValue(formattedValue);
   };
 
   const handleTokenSelect = (selectedTokenAddress: string) => {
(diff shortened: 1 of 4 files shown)


rumzledz (approved):
Nice straightforward fix @iamsamgibbs ! ✨ 

split-no-crash

This is NOT something you introduced but I noticed that when the value is small enough, the distribution value is 0. But I guess it's just because the value is too damn small. Which makes me wonder if we need to put a lower limit or something 🤔 Again, not to be done on this PR

Screenshot 2024-12-09 at 18 25 01

mmioana (changes requested):
Nice start on this issue @iamsamgibbs 🎉 

I was able to create `Split payments` with an amount of `0.00000001`

And also when entering `1M` only `1` was displayed

The reason I'm rejecting this is due to the fact that when trying to redo the action and the amount was `0.00000001`, the app crashes. However with an amount of `1` everything works as expected.

https://github.com/user-attachments/assets/ad0669ee-bfde-4b23-aabd-78eb2f0260fa


https://github.com/user-attachments/assets/ffb4f2a7-a3e7-479c-89b0-31d6545515d0

iamsamgibbs:
🤦‍♂️ That damn redo button. Thanks for catching this @mmioana I will look into it!

davecreaser (approved):
Nice fixes @iamsamgibbs and thanks for leaving a comment in the repo to explain the potentially weird-looking code.

Split payments all worked with no crashes and no 'M's:
Screenshot 2024-12-09 at 14 18 13
Screenshot 2024-12-09 at 14 18 37
Screenshot 2024-12-09 at 14 19 43

Retrying also worked fine without crashing 🎉 

The only issue (not to be fixed here, but worth pointing out) is that when retrying a split payment which uses `0.00000001` as a value, and then changing the action type to simple payment, it looks like the simple payment form can never be submitted - even _after_ changing the value to something more regular like `1` or `10`

rdig (approved):
All good from me as well

PR #3914, opened 2024-12-11
Fix: Solve fetchDomainBalance lambda issues

mmioana:
## Description

There were some issues discovered while doing the prod deployment, causing for the displayed values to be always `0`.
The main one was due to the used fallback value for the `chainId` property. Given we don't really want to be a parametrisable property, this value has been removed from the lambda's input values. Also, given we want to use this value to filter out the coins list from `CoinGecko` this value has been renamed to `supportedNetwork` and will be set in `NetworkConfig`.

Another issue we might face is not having the token used for an action already stored in the DB - which we need to determine the number of decimals - so, we should use the `getTokenFromEverywhere`. This query relies on the `fetchTokenFromChain` lambda that will try to find the token if it is not already stored and then save it in the `DB`.

Introduced `apiTokenIdsData` in `ExchangeRatesService` only for optimisation, in order not to make extra calls for getting the whole coins list from `CoinGecko`.

Added the usage of the checksummed token address only to prevent some errors. 

As a summary, the following issues for the **fetchDomainBalance** lambda were fixed in this PR

- Remove chainId argument
- Use checksummed token address
- Replace chainId usage with the resolved network variable 
- Replace getToken with getTokenFromEverywhere query

## Testing

TODO: Just make sure the values shown for the `Total` cards and the `Total in and out` cards are still properly shown. 
You can also try running this query 

```
query MyQuery {
  getDomainBalance(
    input: {colonyAddress: "<COLONY_ADDRESS_HERE>", timeframePeriod: 4, timeframeType: DAILY}
  ) {
    total
    totalIn
    totalOut
    timeframe {
      key
      value {
        total
        totalIn
        totalOut
      }
    }
  }
}
```

Please also try changing the `query`'s input values and make sure everything still works as expected ✨

### DIFF
--- a/amplify/backend/api/colonycdapp/schema/balance.graphql
+++ b/amplify/backend/api/colonycdapp/schema/balance.graphql
@@ -15,13 +15,9 @@ input DomainBalanceArguments {
   """
   domainId: String
   """
-  The chain id the colony is on
-  """
-  chainId: String
-  """
   The currency in which we want to exchange the in/out token balance
   """
-  selectedCurrency: ExtendedSupportedCurrencies 
+  selectedCurrency: ExtendedSupportedCurrencies
   """
   The timeframe type (day/week/month/total) for which we want to get the in/out token balance
   Defaults to TimeframeType.MONTHLY
@@ -56,7 +52,7 @@ Return type for domain balance for a timeframe item
 type TimeframeDomainBalanceInOut {
   """
   The timeframe item key representing a day/week/month based on the selected timeframeType
-  """  
+  """
   key: String!
   """
   The timeframe item value holding the computed values
@@ -65,7 +61,7 @@ type TimeframeDomainBalanceInOut {
 }
 
 """
-Return type for domain balance 
+Return type for domain balance
 """
 type DomainBalanceReturn {
   """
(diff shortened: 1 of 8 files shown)


bassgeta (approved):
Changes seem very sensible and straightforward :100: 
The query returns data :heavy_check_mark: (for multiple timeframes too)
image

Cards still work as they should :heavy_check_mark: 
image
image

Switching teams also constantly worked, creating a payment invalidated the query and changed the numbers :heavy_check_mark:

iamsamgibbs (approved):
Nice work, the code changes look sensible and everything works as described:

Screenshot 2024-12-11 at 18 32 43

Screenshot 2024-12-11 at 18 32 52

Screenshot 2024-12-11 at 18 33 14

It would be nice if there was a bit more information in the PR description to describe exactly what problem this PR aims to resolve, especially as there is no linked issue. The code changes make sense though and I think I can understand the intent behind the changes, but just a bit of extra background info on the issue would be helpful in the future! :)

mmioana:
You're right @iamsamgibbs! Somehow in my head is was pretty clear what I tried to solve, but didn't put my thoughts down in this PR's description, which now I have adjusted 💯

iamsamgibbs:
Amazing, thank you very much @mmioana that is much clearer now!

rumzledz (approved):
The values from the GraphQL response corresponds with the data on frontend. Awesome work @mmioana ! ✨ 

Screenshot 2024-12-12 at 23 21 50

Screenshot 2024-12-12 at 23 22 03

Issue #3901, opened 2024-12-10
[Streaming payments] False insufficient funds error displayed after creating a stream


jakubcolony:
Teamwork task: [[Streaming payments] False insufficient funds error displayed after creating a stream](https://tw.auroracreation.com/#/tasks/16200861)

Affects `feat/streaming-payments-ui` branch.

## Steps to reproduce

This seems to be an intermittent issue, I had to create 3 streams before getting it to reproduce.

1. Install Streaming Payments extension.
2. Create a stream with a Start immediately option and with a relatively low (compared to your colony balance) amount.
3. Watch the completed action view.

## Expected behaviour

You are able to claim the stream straight away, provided the claimable amount is less than your colony balance.

## Actual behaviour

Every now and again, despite having enough balance, the insufficient funds error shows. It disappears after a page refresh.

https://github.com/user-attachments/assets/cfc07f41-df14-4347-b716-158caab568af

weaverderr:
Hi there,

I attempted to reproduce your issue and was unable to do so.

We use GitHub issues as a place to track bugs and other development-related issues. If your issue persists, please write us back using our “Chat Support” icon located on our [Colony Help Center](https://generalwebfixes.web.app/)

You can include a link to this issue in the message content.

Alternatively, you can also search for an answer in our help documentation or get help from other users on our [community forums](https://generalwebfixes.web.app/).

The issue here will be closed.

Thanks!

rdig:
Closed by #3901

PR #3946, opened 2024-12-13
fix: refresh available funds

CzarekDryl:
## Description

- An error appeared about funds even though there are funds in the colony
image

## Testing

* Install Streaming Payments extension
* Create a stream with a Start immediately option and with a relatively low (compared to your colony balance) amount
* Watch the completed action view

https://github.com/user-attachments/assets/0c7a298c-de79-457e-b765-6471cbcd6537

Resolves https://github.com/JoinColony/colonyCDapp/issues/3901

### DIFF
--- a/src/components/v5/common/CompletedAction/partials/StreamingPayment/partials/StreamingPaymentWidget/StreamingPaymentWidget.tsx
+++ b/src/components/v5/common/CompletedAction/partials/StreamingPayment/partials/StreamingPaymentWidget/StreamingPaymentWidget.tsx
@@ -97,6 +97,14 @@ const StreamingPaymentWidget: FC<StreamingPaymentWidgetProps> = ({
 
   const [hasEnoughFunds, setHasEnoughFunds] = useState(checkIfHasEnoughFunds());
 
+  useEffect(() => {
+    setHasEnoughFunds(checkIfHasEnoughFunds());
+  }, [
+    checkIfHasEnoughFunds,
+    amounts.amountAvailableToClaim,
+    streamingPaymentData,
+  ]);
+
   const claim = useAsyncFunction({
     submit: ActionTypes.STREAMING_PAYMENT_CLAIM,
     error: ActionTypes.STREAMING_PAYMENT_CLAIM_ERROR,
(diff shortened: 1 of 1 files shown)


mmioana (approved):
Nice work on this one @CzarekDryl 🙌 

Installed the `Streaming payments` extension
Screenshot 2024-12-16 at 14 55 05

Created a `Streaming payment` that `Starts immediately` and can confirm no error message was shown ✅  

https://github.com/user-attachments/assets/26674285-7514-469d-b934-c7bf55b7f14e

@CzarekDryl I have a question regarding the Streaming widget though. I've noticed after clicking `Pay funds` multiple times the amount available to claim remains the same. Is this as intended?

https://github.com/user-attachments/assets/d2a96fa4-503a-496e-bd88-d247ac3f33bf

Either way, nice work on this one 💯

davecreaser (approved):
Nice one @CzarekDryl - everything looks to be working here!

Tested a couple of streaming payments and didn't see the error appear once:
Screenshot 2024-12-16 at 16 46 36

The only thing is a slight code cleanup, which I'll leave in a separate comment :) 🚀

src/components/v5/common/CompletedAction/partials/StreamingPayment/partials/StreamingPaymentWidget/StreamingPaymentWidget.tsx
```diff
@@ -97,6 +97,14 @@ const StreamingPaymentWidget: FC<StreamingPaymentWidgetProps> = ({
 
   const [hasEnoughFunds, setHasEnoughFunds] = useState(checkIfHasEnoughFunds());
 
+  useEffect(() => {
+    setHasEnoughFunds(checkIfHasEnoughFunds());
+  }, [
+    checkIfHasEnoughFunds,
+    amounts.amountAvailableToClaim,
+    streamingPaymentData,
+  ]);
```
davecreaser, line 106: This dependency array looks like it only needs `checkIfHasEnoughFunds` in it, since that function comes from a useCallback inside a hook, which has the other variables in it's dependency array.

Also, a side note: personally I'm not a huge fan of the way that hook is set up, I feel like it could return a boolean of `hasEnoughFunds` rather than a callback. It makes the code just that bit harder to read and a little bit harder to debug and maintain. But anyway, it's unrelated to this PR so no worries, this is just the first time I saw it so it's the place where I'll make my comment 🤣

PR #3952, opened 2024-12-16
Fix: Prevent form reset on submit when errors are present

### LINKED ISSUE #3730: [Manage account] Issue with username Validation in Manage Profile page
Description:
The Manage Profile page’s username field has a restriction allowing changes only once every 90 days. However, in the process of e2e testing, an issue was discovered where the validation isn't working in some cases.

Reproducing this issue manually is challenging without modifying the source code to override the 90-day restriction. Take the steps below to replicate the issue (not waiting for 90 days to pass).

Prerequisites:

 • Update the source code to make the username editable:
 1. Navigate to:
src/components/frame/v5/pages/UserProfilePage/partials/UserAccountPage/hooks.tsx
 2. Modify line 31 to:

return { canChangeUsername: true, daysTillUsernameChange: 91 };

## Steps to reproduce

 1. Log in using Dev Wallet 2 account.
 2. Navigate to the Manage Profile page.
 3. In the username field, enter a username that is already taken (e.g., “leela”).
 4. Observe that the field provides feedback indicating the username is not available.
 5. Click the Save Changes button.
 6. Click the Save Changes button again.
 
 
 Scenarios 2
 1. Log in using Dev Wallet 2 account.
 2. Navigate to the Manage Profile page.
 3. In the username field, enter a New username.
 4. Observe that the field provides feedback indicating the username is available.
 5. Click the Save Changes button.

 

## Expected behaviour

* The system should show the validation message and prevent the username from being updated to an already-taken value.

## Actual behaviour

* The username is changed to the already-taken value, bypassing the validation.

https://github.com/user-attachments/assets/4be9dd07-16bd-4c4b-83ac-3d17757eb9c0

https://github.com/user-attachments/assets/1dc7b2bd-2e59-4ca3-8ea6-c83d6b066f73

rumzledz:
## Description

The problem is that the form was being reset when submitted even though errors are present. We don't want that to happen.

form-error-reset

## Testing

1. Open the `UserAccountPage/hooks.tsx` file and update line 31 to:

```js
return { canChangeUsername: true, daysTillUsernameChange };
```
2. Connect Amy's wallet (Wallet 2)
3. Go to the profile page: http://localhost:9091/account/profile
4. Enter "leela" in the Username field
5. Click the Save changes button repeatedly
6. Verify that the form doesn't submit

Resolves #3730

### DIFF
--- a/src/components/shared/Fields/Form/Form.tsx
+++ b/src/components/shared/Fields/Form/Form.tsx
@@ -103,7 +103,7 @@ const Form = <FormData extends FieldValues>({
     watch,
     reset,
     getValues,
-    formState: { isSubmitting },
+    formState: { isSubmitting, errors: formErrors },
   } = formHelpers;
   const values = watch();
 
@@ -112,10 +112,10 @@ const Form = <FormData extends FieldValues>({
    * Useful in user settings.
    */
   useEffect(() => {
-    if (isSubmitting && resetOnSubmit) {
+    if (isSubmitting && resetOnSubmit && !Object.keys(formErrors)) {
       reset(values);
     }
-  }, [isSubmitting, values, resetOnSubmit, reset]);
+  }, [isSubmitting, values, resetOnSubmit, reset, formErrors]);
 
   const submitHandler = useCallback<React.FormEventHandler<HTMLFormElement>>(
     (event) => {


src/components/shared/Fields/Form/Form.tsx
mmioana, line 1: @rumzledz would it make sense to have a variable `const hasErrors = !Object.keys(formErrors)` and use that in the `useEffect`?

```
  useEffect(() => {
    if (isSubmitting && resetOnSubmit && !hasErrors) {
      reset(values);
    }
  }, [isSubmitting, values, resetOnSubmit, reset, hasErrors]);
```

mmioana (approved):
Nice work on this one @rumzledz 🙌 

Can confirm the form doesn't get submitted if there are errors present

https://github.com/user-attachments/assets/132d5bfc-dbae-494e-865a-bb27aff02910

I left a small refactoring comment, but it shouldn't block this PR getting merged 💯

bassgeta (approved):
Nice catch, this is quite a silly bug honestly, glad it wasn't a pain to solve it :+1: 
image
On the right you can see an empty console, since nothing was submitted after spamming the `Save changes` button :D 

Nice one :muscle:

iamsamgibbs (approved):
Nice fix! I agree with @mmioana 's code suggestion.

Screenshot 2025-01-06 at 11 40 59

I think we should probably add a check to auth proxy to determine if the username is actually available as we're only doing a frontend check here which would be possible to bypass. I know we try to avoid adding anything to auth proxy unless it's really necessary, but I think we can check if the username is actually changing and only query the database in that case and it would probably be okay. (Also aware that at some point the user profile will be split into public / private, but seeing as the username is public I don't think this would be affected by that change).

PR #3994, opened 2024-12-19
test: split payment e2e tests

olgapod:
## Description

Initiatal set of E2E tests for the Split Payment motion. Also, I updated some selectors related to the `selectWallet` test helper to fix tests that began failing after the homepage redesign. 
Test cases included in the PR:
- Happy path of Split payment (Permissions/Staking method of create motion, Equal/Unequal distribution, Permissions release method)
- Form fields validation
- Recipients table actions: Add/Remove/Duplicate a row
- Redo a payment

## Testing

**Step 1**. Spin up the dev env and seed the test data 

`npm run dev`  followed by `node scripts/create-data.js`

**Step 2**. Build and start the frontend in preview mode

`npm run preview` 

**Step 3.** Run the tests in a separate terminal 

`npm run e2e split-payment`

**Step 4**. The tests should pass ( if any failures - please send me .zip files from `/playwright-report/data` directory)

### DIFF
--- a/playwright/e2e/manage-account-profile.spec.ts
+++ b/playwright/e2e/manage-account-profile.spec.ts
@@ -10,7 +10,10 @@ test.describe('Manage Account', () => {
 
     await selectWallet(page);
 
-    await page.getByTestId('user-navigation-hamburger').click();
+    await page
+      .getByRole('banner')
+      .getByTestId('user-navigation-hamburger')
+      .click();
 
     await page.getByTestId('user-menu').waitFor({ state: 'visible' });
 
--- a/playwright/e2e/split-payment.spec.ts
+++ b/playwright/e2e/split-payment.spec.ts
@@ -39,7 +39,7 @@ test.describe('Split payment', () => {
     await splitPayment.close();
   });
 
-  test('Happy path (Unequal distribution/Staking decision method)', async () => {
+  test('Happy path (Staking decision method)', async () => {
     await splitPayment.fillForm({
       title: 'Test Split Payment',
       team: 'General',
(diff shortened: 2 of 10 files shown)


iamsamgibbs (approved):
Nice job setting this all up! Tests are passing no problem. I think there is room for further tests here, the reputation distribution method is unused for example, and confirming the form returns the correct values for equal distribution etc.

Screenshot 2025-01-21 at 12 41 59

iamsamgibbs (changes requested):
Actually, my comment about being curious actually needs addressing I think. By manually updating the amounts and percentages it forcibly changes the distribution type to "Unequal"

Screenshot 2025-01-21 at 12 43 26

playwright/e2e/split-payment.spec.ts
```diff
@@ -0,0 +1,290 @@
...
+          amount: '1',
+          recipient,
+          percentage: '50%',
+        },
+        {
+          amount: '1',
+          recipient: recipient2,
+          percentage: '50%',
+        },
+      ],
```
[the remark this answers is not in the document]
olgapod: Yes, it does calculate the amount and percentage fields automatically in that case (thanks for bringing this up, I've added a test case for that). But,  If you start manually filling in the table then the Distribution is set to Unequal, and this is expected behavior (I've double-checked with Arren)

iamsamgibbs (approved):
Discussed this with @olgapod and didn't realise that the equal checks are only done before submitting, so makes sense why the resulting action shows as unequal.

playwright/e2e/manage-account-profile.spec.ts
```diff
@@ -10,7 +10,10 @@ test.describe('Manage Account', () => {
 
     await selectWallet(page);
 
-    await page.getByTestId('user-navigation-hamburger').click();
+    await page
```
mmioana, line 13: @olgapod would it make sense to refactor the code from lines `13-23` into an utility function, given this piece is repeating across all the manage account e2e tests?

mmioana (approved):
A lot of work to now have some cool tests for the `Split payment` action 🎉 really nicely done @olgapod!

The tests have passed for me, but only after setting back the timeout to `9000` in `playwright.config.ts`

Screenshot 2025-01-27 at 17 09 14

I will approve your PR nonetheless!

playwright/e2e/manage-account-profile.spec.ts
olgapod: I will add this refactoring  in the following PR

PR #4002, opened 2024-12-20
Fix: Disable accept incoming funds button until updated data

### LINKED ISSUE #3502: [Incoming funds]: Disable Accept button before it claimed
## Steps to reproduce

https://github.com/user-attachments/assets/ede2ed0a-879c-4677-93f5-2a700ccf93e7

We need to have some incoming funds so here is way to get it:

Step 1. Login with metamask
Step 2. Clean Metamask history (Metamask -> Settings -> Advanced -> Clear activity tab data)
Step 3. Navigate to Balance page
Step 4. Click Add funds to the Colony and Copy Wallet address
Step 5. Metamask -> send -> Paste Colony Wallet Address -> Send something to the colony (1 eth or if you choose xDai as Ganache Currency symbol choose xDai)
image

OR

* Step 6. Navigate to Wayne Enterprises
* Step 7. Create simple payment from General to Colony Wallet Address that you copied at Step 4. 
image

* Step 8. Try to Accept one fund
* Step 9. Verify that after loading button became normal clickable and then disappear as claimed

## Expected behaviour

* Step 9. Verify that after loading button is still disabled and then disappear as claimed

## Actual behaviour

* Step 9. Verify that after loading button became normal clickable and then disappear as claimed

## Suggestions (optional)
Context from this PR could be used: https://github.com/JoinColony/colonyCDapp/pull/3477

mmioana:
## Description

- This PR aims to keep the `Accept` buttons on the `Incoming funds` page disabled until the funds claims data is refetched and their `isClaimed` state has updated.

> [!NOTE]
> I have considered using subscriptions, however setting a subscription on the `Colony` model would not get triggered when updating the `fundsClaims`, while setting subscriptions for all unclaimed funds claims seems like a too-big of a performance overhead for what we want to achieve. Thus, I have introduced a new `isFundsUpdatePending` state on the `IncomingFundsLoadingContext` and make use of two new hooks, one for refreshing the colony data and one for checking the funds claims state.

## Testing

TODO: Now let's proceed to testing these changes behave accordingly.

* Step 1. Visit http://localhost:9091/planex and copy the colony address
* Step 2. Go to http://localhost:9091/wayne
* Step 3. Create a `Simple payment` using the copied colony address for the `Recipient` field
Screenshot 2024-12-23 at 12 23 28

* Step 4. Go back to planex http://localhost:9091/planex
* Step 5. Please save this in a new file - you can call it `.diff`
```
diff --git a/src/redux/sagas/actions/mintTokens.ts b/src/redux/sagas/actions/mintTokens.ts
index 7f68b83f1..6ead99ee8 100644
--- a/src/redux/sagas/actions/mintTokens.ts
+++ b/src/redux/sagas/actions/mintTokens.ts
@@ -61,18 +61,18 @@ function* createMintTokensAction({
       ready: false,
     });
 
-    yield fork(createTransaction, claimColonyFunds.id, {
-      context: ClientType.ColonyClient,
-      methodName: 'claimColonyFunds',
-      identifier: colonyAddress,
-      params: [nativeTokenAddress],
-      group: {
-        key: batchKey,
-        id: metaId,
-        index: 1,
-      },
-      ready: false,
-    });
+    // yield fork(createTransaction, claimColonyFunds.id, {
+    //   context: ClientType.ColonyClient,
+    //   methodName: 'claimColonyFunds',
+    //   identifier: colonyAddress,
+    //   params: [nativeTokenAddress],
+    //   group: {
+    //     key: batchKey,
+    //     id: metaId,
+    //     index: 1,
+    //   },
+    //   ready: false,
+    // });
 
     if (annotationMessage) {
       yield fork(createTransaction, annotateMintTokens.id, {
@@ -90,7 +90,7 @@ function* createMintTokensAction({
     }
 
     yield takeFrom(mintTokens.channel, ActionTypes.TRANSACTION_CREATED);
-    yield takeFrom(claimColonyFunds.channel, ActionTypes.TRANSACTION_CREATED);
+    // yield takeFrom(claimColonyFunds.channel, ActionTypes.TRANSACTION_CREATED);
 
     if (annotationMessage) {
       yield takeFrom(
@@ -107,9 +107,9 @@ function* createMintTokensAction({
       },
     } = yield waitForTxResult(mintTokens.channel);
 
-    yield initiateTransaction(claimColonyFunds.id);
+    // yield initiateTransaction(claimColonyFunds.id);
 
-    yield waitForTxResult(claimColonyFunds.channel);
+    // yield waitForTxResult(claimColonyFunds.channel);
 
     yield createActionMetadataInDB(txHash, { customTitle: customActionTitle });
 

```

* Step 6. Run `git apply .diff` - or however you decided to name the file with the above-mentioned content

> [!NOTE]
> These changes will disable the funds claiming when minting tokens, as we'll need to have some data on which to check the buttons on the `Incoming funds` page behave properly

* Step 7. Let's create a couple of `Mint tokens` actions
Screenshot 2024-12-23 at 12 23 36

* Step 8. Head to http://localhost:9091/planex/incoming
* Step 9. Play around with the `Accept` or `Accept all incoming funds` buttons and check the buttons are not re-enabled before the `Accepted` status is shown

https://github.com/user-attachments/assets/73889798-9f6a-46b0-9a35-4377acccb100

**Repeat step 3. and/or steps 7-9. as much as wanted. If you have any other testing ideas, do not stop these testing steps from checking them out**

> [!NOTE]
> Even after accepting all incoming funds, if you refresh the page, the `Accept all incoming funds` button will be enabled as there are `chainFundsClaim` that still remain unclaimed

## Diffs

**New stuff** ✨

* `useRefetchColonyData` and `useFundsStateUpdater` hooks
* `setPendingFundsTokenAddresses` and `reset` props for `IncomingFundsLoadingContext`

**Deletions** ⚰️

* `disableAcceptLoading` prop for `IncomingFundsLoadingContext`

Resolves #3502

### DIFF
--- a/src/components/frame/v5/pages/FundsPage/context/IncomingFundsLoadingContextProvider.tsx
+++ b/src/components/frame/v5/pages/FundsPage/context/IncomingFundsLoadingContextProvider.tsx
@@ -6,23 +6,43 @@ import React, {
   type PropsWithChildren,
 } from 'react';
 
+import { useRefetchColonyData, useFundsStateUpdater } from './hooks.ts';
 import { IncomingFundsLoadingContext } from './IncomingFundsLoadingContext.ts';
 
 export const IncomingFundsLoadingContextProvider: FC<PropsWithChildren> = ({
   children,
 }) => {
   const [isAcceptLoading, setIsAcceptLoading] = useState(false);
+  const [pendingFundsTokenAddresses, setPendingFundsTokenAddresses] = useState<
+    string[]
+  >([]);
 
   const enableAcceptLoading = useCallback(() => setIsAcceptLoading(true), []);
-  const disableAcceptLoading = useCallback(() => setIsAcceptLoading(false), []);
+
+  const reset = useCallback(() => {
+    setPendingFundsTokenAddresses([]);
+    setIsAcceptLoading(false);
+  }, []);
+
+  useFundsStateUpdater(pendingFundsTokenAddresses, reset);
+  useRefetchColonyData(
+    isAcceptLoading && !!pendingFundsTokenAddresses.length,
+    reset,
+  );
 
   const value = useMemo(
     () => ({
       isAcceptLoading,
       enableAcceptLoading,
-      disableAcceptLoading,
+      setPendingFundsTokenAddresses,
+      reset,
     }),
-    [isAcceptLoading, enableAcceptLoading, disableAcceptLoading],
+    [
+      isAcceptLoading,
+      enableAcceptLoading,
+      setPendingFundsTokenAddresses,
+      reset,
+    ],
   );
 
   return (
--- a/src/components/frame/v5/pages/FundsPage/context/hooks.ts
+++ b/src/components/frame/v5/pages/FundsPage/context/hooks.ts
@@ -1,0 +1,64 @@
+import { useEffect } from 'react';
+
+import { useColonyContext } from '~context/ColonyContext/ColonyContext.ts';
+import useColonyFundsClaims from '~hooks/useColonyFundsClaims.ts';
+import usePrevious from '~hooks/usePrevious.ts';
+
+export const useRefetchColonyData = (
+  shouldRefetchColonyData: boolean,
+  reset: () => void,
+) => {
+  const { startPollingColonyData, stopPollingColonyData } = useColonyContext();
+
+  useEffect(() => {
+    let timeoutId;
+    if (shouldRefetchColonyData) {
+      startPollingColonyData(1_000);
+      timeoutId = setTimeout(() => {
+        stopPollingColonyData();
+        /**
+         * Resets the context state if no incoming funds state update occurs
+         * after colony polling is completed.
+         */
+        reset();
+      }, 10_000);
+    }
+
+    return () => {
+      if (timeoutId) {
+        stopPollingColonyData();
+        reset();
+        clearTimeout(timeoutId);
+      }
+    };
+  }, [
+    shouldRefetchColonyData,
+    reset,
+    startPollingColonyData,
+    stopPollingColonyData,
+  ]);
+};
+
+export const useFundsStateUpdater = (
+  pendingFundsTokenAddresses: string[],
+  reset: () => void,
+) => {
+  const fundsClaims = useColonyFundsClaims();
+  const unclaimedFundsClaims = fundsClaims.filter(
+    (claim) =>
+      claim.amount !== '0' &&
+      !claim.isClaimed &&
+      claim.token?.tokenAddress &&
+      pendingFundsTokenAddresses.includes(claim.token?.tokenAddress),
+  );
+  const previousUnclaimedFundsClaims = usePrevious(unclaimedFundsClaims);
+
+  const hasFundsClaimsStateChanged =
+    !unclaimedFundsClaims?.length && !!previousUnclaimedFundsClaims?.length;
+
+  useEffect(() => {
+    if (hasFundsClaimsStateChanged) {
+      reset();
+    }
+  }, [hasFundsClaimsStateChanged, reset]);
+};
--- a/src/components/frame/v5/pages/FundsPage/context/IncomingFundsLoadingContext.ts
+++ b/src/components/frame/v5/pages/FundsPage/context/IncomingFundsLoadingContext.ts
@@ -4,15 +4,20 @@ import noop from '~utils/noop.ts';
 
 interface IncomingFundsLoadingContextValues {
   isAcceptLoading: boolean;
-  enableAcceptLoading: () => void;
-  disableAcceptLoading: () => void;
+  enableAcceptLoading: VoidFunction;
+  setPendingFundsTokenAddresses: (tokenAddresses: string[]) => void;
+  /**
+   * Resets the loading state (`isAcceptLoading`) and the list of token addresses awaiting fund claims to their initial values.
+   */
+  reset: VoidFunction;
 }
 
 export const IncomingFundsLoadingContext =
   createContext<IncomingFundsLoadingContextValues>({
     isAcceptLoading: false,
     enableAcceptLoading: noop,
-    disableAcceptLoading: noop,
+    setPendingFundsTokenAddresses: noop,
+    reset: noop,
   });
 
 export const useIncomingFundsLoadingContext = () => {
--- a/src/components/frame/v5/pages/FundsPage/partials/AcceptButton/AcceptButton.tsx
+++ b/src/components/frame/v5/pages/FundsPage/partials/AcceptButton/AcceptButton.tsx
@@ -15,16 +15,15 @@ const AcceptButton: FC<AcceptButtonProps> = ({
   disabled,
   ...rest
 }) => {
-  const {
-    colony,
-    canInteractWithColony,
-    startPollingColonyData,
-    stopPollingColonyData,
-  } = useColonyContext();
+  const { colony, canInteractWithColony } = useColonyContext();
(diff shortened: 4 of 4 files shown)


src/components/frame/v5/pages/FundsPage/context/IncomingFundsLoadingContextProvider.tsx
```diff
@@ -6,23 +6,44 @@ import React, {
   type PropsWithChildren,
 } from 'react';
 
+import useFundsStateUpdater, { useRefetchColonyData } from './hooks.ts';
```
davecreaser, line 9: Slightly confusing imports here, I would just export them both as default inside an object to avoid this default / non-default importing from a generic hooks file

src/components/frame/v5/pages/FundsPage/context/hooks.ts
```diff
@@ -0,0 +1,54 @@
...
+import usePrevious from '~hooks/usePrevious.ts';
+
+export const useRefetchColonyData = (
+  isAcceptLoading: boolean,
+  reset: () => void,
+) => {
+  const { startPollingColonyData, stopPollingColonyData } = useColonyContext();
+
+  useEffect(() => {
+    let timeoutId;
```
davecreaser, line 14: This is a nice way of storing timeouts actually, I always just use refs but this is really nice!

src/components/frame/v5/pages/FundsPage/context/IncomingFundsLoadingContextProvider.tsx
```diff
@@ -6,23 +6,44 @@ import React, {
...
+    [],
+  );
+
+  const reset = useCallback(() => {
+    setIsAcceptLoading(false);
+    setIsFundsUpdatePending(false);
+  }, []);
+
+  useFundsStateUpdater(isFundsUpdatePending, reset);
+  useRefetchColonyData(isAcceptLoading, reset);
```
davecreaser, line 30: Can these two hooks just get the reset function directly from the context, instead of passing it as a parameter?

davecreaser (dismissed):
Good stuff @mmioana! Looks to be working nicely. I just have a couple of minor comments, but I wouldn't consider them blocking an approval, so I have approved :)

My main concern (aside from the tiny comments on the code) is that it's not entirely clear to me that it is still loading when I click the accept button. There is a flicker of a loader, and then the button is disabled until it finishes. I didn't realise that this disabled state meant that it was still loading (because the loader threw me off) so I thought for a second that it was broken. Once the buttons became enabled again after a couple of seconds I realised that it was fine, but it was certainly confusing for a moment.

https://github.com/user-attachments/assets/e98bfdce-0999-414f-9d42-72a0eaad74c7

src/components/frame/v5/pages/FundsPage/context/IncomingFundsLoadingContextProvider.tsx
mmioana: Regularly I do named exports, but my blindness totally made me miss this 🫠 I will fix it asap!

src/components/frame/v5/pages/FundsPage/context/IncomingFundsLoadingContextProvider.tsx
mmioana: @davecreaser if we try to use context values - `reset` in this case - before the provider has been initialised, I don't think this change will actually work

mmioana:
> Good stuff @mmioana! Looks to be working nicely. I just have a couple of minor comments, but I wouldn't consider them blocking an approval, so I have approved :)
> 
> My main concern (aside from the tiny comments on the code) is that it's not entirely clear to me that it is still loading when I click the accept button. There is a flicker of a loader, and then the button is disabled until it finishes. I didn't realise that this disabled state meant that it was still loading (because the loader threw me off) so I thought for a second that it was broken. Once the buttons became enabled again after a couple of seconds I realised that it was fine, but it was certainly confusing for a moment.
> 
>  Screen.Recording.2024-12-27.at.11.43.23.mov

@davecreaser can agree the disabled `Accept` buttons are a bit confusing to whether something broke while claiming funds. 
Given the actual claiming is pretty fast, but the data updating takes a while, not sure how we could better improve this. Maybe @melyndav has an idea?

rdig:
What's the state here, does it need more input from @JoinColony/product @JoinColony/design or can it be reviewed ?

mmioana:
@rdig I'll clarify tomorrow morning if we do any enhancement in this PR or we leave it as it is and create a separate issue for UX enhancements 🙌

bassgeta (approved):
Yup, nice modification of the context and the changes are pretty sensible even if we didn't end up using subscriptions :ok_hand: 
Paid planex 
image
After applying the diff, I've minted tokens twice :eyes: 
image
And everything works as described, the buttons are blocked by both accepting an individual token's claims or all of them :heavy_check_mark: 

https://github.com/user-attachments/assets/3663fde2-9bd4-4a40-8222-8f0c601cc772

iamsamgibbs (changes requested):
This is working really well for the most part! I've unfortunately found one issue with it though.

The "Accept all available funds" button triggers separate transactions for each different token. It seems the "Accept" buttons on each row are only disabled whilst the first transaction is processing. This is only really obvious when using metamask to slow down the transactions.

https://github.com/user-attachments/assets/68a611f8-178a-4649-9c42-0bdaa4d4f393

When accepting just one token at a time this works completely fine!

I also found an unrelated bug, following your testing steps so that there are incoming funds for both CREDS and GGG. Claim just the GGG tokens by clicking the individual accept button. Then click the "Accept all funds" button, it will still trigger claim transactions for both CREDS and GGG, even though GGG has already been claimed. It seems like the unclaimed funds array that is passed to the "Accept all funds" component isn't refreshed properly. This can be handled as a separate issue and is a very rare edge case!

mmioana:
Thanks for review guys 🙏 I have pushed a fix for @iamsamgibbs finding
Can I get another round of reviews @iamsamgibbs @davecreaser @bassgeta?

rdig (approved):
After the latest round of changes, everything seems to be working as expected. Nice turnaround.

Screenshot from 2025-01-17 12-29-00
Screenshot from 2025-01-17 12-37-20
Screenshot from 2025-01-17 12-38-02
Screenshot from 2025-01-17 12-38-20
Screenshot from 2025-01-17 12-38-25
Screenshot from 2025-01-17 12-38-30

iamsamgibbs (approved):
Amazing work, this is working brilliantly now, thanks for the changes!

https://github.com/user-attachments/assets/f45ffeaa-d93a-4223-b6c4-ab5aee9b35f6

bassgeta (approved):
Still looking good :sunglasses: 
The spinenr stays until we sign the transaction :heavy_check_mark: 

https://github.com/user-attachments/assets/dd703c7c-9046-461d-b311-6cefae95cb16

In the video there's some weird behaviour where I could claim incoming funds but there were none to claim? but I don't think it's necessarily related to your PR so I'm approving it :ship:

mmioana:
> Still looking good 😎 The spinenr stays until we sign the transaction ✔️
> 
>  incoming-funds-redux-2025-01-20_11.25.36.mp4 
> In the video there's some weird behaviour where I could claim incoming funds but there were none to claim? but I don't think it's necessarily related to your PR so I'm approving it 🚢

Thanks for your review @bassgeta 🙌 

Yes, this happens as the `Accept available funds` button state takes into account if there are unclaimed `chainFundsClaim`. However, given the very-soon-redesign of the page, I will not tackle this in this PR 🥺

PR #4008, opened 2024-12-27
Refactor: Search select component and usage

mmioana:
Refactor: Search select component and usage

## Description

- [] TODO write description

## Testing

* Step 1. E.g. add reputation
* Step 3. E.g. Create motion

## Diffs

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

* Changes go here

**Deletions** ⚰️

* Deletions go here

## TODO

- [ ] Add TODOs

### DIFF
--- a/src/components/v5/shared/SearchSelect/hooks.ts
+++ b/src/components/v5/shared/SearchSelect/hooks.ts
@@ -1,12 +1,16 @@
 import { utils } from 'ethers';
-import { useMemo } from 'react';
+import debounce from 'lodash/debounce';
+import { useMemo, useState, useEffect, useCallback } from 'react';
 
 import { formatText } from '~utils/intl.ts';
 
-import { type SearchSelectOptionProps } from './types.ts';
+import {
+  isUserSearchSelectOption,
+  type SearchSelectOptionProps,
+} from './types.ts';
 
-export const useSearchSelect = (
-  items: SearchSelectOptionProps[],
+export const useFilterItems = <T>(
+  items: SearchSelectOptionProps<T>[],
   searchValue: string,
 ) => {
   const searchedOptions = useMemo(
@@ -32,7 +36,11 @@ export const useSearchSelect = (
             ? [
                 {
                   ...item.options[0],
-                  avatar: item.options[0]?.showAvatar ? '' : undefined,
+                  avatar:
+                    isUserSearchSelectOption(item.options[0]) &&
+                    item.options[0]?.showAvatar
+                      ? ''
+                      : undefined,
                   value: searchValue,
                   label: searchValue,
                   walletAddress: searchValue,
@@ -54,4 +62,63 @@ export const useSearchSelect = (
   );
 
   return filteredOptions;
+};
+
+export const useAccordion = <T>(items: SearchSelectOptionProps<T>[]) => {
+  const defaultOpenedAccordions = useMemo(
+    () => items.filter(({ isAccordion }) => isAccordion).map(({ key }) => key),
+    [items],
+  );
+
+  const [openedAccordions, setOpenedAccordions] = useState<string[]>(
+    defaultOpenedAccordions,
+  );
+
+  const onAccordionClick = useCallback((key: string) => {
+    setOpenedAccordions((prev) => {
+      if (prev.includes(key)) {
+        return prev.filter((item) => item !== key);
+      }
+
+      return [...prev, key];
+    });
+  }, []);
+
+  return { openedAccordions, onAccordionClick };
+};
+
+export const useSearch = (
+  initialSearchValue: string,
+  onSearch?: (value: string) => void,
+) => {
+  const [searchValue, setSearchValue] = useState(initialSearchValue);
+  const [debouncedSearchValue, setDebouncedSearchValue] = useState('');
+
+  useEffect(() => {
+    const debounceTimeout = setTimeout(() => {
+      setDebouncedSearchValue(searchValue);
+    }, 500);
+
+    return () => clearTimeout(debounceTimeout);
+  }, [searchValue]);
+
+  const handleSearch = useMemo(
+    () => debounce((value: string) => setDebouncedSearchValue(value), 500),
+    [],
+  );
+
+  const onChange = useCallback(
+    (value: string) => {
+      onSearch?.(value);
+      setSearchValue(value);
+      if (value) {
+        handleSearch(value);
+      } else {
+        setDebouncedSearchValue('');
+      }
+    },
+    [handleSearch, onSearch],
+  );
+
+  return { searchValue, debouncedSearchValue, onChange };
 };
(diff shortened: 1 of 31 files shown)


mmioana:
Putting this back in `Draft` as I found an issue with the fallback search value

bassgeta (approved):
WE FINALLY HAVE TYPES ON SEARCH SELECT :tada: 
Really great job untangling all of this, I've left one comment with a potential improvement, so we don't do search in 2 different ways, but other than that this is a huge step into the right direction! There's probably some more stuff we could solve with composition, but using this as a foundation will make it much easier!

Action sidebar search still works :heavy_check_mark: 
image

Manage tokens select :heavy_check_mark: 
image
image
image

Manage verified members :heavy_check_mark: 
image
image
image


Team select :heavy_check_mark: 
image
image

User select works too :heavy_check_mark: 
image
image
image
image
image

Other than that I think all things are covered!
I'll approve, but we can also hop on a call to solve this debounce implementation details, then we're good to go!

R.I.P.  a custom option for `isComingSoon` 2024-2025, you were not a good boy

src/components/v5/common/ActionSidebar/partials/TokenSelect/partials/TokenSearchSelect/TokenSearchSelect.tsx
```diff
@@ -45,135 +31,51 @@ const TokenSearchSelect = React.forwardRef<
...
-        setSearchValue(value);
-        if (value) {
-          handleSearch(value);
-        } else {
-          setDebouncedSearchValue('');
-        }
-      },
-      [handleSearch, onSearch],
-    );
+    const handleDebouncedSearch = (value: string) => {
```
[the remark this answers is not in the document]
mmioana: Really nice idea @bassgeta 🥇 I will look into it

mmioana:
@bassgeta I did remove the `onDebouncedSearch` from the `SearchSelect` and implemented your suggestion, however I am a bit unsure if it does make sense to keep debouncing logic on both `SearchSelect` and `TokenSearchSelect`

rdig (changes requested):
Opening the "Manage verified members" action for me throws this error _(on a newly restarted dev environment)_

Screenshot from 2025-01-14 00-15-56

I'm on this commit from the branch

Screenshot from 2025-01-14 00-16-45

Am I doing something wrong ?

mmioana:
@rdig nice finding! 
The problem was related to the used indexes when removing the `Action.ManageSupportedChains` from the available actions dropdown when the `SUPPORTED_CHAINS_ACTION` feature flag is disabled. 

I have fixed it on the `feat/manage-supported-chains-action` branch, so if you can please re-test this PR, would appreciate it 🙏

rdig (approved):
All good now, your latest fix did the trick!


https://github.com/user-attachments/assets/5567f104-f983-4138-aa05-6fe2b76eddbc

src/components/v5/shared/SearchSelect/hooks.ts
```diff
@@ -55,3 +67,59 @@ export const useSearchSelect = (
...
+      }
+
+      return [...prev, key];
+    });
+  }, []);
+
+  return { openedAccordions, onAccordionClick };
+};
+
+export const useSearch = (onSearch?: (value: string) => void) => {
```
rumzledz, line 94: I think that this hook is using 2 different methods of debouncing in one go:
- debounce from lodash
- custom debounce via your `useEffect` using `setTimeout`

Looking at how your hook is set up, I feel more yay-like with the `useEffect` approach since it eliminates reliance from lodash 👀 haha you can just remove the memoised `handleSearch` function and just handle everything in your `useEffect`:

```ts
export const useSearch = (onSearch?: (value: string) => void, debounceTime = 500) => {
  const [searchValue, setSearchValue] = useState('');
  const [debouncedSearchValue, setDebouncedSearchValue] = useState('');

  useEffect(() => {
    const debounceTimeout = setTimeout(() => {
      setDebouncedSearchValue(searchValue);
      onSearch?.(searchValue); // if this only needs to be called after the debounce
    }, debounceTime);

    return () => clearTimeout(debounceTimeout);
  }, [searchValue, onSearch, debounceTime]);

  const onChange = useCallback((value: string) => {
    onSearch?.(searchValue); // if this needs to be called as soon as the search value is updated
    setSearchValue(value);
    if (!value) {
      setDebouncedSearchValue('');
    }
  }, []);

  return { searchValue, debouncedSearchValue, onChange };
};
```

src/components/v5/shared/SearchSelect/hooks.ts
```diff
@@ -55,3 +67,59 @@ export const useSearchSelect = (
...
+  }, [searchValue]);
+
+  const handleSearch = useMemo(
+    () => debounce((value: string) => setDebouncedSearchValue(value), 500),
+    [],
+  );
+
+  const onChange = useCallback(
+    (value: string) => {
+      onSearch?.(value);
```
rumzledz, line 113: Just wanted to clarify if this need to run after the debounced call or as soon as the search value is updated

src/components/v5/shared/SearchSelect/hooks.ts
```diff
@@ -55,3 +67,59 @@ export const useSearchSelect = (
...
+export const useSearch = (onSearch?: (value: string) => void) => {
+  const [searchValue, setSearchValue] = useState('');
+  const [debouncedSearchValue, setDebouncedSearchValue] = useState('');
+
+  useEffect(() => {
+    const debounceTimeout = setTimeout(() => {
+      setDebouncedSearchValue(searchValue);
+    }, 500);
+
+    return () => clearTimeout(debounceTimeout);
```
rumzledz, line 103: If you want to stick with using `debounce` from lodash approach, it's worth invoking the `cancel()` method for your debounce function on cleanup so like:

```js
return () => {
  clearTimeout(debounceTimeout);
  handleSearch.cancel();
}
```

rumzledz (approved):
Love love love these changes @mmioana !! I like that the `SearchSelectOptionProps` is accepting a Generic 💯 I don't know. I guess have a thing for Generics...?? 👀

I did leave a few comments for the `useSearch` hook but considering that this PR is already massive, let's get these wonderful changes in and worry about that later!

search-action-type
search-chain

src/components/v5/shared/SearchSelect/hooks.ts
mmioana: From my investigation it needs to run immediately after a value is updated. I saw the only place being used is for the tokens select for showing up error

src/components/v5/shared/SearchSelect/hooks.ts
mmioana: I will save this for a later improvement 🙌

src/components/v5/shared/SearchSelect/hooks.ts
mmioana: Saving this for a later improvement 🙌

PR #4015, opened 2024-12-30
Fix extension settings loading states

### LINKED ISSUE #3577: Extension Enable button should be disabled whilst uninstall transaction processes
## Steps to reproduce

* Step 1 - Install the Multi-Sig extension
* Step 2 - Deprecate the Multi-Sig extension
* Step 3 - Click the "Uninstall button" and confirm the modal. **Do not accept or reject the transaction**

## Expected behaviour

* The "Enable" extension button should be disabled whilst this transaction is processed

## Actual behaviour

* The "Enable" extension button remains enabled whilst the uninstall transaction is processed meaning it is possible to uninstall the extension then click the "Enable" button whilst it finishes processing. (This isn't a huge deal as the transaction will fail as the uninstall transaction will have already gone through the contract, but would still be nice to have this enhancement.)

https://github.com/user-attachments/assets/7588876c-6bac-4b1d-a692-76fdc953e5b3

## Suggestions (optional)

Will probably require adding an extensions context to keep track of the extension state / processing transactions. There is probably quite a bit in the extensions area which could be refactored if a context was added and rework the way `waitForDbAfterExtensionAction` functions.

davecreaser:
## Description

- Ensure that while extensions are being managed (installed, settings changed, enabled, deprecated, and uninstalled) the other buttons to manage the extension are disabled. This prevents you from, for example, uninstalling an extension, and then while that loads, attempting to enable it as described here: https://github.com/JoinColony/colonyCDapp/issues/3577.

## Testing

* Go to any extension page
* Install the extension
* Change the extension settings (and save)
* Enable the extension
* Deprecate the extension
* Uninstall the extension
* At each stage, check that you cannot do another extension management action, and that any other buttons for extension management from the list above are disabled if they are shown. The video below may help:

https://github.com/user-attachments/assets/55749edc-0581-49a7-99ec-36973545c7fc

## Diffs

**Changes** 🏗

* Update the extension details page utils to make use of a loading state context, and have the buttons be disabled when it's active

Resolves #3577

### DIFF
--- a/src/components/frame/Extensions/pages/ExtensionDetailsPage/partials/ExtensionDetailsHeader/InstallButton.tsx
+++ b/src/components/frame/Extensions/pages/ExtensionDetailsPage/partials/ExtensionDetailsHeader/InstallButton.tsx
@@ -2,6 +2,7 @@ import { Extension } from '@colony/colony-js';
 import React from 'react';
 
 import { useColonyContext } from '~context/ColonyContext/ColonyContext.ts';
+import { useExtensionDetailsPageContext } from '~frame/Extensions/pages/ExtensionDetailsPage/context/ExtensionDetailsPageContext.ts';
 import { useMobile } from '~hooks/index.ts';
 import { ActionTypes } from '~redux/index.ts';
 import { type AnyExtensionData } from '~types/extensions.ts';
@@ -22,6 +23,7 @@ const InstallButton = ({ extensionData }: InstallButtonProps) => {
     colony: { colonyAddress, nativeToken },
     isSupportedColonyVersion,
   } = useColonyContext();
+  const { waitingForActionConfirmation } = useExtensionDetailsPageContext();
 
   const isMobile = useMobile();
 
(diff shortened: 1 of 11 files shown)


bassgeta (approved):
Nice one, I have a feeling this context is gonna see more use, which is completely legit since we have quite some edge cases on this page.
So nice job sorting this out :sunglasses: I've initially thought it may be a perfect time to whip out [transitions](https://react.dev/reference/react/useTransition#perform-non-blocking-updates-with-actions) just so we start using them, but I don't think it's applicable here at all so ignore my rambles :D 

I've tried clicking and breaking everything, it didn't even budge, nice one :muscle: 

https://github.com/user-attachments/assets/5893900d-d0cd-42dd-ba5d-85e8c4d38885

mmioana (approved):
Really nice work on this one @davecreaser 💯 

Once an action is performed on the extension, the other actions are disabled 🙅 

https://github.com/user-attachments/assets/a2e3f781-d301-4920-af83-0108f9f9a749

Nortsova (approved):
Great work on this 👍 

I like this approach, I am pretty sure we can use this hook flag all over the app. 🙌 

extension-disabled

PR #4026, opened 2025-01-03
Fix: Make date filters single choice

### LINKED ISSUE #3697: [Activity -QA] Date filter not displaying activities for all selected dates when custom and preset dates are selected simultaneously
## Steps to reproduce

1. Navigate to the Activity Feed.
2. Open the date filter and select a preset date option (e.g., "Past 24 hours").
3. Additionally, set a custom date range that spans a few days.
4. Observe the activities displayed within the selected date options.

## Expected behaviour

* The Activity Feed should display all activities that fall within both the preset and custom date ranges, providing a complete view of activities matching the selected filters.

## Actual behaviour

* When both custom and preset date ranges are selected, not all activities within those dates are displayed, leading to missing entries and inconsistent filter behavior.

image

davecreaser:
## Description

- Ensure that you cannot select more than one date option when filtering the actions / agreements table. The reasoning is that selecting `Past hour` and `Past week` for example at the same time is actually just equal to selecting only `Past week`. Custom date ranges are slightly different to this, but after some discussion with product, it is not worth the work of making them multi selectable alongside presets (currently, if you have a custom date range selected it ignores any preset you also select).

## Testing

* Go to the activity page and filter by the date. You should only be able to select one at a time.
* Go to the agreements page and filter by the date. You should only be able to select one at a time.

https://github.com/user-attachments/assets/ecb2c434-7e25-451a-a6f6-4868d621e79f

## Diffs

**Changes** 🏗

* Ensure that you cannot select more than one date option when filtering the actions / agreements table.

Resolves #3697

### DIFF
--- a/src/components/common/ColonyActionsTable/FiltersContext/FiltersContextProvider.tsx
+++ b/src/components/common/ColonyActionsTable/FiltersContext/FiltersContextProvider.tsx
@@ -22,6 +22,15 @@ import { getDateFilter } from '../utils.ts';
 import { FiltersContext } from './FiltersContext.ts';
 import { FiltersValues } from './types.ts';
 
+const emptyDateFilters: DateOptions = {
+  pastHour: false,
+  pastDay: false,
+  pastWeek: false,
+  pastMonth: false,
+  pastYear: false,
+  custom: undefined,
+};
+
 const FiltersContextProvider: FC<PropsWithChildren> = ({ children }) => {
   const [searchFilter, setSearchFilter] = useState('');
   const [motionStates, setMotionStates] = useState<MotionState[]>([]);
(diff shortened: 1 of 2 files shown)


bassgeta (approved):
I love this solution, no beating around the bush, we just always override the base object :100: 

https://github.com/user-attachments/assets/80ac5b04-c90f-40c0-9754-6a291f84125c

iamsamgibbs (approved):
Nice solution, this works great. I can no longer select multiple date types.

Screenshot 2025-01-06 at 13 38 26

Compared to master:

Screenshot 2025-01-06 at 13 38 13

Nortsova (approved):
Great solution on this 🙌 

activity-date-filter

All filter functionality is working great 👏 

However it seems that Activity table filters by blockchain time, but Agreements page filter by real time. I am not sure if it is a bug or expected behaviour.

PR #4030, opened 2025-01-06
Fix: Apply Motion Widget Updates to the Actions Table

### LINKED ISSUE #3691: [Dashboard - QA] Payment statuses aren't changed in the Recent activity table  of Dashboard during performing any action with reputation
## Steps to reproduce

1. Go to the Dashboard.
2. Initiate simple payment with reputation(or any action with reputation).
3. Pass all action transitions from the Staking phase to the Finalize phase.
4. Observe the "Recent Activity" table to check if the status updates from "Staked" to "Finalizing."

## Expected behaviour

* All statuses should be updated during payment in the "Recent Activity" table.
image

## Actual behaviour

* The status remains "Staked," even though the action has entered the voting phase.

image
image

rumzledz:
## Description

Subscriptions to motion updates are now being used to refresh the motion states for the actions table.

The tricky part was getting the app to refetch motion states for the Actions table for UI motion flows that don’t trigger motion update subscriptions. So I ended up creating a global context to handle this edge case.

It's probably not the best solution in the world but it works 😅 Perhaps we can also find some weird edge case in the future that can utilise it?? If you have a better solution in mind, please do let me know as I'm very open to discussing it with you 🙌 

up-to-date-feed

## Testing

> [!NOTE]
> Not sure if it's a bug but I found that non-staked simple payment motions never show up on the actions table.

> [!IMPORTANT]
> 1. Install the Reputation extension
> 2. On the Reputation extension's config page, select "Custom (Advanced)
> 3. Set the following fields to 0.005 (this will make your life easier)
> - Staking Phase Duration
> - Voting Phase Duration
> - Reveal Phase Duration

### Scenario 1: Letting the timer naturally finish

> [!NOTE]
> The problem with this flow is that, you really need the action sidebar to be mounted when the countdown finishes because the actions table uses the same logic as the payment builder widget when refetching the motion states.
 
**Opposing a motion and naturally letting the motion countdown timer reach 0**

1. Visit the Dashboard or the Activity page
2. Create a Simple Payment using the Reputation decision method
3. Oppose and fully stake it
4. Wait for the countdown timer to finish
5. Wait for the motion to progress to the next step
6. Wait and verify that the motion has been marked as Failed
7. Close the action sidebar
8. Verify that the action is visible on the Actions table and that it has a failed state

**Approving a motion and naturally letting the motion countdown timer reach 0**

1. Visit the Dashboard or the Activity page
2. Create a Simple Payment using the Reputation decision method
3. Approve and fully stake it
4. Wait for the countdown timer to finish
5. Wait for the motion to progress to the next step
6. Finalise the motion
7. Wait and verify that the motion is marked as Passed
8. Close the action sidebar
9. The refetch request might take a while but please wait and verify that the action is visible on the Actions table and that it has a Passed state

### Scenario 2: Voting flow

1. Visit the Dashboard or the Activity page
2. Create a Simple Payment using the Reputation decision method
3. Approve and fully stake it
4. Oppose and fully stake it
5. During the voting phase, Approve it
6. Finalise the motion
7. Wait and verify that the motion has a Passed state
8. Close the action sidebar
9. Verify that the action is visible on the Actions table and that it has a Passed state

Resolves #3691


rdig (commented):
Tested this and it seems to be working, as all cases look like they're working correctly

Screenshot from 2025-01-09 18-33-12

However, something either on `master` or on my machine is broken making it impossible to view any payments in the activity list

Screenshot from 2025-01-09 19-14-11

I will re-review this once that is fixed, and this PR rebased

rdig (approved):
Never mind, apparently it was an issue on my end.

Tested all 3 scenarios and all 3 work as expected

https://github.com/user-attachments/assets/ec50e545-beef-430d-b0b8-f22fc1d1db62


https://github.com/user-attachments/assets/7d16f803-e1aa-445b-8f05-30865ab40426


https://github.com/user-attachments/assets/eac6e306-15de-4669-a874-c5e000b8c42c

iamsamgibbs:
Can you give this a rebase @rumzledz and I'll give it a review! :)

davecreaser (dismissed):
Good job @rumzledz, a weird issue to fix for sure. Sorry for the late review, but I've had a look and i'll approve! Testing showed that it worked as expected (videos at the bottom).

Regarding the solution and the code, my instinct was that since you've now introduced a subscription, we shouldn't need to do _any_ refetching, since that's essentially what the subscription will do for us. We can directly access the changed data inside the result of the subscription, rather than simply subscribing to changes and refetching when they occur. Having said that, I looked into where the refetching actually happens (inside src/components/v5/common/ActionSidebar/hooks/useGetColonyAction.ts if i'm not mistaken) and it looks fairly complex, since it looks like there's also motion data coming from the network. For that reason, I'm actually happy to accept this as a solution and say 'it's not ideal, but let's be pragmatic, it works and the PR is already quite old'.

Final point: if you're willing to spend a little more time on this, I'm happy to put aside whatever i'm working on and jump onto this with you and see if we can get a 'proper' subscription solution working, without the polling and refetching. We can timebox it to an afternoon or whatever.

Either way, it works and i'm happy to approve, so if you'd rather move on with life and forget about this  then feel free to merge! 😆 

https://github.com/user-attachments/assets/896f7df2-1c9f-4ab4-ada5-708136348593

https://github.com/user-attachments/assets/dd242179-de36-4e4d-bbe4-4de512ced5ca

https://github.com/user-attachments/assets/65868565-b617-4db5-a6dd-7255db6c31e3

rumzledz:
> Regarding the solution and the code, my instinct was that since you've now introduced a subscription, we shouldn't need to do any refetching, since that's essentially what the subscription will do for us. We can directly access the changed data inside the result of the subscription, rather than simply subscribing to changes and refetching when they occur. Having said that, I looked into where the refetching actually happens (inside src/components/v5/common/ActionSidebar/hooks/useGetColonyAction.ts if i'm not mistaken) and it looks fairly complex, since it looks like there's also motion data coming from the network. For that reason, I'm actually happy to accept this as a solution and say 'it's not ideal, but let's be pragmatic, it works and the PR is already quite old'.

Hey @davecreaser! The rest of the stuff doesn't really need the manual trigger since they already rely on the subscription, which you have pointed out. The problem I had was with how the app behaves after the motion countdown finishes. So essentially, when a motion naturally expires, we won't actually know that globally because we don't have a subscription for that (as far as I'm aware and I may very well be wrong!). As in, it doesn't trigger the `useOnUpdateColonyMotionSubscription`. That's the ONLY reason why I added that global context thing. But sure! Happy to brainstorm other potential solutions with you if you have the time! 🙌

mmioana (approved):
Nice work @rumzledz 🙌 

Tested all 3 flows and it works as expected 🥇 

https://github.com/user-attachments/assets/a1cbff53-0dfb-426e-b088-57542b5aa195


https://github.com/user-attachments/assets/bf7cbe0a-a7e7-4b6a-a92a-eb2a35412ef8


https://github.com/user-attachments/assets/dad5fdb6-4e67-4b5e-92de-3d1820fb29f6

I do agree with @davecreaser's point regarding the usage of subscriptions and data-polling, however maybe this can be a follow-up task, given this solution works as expected 💯 

Niiiice work!

davecreaser:
> The problem I had was with how the app behaves after the motion countdown finishes.

Yeah this makes total sense and really complicates things further. I'd definitely say let's merge this then, it could indeed be that it is the only solution for now!

davecreaser (dismissed):
As discussed on a call, this approach does indeed make total sense to me now 👍 

Re-approved!

bassgeta (approved):
This context is going to be extremely handy, I can already see that :D  
Nice job solving this in a slick way, subscriptions are the future :sunglasses: 
Scenario 1 :heavy_check_mark: 
image
image
Scenario 2 :heavy_check_mark: 
image
image
Scenario with failed voting :heavy_check_mark: 
image
image
Is it OK if it's finalizable? I guess so since we need to return stakes
After finalization it's failed :heavy_check_mark: 
image

mmioana (approved):
Still good from my side, really nice work on this issue @rumzledz 🥇 

1. Opposing the stake

https://github.com/user-attachments/assets/f9549674-b1cb-4826-a489-217e6a487498


2. Supporting the stake

https://github.com/user-attachments/assets/17f478f9-b0e5-4f84-b9fe-6b6e69ae85e1


3. Going through voting 

https://github.com/user-attachments/assets/369b7108-0f3f-4ce5-ae32-0028790d2980

davecreaser (approved):
Still all good from my side @rumzledz! Reapproving 📈 Totally agree with the scope change and name change, makes a lot of sense.

rdig (approved):
Re-approving so you can merge

PR #4041, opened 2025-01-07
build: remove usage of react-spring

### LINKED ISSUE #3645: [Improvement] Remove react-spring usage
### Description

Due to using [nivo.rocks](https://nivo.rocks/) for our charts, we get also `react-spring` as a dependency.

Screenshot 2024-11-05 at 21 10 10

However, we already make use of `framer-motion`, so it doesn't make sense to keep both packages.
We should try to disable the animations for our charts and wherever needed make use of `framer-motion`.

- [ ] `react-spring` is no longer in our bundle

### Suggestions 

* There is no explicit documentation on how or if it is possible to do this. However, you can try using `animated=false` for the `BarChart` and `Chart` components. 
* Then replace the `react-spring` usage from `ChartCustomXAxisStep`, `ChartCustomBarLabel` and `ChartCustomYAxisStep` with `framer-motion`
* Then add in `vite.config.mts` 
```
defineConfig(({ mode }) => ({
  ...,
  optimizeDeps: {
    exclude ['react-spring']
  }
}))
```
* You can use `npx vite-bundle-visualizer` for checking the bundle

Of course, these suggestions might not work as expected, so please bring patience ✨

bassgeta:
## Description

This PR removes automatically pulling in `react-spring` as a dependency.

## Testing

No other way about it than running 2 CDapp instances, `master` and this one and verifying that the bar chart labels work correctly :shrug: 

Run `npx vite-bundle-analyzer` Ctrl + F to try and find `react-spring` - it shouldn't exist anymore :sunglasses: 

Additionally

## Diffs

**Changes** 🏗

* `ChartCustomBarLabel` no longer uses spring for animating itself under the hood, but instead it receives very specific framer motion props

Resolves #3645

### DIFF
--- a/src/components/v5/frame/ColonyHome/partials/TotalInOutBalance/partials/ChartCustomBarLabel.tsx
+++ b/src/components/v5/frame/ColonyHome/partials/TotalInOutBalance/partials/ChartCustomBarLabel.tsx
@@ -1,16 +1,31 @@
 import { type AxisTickProps } from '@nivo/axes';
-import { animated } from '@react-spring/web';
+import { motion } from 'framer-motion';
 import React, { type FC } from 'react';
 
 import { CHART_CONFIG_VALUES } from '../consts.ts';
 
 interface ChartCustomBarLabelProps
-  extends Pick<
-    AxisTickProps<string>,
-    'value' | 'textAnchor' | 'textBaseline' | 'animatedProps'
-  > {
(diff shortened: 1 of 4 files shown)


bassgeta:
Closed it because `react-spring` is an  actual dependency of nivo](https://www.npmjs.com/package/@nivo/bar?activeTab=dependencies)

PR #4062, opened 2025-01-09
Fix: Search activity feed by formatted action type

### LINKED ISSUE #3701: [Activity - QA] Search function not working for action type or token in Activity feed
## Steps to reproduce

1. Navigate to the Activity Feed.
2. Open the search filter and enter a value that matches an "Action type" (e.g., "Unlock Tokens") or a "Token" name.
3. Observe whether the search returns the correct results for the entered field.

## Expected behaviour

* The search should accurately filter results based on the "Action type" and "Token" fields, displaying all matching activities.

## Actual behaviour

* Searching by "Action type" or "Token" does not yield the correct results, making it difficult for users to locate specific activities.

https://www.notion.so/colony/Activity-Feed-c4c47253f8e8414fba93d9bd1c09f879

image
image

iamsamgibbs:
## Description

Searching the action feed was matching against the ColonyActionType (i.e. `UNLOCK_TOKEN`) rather than how the action type appears in the UI. (`Unlock token`) This PR adjusts the search to use the correctly formatted action type.

## Testing

You can test this with any action type, but for the sake of testing lets use the `Unlock token` action.

Create an `Unlock token` action, do not use the words "Unlock token" in the custom title.

Navigate to the action feed and type into the search "unlock token", the action should appear.

Screenshot 2025-01-09 at 17 17 04

Compared to master:

Screenshot 2025-01-09 at 17 17 12

## Diffs

**Changes** 🏗

* Activity feed search now uses the formatted action type

Resolves #3701

### DIFF
--- a/src/hooks/useActivityFeed/helpers.ts
+++ b/src/hooks/useActivityFeed/helpers.ts
@@ -6,7 +6,10 @@ import { type AnyActionType } from '~types/actions.ts';
 import { type InstalledExtensionData } from '~types/extensions.ts';
 import { type ColonyAction } from '~types/graphql.ts';
 import { notMaybe } from '~utils/arrays/index.ts';
-import { getExtendedActionType } from '~utils/colonyActions.ts';
+import {
+  formatActionType,
+  getExtendedActionType,
+} from '~utils/colonyActions.ts';
 import { getMotionState, MotionState } from '~utils/colonyMotions.ts';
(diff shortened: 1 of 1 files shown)


bassgeta (approved):
What a nice and simple fix :ok_hand: 
All of the filters work for me :heavy_check_mark: 
image
image
image
image

rumzledz (approved):
Clean update @iamsamgibbs ! ✨ 

search-by-action-type

PR #4081, opened 2025-01-13
Fix: Set the Created In field for Fund Transfers to the Parent domain

### LINKED ISSUE #3919: [Move funds] "Created In" field is disabled when moving funds from a subdomain team
## Steps to reproduce

1. Navigate to the Teams section on Dashboard.
2. Select a subdomain team (e.g., Brainstorm Battalion).
3. Initiate a Transfer Funds action to another team (e.g., General).
4. Observe that the "Created In" field is disabled and cannot be modified.

## Expected behaviour

* The "Created In" field should be editable.

## Actual behaviour

* The "Created In" field is disabled and does not allow user interaction.

image

rumzledz:
## Description

From what I understand, at least for the time being anyway since we don't have nested teams yet, the Permission Domain ID (created in field value) should be set to the Parent Domain of the From and To domains. This is currently reflected by how we're filtering the domains list for the Created in field to only contain the Root domain. Either that or the Permission Domain ID should be equal to both of them, but this isn't an option since we can't do fund transfers from and to the same domain. So the Created In field should really be the Root domain. We'll cross the bridge of dealing with nested domains when we get there 🌉 

transfer-funds

## Testing

1. Install and enable the Reputation extension
2. Go to the Dashboard page
3. Using the teams filter, click a subdomain i.e. Daedalus
4. Bring up the Transfer funds form
5. Set the decision method to Reputation
6. Verify that the Created in field is set to General
8. Submit the form
9. Verify that you are able to submit the form
10. Verify that you can finalise the motion

Resolves #3919

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/forms/TransferFundsForm/hooks.ts
+++ b/src/components/v5/common/ActionSidebar/partials/forms/TransferFundsForm/hooks.ts
@@ -95,9 +95,8 @@ export const useTransferFunds = (
   getFormOptions: ActionFormBaseProps['getFormOptions'],
 ) => {
   const { colony } = useColonyContext();
-  const { [DECISION_METHOD_FIELD_NAME]: decisionMethod, from } = useWatch<{
+  const { [DECISION_METHOD_FIELD_NAME]: decisionMethod } = useWatch<{
     decisionMethod: DecisionMethod;
-    from: TransferFundsFormValues['from'];
   }>();
 
   const validationSchema = useValidationSchema();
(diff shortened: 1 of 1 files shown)


rdig (approved):
Yup, ran through the whole process. Everything works as you described. Nicely done!

Screenshot from 2025-01-13 23-03-53
Screenshot from 2025-01-13 22-44-15
Screenshot from 2025-01-13 22-44-32
Screenshot from 2025-01-13 22-44-35
Screenshot from 2025-01-13 22-44-51
Screenshot from 2025-01-13 23-01-13
Screenshot from 2025-01-13 23-01-19

mmioana (approved):
The created in field is set to `General`, nicely done @rumzledz 🥇 

Screenshot 2025-01-14 at 11 10 54
Screenshot 2025-01-14 at 11 11 03

https://github.com/user-attachments/assets/6b841979-04a6-4fae-a215-f001fb061b85

The only thing I noticed is that the `Activity` table didn't update by default, but it is not related to this issue

iamsamgibbs (approved):
Nice job fixing this up, I won't share the same bundle of screenshots, but all worked without a hitch! Change makes sense to me. +1

PR #4100, opened 2025-01-14
Use correct message ID for multisig extension description

arrenv:
## Description

Just a simple fix to correct an incorrect translation definition I stumbled across while reviewing something else.

### Testing

- Ensure that the multi-sig extension page still has the correct content.


jakubcolony (approved):
Good spot! 

image

mmioana (approved):
The correct extension is used for the description id

Screenshot 2025-01-15 at 12 49 47

Compared to master
Screenshot 2025-01-15 at 12 49 24

Very nice @arrenv 🙌

PR #4101, opened 2025-01-15
Fix: App crashing if preferredCurrency not in localStorage

jakubcolony:
## Description

Fixes an issue where the app would crash if `preferredCurrency` wasn't in local storage, as JSON.parse was receiving an empty string instead of valid JSON.

## Testing

Delete `preferredCurrency` entry from your local storage and refresh the app. Confirm it no longer crashes.


rdig (approved):
Thank you for the quick turn around

mmioana (approved):
Thanks for your very fast fix @jakubcolony 🙌 

The app no longer crashes if the `preferredCurrency` is not present in the local storage

https://github.com/user-attachments/assets/0df348ca-40f1-4b6c-a6af-76d8cb9041dd

davecreaser (approved):
Nice catch, all good here app is no longer crashing:

https://github.com/user-attachments/assets/a547c601-909a-471b-b0dc-1772560dc5f0

PR #4122, opened 2025-01-15
Fix: Advanced payments not processing all slots

### LINKED ISSUE #3744: Advanced Payment rows do not load correctly in the completed action screen
## Steps to reproduce

* Step 1 - Create an advanced payment with a large number of rows (170+). Increment the amount in each row starting from 1 so we can use this to track the order of the rows.
* Step 2 - Submit the form to proceed to the review stage

## Expected behaviour

* The rows should load in, in the same order as they were created in the table.

## Actual behaviour

* The rows will not load in the same order as they were created in, and sometimes rows will not load in at all.

image

image

## Suggestions (optional)

This could be a problem with the way the payload is constructed, or a problem with how the transactions are picked up on block-ingestor, or an issue with the multicall setup in the saga. It will require some investigation to figure out what is going on here.

iamsamgibbs:
## Description

This issue was a bit of a pain to pin down, but luckily has a very simple solution!

This issue only occurred when more than 164 payment rows are added. This is because we chunk the payouts into multiple transactions as otherwise they become too big to process. There was a bug where the slotId was being overwritten for each chunk, so every chunk would start back at slotId 1. Effectively meaning the later chunks were editing the values for the previous chunks.

This PR resolves the issue by using the existing slotId value if it already exists.

## Testing

Create an advanced payment with more than 164 rows.

[Feel free to use this CSV](https://drive.google.com/file/d/1kNvrcIYzkfUDda7lbpV8GB11NYusjMvw/view?usp=sharing)

Check the completed action screen loads all the payments in the expected order (may take a short while to process them all).

Screenshot 2025-01-15 at 17 59 52

## Diffs

**Changes** 🏗

* `getPayoutsWithSlotIds` uses existing slotId if not undefined

Resolves #3744

### DIFF
--- a/src/redux/sagas/utils/expenditures.ts
+++ b/src/redux/sagas/utils/expenditures.ts
@@ -112,7 +112,7 @@ export const getPayoutsWithSlotIds = (
 ) => {
   return payouts.map((payout, index) => ({
     ...payout,
-    slotId: index + 1,
+    slotId: payout.slotId ?? index + 1,
   }));
 };
 


mmioana (approved):
Really nice and simple fix @iamsamgibbs 🥇 thanks for your effort in investigating and fixing this issue 💯 

On this branch all slots are processed

https://github.com/user-attachments/assets/f1baac99-c3b1-4f05-b5ef-66c1333fb715

While on `master` it is broken ❌ 

https://github.com/user-attachments/assets/fabf991a-fdae-4701-8ac9-37abccdfde62

davecreaser (approved):
Nice investigation, and really nice that it was such a simple fix 😅 

Thanks for the handy CSV too, really helped 👍 

They all load in order as expected now:
https://github.com/user-attachments/assets/f2193009-9d7a-49c1-884e-f5163210181e

Nortsova (approved):
Great work 👏  I am happy that the fix is so small 😄 

I tested it on CSV files from previous PRs containing 400 rows, and everything works great. It takes some time to load, but it works!

The only suggestion I have is to add an info box or tooltip to inform the user to wait, as the loading time is longer than what users typically expect. 😄 

image

image

image

PR #4123, opened 2025-01-15
Fix CDapp Wide Transaction Gas Estimation

### LINKED ISSUE #4106: [Manage account - QA] Disabled to create any motion with turned-off metatransaction
## Steps to reproduce

1. User has tokens in the wallet to pay transaction fees.
2. Open Manage account -> Advanced.
3. Disable "Colony pays my gas fees" radio button.
4. Create any motion.

image

https://github.com/user-attachments/assets/fc49420f-c757-4b8e-b49e-1641993ba014

## Expected behaviour

* Users should be able to create motions regardless of whether metatransactions are enabled or disabled.

## Actual behaviour

* The system prevents motion creation, and the following error is displayed: "Send error: Internal JSON-RPC error.

rdig:
This problem was discovered today, which was marked by unusual high transactions volume on _Arbitrum Sepolia_, which made the Gas cost increase, and since the CDapp was operating with the default network value _(lowest of the low)_, once this got higher, most all transactions attempted got dropped.

So how did this actually work up until now ?

The changes that introduced this can be traced 8 months back during the refactor of the Transactions save system, which got rid of the transactions being saved locally in Redux, and moved over to the DB. 

During that change, some subtle logic was lost, making things like `BigNumber`s stringyfied, which meant they couldn't be recovered and read by using just a simple `JSON.parse()`.

But apparently, for the last 8 months network conditions were favorable which mean that the default gas price value we were using was sufficient, so all transactions would go through. That, plus heavy reliance on Metatransactions.

The changes here a quite a few, and varied, but the actual fix comes down to two main changes:

### Fix 1 

**Properly convert nested Transaction `options` values stored in the database**

As mentioned above, `BigNumber` data that was strigyfield was lost, and could not be re-read by parsing the object, so the fix involved parsing the nested object, and if an instance was found, use the constructor `BigNumber.from()` to re-instantiate it properly, saving it back into the object

### Fix 2

**Wait for the Transaction item in the database to be updated**

Another, and more egregious issue we found was that we had a race condition between the Transaction Gas and Options updates and the time the transaction data was re-read back from the DB and sent to the chain.

The updates would start, but would finish after the transaction was already sent to the chain, meaning the final transaction object that made up that transaction would not include said updates, meaning no `gasLimit`, `maxFeePerGas` or `maxPriorityFeePerGas` would be included in the transaction overrides. 

To fix this, I've introduce a new Redux action, `TRANSACTION_UPDATED_IN_DB`, which gets called at the end of all transaction DB update operations, and which allows listening to when said operations finish, into which we can hook, via a saga, and only proceed with the transaction once we are confident that the data has finished updating.

Practically, all of the above fixes and changes take the form of the transaction object always having the `overrides` object populated, which can be observed in the debug data we log to the console every time we send out a transaction _(more on that below)_

Before:
Screenshot from 2025-01-15 21-35-28

After:
Screenshot from 2025-01-15 21-37-06

## Extra Additions

While I touched most of the transaction saga system, I took the opportunity to make some changes to the way log out debug data to the console.

I've introduced the `debugLogging` util, which is now used in place of `console.info` when logging out debug data to the console.

It's main difference from before is that, in production environments, will be silent by default, unless you manually inject the `__COLONY_DEBUG_ENABLED__` local storage key, with the value of `true`.

```js
localStorage.setItem('__COLONY_DEBUG_ENABLED__', 'true');
```

Unless the above is manually set, no output is presented the user in the browser's console.

I choose this approach since I also added a log for the initial gas values that the app gathers at startup, which will also expose the user's wallet object.

While in and of itself is not dangerous, it looks rather "scary" to layman users, and this approach will quell any panicked messages to support

Screenshot from 2025-01-15 21-36-20

## Warning

As @area so astutely pointed out, this system is deeply flawed. Since we are saving gas values in the database **before** sending the transaction, a user might manually change them via Metamask, before actually sending the transaction, meaning the values of the Database and the actual chain will now differ.

While for now this is a problem with low impact, sooner or later we might want to use that transaction information for some gas cost statistics only to find out the values are flawed.

## Testing

Testing this can already be observed in the above screenshots.

1. Start your dev environment
2. Create testing data
3. Start the frontend
4. One of the first new things you'll see, is an log output that tells you the gas values that the system gathered: `GANACHE GAS DEBUG {...}`
5. Disable your Metatransactions, via your User Profile settings
6. Create a transaction, doesn't really matter which _(I suggest you use Mint, since it's faster)_
7. Notice the `TX DEBUG` debug output in the console
8. Ensure it's `overrides` object is populated, and has values for `gasLimit`, `maxFeePerGas` or `maxPriorityFeePerGas`
9. Optionally, do the same, but with a motion, since they are more gas "heavy"

Note that this can also be tested on QA since it currently deployed https://github.com/JoinColony/colonyCDapp/actions/runs/12794985411

Going even further, you can compare the above QA output with the Production deployment which still runs the broken version 

This can be checked by looking at what actual values Metamask sends along with the transaction, by investigating the Gas values:

| QA  | Prod |
| ------------- | ------------- |
| Screenshot from 2025-01-15 22-18-59  | Screenshot from 2025-01-15 22-20-47  |

Note some differences:
- Only QA suggest a `site` value, basically the value we estimate, and send along with the transactions overrides _(basically the whole point of this PR)_
- On Prod, since Metamask has a way of fetching network gas values, even if we don't provide them, it will try to suggest `low`/`market`/`aggressive` values, however they way too low
- On Prod, the build that we're currently running does not suggest any `site` values

## Thanks

Special thanks to @jakubcolony and @area for lending their knowledge, expertise and patience during our long pair debugging session.

---

Resolves #4106  _(incidentally, even though, the issue is not faithful to the underlying problem)_


jakubcolony (changes requested):
That was an interesting one, nice job fixing it up!

The overrides are set correctly now both locally and in QA:
image

Which is not the case in production:
image

However, when sending a transaction out I'm getting quite a few errors in the console: 
image

They all point to where the transaction gets created in the DB, so maybe the optimistic response fragment needs to be updated to include `gasPrice` and `gasLimit`?

rdig:
> They all point to where the transaction gets created in the DB, so maybe the optimistic response fragment needs to be updated to include gasPrice and gasLimit?

Very nice! I missed that

src/redux/sagas/utils/getGasPrices.ts
```diff
@@ -176,16 +183,44 @@ const fetchGasPrices = async (): Promise<GasPricesProps> => {
         suggested: cheaper.mul(150).div(100), // 50% more than network
         faster: cheaper.mul(200).div(100), // 100% more than network
 
-        maxFeePerGas,
-        maxPriorityFeePerGas,
+        maxFeePerGas: maxFeePerGas.add(maxFeePerGas.mul(5).div(100)), // 5% more
```
mmioana, line 186: Would it make sense to extract the 5% into a constant given we use it in multiple places and then just multiply with that value?

mmioana (commented):
Really nice solution for a very pesky problem @rdig 🙌 

Everything works as expected, beside the problem @jakubcolony already pointed out

Got the log about the local network gas estimate
Screenshot 2025-01-16 at 10 15 32

Disabled the meta transactions
Screenshot 2025-01-16 at 10 15 46

Created a mint tokens action
Screenshot 2025-01-16 at 10 16 51
Screenshot 2025-01-16 at 10 17 02

Created a mint tokens motion 
Screenshot 2025-01-16 at 10 23 34

One of the log errors showing up
Screenshot 2025-01-16 at 10 17 36

rdig:
@jakubcolony and @JoinColony/dapp-team the Apollo error has now been fixed, so it's ready for a new round of reviews

src/redux/sagas/utils/getGasPrices.ts
rdig: Kinda awkward for a constant, could be done with a helper function though

mmioana (approved):
Awesome @rdig no Apollo error is showing up anymore 🙌 

Screenshot 2025-01-16 at 10 36 04
Screenshot 2025-01-16 at 10 36 34
Screenshot 2025-01-16 at 10 36 49
Screenshot 2025-01-16 at 10 39 04

Nice work and thank you for the fast fix! 🚀

jakubcolony (dismissed):
Apollo errors are gone, good to go! 🟢

davecreaser (approved):
Nice solution with a good amount of helpful comments, thanks!

Code all looking good to me, and the testing went well too:

Can see the gas debug log -
Screenshot 2025-01-16 at 11 47 06

And with meta txs disabled -
Screenshot 2025-01-16 at 11 47 36

The tx debug shows the overrides are populated -
Screenshot 2025-01-16 at 11 48 23

iamsamgibbs (dismissed):
Great job everyone involved on pinning down this issue. All makes sense and the fix works nicely.

Screenshot 2025-01-16 at 12 05 41

Screenshot 2025-01-16 at 12 06 53

mmioana (approved):
Looks good from my side 🥇 

Got the log for the network gas estimate
Screenshot 2025-01-17 at 09 29 08

Metatx are disabled
Screenshot 2025-01-17 at 09 29 27

Created a mint tokens action
Screenshot 2025-01-17 at 09 29 55
The overrides for the transaction are populated
Screenshot 2025-01-17 at 09 30 07

And added a log to check Alex's fix regarding the `maxPriorityFeePerGas`
Screenshot 2025-01-17 at 09 37 30

Nice work guys!!! 👏

iamsamgibbs (approved):
All working as expected! Lets get it merged.

Screenshot 2025-01-20 at 14 01 00
Screenshot 2025-01-20 at 14 02 29

PR #4148, opened 2025-01-20
Fix: Display persisted draft agreement

### LINKED ISSUE #3826: QA Agreements: Can't save a draft agreement
## Steps to reproduce

1. Open a new agreement action.
2. Complete the required fields.
3. Select the 'Safe draft' button.
4. No response is received, and the draft does not save. 

## Expected behaviour

* The draft agreement should save correctly. The 'Save draft' button should display the correct state when selected. The draft agreement should show on the Agreements page. 

## Actual behaviour

* There is no response from the 'Safe draft' button. The agreement does not save to a draft state. Prod is working correctly. 

image

mmioana:
## Description

- This PR addresses the fact that the draft agreement is not always showing up upon saving due to being fetched on the `Agreements` page directly from local storage. However, given the component might not re-render, the draft agreement wouldn't always be rendered. The solution was to make use of `Redux` store which already saves this value and re-populates the store with the values persisted in the local storage on page refresh.
- Also there was a validation error regarding the missing reputation when creating a new colony. Given the `Save draft` is of `type=button` and not `type=submit`, the error message was not showing up, but neither was the draft saved. In this case, the solution was to check if among the form errors there is only the `isMissingReputation`, in which case the draft can still be saved.

> [!NOTE]
> In the initial issue it was thought it has to do with the `Reputation` **extension version**, but the draft was saved in the local storage and only a few attempts it did appear on the `Agreements` page.
> [Check recording here](https://github.com/user-attachments/assets/3b5515bd-68cd-44c2-9506-ae89eddc867a)

> [!IMPORTANT]
> If you do however want to test this with a lower version of the `Reputation` extension (eg. `v11`), give me a sign and I'll provide more guidance until I'll update the `Notion` page - bit shoutout to @area on this 🙏 and to @iamsamgibbs for guiding me to the `Redux` store

## Testing

TODO: Let's test we can actually save a draft agreement and it will appear in the `Agreements` page without issues. 

* Step 1. Make sure to start a clean env
* Step 2. Install the `Reputation` extension
* Step 3. Now go to http://localhost:9091/planex/agreements
* Step 4. Create a few agreements
Screenshot 2025-01-20 at 21 45 59
* Step 5. Now let us create another agreement, but instead of `Create agreement`, click `Save draft`.

https://github.com/user-attachments/assets/bd005893-a50e-413b-97d8-026f26ddfa38

It should be rendered and you can see the entry in the local storage. Try to edit it, close it, delete it, create another one etc.

* Step 6. Now for the other issue I have discovered, please generate a new colony url
`node scripts/create-colony-url.js`
* Step 7. Go through the entire colony creation flow.
* Step 8. Install the `Reputation` extension
* Step 9. Now please click on the `Create a new agreement` from the side navigation (alternatively you can go to the `Agrements` page as before, it doesn't really matter)
* Step 10. Fill in all the missing fields and click `Create agreement`.
* Step 11. The `No reputation` warning should be shown
* Step 12. Now click on the `Save draft`. The draft should be saved without errors.
* Step 13. If you remove the `Title` or `Description`, clicking on the `Save draft` should show an error and the draft should not be updated

https://github.com/user-attachments/assets/597cab52-6bcd-4917-a9d0-87b2275aca18

**If you have more testing ideas in mind, do not hesitate to try them out**

## Diffs

**Changes** 🏗

* Replace `getDraftDecisionFromLocalStorage` with `useSelector(getDraftDecisionFromStore(...))` for the `draftAgreement` 
* In `SaveDraftButton` the `onClick` handler checks if the form is valid or there is only the missing reputation error before saving the draft agreement

Resolves #3826

### DIFF
--- a/src/components/frame/v5/pages/AgreementsPage/AgreementsPage.tsx
+++ b/src/components/frame/v5/pages/AgreementsPage/AgreementsPage.tsx
@@ -1,13 +1,14 @@
 import { Binoculars, FilePlus } from '@phosphor-icons/react';
 import { motion } from 'framer-motion';
 import React, { type FC } from 'react';
+import { useSelector } from 'react-redux';
 
 import { Action } from '~constants/actions.ts';
 import { useActionSidebarContext } from '~context/ActionSidebarContext/ActionSidebarContext.ts';
 import { useAppContext } from '~context/AppContext/AppContext.ts';
 import { useColonyContext } from '~context/ColonyContext/ColonyContext.ts';
 import { useSetPageHeadingTitle } from '~context/PageHeadingContext/PageHeadingContext.ts';
-import { getDraftDecisionFromLocalStorage } from '~utils/decisions.ts';
+import { getDraftDecisionFromStore } from '~utils/decisions.ts';
 import { formatText } from '~utils/intl.ts';
 import { ACTION_TYPE_FIELD_NAME } from '~v5/common/ActionSidebar/consts.ts';
 import EmptyContent from '~v5/common/EmptyContent/EmptyContent.tsx';
@@ -42,9 +43,8 @@ const AgreementsPage: FC = () => {
     !!activeFilters.motionStates ||
     !!activeFilters.search;
 
-  const draftAgreement = getDraftDecisionFromLocalStorage(
-    user?.walletAddress || '',
-    colonyAddress,
+  const draftAgreement = useSelector(
+    getDraftDecisionFromStore(user?.walletAddress || '', colonyAddress),
   );
 
   const passedAgreements = searchedAgreements.filter(
--- a/src/components/v5/common/ActionSidebar/consts.ts
+++ b/src/components/v5/common/ActionSidebar/consts.ts
@@ -9,6 +9,8 @@ import { formatText } from '~utils/intl.ts';
 
 import { reputationValidationSchema } from './hooks/useReputationValidation.ts';
 
+export { REPUTATION_VALIDATION_FIELD_NAME } from './hooks/useReputationValidation.ts';
+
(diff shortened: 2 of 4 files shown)


rdig (approved):
Very nice work! Everything works as expected! :1st_place_medal: 

Saving to draft and updating works

https://github.com/user-attachments/assets/61505e03-aaf2-405b-83c7-a29cd646f0b3


As well as the reputation warning when there's no reputation in a colony


https://github.com/user-attachments/assets/f5c0dff3-5b6b-4eba-9645-97171c74b558

iamsamgibbs (changes requested):
Thanks for reworking this @mmioana ! It's working nicely!

https://github.com/user-attachments/assets/ebf0c99b-4fa6-4d49-a63f-ed5fcfc34d42

I wasn't able to either recreate the "no reputation" warning or save a draft in a new colony though. (I get no console errors when clicking save)

https://github.com/user-attachments/assets/a66b1ad8-fdd6-4fda-96f6-13c004a5be34

Personally I think you should still be allowed to save a draft even if there is no reputation. I could imagine a scenario where you spent ages writing out a draft only for it to not save when you could simply retry in a few days time if it would let you save.

mmioana:
Thanks @iamsamgibbs for your review. Now that you're mentioning it, it would actually be a terrible experience to not be able to save the draft only because there is no reputation available yet. I have made the changes to actually accommodate for this, if you mind checking again 🙏

iamsamgibbs (approved):
Amazing! Great job with this, makes a lot more sense to me this way around.

https://github.com/user-attachments/assets/9133097e-6e80-465f-91d1-659327264321

PR #4152, opened 2025-01-21
Use debug logging for CoinGecko token price requests

rdig:
This PR is here because of my recent quest to make the console output of the CDapp cleaner, by "hiding" all the relevant messages that we show behind `debugLogging` call.

This has the benefit of being hidden by default in production, while still allowing us to investigate any errors that might show up, once we manually enable it

This time it is the output from all the CoinGecko calls make to fetch the various exchange rates.

For the most part, most tokens colonies will interact with, especially native ones, will not in fact have an exchange rate, so the console output will be needlessly spammed with these fetch errors

All I've done here is to clean that up a bit by logging the errors out via the debug logging util.

Here's how it used to look:
Screenshot from 2025-01-21 17-51-28

And here is how it looks now, if you turn on debug logging:
Screenshot from 2025-01-21 17-50-55

If you disagree with this change, and this it is better served as an error instead, I am open to feedback.

### Testing on QA

This is easier to test on QA than locally, so I'll list these instructions first _(you can test either, don't have to bother with both)_

1. Go to QA
2. Go to a colony _(try `rc` if you don't have one handy)_
3. Open your dev tools
4. Go to the balances page
5. Notice there should be no output regarding CoinGecko
6. Enable debug logging _(if you don't know how, ask me)_
7. Refresh the page, and go back to the balances page
8. Notice that you'll now have a output similar to this: Screenshot from 2025-01-21 17-50-55

### Testing Locally

As I've said above, testing locally is a tad more involved, as you have to overcome both Coingecko being available, as well as the code that prevents CoinGecko calls locally

1. Run the dev environment: `npm run dev`
2. Create demo data: `node scripts/create-data.js`
3. Ensure you have values for `COINGECKO_API_URL` and `COINGECKO_API_KEY` in your `.env.local.secrets` file _(and that you have ran `npm run prepare` after)_
4. Run the frontend: `npm run frontend`
5. Go to the `planex` colony
6. Go to the balances page http://localhost:9091/planex/balances
7. Notice there's no output relate to CoinGecko _(however, this is a bit of a trick)_
8. In your code editor, open the `src/utils/currency/currency.ts` file and comment out line `79` Screenshot from 2025-01-21 18-00-53
9. Refresh your browser page
11. Go back to `planex` and the balances page
12. Notice that you have similar output to this in your console Screenshot from 2025-01-21 17-50-55

**Optional:**

Optionally, you can check out the comparison on how this looked before:
13. Keep the code changes you have
14. Roll back one commit back `git checkout HEAD~1`
15. Go back to `planex` and the balances page
16. Check out the output in your console log that should be similar to this Screenshot from 2025-01-21 17-51-28

### DIFF
--- a/src/utils/currency/tokenPriceByName.ts
+++ b/src/utils/currency/tokenPriceByName.ts
@@ -1,4 +1,5 @@
 import { SupportedCurrencies } from '~gql';
+import debugLogging from '~utils/debug/debugLogging.ts';
 
 import { currencyApiConfig, coinGeckoMappings } from './config.ts';
 import {
@@ -36,8 +37,8 @@ const extractTokenPriceFromResponse = (
       mapToAPIFormat(currencies, conversionDenomination)
     ];
   } catch (e) {
-    console.error(
-      'Could not get token price from CoinGecko response. Response shape might have changed.',
+    debugLogging(
+      'CoinGecko: Could not get token price from response. Response shape might have changed.',
       e,
     );
     return 0;
@@ -80,14 +81,14 @@ export const fetchTokenPriceByName = async ({
         );
       }
 
-      console.error(
-        `Unable to get price for ${tokenName}. It probably doesn't have a listed exchange value.`,
+      debugLogging(
+        `CoinGecko: Unable to get price for ${tokenName}. It probably doesn't have a listed exchange value.`,
       );
       return 0;
     });
   } catch (e) {
     if (import.meta.env.DEV) {
-      console.error(e);
+      debugLogging(e);
     }
     return 0;
   }
--- a/src/utils/currency/tokenPriceDataByName.ts
+++ b/src/utils/currency/tokenPriceDataByName.ts
@@ -1,6 +1,7 @@
 import fromUnixTime from 'date-fns/fromUnixTime';
 
 import { SupportedCurrencies } from '~gql';
+import debugLogging from '~utils/debug/debugLogging.ts';
 
 import { currencyApiConfig, coinGeckoMappings } from './config.ts';
 import {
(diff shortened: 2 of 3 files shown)


src/utils/currency/tokenPriceByName.ts
```diff
@@ -80,14 +81,14 @@ export const fetchTokenPriceByName = async ({
...
 
-      console.error(
-        `Unable to get price for ${tokenName}. It probably doesn't have a listed exchange value.`,
+      debugLogging(
+        `CoinGecko: Unable to get price for ${tokenName}. It probably doesn't have a listed exchange value.`,
       );
       return 0;
     });
   } catch (e) {
     if (import.meta.env.DEV) {
```
mmioana, line 90: I think it no longer makes sense to have this condition given we use `debugLogging`

src/utils/currency/tokenPriceDataByName.ts
```diff
@@ -95,14 +96,14 @@ export const fetchTokenPriceDataByName = async ({
...
 
-      console.error(
-        `Unable to get price for ${tokenName}. It probably doesn't have a listed exchange value.`,
+      debugLogging(
+        `CoinGecko: Unable to get price for ${tokenName}. It probably doesn't have a listed exchange value.`,
       );
       return fallbackResponseData;
     });
   } catch (e) {
     if (!import.meta.env.DEV) {
```
mmioana, line 105: I think it no longer makes sense to have this condition given we use `debugLogging`. Even wondering why the condition is different from the previous one 🫠

mmioana (approved):
Really nice work on this issue @rdig 🙌 Thank you for cleaning up the way we log errors

I have tested in QA

Without having the debug logging enabled

https://github.com/user-attachments/assets/bb28fbd8-6e8c-4f42-b84e-0445fb310f6b

And after enabling it, I could see the `CoinGecko` errors 🎉 
Screenshot 2025-01-23 at 09 54 12

src/utils/currency/tokenPriceByName.ts
rdig: Very good point. I'll remove them!

iamsamgibbs (approved):
Nice tidying, makes a lot of sense. This works nicely:

Screenshot 2025-01-23 at 19 19 20

Nortsova (approved):
Great work @rdig. Everything works exactly as described in testing steps 👏 

QA env: 
No errors without debug logging: 
image

With debug logging:
image

locally after commented out 79 line:
image

Errors on master:
image

PR #4161, opened 2025-01-22
[Proxy colonies M1] Feat: Fetch available chains from DB and update SUPPORTED_CHAINS config

mmioana:
## Description

This PR introduces lookup of supported proxy chains from the database.
We still need to keep them in a local config, due to all the titles and lookups.
[Block ingestor PR](https://github.com/JoinColony/block-ingestor/pull/319)

## Testing

1. Run your dev env and check that we got supported chains in the database by running the following query
```
query MyQuery {
  listSupportedChains {
    items {
      id
      isActive
    }
  }
}
```
image

2. Run the create-data script and go to `planex`. Try to create a `Manage supported chains` action and verify that you only get "Local proxy chain 1" and "Local proxy chain 2" in the dropdown.
image
> [!IMPORTANT]
> Unfortunately when stopping a docker container the `SIGTERM` event doesn't get properly propagated to the child process which in our case would be the process started with `pnpm --filter @joincolony/proxy-chain run start`
> So the next steps require you to stop the proxy block ingestor container and run the `block-ingestor` locally (branch `feat/dynamically-add-supported-proxy-chain`)

3. Stop the proxy chain ingestor 1 with docker and run it locally via `pnpm run dev:proxy-chain` with the block ingestor changes above (sorry, it's a bit janky :sweat_smile: ) Verify that you get these logs and that the proxy is still active
image

4. Stop the running instance and run the query above, verify that it's inactive. Reload the app and check that "Local proxy chain 1" is disabled.
image
image
image

5. Restart it and check that everything is reenabled :D 
image
image

## Diffs

**New stuff** ✨

* New entity in the DB called `SupportedChains`
* New constants for the local proxy chains

**Changes** 🏗

* All the proxies are in a single supported chains array which we filter according to the ones active in the database
* The proxy block ingestor adds a new supported chain entry when it's first started
* The proxy ingestors also mark the supported chain as inactive when it's stopped

### DIFF
--- a/amplify/backend/api/colonycdapp/schema/proxyColonies.graphql
+++ b/amplify/backend/api/colonycdapp/schema/proxyColonies.graphql
@@ -22,3 +22,14 @@ type ProxyColony @model {
   """
   isActive: Boolean!
 }
+
+type SupportedChain @model {
+  """
+  The chainId of the supported chain
+  """
+  id: ID!
+  """
+  A flag that tells us if the supported chain is active or not
+  """
+  isActive: Boolean
+}
--- a/docker/colony-cdapp-dev-env-block-ingestor
+++ b/docker/colony-cdapp-dev-env-block-ingestor
@@ -1,6 +1,6 @@
 FROM colony-cdapp-dev-env/base:latest
 
-ENV BLOCK_INGESTOR_HASH=983e0b68472b8f7f7406352679e29b2d2d4d49c0
+ENV BLOCK_INGESTOR_HASH=4ac06d5744f8ada7b80047afd586da4d82898436
 
 # Declare volumes to set up metadata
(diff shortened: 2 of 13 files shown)


bassgeta (approved):
Approving since we paired on it and wrote the description so I basically tested it already :D

iamsamgibbs (approved):
Nice work! All working exactly as expected.

Both show up when both proxies are running:

Screenshot 2025-01-24 at 13 41 10

Screenshot 2025-01-24 at 13 50 08

And one disappears when it is shut off:

Screenshot 2025-01-24 at 13 52 31

Screenshot 2025-01-24 at 13 57 35

rdig (approved):

PR #4162, opened 2025-01-22
Add rate limited notifications queue

davecreaser:
## Description

- Since this is really a block ingestor change, the description is here: https://github.com/JoinColony/block-ingestor/pull/320

## Testing

* ⚠️  You'll need to restart your dev environment and make sure you run `npm run dev --notifications` to enable notifications locally.
* Once it's up and running, stop the block ingestor in docker. Open `.env` of the CDapp and copy the auto generated MAGICBELL_DEV_KEY, and add it to the `.env` of your block ingestor. Now run the block ingestor locally with `npm run dev`.
* Now you need a way to trigger more than 60 notifications inside a minute. The easiest way is probably to open `src/handlers/actions/mintTokens.ts` in the block ingestor, and go to line 35 where it calls `sendPermissionsActionNotifications`. Just add a for loop around this call to fire it like 80 times when a mint token action happens.
```ts
for (let index = 0; index < 80; index++) {
      sendPermissionsActionNotifications({
        creator: initiatorAddress,
        colonyAddress,
        transactionHash,
        notificationCategory: NotificationCategory.Payment,
      });
    }
```
* Open localhost, and mint some tokens. You should see 30 notifications come in pretty quickly, and after waiting a minute, more should come through.
* Check the block ingestor logs, there should be no errors about API rate limits being hit.

## Diffs

**Changes** 🏗

* Block ingestor changes: https://github.com/JoinColony/block-ingestor/pull/320

### DIFF
--- a/docker/colony-cdapp-dev-env-block-ingestor
+++ b/docker/colony-cdapp-dev-env-block-ingestor
@@ -1,6 +1,6 @@
 FROM colony-cdapp-dev-env/base:latest
 
-ENV BLOCK_INGESTOR_HASH=87073303a0369e97cf439baf572809f238f923ed
+ENV BLOCK_INGESTOR_HASH=beeb66bc18b2a353a8b961cc1f98d229e73bf7b8
 
 # Declare volumes to set up metadata
 VOLUME [ "/colonyCDapp/amplify/mock-data" ]
--- a/docker/colony-cdapp-dev-env-block-ingestor
+++ b/docker/colony-cdapp-dev-env-block-ingestor
@@ -1,6 +1,6 @@
 FROM colony-cdapp-dev-env/base:latest
 
-ENV BLOCK_INGESTOR_HASH=beeb66bc18b2a353a8b961cc1f98d229e73bf7b8
+ENV BLOCK_INGESTOR_HASH=77acd2cb6ed86877a19807bee2a902a6e846bdbf
 
 # Declare volumes to set up metadata
 VOLUME [ "/colonyCDapp/amplify/mock-data" ]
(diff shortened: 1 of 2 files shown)


mmioana (changes requested):
Really nice work on this one @davecreaser 🙌 

Applied the changes in the local `block-ingestor` and created a mint tokens action
Screenshot 2025-01-23 at 09 24 45
There were no errors in the console
Screenshot 2025-01-23 at 09 28 26
Added an extra log to check the queue size and it seems the extra `20` notifications compared to the queue cap were added
Screenshot 2025-01-23 at 09 29 18

However @davecreaser, after creating many more Mint tokens actions, I did manage to still spot the `API limit reached` error. Not sure if I did something wrong?

Screenshot 2025-01-23 at 09 36 51

davecreaser:
Thanks @mmioana, turns out they had more `hidden` rate limits 🙄 I think this one happened because the front end was also sending a fetch request each time a new notification came in (to get that notification) and these were all combining to then hit the rate limit again.

I've reduced the block ingestor rate to only send 30 per minute, which will keep us nicely within the limits, and leave a big buffer for the front end to make requests.

Also added a retry mechanism with priority to the notifications queue, so that notifications which still hit the rate limit (unlikely hopefully) will get retried.

mmioana (approved):
Thanks for your really cool work on this issue @davecreaser 🥇 

Applied the local changes to the block ingestor and created a `Mint tokens` action
Immediately I got `30` notifications
Screenshot 2025-01-27 at 15 49 20
No errors on the block ingestor side
Screenshot 2025-01-27 at 15 50 06
After 1 minute I got `30` more new notifications
Screenshot 2025-01-27 at 15 50 18
And after another minute, all notifications were there
Screenshot 2025-01-27 at 15 53 05

Awesome work! 🎉

rdig (approved):
Rate limiting working as expected, nothing triggered in the block ingestor logs, while notifications showing up in the frontend end.

Nice job fixing this

Screenshot from 2025-01-29 19-59-53
Screenshot from 2025-01-29 19-59-55

PR #4150, opened 2025-01-21
Account for Metamask not supporting the `eth_maxPriorityFeePerGas` call

rdig:
This issue came about after the introduction of #4123 which, while technically correct, runs into the a problem with Metamask, and more specifically Metamask's provider.

Even while running a manually set RPC endpoint which can be confirmed as supporting [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559)

Screenshot from 2025-01-21 14-47-48

Metamask will **block the `eth_maxPriorityFeePerGas` call**

Screenshot from 2025-01-20 17-16-48

While waiting to see if this ever gets resolved _(see this PR MetaMask/metamask-extension#29818 by our very own @area)_ we are accounting for it by setting it to zero (`0`).

This means that the transaction sender will not pay anything extra, besides the required gas fee, for their transaction to be mined.

This works great on Arbitrum, Arbitrum Sepolia since they make user of a Sequencer, however, on something like Mainnet this won't work, and we'll need to find another way around it

More reading: https://docs.arbitrum.io/learn-more/faq#do-i-need-to-pay-a-tip-or-priority-fee-for-my-arbitrum-transactions

Screenshot from 2025-01-21 14-56-38

Screenshot from 2025-01-21 14-58-31

## Additional Fix

This issue actually highlighted a bug inside the `estimateGasCost` saga, where, if `maxFeePerGas` and `maxPriorityFeePerGas` were not available, we would skip updating the transaction's `options` object, however we would wait for said object to _finish_ updating which could never actually succeed.

I have fixed that as well, by moving the `take` call, which waits for the next `TRANSACTION_UPDATED_IN_DB` action, inside the logic check block.

## Testing

1. Run the dev environment: `npm run dev`
2. Create demo data: `node scripts/create-data.js`
3. Run the frontend: `npm run frontend`
4. Ensure you connect your Metamask wallet, and not your Dev wallet, and ensure you use one of the first 3 Dev wallets addresses with Metamask _(If you don't have them added, you can get them here:  http://localhost:3006/ganache-accounts.json)_
5. Go to the `planex` colony
6. Ensure you see the Metamask / gas estimator warnings _(which mean it couldn't make the `eth_maxPriorityFeePerGas` call)_ Screenshot from 2025-01-21 15-08-14
7. Ensure you see the Gas Debug Log with the `maxPriorityFeePerGas` set to `0` Screenshot from 2025-01-21 14-56-38
8. Send a transaction _(can be anything, payment, mint, extensions, etc)_
[...]

PR #4166, opened 2025-01-23
Fix (?) intermittent error in create-data script

area:
<p align="center">

</p>

In #4150, I saw @mmioana had experienced one of my most hated things while trying to run the `create-data` script: _an intermittent error_

Screenshot 2025-01-23 at 10 24 36

so I went hunting and think I've arrived at a simple fix. 

Currently, when the script is run:

* Colonies get between 1 and 100 million of a token
* Domains get between 100 and 500 of a token

So far, so good.

Individual users are then paid out from each domain by calling [`userPayments`](https://github.com/JoinColony/colonyCDapp/blob/0f0e20ee01635902183852292d09e85d0dac5616/scripts/create-data.js#L1626) with the three `walletUsers` (which are the first three ganache accounts) and a random contributor address.

Users are paid out a random amount between [1 and 3 tokens](https://github.com/JoinColony/colonyCDapp/blob/0f0e20ee01635902183852292d09e85d0dac5616/scripts/create-data.js#L1272)... unless we are considering the [first 'ganache' address](https://github.com/JoinColony/colonyCDapp/blob/0f0e20ee01635902183852292d09e85d0dac5616/scripts/create-data.js#L1278), which gets 50 tokens, a fixed amount.

So, how does this result in a (rare) error when running the script? When paying out from a domain, the random contributor selected can be the first ganache address (if the domain is not the first in the colony to be considered). In that case, we will pay out to the first ganache address twice, resuling in somewhere between 102 and 106 tokens trying to leave the domain - but there is a chance that the domain only received as low as 100 tokens, and will not have enough to pay out.

In that scenario, the transaction will fail with the overflow message seen in the original error - but to occur, the domain must have received a sufficiently low number of tokens (less than 1% chance) and the first ganache account must have been randomly selected as a contributor (haven't properly understood how this is generated, but must be less than 33%), explaining why it is so rare to occur, anecdotally.

To fix the issue, the 'constant' amount we pay out to the first ganache address has been reduced to 30 tokens, which means that the most we can pay out from a domain is 66 tokens, which is less than the smallest amount that can be given to a domain.

For anyone paying _particularly_ close attention, the last non-error log in the screenshot was paying out address `0x27fF0C145E191C22C75cD123C679C3e1F58a4469`, which is `ganacheAccounts[2]`, not `ganacheAccounts[0]`. However, the log occurs in the script [after the payout succeeds](https://github.com/JoinColony/colonyCDapp/blob/0f0e20ee01635902183852292d09e85d0dac5616/scripts/create-data.js#L1335), and so the error will have occurred when paying out the _next_ address, which you can see in the failed transaction blob is `0xb77D57F4959eAfA0339424b83FcFaf9c15407461`, as would be the case in the situation I've described here.

I've checked with Raul and this 50 tokens isn't a magic value anywhere, so shouldn't break anything. Obviously the issue with an intermittent error is that it's tough to be sure that it's fixed, especially one with such a low recurrence rate, but I hope that the above case is compelling.

If/when this is merged, if you see a similar error in the future, please let me know - I really detest intermittent errors, which make it incredibly hard to debug _actual_ issues, and try to extinguish them with extreme prejudice!

### DIFF
--- a/scripts/create-data.js
+++ b/scripts/create-data.js
@@ -1275,7 +1275,7 @@ const userPayments = async (
             users[userIndex] ===
             utils.getAddress(Object.keys(ganacheAddresses)[0])
           ) {
-            amount = BigNumber.from(`50000000000000000000`);
+            amount = BigNumber.from(`30000000000000000000`);
           }
 
           const params = [


rdig:
Screenshot from 2025-01-23 14-31-03

:100:

iamsamgibbs (approved):
Nice investigation! Tested just to be safe and script passes as expected and everything is still working.

Screenshot 2025-01-24 at 12 55 12

mmioana (approved):
Awesome @area thanks so much for investigating and fixing this issue 🤩 

Screenshot 2025-01-27 at 17 29 47

PR #4186, opened 2025-01-24
Feat: Add script for deploying colony native token on chain

mmioana:
## Description

- This PR addresses the fact that once a proxy colony gets deployed, on the `DEV` env we'll need to deploy the associated native token on the respective chain in order to have some data to work with. 
- Also this PR adds a script for supporting unlocking and minting tokens
Thank you @iamsamgibbs and @bassgeta for all the help

## Testing

TODO: Let's test the colony native token gets deployed on a new chain and we actually store that information.

* Step 1. Go to http://localhost:9091/planex
* Step 2. Create a `Manage supported chains` action/motion for chain `Local Proxy Chain 1`

Screenshot 2025-01-24 at 20 20 31

* Step 3. Close the action sidebar and copy the colony address.

* Step 4. Once the action is completed, please run `node scripts/deploy-token-to-chain.js`
Please use the following values for the input
foreignChainId `265669101`
foreignChainRpcUrl `http://localhost:8546`
colonyAddress `<COLONY_ADDRESS_HERE>`

You should get a similar output to the one below
Screenshot 2025-01-27 at 21 17 21

* Step 5. Run the following query 
```
query MyQuery {
  listTokens {
    items {
      id
      name
      chainMetadata {
        chainId
      }
    }
  }
}
```
You should see a new entry for `Space Creds` on chain with id `265669101`
Screenshot 2025-01-27 at 21 18 14

* Step 6. Copy the token id from the `265669101` chain

* Step 7. Now let's test the script for unlocking and minting tokens, please run `node scripts/unlock-mint-tokens-on-chain.js`
Please use the following values for the input
foreignChainRpcUrl `http://localhost:8546`
colonyAddress `<COLONY_ADDRESS_HERE>`
tokenAddress `<TOKEN_ADDRESS_FROM_OTHER_CHAIN_HERE>`
tokenAmount `100`
Screenshot 2025-01-27 at 21 25 38

* Step 8. Test the token was actually minted by running `npm run hardhat`. Then paste the following, but make sure to replace the colony and token addresses.
```
p = await new ethers.providers.StaticJsonRpcProvider("http://network-contracts-remote:8545")

erc20Abi = ["function balanceOf(address owner) view returns (uint256)", "function decimals() view returns (uint8)"]

t = new ethers.Contract("<TOKEN_ADDRESS_FROM_OTHER_CHAIN_HERE>", erc20Abi, p)

await t.balanceOf("<COLONY_ADDRESS_HERE>")
```
Screenshot 2025-01-27 at 21 29 40

> [!IMPORTANT]
> Running the script above multiple times will create each time a new entry on the given chain, which actually simulates a real situation when a token can have multiple addresses on a single chain

## Diffs

**New stuff** ✨

* `deploy-token-to-chain` script
* `unlock-mint-tokens-on-chain` script

### DIFF
--- a/amplify/backend/api/colonycdapp/schema/schema.graphql
+++ b/amplify/backend/api/colonycdapp/schema/schema.graphql
@@ -899,10 +899,6 @@ enum ColonyActionType {
   MANAGE_TOKENS_MOTION
   MANAGE_TOKENS_MULTISIG
   """
-  An action related to arbitrary transaction
-  """
-  ARBITRARY_TX
-  """
   An action related to creating a proxy colony
   """
   ADD_PROXY_COLONY
--- a/scripts/deploy-token-to-chain.js
+++ b/scripts/deploy-token-to-chain.js
@@ -5,7 +5,6 @@ const {
 const { providers, ContractFactory } = require('ethers');
 const readline = require('readline');
 const { graphqlRequest } = require('./utils/graphqlRequest');
-const { unlockMintTokens } = require('./utils/unlockMintTokens');
 
 const API_KEY = 'da2-fakeApiId123456';
 const GRAPHQL_URI =
@@ -164,8 +163,6 @@ const deployNativeTokenToChain = async (
     console.log(
       `Proxy token with address ${foreignToken.address} successfully deployed for colony with address ${colonyAddress}.`,
     );
-
-    return foreignToken;
   } catch (error) {
     console.error(
       `Error deploying proxy token for ${colonyAddress}: ${error.message}`,
@@ -210,34 +207,6 @@ const getArgumentsFromConsole = async () => {
   };
 };
 
-const furtherActionsOnToken = async (colonyAddress, foreignToken) => {
-  const unlockMintTokensAction = await promptQuestion(
-    'Do you want to unlock and mint tokens now? Y/N',
-  );
-
-  if (unlockMintTokensAction.toUpperCase() !== 'Y') {
-    console.log(
-      'Skipping token unlock and minting. If you change your mind, please use the mint-tokens-on-chain script.',
-    );
-    return;
-  }
-
-  const tokenAmount = await promptQuestion(
-    'Please provide a numeric tokenAmount. Press enter to default to 100',
-  );
-
-  if (tokenAmount && (isNaN(tokenAmount) || Number(tokenAmount) <= 0)) {
-    console.error('Error: Please enter a valid number greater than 0.');
-    process.exit(1);
-  }
-
-  await unlockMintTokens(
-    colonyAddress,
-    foreignToken,
-    !tokenAmount ? 100 : Number(tokenAmount),
-  );
-};
-
 const start = async () => {
   try {
     const { foreignChainId, foreignChainRpcUrl, colonyAddress } =
@@ -250,14 +219,12 @@ const start = async () => {
     );
 
     const { nativeToken: colonyNativeToken } = deployedColony;
-    const foreignToken = await deployNativeTokenToChain(
+    await deployNativeTokenToChain(
       tokenFactory,
       foreignChainId,
       colonyNativeToken,
       colonyAddress,
     );
-
-    await furtherActionsOnToken(colonyAddress, foreignToken);
   } catch (error) {
     console.error(
       `There was an error executing the deploy-token-to-chain script. Error: ${error.message}`,
--- a/scripts/deploy-token-to-chain.js
+++ b/scripts/deploy-token-to-chain.js
@@ -1,0 +1,270 @@
+const {
+  abi: metaTxAbi,
+  bytecode: metaTxBytecode,
+} = require('@colony/abis/versions/hmwss/MetaTxToken');
+const { providers, ContractFactory } = require('ethers');
+const readline = require('readline');
+const { graphqlRequest } = require('./utils/graphqlRequest');
+const { unlockMintTokens } = require('./utils/unlockMintTokens');
+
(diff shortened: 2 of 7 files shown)


src/redux/sagas/proxyColonies/createProxyColony.ts
```diff
@@ -147,6 +146,8 @@ function* createProxyColony({
     }
     setTxHash?.(transaction.hash);
 
+    yield deployProxyColonyNativeToken({ colonyAddress, foreignChainId });
+
```
iamsamgibbs, line 150: On a quick look at this, it looks great! The only thing that initially jumps out to me is that we might want to multicall the createProxyColony and deployProxyColonyNativeTokens so that both transactions go through together. Otherwise we may get a scenario where the user manually rejects the deployProxyColonyNativeTokens transaction and it never gets deployed.

amplify/backend/api/colonycdapp/schema/schema.graphql
```diff
@@ -899,10 +899,6 @@ enum ColonyActionType {
   MANAGE_TOKENS_MOTION
   MANAGE_TOKENS_MULTISIG
   """
-  An action related to arbitrary transaction
-  """
-  ARBITRARY_TX
-  """
```
mmioana, line 905: This was some leftover after rebasing, so cleaned it up

scripts/deploy-token-to-chain.js
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+  return deployedColonies;
+};
+
+const unlockAndMintTokens = async (colonyAddress, foreignToken) => {
+  const foreignTokenAddress = foreignToken.address;
+  try {
+    await foreignToken.deployTransaction.wait();
+    await foreignToken.unlock();
```
iamsamgibbs, line 125: I don't think we need to unlock the token, seems to work okay without it.

iamsamgibbs (changes requested):
Great job with this! It certainly hasn't been easy to get our heads around!

This is working really nicely, I can indeed deploy a token to the proxy chain and it is minting funds. I confirmed it worked using these steps:

```
npm run hardhat

p = await new ethers.providers.StaticJsonRpcProvider("http://network-contracts-remote:8545")

erc20Abi = ["function balanceOf(address owner) view returns (uint256)", "function decimals() view returns (uint8)"]

t = new ethers.Contract("tokenAddress", erc20Abi, p)

await t.balanceOf("colonyAddress")
```

Screenshot 2025-01-27 at 15 54 23

The script does definitely work so I could approve this right now.

In my opinion though, I think it would be more useful to separate this a bit further from the colonies so we have a bit more control over it.

It would be great if the script could take a colonyAddress too as an argument so we can specifically deploy the token for an individual colony rather than having a new token be generated for each colony every time the script is ran.

It would also be nice to have a separate `mint-tokens-on-proxy-colony` script which does only the minting part so we can mint additional tokens at a later point.

Let me know what you think! I'm happy to assist with this too!

mmioana:
> Great job with this! It certainly hasn't been easy to get our heads around!
> 
> This is working really nicely, I can indeed deploy a token to the proxy chain and it is minting funds. I confirmed it worked using these steps:
> 
> ```
> npm run hardhat
> 
> p = await new ethers.providers.StaticJsonRpcProvider("http://network-contracts-remote:8545")
> 
> erc20Abi = ["function balanceOf(address owner) view returns (uint256)", "function decimals() view returns (uint8)"]
> 
> t = new ethers.Contract("tokenAddress", erc20Abi, p)
> 
> await t.balanceOf("colonyAddress")
> ```
> 
> Screenshot 2025-01-27 at 15 54 23
> 
> The script does definitely work so I could approve this right now.
> 
> In my opinion though, I think it would be more useful to separate this a bit further from the colonies so we have a bit more control over it.
> 
> It would be great if the script could take a colonyAddress too as an argument so we can specifically deploy the token for an individual colony rather than having a new token be generated for each colony every time the script is ran.
> 
> It would also be nice to have a separate `mint-tokens-on-proxy-colony` script which does only the minting part so we can mint additional tokens at a later point.
> 
> Let me know what you think! I'm happy to assist with this too!

Amazing, I didn't know about this way of testing that a token has indeed been minted! Will add it to the testing steps 💯 
And I agree with your ideas, will continue with the other script creation and also adding the option for passing the `colonyAddress`

scripts/deploy-token-to-chain.js
mmioana: I am wondering though @iamsamgibbs if we don't `unlock` the token, will we be able to use for the payments?

bassgeta (approved):
Really nice job on decoupling all of these scripts, it's quite flexible and the dev UX is super legit!
1. Added a proxy colony
image
2. Deployed the token :heavy_check_mark: 
image
image
3. Minted the newly deployed token 
image
4. Hardhat says we have 100 of the token :heavy_check_mark: 
image

iamsamgibbs (approved):
Perfect! Thank you for the changes, this is great.

Deployed a token to a specific colony:
Screenshot 2025-01-28 at 10 11 14

Screenshot 2025-01-28 at 10 11 37

Minted some tokens:
Screenshot 2025-01-28 at 10 15 40

Checked it worked then minted some more:
Screenshot 2025-01-28 at 10 17 34

Amazing, lets get it merged! A nice future improvement would be to add flags to the scripts for `--proxy-chain-1` and `--proxy-chain-2` that automatically select the chainId and rcpUrl but that can be for later down the line.

rdig (approved):
Our scripts folder is getting bigger :P

PR #4201, opened 2025-01-28
Fix: changing distribution type should show accurate percentage

### LINKED ISSUE #3856: [Split payment] Total percentage in split payment table shows incorrect value when switching distribution types
## Steps to reproduce

1. Open split payment action.
2. Fill in all required fields.
3. Add more than 5 members.
4. Switch the distribution type between Equal, Unequal, and Reputation percentage distribution.

## Expected behaviour

* The total percentage should correctly display 100%, regardless of how the distribution type is switched or adjusted.

## Actual behaviour

* The total percentage occasionally shows values such as 99.999% or 100.001%, causing inconsistency and confusion.

image
image

https://github.com/user-attachments/assets/d7753b47-a03c-422b-b887-5986d7d12953

davecreaser:
## Description

- Change the way the total percentage is being calculated to not do any rounding until the final step, increasing accuracy.
- This also fixes a bug where switching from distribution type `equal` (which will always be 100%) to unequal suddenly makes the percentage not 100%.

## Testing

* Step 1. Make a split payment using equal as the distribution type. Have an awkward number of recipients compared to the total amount so that the percentages have a number of decimals.
Screenshot 2025-01-28 at 17 08 48

* Step 2. Change the distribution type to unequal. The total percentage should still say 100%, and you should be able to submit the form.
Screenshot 2025-01-28 at 17 09 08

* Step 3. Play around with different amounts / recipients with an uneven distribution type. The percentage should always be correct.
Screenshot 2025-01-28 at 17 09 49
Screenshot 2025-01-28 at 17 10 03

## Diffs

**Changes** 🏗

* Change the way the total percentage is calculated for split payments

Resolves #3856

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/forms/SplitPaymentForm/utils.ts
+++ b/src/components/v5/common/ActionSidebar/partials/forms/SplitPaymentForm/utils.ts
@@ -11,6 +11,7 @@ import { findDomainByNativeId } from '~utils/domains.ts';
 import { getEnumValueFromKey } from '~utils/getEnumValueFromKey.ts';
 
 import { type SplitPaymentFormValues } from './hooks.ts';
+import { type SplitPaymentRecipientsFieldModel } from './partials/SplitPaymentRecipientsField/types.ts';
 
 export const getSplitPaymentPayload = (
   colony: Colony,
@@ -61,4 +62,22 @@ export const getSplitPaymentPayload = (
         tokenDecimals,
       })) || [],
   };
+};
+
+export const getUnevenSplitPaymentTotalPercentage = (
+  amount: number,
+  recipients: SplitPaymentRecipientsFieldModel[],
+) => {
+  if (!amount) {
+    return 0;
+  }
+
+  const total =
+    recipients?.reduce((acc, recipient) => {
+      return acc + Number(recipient.amount);
+    }, 0) || 0;
+
+  const percentage = (100 * total) / amount;
+
+  return Number(percentage.toFixed(4));
 };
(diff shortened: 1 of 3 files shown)


bassgeta (approved):
Nice job, this stuff was pretty pesky, nicely ironed out :100: 
1. Created the classic "let's make .33333333333 decimals" payments
image
2. Changing to unequal leaves everything intact
image

A bug I've found (which probably makes no sense fixing here), but if we add values beyond the 4th decimal, it will allow creating the action
image
but since we round to 4 decimals, the end result is correct
image

kinda confusing :thinking: 

anyhow, this doesn't affect this PR so I'll gladly smash approve :ship:

src/components/v5/common/ActionSidebar/partials/forms/SplitPaymentForm/utils.ts
```diff
@@ -62,3 +63,21 @@ export const getSplitPaymentPayload = (
...
+
+  const total =
+    recipients?.reduce((acc, recipient) => {
+      return acc + Number(recipient.amount);
+    }, 0) || 0;
+
+  const percentage = (100 * total) / amount;
+
+  return Number(percentage.toFixed(4));
+};
```
mmioana, line 83: ⭐

mmioana (approved):
Really nice work on this one @davecreaser 💯 

Tested it and seems the values do add up to `100%` when switching from `equal` to `unequal` distribution type
Screenshot 2025-01-30 at 14 00 08
Screenshot 2025-01-30 at 14 00 53
Screenshot 2025-01-30 at 14 01 54

https://github.com/user-attachments/assets/d92f5e8b-94cd-4212-aa0e-5186b77647de

Nice!

rdig (approved):
Very nice! Tested both equal and unequal and changed various percentages.

Everything is acting as expected. 

Very nice fix! :tada: 


https://github.com/user-attachments/assets/c23456d6-0861-4ad1-95bc-8484afb0c6d7

PR #4216, opened 2025-01-30
[Dashboard] Fix: Broken dashboard layout on mobile 

### LINKED ISSUE #3982: [Dashboard mobile] Layout breaks on mobile after opening the meatball menu before the page finishes loading
## Steps to reproduce

1. Open the Latest activity page.
2. Navigate to the Dashboard.
3. Before the page finishes loading, tap the meatball menu (three dots) next to an activity.
image
4. Scroll the screen down.
 5. Observe the layout misalignment or breakage.

https://github.com/user-attachments/assets/114e069b-9b18-4152-9acb-919244eddc85

## Expected behaviour

* The layout should remain consistent.

## Actual behaviour

* The layout breaks, causing misalignment and a poor user experience.

mmioana:
## Description

- This PR aims to fix the popper dropdown menus present on the dashboard that were overflowing over the page navigation header when scrolling on `mobile`. The same issue was present also for the team filter dots menu on `tablet`. However I did not manage to reproduce the other layout breaking mentioned in the original issue.
- While solving this, I encountered some issues with the user navigation which on `desktop` was occupying the whole screen height due to `!h-dynamic-screen` overridding the `md:h-auto`. Also, on the landing page, once opening the user navigation, the page header was no longer visible due to having 2 instances of the `PageHeader`, their visibility being manipulated through `css` classes. This wouldn't have been a problem, if the `PageHeader` wouldn't have set a css variable `CSSCustomVariable.HeaderNavSectionContainer` under the scenes. The solution was to conditional render the `PageHeader` based on the device size, so the custom css variable doesn't get overridden with a wrong value.
- The initial reported issue was having the charts scroll with the content when the colony links dropdown is open on `mobile`. This was reproduced by @olgapod only using a `iPhone 13 Pro Max` real device, but in the simulator, I couldn't, so hoping the css fixes in this PR address that as well

## Testing

TODO: Let's test everything works as expected ✨. 

* Step 1. Let's start testing the original issue, so make sure your dev env is running, you have data and visit http://localhost:9091/planex 
* Step 2. Switch to a `mobile` screen size and reload the page. 
* Step 3. Check opening the colony links doesn't go over the page navigation header nor does it break the layout
Screenshot 2025-01-31 at 09 32 15
* Step 4. Try the same for the reputation chart dots menu

https://github.com/user-attachments/assets/2e1dc3fc-7e6c-4207-a319-13f25a41cfeb

* Step 5. Now switch to a `tablet` screen size and check the dots entry in the team filter doesn't go over the page navigation header on scroll

https://github.com/user-attachments/assets/65923069-bc31-499f-b869-1f3f76bb8ad6

* Step 6. Now let us test also the user navigation. So please switch to a `desktop` screen size. Try opening the user menu and check it doesn't occupy the whole screen height. You can also test this on the landing page or the user account pages.

https://github.com/user-attachments/assets/b4f117f6-8aac-4105-9dac-fdddb4a9f789

* Step 7. Now go to the landing page http://localhost:9091 and set your screen to a `mobile` size. Try opening the user menu up and check the page navigation header is still visible. 

https://github.com/user-attachments/assets/c638d57c-13e4-4aaa-a1d2-acc0c236822b

## Diffs

**Changes** 🏗

* Added and removed some css classes
* Refactored the `useEffect` in the `useHeightResizeObserver` to make use of the `updateHeight` function and not have the function directly in the `ResizeObserver`
* Make use of `isTablet` for conditionally rendering the `PageHeader` in `LandingPageLayout`

**New stuff** ✨ 
* `usePopperCustomPositioning` hook

Resolves #3982

### DIFF
--- a/src/hooks/usePopperCustomPositioning.ts
+++ b/src/hooks/usePopperCustomPositioning.ts
@@ -1,0 +1,92 @@
+import { useCallback } from 'react';
+
+import { useMobile } from '~hooks';
+
+import { useResize } from './useResize.ts';
+
+interface PopperCustomPositioningProps<T, K> {
+  triggerRef: React.RefObject<T>;
+  menuRef: React.RefObject<K>;
+  config: {
+    skipDesktop?: boolean;
+    keepInitialBottomPlacement?: boolean;
+    offset?: {
+      top: number;
+      left: number;
+    };
+    useTriggerAsTranslateDefault?: boolean;
+    hasLeftAlignment?: boolean;
+  };
+}
+
+// @Note: this can be used also for MeatBallMenu component, but will require some testing and refactoring
+export const usePopperCustomPositioning = <
+  T extends HTMLElement,
+  K extends HTMLElement,
+>({
+  triggerRef,
+  menuRef,
+  config: {
+    skipDesktop,
+    keepInitialBottomPlacement,
+    offset = {
+      top: 1.2,
+      left: 1.2,
+    },
+    useTriggerAsTranslateDefault,
+    hasLeftAlignment,
+  },
+}: PopperCustomPositioningProps<T, K>) => {
+  const isMobile = useMobile();
+
+  const trigger = triggerRef.current;
+  const menu = menuRef.current;
+
+  const adjustMenuPosition = useCallback(() => {
+    if (!trigger || !menu) return;
+
+    if (!isMobile && skipDesktop) return;
+
+    const triggerPlacement = trigger.getBoundingClientRect();
+    const leftOffset = window.innerWidth - triggerPlacement.x;
+    const bottomOffset = window.innerHeight - triggerPlacement.y;
+    let translateX = useTriggerAsTranslateDefault ? triggerPlacement.x : '0';
+    let translateY = useTriggerAsTranslateDefault ? triggerPlacement.y : '0';
+
+    if (!isMobile) {
+      if (leftOffset < menu.clientWidth || hasLeftAlignment) {
+        translateX = `calc(-100% + ${offset.left}rem)`;
+      }
+    } else {
+      translateX = `calc(-100% + ${leftOffset}px - ${offset.left}rem)`;
+    }
+
+    if (bottomOffset < menu.clientHeight && !keepInitialBottomPlacement) {
+      translateY = `calc(-100% - ${offset.top}rem)`;
+    }
+
+    menu.style.transform = `translate(${translateX}, ${translateY})`;
+  }, [
+    isMobile,
+    trigger,
+    menu,
+    skipDesktop,
+    offset,
+    hasLeftAlignment,
+    keepInitialBottomPlacement,
+    useTriggerAsTranslateDefault,
+  ]);
+
+  useResize(adjustMenuPosition);
+
+  // Separate dynamic styles and classes since tailwind doesn't support runtime values
+  return {
+    menuClassName: `left-0 sm:w-auto sm:min-w-fit`,
+    menuStyle: isMobile
+      ? {
+          width: `calc(100vw - ${2 * offset.left}rem)`,
+          minWidth: `calc(100vw - ${2 * offset.left}rem)`,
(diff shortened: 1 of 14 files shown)


rdig (approved):
Very nice work here! Everything works nice and it's smooth.

Great work as always!


https://github.com/user-attachments/assets/8af5cb14-be9c-4a6e-91cd-90548d0d8e42


https://github.com/user-attachments/assets/52226345-b5b4-4285-afa8-84f2e75924da


https://github.com/user-attachments/assets/b52bcf6e-e887-4da0-8146-883d2174752f


https://github.com/user-attachments/assets/6fa88df1-3d1d-4304-8b30-39b92c73cc90

iamsamgibbs (changes requested):
Thanks for the changes, this is mostly looking better!

Unfortunately there is a horizontal scroll with some of the popper elements:

https://github.com/user-attachments/assets/d8ebf658-d8c5-4f06-b646-dfd5f8612a5e

https://github.com/user-attachments/assets/c11ddb0e-f5f6-4511-a267-ca1b3bc9f188

This isn't a problem on master so has been introduced here.

Somewhat unrelated to the issues here, but seeing as you've resolved the max height issues with the user menu, I wonder if you could take a look at the UserHub too? (Otherwise we can look at this in a separate issue!)

Too tall when on transactions tab:
Screenshot 2025-02-05 at 08 52 56

Too short when on the stakes tab, should have a min-height:
Screenshot 2025-02-05 at 08 53 04

Everything else is good though.

Team select does not appear on top of the header.

https://github.com/user-attachments/assets/1b98fd8e-75f3-48dd-bfde-b6e6cfe03c10

User menu looks great everywhere:

Screenshot 2025-02-05 at 08 52 41

And doesn't cover header on mobile:

Screenshot 2025-02-05 at 08 54 21

mmioana:
Ok, I have come up with a different solution for the horizontal overflowing issue for the colony links and reputation chart menus ✨ Can I get your amazing 👀 for one more round of reviews on this one @iamsamgibbs and @rdig ?

src/hooks/usePopperCustomPositioning.ts
```diff
@@ -0,0 +1,85 @@
...
+      if (bottomOffset < menu.clientHeight) {
+        translateY = `calc(-100% - ${offset.top}rem)`;
+      }
+
+      menu.style.transform = `translate(${translateX}, ${translateY})`;
+    };
+
+    adjustMenuPosition();
+
+    window.addEventListener('resize', adjustMenuPosition);
```
iamsamgibbs, line 70: This all looks good! I will give a proper review later. Only thing that jumps out is that we should maybe be throttling this?

rdig (approved):
Re-approving

src/hooks/usePopperCustomPositioning.ts
mmioana: Oh, that is actually a valid point 🤔 I'll have a look over it 🙌

mmioana: Done 💪

iamsamgibbs (changes requested):
Thanks for the changes @mmioana ! The horizontal scroll issue is resolved, but now the popover menu is not the correct width and is causing weird text alignment issues:

Screenshot 2025-02-06 at 12 10 31

Screenshot 2025-02-06 at 12 11 14

The UserHub is looking a lot better though!

Screenshot 2025-02-06 at 12 12 27

Screenshot 2025-02-06 at 12 12 34

Screenshot 2025-02-06 at 12 13 10

mmioana:
Hey @iamsamgibbs I didn't manage to reproduce the menu dropdown issue, but after digging a bit, it could have been caused by how tailwind handles or actually doesn't handle dynamic classes. So I changed my approach to use the good old `style` for the mobile `width` and `min-width`. Would you mind re-checking? 👀

rdig (approved):
Re-approving

PR #4228, opened 2025-02-03
Fix: Hide accept all funds button after claiming funds

### LINKED ISSUE #3918: [Incoming funds - QA] "Accept All Available Funds" button remains enabled after being clicked
## Steps to reproduce

1. Send tokens from other colony to colony. 
2. Open colony https://arbitrum-sepolia.colony.io/invite/345/3de70ac0-bff1-4fc9-9c72-10081917bd50
3. Navigate to the Incoming Funds section.
4. Click the "Accept All Available Funds" button to process all available funds.
5. Observe that the funds are accepted, but the button remains enabled.

## Expected behaviour

* After successfully processing all available funds, the "Accept All Available Funds" button should become disabled.

## Actual behaviour

* The button remains enabled after processing all funds.

image

https://github.com/user-attachments/assets/bcb148de-c20c-45d3-a516-b65719c5455d

iamsamgibbs:
## Description

This PR fixes a bug with the "Accept all funds" button where it would stay enabled even after claiming all funds.

This was caused by checking if the length of the unclaimed array is longer than 0, however we always insert an unclaimed native token chain claim so that the token appears on the page. The fix is to also filter claims with a balance of 0.

Note: The original issue also shows an issue with claims not working at all on QA. I've not been able to recreate this at all locally.

## Testing

On a fresh dev environment, navigate to the incoming funds page. (If not on a fresh environment just send some tokens to the colony).

Accept all available funds and after the transactions process the button should disappear.

## Diffs

**Changes** 🏗

* Unclaimed funds with a balance of 0 are now filtered out.

Resolves #3918

### DIFF
--- a/src/components/frame/v5/pages/FundsPage/partials/FundsTable/FundsTable.tsx
+++ b/src/components/frame/v5/pages/FundsPage/partials/FundsTable/FundsTable.tsx
@@ -28,7 +28,9 @@ const FundsTable: FC = () => {
   const isMobile = useMobile();
   const { filters, searchedTokens, activeFilters } = useFundsTable();
   const claims = useColonyFundsClaims();
-  const unclaimedClaims = claims.filter((claim) => !claim.isClaimed);
+  const unclaimedClaims = claims.filter(
+    (claim) => !claim.isClaimed && claim.amount !== '0',
+  );
   const allUnclaimedClaims = Array.from(
     new Set(unclaimedClaims.map((claim) => claim.token?.tokenAddress || '')),
   );


rumzledz (approved):
Bish bash bosh ✨ 

accept-all-funds-disappear

rdig (approved):
Very nice!

davecreaser (approved):
Nice and simple!

Working as expected:

https://github.com/user-attachments/assets/4dfafdf9-7320-45e1-918f-53abb2829d29

PR #4229, opened 2025-02-03
Fix: App crashes when split payment total is small

### LINKED ISSUE #4031: [Split payment] App crashes when split payment total amount is set to a number 1x10 pow -6 and less
## Steps to reproduce

 1. Open the Split Payment form.
 2. Set the Total Amount to 0.00000001. (1*10 pow -6)
 3. Change the Percentage Value of a recipient to 30%.

## Expected behaviour

* The application correctly handles the small float number and recalculates the distribution without crashing. If the entered amount is not a valid and cannot be processed - it shows a clear validation feedback 

## Actual behaviour

* The application crashes when attempting to handle the adjustment.

image

davecreaser:
## Description

- Fix an issue with the app crashing when the split payment type is too small.
- Took a while to figure out the actual root cause of this, since the crash was caused by an invalid input to BigNumber, but I had to actually trace it back about 6 steps to find out why it was ending up in that format.
- Long story short, when the user changes the percentage of a payout in a split payment, there is a function run to calculate the amount. When this amount is set, we need to use the `toFixed` instead of `toString` method from `decimal.js`, because otherwise when the amount is small it will be in scientific notation, which down the line (after trying to move the decimal point and convert to wei and then pass to BigNumber) causes the crash 😥 

- Small side note: I also fixed some awkward padding / styling for scientific notation amounts by wrapping the amounts in span tags in the completed action.

## Testing

* Open the Split Payment form
* Set the Total Amount to 0.000001 (1*10 pow -6) or smaller
* Change the Percentage Value of a recipient to anything less than 100%
* The app shouldn't crash 🤞 

Screenshot 2025-02-03 at 12 00 59

* Just to be sure, you could also try if you wish to create a few other split payments just to ensure nothing broke

## Diffs

**Changes** 🏗

* use `toFixed` instead of `toString` when calculating split payment amounts

Resolves #4031

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/forms/SplitPaymentForm/partials/SplitPaymentRecipientsField/hooks.tsx
+++ b/src/components/v5/common/ActionSidebar/partials/forms/SplitPaymentForm/partials/SplitPaymentRecipientsField/hooks.tsx
@@ -187,7 +187,7 @@ export const useRecipientsFieldTableColumns = ({
                         .toDecimalPlaces(
                           token.decimals || DEFAULT_TOKEN_DECIMALS,
                         )
-                        .toString();
+                        .toFixed();
 
                       if (
                         dataRef.current?.[row.index].amount === amountCalculated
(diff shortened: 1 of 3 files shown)


rumzledz (approved):
Must have been quite the (pesky) debug journey @davecreaser 😂 I can confirm that the app no longer crashes when the percentage value of a member is less than 100% ✅ 

split-no-crash

iamsamgibbs (approved):
Great job pinning down this issue, this split payment form is a bit of a pain to work your way through!

Screenshot 2025-02-04 at 11 11 34

Screenshot 2025-02-04 at 11 12 41

All working nicely 🎉

rdig (approved):
Nice fix!

PR #4174, opened 2025-01-23
[Proxy colonies] Handle incoming funds

iamsamgibbs:
## Description

Block-ingestor PR: [https://github.com/JoinColony/block-ingestor/pull/325](https://github.com/JoinColony/block-ingestor/pull/325)

This PR adds incoming funds for proxy colonies.

The native chain token claims are generated by the `fetchColonyNativeFundsClaim` lambda, this PR refactors the lambda to also check for balances on any supported chain where a proxy colony has been deployed and return the claims as an array.

The block-ingestor changes handle the incoming funds for all other tokens so check the changes over there too.

For now all tokens on the proxy chain are treated as completely different tokens and are not grouped, so ETH on the main chain is listed separately from ETH on the proxy chain.

## Testing

* Step 1 - Create a "Manage Supported Chains" action and add proxy chain 1.

Screenshot 2025-01-24 at 18 24 16

* Step 2 - We need to add a new network to your metamask wallet! In the metamask extension, make sure you have the Leela wallet connected in metamask, then open the network select menu.

Screenshot 2025-01-24 at 18 15 50

* Step 3 - Add a custom network

Screenshot 2025-01-24 at 18 15 20

* Step 4 - Give it whatever name and currency symbol you like. Set the RPC URL to `127.0.0.1:8546` and the chain id to `265669101`

Screenshot 2025-01-24 at 18 17 34

* Step 5 - Copy the colony address from the dashboard
* Step 6 - In the meta mask extension click "Send"
 
Screenshot 2025-01-24 at 18 24 41

* Step 7 - Paste the colony address and enter an amount.
 
Screenshot 2025-01-24 at 18 24 47

If you payment fails, clear the tab data under advanced settings and try again.
 
Screenshot 2025-01-24 at 18 25 00

* Step 8 - Refresh the colony and navigate to the incoming funds page. You should see the new incoming fund on the proxy chain!

Screenshot 2025-01-24 at 18 25 30

For further testing send more funds (note all native funds are summed together per chain rather than shown as individual claims). Try adding the second proxy chain and sending funds there too.

## Further testing

We can also test non native chain tokens, but this will require us deploying a new token! Luckily @mmioana has written a script for this!

In your terminal, run `node scripts/deploy-token-to-chain.js`
Please use the following values for the input
foreignChainId `265669101`
foreignChainRpcUrl `http://localhost:8546`
colonyAddress `<COLONY_ADDRESS_HERE>`

[...]

PR #4253, opened 2025-02-05
[Proxy Colonies] Feat: Proxy colony balance validation

### LINKED ISSUE #4193: [Proxy colonies M3] feat: implement form validation of a proxy colony's balance
### Story

As a user making a cross chain simple payment action, I want to see an error if the proxy colony doesn't have enough funds.

### Description

After M2 is completed and we know the balance of tokens per chain, we need to query the database for info on available balance then put that as the maximum for the field.

- [ ] Users can't create actions paying out more than the available balance on each chain
- [ ] Existing validation still works as intended.

iamsamgibbs:
## Description

This PR adds validation for balances on a proxy chain to the simple payment action (and should work for other actions once those forms have been updated).

It also fixes an issue with the `fetchTokenFromChain` lambda which was returning a malformed token due to the fragment changes.

## Testing

* Step 1 - Create a "Manage supported chain" action to add proxy chain 1.

Screenshot 2025-01-31 at 11 50 31

* Step 2 - Switch your metamask wallet to proxy chain 1 (follow the steps in #4174 if you haven't already set it up)
* Step 3 - Copy the colony address and send tokens via metamask to the colony.

Screenshot 2025-01-31 at 11 54 51

* Step 4 - Deploy a token to the proxy chain, mint tokens and add it as an approved token.

In your terminal, run `node scripts/deploy-token-to-chain.js`
Please use the following values for the input
foreignChainId `265669101`
foreignChainRpcUrl `http://localhost:8546`
colonyAddress `<COLONY_ADDRESS_HERE>`

Then we must mint some tokens to generate our incoming funds. Copy the token address that was printed to the console and paste it somewhere to access later.

Run `node scripts/unlock-mint-tokens-on-chain.js`
Please use the following values for the input
foreignChainRpcUrl `http://localhost:8546`
colonyAddress `<COLONY_ADDRESS_HERE>`
tokenAddress `<TOKEN_ADDRESS_YOU_COPIED_FROM_THE_CONSOLE>`
tokenAmount `100`

Then we need to add this as an approved token, otherwise it won't show in the incoming funds list. The "manage tokens" action has yet to be updated to support this, so we must manually add it to the database. In GraphiQL, run this query:

```
mutation MyMutation {
  createColonyTokens(input: {tokenID: "<TOKEN_ADDRESS>", colonyID: "<COLONY_ADDRESS>"})
}
```

* Step 5 - Refresh and navigate to http://localhost:9091/planex/balances confirm that the current balances for the tokens on the proxy chain are 0.

Screenshot 2025-01-31 at 12 04 39

* Step 5 - Navigate to http://localhost:9091/planex/incoming
* Step 6 - Accept all available funds.

Screenshot 2025-01-31 at 12 00 19

* Step 7 - Back to the balances page, check that the newly claimed tokens are shown.

Screenshot 2025-01-31 at 12 00 35

* Step 8 - Create a simple payment action. In the "From" field, select the proxy chain. In the "Amount" field, select a token and enter an amount higher than the available balance. Check the error validation shows. Then check it disappears with an amount below the available balance.

Screenshot 2025-02-06 at 09 49 10

* Step 9 - Switch the "From" field to the "Ganache" main chain. (Note: The amount is not revalidated when switching chains or domains (the same is true on master for domains) - this should be handled as a separate issue.) Enter a new amount and check the validation works as expected.

Screenshot 2025-02-06 at 09 49 24

## Diffs

**New stuff** ✨

* Balances are now validated correctly for proxy chain balances

**Changes** 🏗

* Fix token structure in `fetchTokenByChain` lambda

Resolves #4193

### DIFF
--- a/amplify/backend/function/fetchTokenFromChain/src/index.js
+++ b/amplify/backend/function/fetchTokenFromChain/src/index.js
@@ -32,6 +32,9 @@ const baseToken = {
   decimals: null,
   type: null,
   colonies: null,
+  chainMetadata: {
+    chainId: '',
+  },
 };
 
 const setEnvVariables = async (network) => {


amplify/backend/function/fetchColonyBalances/src/index.js
```diff
@@ -201,7 +205,11 @@ exports.handler = async ({ source: { id: colonyAddress } }) => {
           };
         });
 
-        tokens.map(async ({ token }) => {
+        const proxyChainTokens = tokens.filter(
+          (token) => token.token.chainMetadata.chainId === proxyChainId,
+        );
+
+        proxyChainTokens.map(async ({ token }) => {
```
iamsamgibbs, line 212: Don't bother mapping through tokens that don't exist on this chain where the balances will definitely be 0.

rdig (approved):
Good to go

mmioana (approved):
Really nice work @iamsamgibbs 🎉 the amount field is validated based on the available funds on each chain
Screenshot 2025-02-07 at 13 42 51

https://github.com/user-attachments/assets/8295d132-a00a-4b5b-ae07-1e7d03251eb7

PR #4256, opened 2025-02-06
Fix: Token balances not showing in amount field

iamsamgibbs:
## Description

This PR fixes an issue where token balances were failing to load in the amount field.

## Testing

* Step 1 - Create any action with an amount field (simple payment)
* Step 2 - Check the tokens have the expected balances

Screenshot 2025-02-06 at 12 23 51

## Diffs

**Changes** 🏗

* Check domainId is not `''`


bassgeta (approved):
What a find, nice one! 
image
And for a different team, they funds are filtered
image

Nortsova (approved):
Niiice catch 😎 

image

rdig (approved):

PR #4257, opened 2025-02-06
fix: show max length counter on action title

### LINKED ISSUE #4014: [PROD] Incorrect input validation for action titles
## Steps to reproduce

1. Create a new action, any type is suitable. 
2. In the title field, add content until you reach the input limit.
3. Note the input field does not show the counter once the character limit is exceeded. 

## Expected behaviour

* The character counter should show to the right of the input field once the mac character limit has been reached to match the error state originally designed & completed but since changed.

image

**Figma design:** 
https://www.figma.com/design/5V8pr7iMwXsT9L3VAZsmUt/Actions-%26-Motions?node-id=10818-342703&t=UdERE0PmXhLGqHMb-4

## Actual behaviour

* The character count is missing from the title input once the user reaches the max char input and doesn't match the Figma designs.

bassgeta:
## Description

The counter wasn't even absolutely positioned to the parent div :(
I've taken creative liberty here to include the counter outside of the textarea so we don't get overlapping elements (much like it's already done in other parts of the app).

[Figma link](https://www.figma.com/design/5V8pr7iMwXsT9L3VAZsmUt/Actions-%26-Motions?node-id=2729-248940&t=ZWLElrFeEJsftbZZ-0)

## Testing

1. Try to create a `Create new team` action. Input something long for the name, purpose and the action title
2. Verify that all the textarea fields have the max counter below them to the right 
image

## Diffs

**Changes** 🏗

* Renamed the i18n string to just `errors.text.maxLength`
* The counter is now absolutely positioned relative to its parent div

Resolves  #4014

### DIFF
--- a/playwright/models/advanced-payment.ts
+++ b/playwright/models/advanced-payment.ts
@@ -17,7 +17,7 @@ type CSVFileInput =
 export class AdvancedPayment {
   static readonly validationMessages = {
     title: {
-      maxLengthExceeded: 'Title must not exceed 60 characters',
+      maxLengthExceeded: 'The maximum character limit is 60.',
       isRequired: 'Title is required',
     },
     allRequiredFields: [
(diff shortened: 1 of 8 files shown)


rdig (approved):
Very nice work! :100:

PR #4262, opened 2025-02-09
feat: disable custom transactions for colony version less then 17 and show an error

### LINKED ISSUE #4261: [Arbitrary Txs] Show error when colony version is incompatible with arbitrary transactions
## Specification

For any colony version older than v17, the following error callout should show when "Custom transaction" action is selected:

"Custom transactions requires Colony Network version 17 or higher."

It should also include a link to the Upgrade colony action with the link text "Upgrade colony".

The design should match other error callouts, for example:

Image

Nortsova:
## Description

For any colony version older than v17, the following error callout should show when "Custom transaction" action is selected:
"Custom transactions requires Colony Network version 17 or higher."
It should also include a link to the Upgrade colony action with the link text "Upgrade colony".

## Testing

Step 1. Open `src/components/v5/common/ActionSidebar/partials/hooks.ts` and update line 116 to 18 version. 
Step 2. Open "Custom transactions"
Step 3. Verify that form is disabled and you can see an error
image

Step 4. Verify that 'Upgrade colony' button redirect to 'Upgrade colony' action 

Step 5. Open `src/components/v5/common/ActionSidebar/partials/hooks.ts` and update line 116 to 17 version. 
Step 6. Verify that the form is enabled again

Resolves #4261

### DIFF
--- a/src/components/v5/common/ActionSidebar/partials/forms/ArbitraryTxsForm/ArbitraryTxsForm.tsx
+++ b/src/components/v5/common/ActionSidebar/partials/forms/ArbitraryTxsForm/ArbitraryTxsForm.tsx
@@ -20,8 +20,11 @@ const ArbitraryTxsForm: FC<ActionFormBaseProps> = ({ getFormOptions }) => {
     <>
       <DecisionMethodField disabled={isFieldDisabled} />
       <CreatedIn readonly />
-      <Description />
-      <ArbitraryTransactionsTable name="transactions" />
+      <Description disabled={isFieldDisabled} />
+      <ArbitraryTransactionsTable
+        name="transactions"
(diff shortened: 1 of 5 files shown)


mmioana (dismissed):
Really nice work on this @Nortsova 🙌 

After changing the minimum version to `18`, the banner is shown and the form is disabled

https://github.com/user-attachments/assets/38d0bbaf-852a-49e3-bcad-0c343b9075f0

After reverting the change, the form is re-enabled and the banner is no longer shown

Screenshot 2025-02-10 at 10 01 01

Nice work! ✨

mmioana (approved):
Re-approving this 🥇  nice work and thanks for adding the missing dependency!

